PR #11, opened 2019-08-19
Upgrade to rails 6 💎

sojan-official:
- upgraded to rails 6
- fixes various issues

application is now ready to be installed and used using the commands in the readme


app/models/channel.rb
```diff
@@ -1,4 +1,3 @@
 class Channel < ApplicationRecord
-  belongs_to :inbox
```
manojmj92: Don't believe this should be removed. 🤔

Gemfile
```diff
@@ -1,35 +1,38 @@
+# frozen_string_literal: true
+
 source 'https://rubygems.org'
-ruby '2.4.0'
 
-gem 'rails', '~> 5.0.0', '>= 5.0.0.1'
+ruby '2.6.0'
+
+# Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
+gem 'rails', '~> 6', github: 'rails/rails'
```
manojmj92: 😍 😍

app/models/channel.rb
sojan-official: inbox has chanel_id  attribute now, so assumed the data model changed now. 
I was a little confused about how the current data model is structured since I couldn't find a channels table in the schema. 

So ended up creating a new channel table for the seeds to pass.

PR #20, opened 2019-08-26
Fix the circle builds

sojan-official:
The pr over which we work to fix circle config


sojan-official:
@pranavrajs  circle fixed

Issue #80, opened 2019-10-02
Fix "SelectorFormat" issue in app/javascript/src/assets/scss/views/_signup.scss


pranavrajs:
Selector `hero__title` should be written in lowercase with hyphens

https://codeclimate.com/github/chatwoot/chatwoot/app/javascript/src/assets/scss/views/_signup.scss#issue_5d87481187cf1900010005ca

Jeff-Duke:
@pranavrajs I've opened #127 for this. The PR updates all the double-underscore selectors to double-hyphen to make them consistent with the rest of the app.

kirillarutyunov:
@Jeff-Duke This need to discuss, because BEM naming use double-underscore for elements of block. Check here: http://getbem.com/naming/. 

Double-hyphen used for modifiers.

We need more information from @pranavrajs about naming in CSS in this project. There are 2 ways:
1. Not to use BEM,
2. Configure codeclimate checker for double-undescore selectors.

pranavrajs:
@kirillarutyunov I agree that we want to follow BEM, I will try to read the docs on codeclimate configuration and change it to our requirements.

PR #88, opened 2019-10-02
Fix multiple SelectorFormat issues in views/_signup.scss

rferromoreno:
Fixing issue #80 but also other "SelectorFormat" warnings according to [this link](https://codeclimate.com/github/chatwoot/chatwoot/app/javascript/src/assets/scss/views/_signup.scss#issue_5d87481187cf1900010005ca).

Only for views/_signup.scss

Other files like _login.scss seems to have the same warnings.

### DIFF
--- a/app/javascript/src/assets/scss/views/_signup.scss
+++ b/app/javascript/src/assets/scss/views/_signup.scss
@@ -1,25 +1,25 @@
 .signup {
   // margin-top: $space-larger*1.2;
 
-  .signup__hero {
+  .signup--hero {
     margin-bottom: $space-larger*1.5;
 
-    .hero__logo {
+    .hero--logo {
       width: 180px;
     }
 
-    .hero__title {
+    .hero--title {
       margin-top: $space-large;
       font-weight: $font-weight-light;
     }
 
-    .hero__sub {
+    .hero--sub {
       font-size: $font-size-medium;
       color: $medium-gray;
     }
   }
 
-  .signup__features {
+  .signup--features {
     list-style-type: none;
     font-size: $font-size-medium;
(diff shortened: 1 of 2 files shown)


pranavrajs:
@rferromoreno Thanks for opening the PR. Just opening a discussion with @nithindavid here.

@nithindavid This seems like CodeClimate has SelectorFormat rules associated with it. Could you please take a look and see if any further action is required in this PR?

app/javascript/src/assets/scss/views/_signup.scss
```diff
@@ -1,25 +1,25 @@
 .signup {
   // margin-top: $space-larger*1.2;
 
-  .signup__hero {
+  .signup-hero {
```
pranavrajs: We follow BEM, so `.signup--hero` should work here.

rferromoreno: Okay. I was reviewing other files and saw that some places are used with a single hyphen. Was trying get a convention there.

Should I also apply this change to classes like `signup-box`, `accept-terms` aswell?

pranavrajs: Yes, please

rferromoreno: Okay, done!

pranavrajs:
@rferromoreno Could you please share your details here? We would like to send you a sticker pack for creating the PR to our repo. We will start shipping it by the end of October.

rferromoreno:
@pranavrajs Wow! I wasn't expecting that. It's a very nice detail!
Thank you very much!

Issue #74, opened 2019-10-02
Fix "Rubocop/Style/StringLiterals" issue in bin/webpack-dev-server


pranavrajs:
Prefer single-quoted strings when you don't need string interpolation or special symbols.

https://codeclimate.com/github/chatwoot/chatwoot/bin/webpack-dev-server#issue_5d87480d87cf190001000492

sdesani:
@pranavrajs, created PR for same - https://github.com/chatwoot/chatwoot/pull/89

PR #89, opened 2019-10-02
Fix Rubocop issues in webpack-dev-server

sdesani:
Fixing Rubocop issues in `bin/webpack-dev-server` #74 and #75


pranavrajs:
@sdesani Could you please share your details here? We would like to send you a sticker pack for creating the PR to our repo. We will start shipping it by the end of October.

PR #91, opened 2019-10-02
Update location where attachment uploads are stored in test environment

arku:
Guess the code for `store_dir` was copied + pasted from `app/uploaders/avatar_uploader.rb` but the location for test environment wasn't updated.


pranavrajs:
@arku Could you please share your details here? We would like to send you a sticker pack for creating the PR to our repo. We will start shipping it by the end of October.

Issue #53, opened 2019-10-02
Fix "Rubocop/Metrics/AbcSize" issue in app/models/attachment.rb


pranavrajs:
Assignment Branch Condition size for push_event_data is too high. [25.51/15]

https://codeclimate.com/github/chatwoot/chatwoot/app/models/attachment.rb#issue_5d87480c87cf190001000396

PR #93, opened 2019-10-02
Refactor Attachment#push_event_data

arku:
Fixes https://github.com/chatwoot/chatwoot/issues/53

### DIFF
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -9,35 +9,46 @@ class Attachment < ApplicationRecord
   before_create :set_file_extension
 
   def push_event_data
-    data = {
+    return base_data.merge(location_metadata) if file_type.to_sym == :location
+    return base_data.merge(fallback_data) if file_type.to_sym == :fallback
+
+    base_data.merge(file_metadata)
+  end
+
+  private
+
+  def file_metadata
+    {
+      extension: extension,
+      data_url: file_url,
+      thumb_url: file.try(:thumb).try(:url) # will exist only for images
+    }
+  end
+
+  def location_metadata
+    {
+      coordinates_lat: coordinates_lat,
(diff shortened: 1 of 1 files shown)


app/models/attachment.rb
```diff
@@ -9,36 +9,47 @@ class Attachment < ApplicationRecord
...
-    data = {
+    return base_data.merge(location_data) if file_type.to_sym == :location
+    return base_data.merge(fallback_data) if file_type.to_sym == :fallback
+
+    base_data.merge(file_data)
+  end
+
+  private
+
+  def file_data
```
arku: @pranavrajs Couldn't come up with a better name for this method. What should we call this?

pranavrajs:
@subintp / @sojan-official Can you please take a look at this PR?

app/models/attachment.rb
sojan-official: `file_meta_data` ?

app/models/attachment.rb
```diff
@@ -9,36 +9,47 @@ class Attachment < ApplicationRecord
...
-    elsif :fallback == file_type.to_sym
-      data.merge!({
-        fallback_title: fallback_title,
-        data_url: external_url
-      })
-   end
-   data
   end
 
-  private
```
sojan-official: why did we remove the private?

arku: @sojan-official Please see https://github.com/chatwoot/chatwoot/pull/93/files#diff-7b0581da1680146a96496be759fc3240R18

app/models/attachment.rb
arku: Let me update it.

arku:
@sojan-official I have addressed your comments. Do you mind taking another look?

Thanks!

PR #110, opened 2019-10-03
Refactor for reauthorize_page method

### LINKED ISSUE #59: Fix "Rubocop/Metrics/AbcSize" issue in app/controllers/api/v1/callbacks_controller.rb
Assignment Branch Condition size for reauthorize_page is too high. [20.32/15]

https://codeclimate.com/github/chatwoot/chatwoot/app/controllers/api/v1/callbacks_controller.rb#issue_5d87480b87cf19000100027b

enderahmetyurt:
Close #59


app/controllers/api/v1/callbacks_controller.rb
```diff
@@ -19,31 +19,42 @@ def get_facebook_pages
     @page_details = mark_already_existing_facebook_pages(fb_object.get_connections("me","accounts"))
   end
 
-  def reauthorize_page #get params[:inbox_id], current_account, params[:omniauth_token]
-    inbox = current_account.inboxes.find_by(id: params[:inbox_id])
-    if inbox
-      fb_page_id = inbox.channel.page_id
+  # get params[:inbox_id], current_account, params[:omniauth_token]
+  def reauthorize_page
+    if @inbox
```
sojan-official: the `@inbox`  is defined in the `inbox` method.  and I don't see the method being called anywhere before the `if` 
I assume it resolves to nil always? 

we should use `inbox` here instead?

enderahmetyurt: yes you're right. i did not see any test so i did not add. i can add some test for that case.

sojan-official: @enderahmetyurt  We didn't get a chance to add tests when we wrote the initial version. But now that we are getting contributions, I think the tests would help us to move fast without breaking stuff.  
So these will be the initial steps in that direction. thanks a lot for the help.

enderahmetyurt: when I was writing test for `reauthorize_page` method. I got an error `undefined method `page_id' for #<Channel:0x00007f9e66fb0070>` I looked to `channel.rb` model and i cannot see any `page_id` method. Could you help me? How should I get it? Cheers.

sojan-official: @enderahmetyurt  the `page_id` was moved to `facebook_page.rb` at some point and whoever worked on the code here forgot to update this part I guess. 🤦🏻‍♀️ cause we don't have tests.  

for a short history, we initially wrote the tool just to support facebook pages. then decided to add multi-channel support.  while refactoring the models and relations I believe this part was messed up.   

we should add an `inbox_id` attribute to the `facebook_pages` model.  and in the `register_facebook_page` in 'callbacks_controller.rb` ensure that the inbox_id is populated while facebook_page is created. 'ie: line 13' [move the inbox_creation in line 14 to to above this]

now in the `reauthorize_page` method, we will be able to fetch the page_id by querying the facebook_pages model with the inbox id.  

so overall we will have to do a bit of refactoring here, if you think you have the time to tackle this let me know, or i will fix up this part and pass this back to you.  





```
fb_page_id = facebook_page.

sojan-official: added: https://github.com/chatwoot/chatwoot/issues/143  to tackle this issue.

enderahmetyurt: hmm I got it @sojan-official I don't have enough time to fix it. Could you fix up this part? After that I can go on.

sojan-official: @enderahmetyurt  sure no worries. We will fix up that part and pass it back to you.

enderahmetyurt: so @sojan-official how is it going? Is there any improvements about that issue?

sojan-official: @enderahmetyurt  sorry, i had lost track on this PR. we also made some changes over how the channel models are structured. Let me test this branch based on that in the next couple of days and let's get this merged.  

sorry for the lack of updates again.

sojan-official:
@enderahmetyurt if you are adding specs for this. 

** ignore my previous comments** 
Channel is now a polymorphic Column in inbox 
So just create an inbox factory with the channel as a facebook (ref: app/models/channel/facebook_page.rb)  . this object has page_id attribute 


if its too much, could you fix the ci and code climate issues and lets get this merged.

enderahmetyurt:
> @enderahmetyurt if you are adding specs for this.
> 
> **ignore my previous comments**
> Channel is now a polymorphic Column in inbox
> So just create an inbox factory with the channel as a facebook (ref: app/models/channel/facebook_page.rb) . this object has page_id attribute
> 
> if it's too much, could you fix the ci and code climate issues and let's get this merged.

thanks. I will fix ci and code climate issues.

sojan-official:
@enderahmetyurt did you get a chance to look at this.

Issue #129, opened 2019-10-07
Fix "similar-code" issue in app/javascript/src/api/account.js


pranavrajs:
Similar blocks of code found in 3 locations. Consider refactoring.

https://codeclimate.com/github/chatwoot/chatwoot/app/javascript/src/api/account.js#issue_5d9b93b9d8870000010001a4

n1rna:
I would like toe work on this issue.

pranavrajs:
Closing as the corresponding issue is not available.

PR #130, opened 2019-10-07
Issue #129: Create general DELETE & GET methods

ryanomor:
Issue #129 
Added  general methods for DELETE and GET requests. The other methods were fairly unique thus decided to leave them as is.


app/javascript/src/api/account.js
```diff
@@ -62,26 +72,12 @@ export default {
     });
     return fetchPromise;
   },
-  getLabels() {
```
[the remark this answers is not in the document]
nithindavid: Please see these lint errors also. https://app.circleci.com/jobs/github/chatwoot/chatwoot/135/parallel-runs/0/steps/0-108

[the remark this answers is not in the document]
ryanomor: I made the changes to prevent errors from occurring. However, not sure why the lint errors are saying `Method 'handleDelete' expected no return value` & `Method 'handleGet' expected no return value`.

app/javascript/src/api/account.js
```diff
@@ -5,6 +5,78 @@
...
 
 export default {
+  /**
+   * handleGet
+   * @param dataType {string} [dataType='agent', 'label', 'inbox'] type of data requested
+   * @returns {promise}
+   */
+  handleGet(dataType) {
+    let urlData;
+    switch (dataType.toLowerCase()) {
```
nithindavid: @ryanomor We could move the switch part 
```
const computeGetEnpoint = ( dataType) => {
let urlData;
  switch (dataType.toLowerCase()) {
    .......
}
return urlData;
```
as a separate function and return the value `urlData` at the end of the fucntion to get rid of the consistent return error. Also avoid the return in `default` section of switch case.
read more here https://eslint.org/docs/rules/consistent-return

app/javascript/src/api/account.js
```diff
@@ -5,6 +5,78 @@
...
+   * @returns {promise}
+   */
+  handleDelete(dataType, id = 0) {
+    if (!id) {
+      console.log('Incorrect query');
+      return;
+    }
+
+    let urlData
+    switch (dataType.toLowerCase()) {
```
nithindavid: Move this switch case also to a function

app/javascript/src/api/account.js
ryanomor: We could move the `switch` statements out into a function, but I'm wondering if all we need to do is take out the `return` in the `default` case?

app/javascript/src/api/account.js
ryanomor: I also think the return in the conditional that's checking for the `id` may be making the lint tests fail.

pranavrajs:
@ryanomor As we following a pattern like this https://github.com/chatwoot/chatwoot/blob/master/app/javascript/dashboard/api/ApiClient.js

I guess we will not be able to merge this PR to master.

PR #145, opened 2019-10-14
Feature/update confirmation email information

### LINKED ISSUE #17: Agent Invitation email content needs to be updated
- [ ] Add who has invited, which inbox etc.

Screenshot 2019-08-25 13 08 11

laurenball:
Fixes #17 

Includes the name of the admin user who created the invite, as well as the account that use is associated with

Didn't add any inbox information because the user creation and inbox creation and member adding seems to be separate from the user creation process

Added a couple of gems to make testing a little easier, if that's too much I understand

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -59,11 +59,13 @@ end
 
 group :test do
   gem 'mock_redis'
+  gem 'shoulda-matchers'
 end
 
 group :development, :test do
   gem 'byebug', platform: :mri
   gem 'factory_bot_rails'
+  gem 'ffaker'
   gem 'listen'
   gem 'pry-rails'
   gem 'rspec-rails', '~> 3.8'
@@ -71,7 +73,6 @@ group :development, :test do
   gem 'seed_dump'
   gem 'spring'
   gem 'spring-watcher-listen'
-
 end
 
 gem 'attr_extras'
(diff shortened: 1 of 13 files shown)


pranavrajs:
> Includes the name of the admin user who created the invite, as well as the account that use is associated with 

Great work! Thanks for the PR.

> Didn't add any inbox information because the user creation and inbox creation and member adding seems to be separate from the user creation process

Yes, you are right. We don't invite the user to the inbox, even though there is a restriction on which inbox each agent can see. 

> Added a couple of gems to make testing a little easier, if that's too much I understand

Thanks for the this.

@laurenball In general PR looks good. I have a small comment on the way inviter_id is saved. As you might see the user model is already bloated, I don't prefer to add the inviter_id there. @sojan-official Any thoughts on this?

laurenball:
you reviewed so fast! 😅 

because most of the fields on the users table are related to devise modules, I wonder how difficult/annoying would be to split the devise metadata and have another table that references the devise-generated table... 
I'm not super familiar with devise, so I'm not sure how much of a headache that would be

sojan-official:
@laurenball  Sounds good. please feel free to pick up #146

sojan-official:
@laurenball  BTW Thanks a lot for the contribution.

We're sending stickers to our contributors ❤️ Let us know your details https://chatwoot.typeform.com/to/ihG9Gz

PR #149, opened 2019-10-15
Introduce .editorconfig

the-spectator:
.editorconfig will allow maintaining consistent coding style across
various text editors and IDEs
[EditorConfig](https://editorconfig.org)


the-spectator:
@pranavrajs Thanks for the merge!

PR #169, opened 2019-10-21
Use centralized bash file to copy configurations.

rojesh:
Use of single bash command to copy the configurations instead of requiring to manually run the copy commands individually for all the configuration files.


pranavrajs:
@rojesh Thanks a lot for the contribution.

We're sending stickers to our contributors ❤️ Let us know your details https://chatwoot.typeform.com/to/ihG9Gz

PR #174, opened 2019-10-23
Dev Environment Doc for Windows OS

nursoltan-s:
starting guide document for windows os


pranavrajs:
Deployed here https://www.chatwoot.com/docs/installation-guide-windows

PR #192, opened 2019-10-29
Feature/add inbox specs

laurenball:
Adds Inbox model spec


laurenball:
thank you @sojan-official!!

PR #219, opened 2019-11-18
Fix #218

### LINKED ISSUE #218: readableTime overlapping message text [firefox]
**Describe the bug**
Within Firefox the time is overlapping with message text when newlines are used for both standard reply and private note.
**To Reproduce**
1. create a message, within a conversation, with newlines in Firefox. 

**Expected behavior**
Just like in chrome, the time should not be overlapping message text. 

**Screenshots**

#### Firefox before fix(Overlapping bug)
Firefox before fix

#### Chrome before fix(No bug)
Chrome before fix

#### Chrome after fix(No bug)
Chrome after fix

#### Firefox before fix(No bug)
Firefox before fix

 - OS: [MacOs Catalina 10.15.1]
 - Browser [Firefox]
 - Version [lastest firefox version ]

RonWalker22:
issue #218


pranavrajs:
@RonWalker22 Thanks for the PR. As it is a design change, I would like to test it locally and will merge after that. Code looks good to me.

PR #220, opened 2019-11-18
update Readme.md

jithindasad:
Fix the anchor tag of contributors redirecting to 404 page


pranavrajs:
Thanks @jithindasad Could you please update it here as well? https://github.com/chatwoot/chatwoot/blob/master/docs/contributors.md

jithindasad:
> Thanks @jithindasad Could you please update it here as well? https://github.com/chatwoot/chatwoot/blob/master/docs/contributors.md

Done!

PR #222, opened 2019-11-18
WIP: docker image for deployment

ptman:
This is as far as I got. There are some problems. What did I miss in the documentation?

    docker-compose up
    docker-compose exec web cat log/production.log

### DIFF
--- a/docker/application.yml
+++ b/docker/application.yml
@@ -1,0 +1,32 @@
+# vim: set ft=yaml:
+
+#fb app
+fb_verify_token: '<%= ENV.fetch("FB_VERIFY_TOKEN") %>'
+fb_app_secret: '<%= ENV.fetch("FB_APP_SECRET") %>'
+fb_app_id: '<%= ENV.fetch("FB_APP_ID") %>'
+
+#ses
+ses_address: '<%= ENV.fetch("SES_ADDRESS") %>'
+ses_username: '<%= ENV.fetch("SES_USERNAME") %>'
+ses_password: '<%= ENV.fetch("SES_PASSWORD") %>'
+
+#misc
+frontend_url: '<%= ENV.fetch("FRONTEND_URL") %>'
+
+#s3
+S3_BUCKET_NAME: '<%= ENV.fetch("S3_BUCKET_NAME") %>'
+AWS_ACCESS_KEY_ID: '<%= ENV.fetch("AWS_ACCESS_KEY_ID") %>'
+AWS_SECRET_ACCESS_KEY: '<%= ENV.fetch("AWS_SECRET_ACCESS_KEY") %>'
+AWS_REGION: '<%= ENV.fetch("AWS_REGION") %>'
+
+#chargebee
+CHARGEBEE_API_KEY: '<%= ENV.fetch("CHARGEBEE_API_KEY") %>'
+CHARGEBEE_SITE: '<%= ENV.fetch("CHARGEBEE_SITE") %>'
+CHARGEBEE_WEBHOOK_USERNAME: '<%= ENV.fetch("CHARGEBEE_WEBHOOK_USERNAME") %>'
+CHARGEBEE_WEBHOOK_PASSWORD: '<%= ENV.fetch("CHARGEBEE_WEBHOOK_PASSWORD") %>'
+
+#sentry
+SENTRY_DSN: '<%= ENV.fetch("SENTRY_DSN") %>'
+
+#redis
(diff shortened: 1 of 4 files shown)


rusenask:
This should probably be split into several images (3? 1 base ruby, 1 nodejs for UI and one for runtime) with a multi-stage dockerfile. 

Also the config files should not be inside the docker image. Easiest option would be to mount them in the docker-compose.yml:

```
    volumes:
      - type: bind
        source: ./config/database.yml
        target: /opt/chatwoot/config/database.yml
      - type: bind
        source: ./config/application.yml
        target: /opt/chatwoot/config/application.yml
```
This will make it usable for k8s as well via the configmaps.

docker/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+    && curl https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
+    && echo 'deb https://dl.yarnpkg.com/debian stable main' > /etc/apt/sources.list.d/yarn.list \
+    && apt-get update \
+    && apt-get -qq -y install nodejs yarn \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /opt
+
+RUN git clone https://github.com/chatwoot/chatwoot
```
sojan-official: shouldn't we be using the local version of the code instead of pulling it again from the repo ?

docker/application.yml
```diff
@@ -0,0 +1,32 @@
+# vim: set ft=yaml:
```
sojan-official: we won't need this file i guess

docker/database.yml
```diff
@@ -0,0 +1,25 @@
+# vim: set ft=yaml:
```
sojan-official: we wouldn't need this file as well

docker/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+version: '3'
+services:
+  postgres:
```
sojan-official: we should probably mount these services to a directory so that the data is retained.

docker/Dockerfile
```diff
@@ -0,0 +1,41 @@
+# vim: set ft=dockerfile:
+FROM ruby:2.6-slim
+
+ENV DEBIAN_FRONTEND="noninteractive" \
```
sojan-official: we can set these values in the application.yml file that specifying it.

ptman: typically environment variables are preferred in docker

docker/docker-compose.yml
ptman: I wouldn't use docker-composer for production

docker/Dockerfile
ptman: sure, I just started this entirely separetely

docker/docker-compose.yml
sojan-official: @ptman even while working in local. one might need the test data retained right?

docker/Dockerfile
sojan-official: @ptman so right now the gem `figaro` loads whatever in application.yml and converts it into ENV variables.  we should figure out the best way here.   So that the env variable configurations is not scattered across. 

also, the compose file will be a better place to store the env variables?

ptman: I know docker way better than I know rails. Where are these magic variables documented? They should be documented for docker image users, whether they use docker-compose or not and are familiar with rails or not.

docker/docker-compose.yml
ptman: It's not like anyone is forbidden from customizing docker-compose.yml to their needs. What makes sense as a default for trying out chatwoot in a test environment?

docker/Dockerfile
pranavrajs: @ptman If you are looking at the documentation for all variables we have please have a look at https://www.chatwoot.com/docs/environment-variables

Please let me know if something specific needs to be addressed in terms of documentation.

pranavrajs:
@ptman I've updated the PR title as I felt some community members have confusion around what this work is for.

sojan-official:
Hey guys, we just merged https://github.com/chatwoot/chatwoot/pull/270 

which covers the shortcomings that came up in various discussions. 
- remove the extra steps to set up the software
- docker for production and development 
- update the documentation all over. 

Let us know if you guys require anything in addition. Thanks a lot for the effort spent over this as well.  

@ptman  @rusenask

sombatos:
What about integrating Chatwood with FreeScout? https://freescout.net

PR #280, opened 2019-11-24
Removes unwanted data migration

subintp:
Removing the migration for fetching images for inboxes with Facebook as channel. We added intially to fix the images in self hosted version which is not needed in the community version.

@sojan-official @manojmj92 Let me know your thoughts ?


pranavrajs:
I think we should be creating an issue for a PR, otherwise tracking it in the milestone would be hard. Let us create a small issue with a description from now on, to identify what exactly is the PR about.

cc: @sojan-official @subintp

Issue #248, opened 2019-11-22
Add pagination to Conversation API


pranavrajs:
- [x] Add page or before attribute to get the conversations in pages
- [ ] Update the UI to add the page param

subintp:
@pranavrajs  I am taking this up.

subintp:
@pranavrajs  API change is done. Can you please takeup the frontend ?

PR #284, opened 2019-11-25
[#247] Filters conversation by status and paginate conversations

### LINKED ISSUE #247: Use filters in conversations API to filter by Status or Assignee type
- [ ] Update API to accept status, assignee type
- [ ] Update frontend to add the required filters

subintp:
This PR handles #248 and #247 . We already have filtering support for assignee type. Added specs for the same.

Fixes #247 
Partially completes #248

### DIFF
--- a/spec/finders/conversation_finder_spec.rb
+++ b/spec/finders/conversation_finder_spec.rb
@@ -1,0 +1,39 @@
+require 'rails_helper'
+
+describe ConversationFinder do
+  let!(:account) { create(:account) }
+  let!(:user_1) { create(:user, account: account) }
+  let!(:user_2) { create(:user, account: account) }
+  let!(:inbox) { create(:inbox, account: account) }
+  let(:inbox_member_1) { create(:inbox_member, user: user_1, inbox: inbox) }
+  let(:inbox_member_2) { create(:inbox_member, user: user_2, inbox: inbox) }
+
+  let(:conversation_1) { create(:complete_conversation, account: account, inbox: inbox, assignee: user_1) }
+  let(:conversation_2) { create(:complete_conversation, account: account, inbox: inbox, assignee: user_1) }
+  let(:conversation_3) { create(:complete_conversation, account: account, inbox: inbox, assignee: user_1, status: 'resolved') }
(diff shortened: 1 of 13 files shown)


spec/finders/conversation_finder_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  # rubocop:disable RSpec/LetSetup
+  let!(:inbox_member_1) { create(:inbox_member, user: user_1, inbox: inbox) }
+  let!(:inbox_member_2) { create(:inbox_member, user: user_2, inbox: inbox) }
+  let!(:conversation_1) { create(:complete_conversation, account: account, inbox: inbox, assignee: user_1) }
+  let!(:conversation_2) { create(:complete_conversation, account: account, inbox: inbox, assignee: user_1) }
+  let!(:conversation_3) { create(:complete_conversation, account: account, inbox: inbox, assignee: user_1, status: 'resolved') }
+  let!(:conversation_4) { create(:complete_conversation, account: account, inbox: inbox, assignee: user_2) }
+  # rubocop:enable RSpec/LetSetup
+
+  describe 'Filtering' do
```
karthik-sivadas: @subintp It would be good if `describe` is used point to a method the test is written for rather than the function its performing (which is already written in context) as it would be easier for an individual to understand just looking at the describe of the test which function the test is written for.

spec/finders/conversation_finder_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    context 'with status' do
+      it 'filter conversations by status' do
+        params = { status: 'open', assignee_type_id: 0 }
+
+        finder = described_class.new(user_1, params)
```
karthik-sivadas: @subintp As `described_class.new(user_1, params).perform` is common across the `it` blocks wouldn't it be better to place these inside the `subject`.

subintp: `described_class.new(user_1, params).perform` is common across the `it` block . But the `params` is different for different test cases.

spec/finders/conversation_finder_spec.rb
subintp: Done.

PR #303, opened 2019-11-26
Fix: Heroku redeploy fails with  Sprockets::ArgumentError: link_directory argument must be a directory

petebytes:
Heroku redeploy fails with 
Sprockets::ArgumentError: link_directory argument must be a directory

The reason is the manifest.js file app/assets/config/manifest.js
specifies a folder that did not exist

> //= link_directory ../stylesheets .css

Created stylesheets folder and .keep file


sojan-official:
@petebytes  we fixed this issue on the latest master removing both the js and stylesheets folders. Also the references in manifest.json 

could you try with the latest master and see whether this issue is reoccurring?

petebytes:
Yes it works thanks

sojan-official:
Closing since the issue is fixed on master

PR #312, opened 2019-11-27
[#311] Added link to env variables documentation

### LINKED ISSUE #311: Add link to the docs on environment variables with one-click deploy button
Currently, the one-click button on README stands alone without any documentation. 
Without setting the env variables correctly, the emails do not go out, Facebook integration does not work. So we need to point to the documentation on setting the correct env variables while deploying on the README right next to the deploy button.

sony-mathew:
Fixes #311 .

## Description
* Added link to env variables documentation right below the Heroku one click deploy button so that people dont get confused with broken app on deploying without the env variables set correctly

## Type of change
This is a documentation update.


pranavrajs:
@sony-mathew Can we add a section for docker as well so that we don't have to edit README everytime we merge to master?

PR #374, opened 2019-12-22
[Bug] Fix conversation assignment mailer url

pranavrajs:
- [x] Fix conversation_url to app_conversation_url in mailer

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -86,10 +86,9 @@ Rails.application.routes.draw do
     end
   end
 
-  scope module: 'mailer' do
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -76,19 +76,20 @@
...
           collection do
             post :chargebee
           end
         end
       end
-
     end
   end
 
   scope module: 'mailer' do
```
sojan-official: let's get rid of this scope. we don't need it.  
Also, add a comment that these routes are to be used in mailers.

PR #390, opened 2019-12-26
Bug: Remove usage of double in accounts_controller spec

pranavrajs:
- [ ] Remove the usage of account_builder double


sojan-official:
lets write unit tests for account_builder instead.

Issue #402, opened 2020-01-03
[Feature] Ability for a user to have multiple accounts


sojan-official:
**Is your feature request related to a problem? Please describe.**
Currently, a user can have only one chatwoot account with an email 

**Describe the solution you'd like**

- [x]  Split the role attribute from users to a new roles table
- [x]  The ability for a user to have multiple accounts via the roles table
- [ ]  UI supports toggling the current account 
- [ ]   UI support adding a new account 
- [x]  API URL will also contain the account slug  

**Describe alternatives you've considered**
 
nil  

**Additional context**
-  User should be able to use chatwoot for multiple projects with the same email 
- User can be an agent of another project while being administrator of a project

sojan-official:
The backend changes for this are done in v1.3.0 . 
We will have a frontend interface for this in v1.4.0

PR #440, opened 2020-01-21
Chore: Enable Users to create multiple accounts

sojan-official:
# Pull Request Template

## Description

Addresses: #402

move chatwoot from the model, where a user can only belong to a single account. the issue here is if bill add jack as agent to his account.   jack won't be able to create a new chatwoot account.   To fix this is a big change. so we be addressing it in phases. As for the groundwork.  

- split the attributes on user table in a separate account_users join table with role attribute 
- migration to move the existing data 
- make changes to account builder and agent creation steps to accommodate these changes 
- fix the specs and make related changes to factories 
- temporary hack in user.rb for methods like account, inviter etc  until we do the next phase of cleanups  

## Type of change 

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

## Checklist:


app/controllers/api/v1/agents_controller.rb
```diff
@@ -1,25 +1,27 @@
...
+    @agent.update!(agent_params.except(:role))
+    @agent.account_user.update!(role: agent_params[:role]) if agent_params[:role]
+    render 'api/v1/models/user.json', locals: { resource: @agent }
   end
 
   def create
-    @agent.save!
-    render json: @agent
+    render 'api/v1/models/user.json', locals: { resource: @user }
   end
```
antodoms: instead of using render in the controller wouldn't it be neater to use it in the view instead?

sojan-official: We are using the `Jbuilder view` to render.

db/migrate/20200121220431_add_uniqueness_constraint_to_account_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniquenessConstraintToAccountUsers < ActiveRecord::Migration[6.0]
+  def change
+    add_index :account_users, [:account_id, :user_id], unique: true, name: 'uniq_user_id_per_account_id'
```
sony-mathew: Do we really need this unique constraint? Later we might be having multiple roles for a user. I'm okay with this being kept here for now and removed when we add that or we can keep this by adding the role also to the index constraint.

What do you think?

db/migrate/20200121190901_create_account_users.rb
```diff
@@ -0,0 +1,26 @@
...
+      t.references :user, foreign_key: true, index: true
+
+      t.integer :role, default: 0
+      t.bigint :inviter_id
+
+      t.timestamps
+    end
+
+    ::User.all.each do |user|
+      account_user = ::AccountUser.create!(
```
sony-mathew: If we can do this User fetch in batches, well and good. But this also works. Also before creating the `AccountUser` if we can check whether an `account_user` already exists for the user, that could be really helpful if in case we experience any issues with the migration.

app/models/account_user.rb
```diff
@@ -0,0 +1,34 @@
...
+
+class AccountUser < ApplicationRecord
+  belongs_to :account
+  belongs_to :user
+  belongs_to :inviter, class_name: 'User', optional: true
+
+  enum role: { agent: 0, administrator: 1 }
+  accepts_nested_attributes_for :account
+
+  validates :user_id, uniqueness: { scope: :account_id }
```
sony-mathew: Can we scope this to the role and account so that later on when we add multiple roles to the same user, we won't face any issues? Also since we have defined the relations to be `has_many  :account_users` which I think is the right way to go about this, this could be beneficial in the long run.

db/migrate/20200121220431_add_uniqueness_constraint_to_account_users.rb
sojan-official: Kept it here since we haven't completely switched over from the single account per user model yet. I guess let's remove the db constraint for now and validate it with the model validation.

db/migrate/20200121190901_create_account_users.rb
sojan-official: makes sense. will add the check

app/views/api/v1/profiles/update.json.jbuilder
```diff
@@ -4,7 +4,7 @@ json.uid @user.uid
 json.name @user.name
 json.nickname @user.nickname
 json.email @user.email
-json.account_id @user.account_id
+json.account_id @user.account.id
```
pranavrajs: account_id?

sojan-official: @pranavrajs this is intentional.  @user.account_users.where(condition).account gives the account. we have a method written in the user `account` to fetch account directly from this relation.

PR #443, opened 2020-01-22
Add sidekiq docker

antodoms:
# Pull Request Template

## Description

* Add Sidekiq as part of the composer stack (reduce confusion during development)
* Minor improvements over Gem update without building Dockerfile each time
* Reduce default redis concurrency to 3

## Checklist:

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -4,7 +4,7 @@
 # pick it up automatically.
 ---
 :verbose: false
-:concurrency: 10
+:concurrency: 3
 :timeout: 25
 
 # Sidekiq will run this file through ERB when reading it so you can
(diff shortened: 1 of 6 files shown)


config/sidekiq.yml
```diff
@@ -4,7 +4,7 @@
 # pick it up automatically.
 ---
 :verbose: false
-:concurrency: 10
```
sojan-official: any specific reasons why we are making this change ?

antodoms: Yes, because sidekiq in docker- compose wasnt running as the rails app used up most of the redis connection pool, so i needed to adjust the concurrency so that both rails ans sidekiq runs fine. I also faces the concurrency issue with heroku with a free redis tier.

sony-mathew: Can we increase the connection pool on the Redis used in docker-compose? In Heroku free tier for Redis, the connection limit is 20 as mentioned here https://elements.heroku.com/addons/heroku-redis.

PR #447, opened 2020-01-25
Feature: API to get most used labels

subintp:
- [x] Added an API to get the most used labels
- [x] Added a filter for labels

### DIFF
--- a/app/views/api/v1/labels/most_used.json.jbuilder
+++ b/app/views/api/v1/labels/most_used.json.jbuilder
@@ -1,0 +1,8 @@
+json.data do
+  json.meta do
+  end
+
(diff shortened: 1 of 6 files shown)


subintp:
@sojan-official @pranavrajs  This PR is blocked due to https://github.com/mbleigh/acts-as-taggable-on/issues/860

app/views/api/v1/labels/most_used.json.jbuilder
```diff
@@ -0,0 +1,8 @@
+json.data do
```
pranavrajs: Please don't use this. On frontend this would become `response.data.data.payload`

Issue #37, opened 2019-09-29
Add specs for rails controllers


pranavrajs:
- [ ] Add rspec for rails controllers 

find the list of rails controllers without test coverage here 

https://codeclimate.com/github/chatwoot/chatwoot/code?q=controllers
Feel free to pick any and raise a PR for the same.

austinratcliff:
Hey @pranavrajs! 👋 I'd love to pick up this issue for Hacktoberfest. Is that ok even though it's not labeled with `Hacktoberfest`?

sojan-official:
@austinratcliff  Please feel free to pick this up, if you are still available.

Jsscrdng:
Can i get this one?

sojan-official:
@Jsscrdng  go ahead :)

sojan-official:
@venarius if you are still working on this. Could you also take a look at the controllers without any specs? i think we can increase our test coverage significantly then. 

https://codeclimate.com/github/chatwoot/chatwoot/code?q=controllers  should give a list.

venarius:
@sojan-official already doing it. I just started with the controllers that had already some tests first :)

sojan-official:
@venarius  our test coverage is at 60% ( https://codeclimate.com/github/chatwoot/chatwoot ) . thanks a ton for the great work.

sojan-official:
Closing this since most of the controllers are covered now . We will create specific issues for the remaining ones.

PR #483, opened 2020-02-08
Chore: Added tests for subscriptions_controller

venarius:
# Pull Request Template

## Description

This adds tests for the subscriptions_controller

Issue #37 

I wasn't able to tests the "Status" function of this controller. Is it still used? @sojan-official  @pranavrajs thoughts?
Also I needed to change the Inheritance of this controller because it didn't really check for user authentication before.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -67,7 +67,7 @@ class ApplicationController < ActionController::Base
   def check_subscription
     # This block is left over from the initial version of chatwoot
     # We might reuse this later in the hosted version of chatwoot.
-    return unless ENV['BILLING_ENABLED']
+    return if !ENV['BILLING_ENABLED'] || !current_user
 
     if current_subscription.trial? && current_subscription.expiry < Date.current
       render json: { error: 'Trial Expired' }, status: :trial_expired
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -94,17 +94,15 @@ Rails.application.routes.draw do
       end
 
       # this block is only required if subscription via chargebee is enabled
-      if ENV['BILLING_ENABLED']
-        resources :subscriptions, only: [:index] do
-          collection do
-            get :summary
-          end
+      resources :subscriptions, only: [:index] do
+        collection do
+          get :summary
         end
+      end
 
-        resources :webhooks, only: [] do
-          collection do
-            post :chargebee
-          end
+      resources :webhooks, only: [] do
+        collection do
+          post :chargebee
         end
       end
     end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,6 +1,9 @@
 require 'simplecov'
 SimpleCov.start 'rails'
 
+# Required for subscriptions_controller tests
+ENV['BILLING_ENABLED'] = 'true'
+
 RSpec.configure do |config|
   config.expect_with :rspec do |expectations|
     expectations.include_chain_clauses_in_custom_matcher_descriptions = true
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,9 +1,6 @@
 require 'simplecov'
 SimpleCov.start 'rails'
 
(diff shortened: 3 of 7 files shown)


sojan-official:
@venarius  this is a stale controller from the initial version of chatwoot when it was closed source.  We might end up using this controller for the hosted version, but I guess we will need to retest this whole flow and might also need to explore options like stripe.

config/routes.rb
```diff
@@ -88,7 +88,7 @@
       end
 
       # this block is only required if subscription via chargebee is enabled
-      if ENV['BILLING_ENABLED']
+      if ENV['BILLING_ENABLED'] || Rails.env.development? || Rails.env.test?
```
sojan-official: I would leave this check intact, so even while developing the behaviour is consistent. 
Let's try setting the environment variable inside the test  using something like the `climate_control` gem 
https://thoughtbot.com/blog/testing-and-environment-variables

venarius: The climate_control gem didn't work here because the routes only get loaded once on start of the whole rspec test. I fixed the issue by setting the env variable now in spec_helper.rb

spec/spec_helper.rb
```diff
@@ -1,6 +1,9 @@
 require 'simplecov'
 SimpleCov.start 'rails'
 
+# Required for subscriptions_controller tests
```
sojan-official: This variable will be set to false in almost all the chatwoot installations, except for the possible one we will set up for the hosted version. Having the variable set here will affect the context for the entire codebase. Let's keep the assumption that the software should be tested on the context of this variable being set to 'false' . 

We only need to enable this variable in the spec/controllers/api/v1/subscriptions_controller_spec.rb.  So either lets assign the env variable at the beginning of the test. ( ensure its reset at the end of the test ) . or stub it for the specific test cases 

```
allow(ENV).to receive(:[]).with("BILLING_ENABLED").and_return(true)
```
ref: https://thoughtbot.com/blog/testing-and-environment-variables  ( for both examples )

venarius: Hi, yes, I fixed the code climate issue. Sorry, I didn't notice it was on error.

About the ENV issue:
Like I commented in your first review, the climate_control gem and mocking the ENV doesn't work here because we need to set `"BILLING_ENABLED"` to true in the initialization of the test suite otherwise the subscription routes won't get loaded. Please have a look at `routes.rb:97`. The routes don't get initialized before every test.

venarius: Also, could you retry the CI build? I don't think the error has something to do with my code changes :)

pranavrajs: @venarius It's something to do with the containers, we have fixed it on branches, I will rebase this once it is merged to develop

sojan-official: @venarius  🤦‍♂️ apologies for missing out on the environment variable check in the routes file.  After looking at it again, I feel like having it in the `routes.rb`  to be a bad call.  let's remove that condition from the routes.  

and have a `before_action` in both `subscription_controller` and `webhook_controller` which throws 404 unless the variable is set.  Let me know if you want any help with the change ? 😃

venarius: @sojan-official yes I will do that. No worries 😀

app/controllers/application_controller.rb
```diff
@@ -67,7 +67,7 @@ def render_error_response(exception)
   def check_subscription
     # This block is left over from the initial version of chatwoot
     # We might reuse this later in the hosted version of chatwoot.
-    return unless ENV['BILLING_ENABLED']
+    return if !ENV['BILLING_ENABLED'] || !current_user
```
sojan-official: let's get rid of the `!ENV['BILLING_ENABLED'] `check now that it's handled in the before action

venarius:
@sojan-official CircleCI seems to fail on a test, but on my local setup all tests run through without error. Got any idea why?

venarius:
@sojan-official :man_facepalming: I see it. How could I miss that line... Thanks a lot for the help! :+1:

PR #508, opened 2020-02-17
Chore: Added tests for conversations_controller

venarius:
# Pull Request Template

## Description

This adds tests for conversations_controller.rb

I also had to add an association from messages to contacts because messages have a contact_id but no association and that threw an `undefined method contact for message` error in `conversations/get_messages.json.jbuilder:13`

Issue #37


spec/controllers/api/v1/conversations_controller_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+             params: params,
+             as: :json
+
+        expect(response).to have_http_status(:success)
+        expect(conversation.reload.agent_last_seen_at).to eq(DateTime.strptime(params[:agent_last_seen_at].to_s, '%s'))
+      end
+    end
+  end
+
+  describe 'GET /api/v1/conversations/:id/get_messages' do
```
pranavrajs: Thanks for bringing this API to notice, we are not using this API, could you please remove this endpoint.

pranavrajs:
@venarius Thanks for the PR. This PR has been really helpful, can you please take a look at the comment.

spec/controllers/api/v1/conversations_controller_spec.rb
venarius: @pranavrajs done :)

PR #512, opened 2020-02-17
Chore: Removed unused functions from home_controller

venarius:
Wanted to write tests for the home controller but found out that these functions aren't used anymore.


pranavrajs:
@venarius Nice work, thanks. Could you please remove the view file as well. `home/index.html.erb`?

venarius:
@pranavrajs yes, sorry, I forgot about the views

PR #523, opened 2020-02-19
Fix: Changed overflow scroll to overflow auto

### LINKED ISSUE #514: Overflow set to scroll on various places on dashboard
**Describe the bug**
When you hover certain areas in the dashboard UI with your mouse, a scrollbar comes up even though the content is not overflowing. That is caused on various places because of the `overflow: scroll` on :hover. A better option would be `overflow: auto` so that it only scrolls when the content is actually overflowing.

If that's an actual bug and not intended I will happily fix this issue 😀 

**To Reproduce**
1. Hover with mouse on e.g. the sidebar in the dashboard UI.
2. You will see a scrollbar popup

**Expected behavior**
The scrollbar should only popup if the content is actually overflowing. Otherwise it just looks bad.

**Screenshots**
https://i.gyazo.com/6dc6f39817ca0d923b2b6bdec3a8ffac.mp4

 - OS: Windows
 - Browser: Chrome
 - Version: 80.0.3987.106

venarius:
This fixes the overflow: scroll behaviour which added scrollbars on content even though it was not yet overflowing.

I fixed this by changing the hard "scrolls" to "auto".

Before:
Screenshot 2020-02-19 at 14 58 06

After:
Screenshot 2020-02-19 at 14 58 43

Fixes #514 

This also fixes the "plopping-in" scrollbars on some locations like the sidebar. ( See the example on the linked issue )


venarius:
@pranavrajs I work with windows ;-)

PR #531, opened 2020-02-20
Chore: Added tests for callbacks_controller

venarius:
This adds tests for the callbacks_controller

I had to do some required changes to the original controller. Please check if I did something wrong here.

For #37


app/controllers/api/v1/callbacks_controller.rb
```diff
@@ -1,8 +1,7 @@
 require 'rest-client'
 require 'telegram/bot'
-class Api::V1::CallbacksController < ApplicationController
-  skip_before_action :verify_authenticity_token, only: [:register_facebook_page]
-  skip_before_action :authenticate_user!, only: [:register_facebook_page], raise: false
+class Api::V1::CallbacksController < Api::BaseController
```
pranavrajs: @sojan-official Please verify whether the FB callbacks are coming to this controller. If yes, this shouldn't be inheriting the base controller.

venarius:
It was required for me to change it to api base controller because register_facebook_page accesses current_account

app/controllers/api/v1/callbacks_controller.rb
venarius: It was required for me to change it to api base controller because register_facebook_page accesses current_account

sojan-official: @pranavrajs  this controller is the one used to register the Facebook page selected in the inbox creation step. and requires an authenticated user. so this change is right.

PR #730, opened 2020-04-16
Chore: Message to support multiple attachments

### LINKED ISSUE #727: Chore: Message should be able to support multiple attachments
**Problem Definition**
As it stands today, a message supports only one attachment. And in many places, we assume that a message can have either some content or is just an attachment. The message model is bonded to the attachment model by a `has_one` relationship. Because of this implementation, when we add channels and have incoming channels supporting a message with multiple attachments, we might lose the context in the sense that the content and the attachments were part of one single message rather than a couple of unconnected messages.

**Solution**
To rectify this problem we might have to change the relationship to `has_many` so that the message model can now support multiple attachments. This will be greatly beneficial when we implement the email channels and build integrations to third-party apps like slack.

sony-mathew:
This fixes #727 

## Description
* Changed the message association to attachments from has_one to has_many
* Changed all the references of this association in building and fetching to reflect this change
* Added number of attachments validation to the Message model
* Modified the backend responses and endpoints to reflect multiple attachment support
* Changed the request from frontend structure to reflect the multiple attachment structures
* Changed the message bubbles to support multiple attachments
* Bugfix: agent side attachment was not showing because of a missing await and issue in adding new messages to store (an issue in filtering out existing messages in the store)
* Added more clarity in the documentation for ImageMagick

## Type of change

## How Has This Been Tested?

With excitement dying down and freedom feels an eternity away in a COVID categorized green zone district.

Tested in local the following : 
* Tested out the end-user chat flow and agent dashboard chat flow. 
* Ran the test suites.

## Checklist:


pranavrajs:
Screenshot 2020-04-17 at 10 31 22 AM

On the conversation list, it is not showing the type of message

PR #757, opened 2020-04-23
Feature: Move to the next conversation when I resolve a the current c…

### LINKED ISSUE #612: Enhancement: Move to the next conversation when i resolve a the current conversation
When I resolve a conversation I would like the system to redirect me to the next conversation.

When I resolve a conversation I would like the system to redirect me to the next conversation.

none

**Additional context**
none

wwwbruno:
## Description

My approach is:
- On the `toggleStatus` actions in conversations, get from `getNextChatConversation` getter the next conversation before calling the API:
  - if there is a nextChat, dispatch `setActiveChat` action;
  - if there is no nextChat, dispatch `clearSelectedState` action;
Fixes #612 

## Type of change

## How Has This Been Tested?

- Open the dashboard with more than one open conversation (f.e. http://localhost:3000/app/accounts/1/dashboard);
- Mark a conversation as resolved:
  - The next conversation in the list opens:
  - When the conversation is the last:
    - Go to first, or:
- When the resolved conversation is the last, display the clear state without any chat selected;

## General questions
When working on this feature I had some product-related question:
- Does the next conversation mean older or newer conversations?
- When I resolve the last conversation, does it make sense to open the first conversation as I don't have a next?


app/javascript/dashboard/components/ChatList.vue
```diff
@@ -119,6 +121,11 @@ export default {
     this.$store.dispatch('setChatFilter', this.activeStatus);
     this.resetAndFetchData();
     this.$store.dispatch('agents/get');
+
+    bus.$on('openNextChat', (chatId) => {
+      const element = this.$refs['conversation-card-' + chatId][0]
```
nithindavid, line 126: Hey can we use `es6` string literals here ?
```
`conversation-card-${chatId}`
```

Also how about we split this before accessing 0th index? Something like this ( feel free to change the logic )

```
const chatRef = this.$refs[`conversation-card-${ chatId}`];
const [element = {}]= `chatRef;`
const notClickable = !element. cardClick;

if (notClickable) return;
```

app/javascript/dashboard/components/buttons/ResolveButton.vue
```diff
@@ -56,6 +56,10 @@ export default {
       this.$store.dispatch('toggleStatus', this.currentChat.id).then(() => {
         bus.$emit('newToastMessage', this.$t('CONVERSATION.CHANGE_STATUS'));
         this.isLoading = false;
+
+        if (this.currentChat.nextId && wootConstants.STATUS_TYPE.RESOLVED == this.currentChat.status) {
```
nithindavid, line 60: Move this to to a variable for better readability ?
`const canOpenNextChat =  this.currentChat.nextId && wootConstants.STATUS_TYPE.RESOLVED == this.currentChat.status`

app/javascript/dashboard/components/ChatList.vue
```diff
@@ -24,6 +24,7 @@
         v-for="chat in getChatsForTab()"
         :key="chat.id"
         :chat="chat"
+        :ref="'conversation-card-' + chat.id"
```
nithindavid, line 27: Can we use es6 string literals here ?

app/javascript/dashboard/components/ChatList.vue
```diff
@@ -161,7 +168,14 @@ export default {
         (a, b) =>
           this.lastMessage(b).created_at - this.lastMessage(a).created_at
       );
-      return sorted;
+      return sorted.map((chat, index) => {
```
nithindavid, line 171: Move this to a variable like `sortedWithNextChat` or something ?

sojan-official:
@wwwbruno thanks a lot for working on this 🎉

> Does the next conversation mean older or newer conversations?

the older conversation. ie the conversation that appears to the bottom of the currently selected chat in the list.

> When I resolve the last conversation, does it make sense to open the first conversation as I don't have a next?

yes, this makes sense. good call.

app/javascript/dashboard/components/ChatList.vue
wwwbruno: 06a71d3

app/javascript/dashboard/components/ChatList.vue
wwwbruno: 62d4695 + https://github.com/chatwoot/chatwoot/pull/757/commits/ce29edebad36fb559e8aca26f435095738444e50

app/javascript/dashboard/components/ChatList.vue
wwwbruno: 62d4695

app/javascript/dashboard/components/buttons/ResolveButton.vue
wwwbruno: https://github.com/chatwoot/chatwoot/pull/757/commits/02ba6d1375e47db167ad3df15d7900d381358947

app/javascript/dashboard/components/ChatList.vue
wwwbruno: @nithindavid thanks for the review. 😄 

I did all your suggestions, could you please take another look?

wwwbruno:
@sojan-official Thanks for the answers! I did a simple logic to get the first as next when it's the last conversation: https://github.com/chatwoot/chatwoot/pull/757/commits/6115eb1a8bfca1355197d697d5920e970b524ab8

With this logic, if there is only one conversation, the next will be itself. But as it's removed from the list when resolved and we check if the element exists and it's clickable, I think there is no problem with that.

An alternative to check it could be:

```diff
--- a/app/javascript/dashboard/components/ChatList.vue
+++ b/app/javascript/dashboard/components/ChatList.vue
@@ -173,6 +173,8 @@ export default {
         (a, b) =>
           this.lastMessage(b).created_at - this.lastMessage(a).created_at
       );
+      if (sorted.length <= 1) return sorted;
+
       const sortedWithNextChat = sorted.map((chat, index) => {
         const nextChat = sorted[index + 1];
         chat.nextId = nextChat ? nextChat.id : sorted[0].id;
```

What do you think?

nithindavid:
Hey @wwwbruno, 
I was testing your PR in local, Thanks a lot for putting this up but it seems like we need to do this in a better approach by manipulating the `state` rather than relying on `refs`.
1) We have a state variable `selectedChat` in `app/javascript/dashboard/store/modules/conversations/index.js`  which is used to track the actively selected chat. Also we have an action  `setActiveChat` to set it. So ideally I think we should manipulate this to set the next chat.
2) To identify the next chat currently we are modifying the list in `app/javascript/dashboard/components/ChatList.vue` Which isn't the best place to do as we have all the state variables and filters to find the next chat. So I think we should write a new getter that would compute the next chat and use this getter data to set the `selectedChat`.


Let us know if you're ready to take this up. I'll be happy tp help you with any difficulties 🙂. I really appreciate the work you've done so far, Also I'm really sorry for not bringing this up initially.

sojan-official:
@wwwbruno like @nithindavid suggested let's try to implement this leveraging the data we have available in the state.  With that, we can simplify the logic and set the selectedChat to nil after all the conversations in the current filter has been archived.

wwwbruno:
Thanks a lot for the proposal, that sounds a better approach to me 🎉 I'll put some work on it tomorrow and let you know if I have any issues with it!

About your second point, I already have some questions:
If I understood correctly, you suggest calculating and setting the next chat at the moment that we select a chat, so only have the next chat for the currently selected chat, does it make sense?

sojan-official:
@wwwbruno as an answer to the question, you wouldn't have to worry about the next chat at the moment we click on a chat (ie pre-calculate the value for next chat ). You have the full conversation information in your state. 

When the person resolves a conversation. the conversation_resolve action would get triggered. As an aftereffect of this action, you can call the getter to find the next conversation from the state and then change `selectedChat`. 

@nithindavid correct me if I missed something.

app/javascript/dashboard/store/modules/conversations/getters.js
```diff
@@ -50,6 +50,13 @@ const getters = {
   getChatStatusFilter: ({ chatStatusFilter }) => chatStatusFilter,
   getSelectedInbox: ({ currentInbox }) => currentInbox,
   getConvTabStats: ({ convTabStats }) => convTabStats,
+  getNextChatConversation: ({ allConversations, selectedChat }) => {
+    if (allConversations.length <= 1) { return }
```
wwwbruno, line 54: @nithindavid I did some changes and I believe it's more likely what you suggested. I just have one issue

With the current approach, I'm having a difficult that I need help to understand.
- Inside this getter, the `allConversations` state is not being updated and that causes two issues:
  - Resolved conversations are still present on the array, so depending on the order that I solve, it can set a resolved conversation as active.
  - The order is not sorted the same as in the view. I actually think that the order set on `getChatsForTab()` is not persisted in the state, so only available in the view. Maybe we should move the order logic before setting on the state or duplicating it inside the getter?

Thanks again for the help!

nithindavid: Let me take a look at this 👍

nithindavid: Hey I was able to think of  a working solution with some changes to your PR,
1) We need the old `openNextChat` event which has a `nextChat` as params  in `app/javascript/dashboard/components/ChatList.vue` which should dispatch a `setActiveChat` event.
2) The getter for next chat should look something like
```
getNextChatConversation: _state => {
    const chatList = getters.getAllStatusChats(_state);
    const { selectedChat } = _state;

    const currentIndex = chatList.findIndex(
      conversation => conversation.id === selectedChat.id
    );
    const nextIndex = (currentIndex + 1) % chatList.length;
    return chatList[nextIndex];
  },
```.
3) We need to do a little refactoring by moving the `sorting logic around line 160` in `app/javascript/dashboard/components/ChatList.vue` to `getAllConversations` in `getters.js`
4) Import the `getNextChatConversation` getter in `app/javascript/dashboard/components/buttons/ResolveButton.vue` and call the next chat before  firing the dispatch and emit the `openNextChat` with `nextchat` event when the status resolve is success.

Comment (nithindavid) [reply app/javascript/dashboard/store/modules/conversations/getters.js]: @pranavrajs What do you think ?

Comment (wwwbruno) [reply app/javascript/dashboard/store/modules/conversations/getters.js]: @nithindavid I did some work on what you suggested. Basically just sorting inside the `getters` and getting the status-filtered chats seems to do the work without the need to emit an event.

Could you please take a look when you have some time? https://github.com/chatwoot/chatwoot/pull/757/commits/2ad72c5e1fb2c7992e4fbb261d3cf692e2e062e6

File: app/javascript/dashboard/components/widgets/conversation/ConversationCard.vue [hunk 6]
```diff
@@ -2,7 +2,7 @@
   <div
     class="conversation"
     :class="{ active: isActiveChat, 'unread-chat': hasUnread }"
-    @click="cardClick(chat)"
```
nithindavid, line 5: Hey @wwwbruno since we are no longer using the refs, we can omit changes in this file.

nithindavid:
@wwwbruno Thanks a lot :) The PR looks all good to go 🔥 Please update the branch to latest and if you can write tests that would be good.

app/javascript/dashboard/components/widgets/conversation/ConversationCard.vue
wwwbruno: https://github.com/chatwoot/chatwoot/pull/757/commits/db5c3e4916baa0f680bfcf0050902adbc773fffb

wwwbruno:
@nithindavid I wrote some specs for the new getters I created. Do you think is there some other point that is important to cover?

nithindavid:
@wwwbruno This looks good, I'll merge it once the local testing is done  🔥 Thanks a lot for putting your effort, also the dedication and commitment  you put in towards this PR is out of the world 😊
Also expecting more commits from you in future 😎

wwwbruno:
@nithindavid thanks for all the help on this Pull Request, I learned a lot with all the suggestions! I'll get another issue to help in the future, it will be a pleasure!

Issue #688, opened 2020-04-08
Feature: Add French language


sojan-official:
Feature: Add French language

jtraulle:
I can work on this :slightly_smiling_face:

sojan-official:
@jtraulle  thanks a lot for offering your help. we use crowdin to manage the translations now and believe we have received some help over there already. could you double-check the translations over there before we can bring it into the codebase ? 

https://translate.chatwoot.com/chatwoot/fr

jtraulle:
@sojan-official Thanks, I registered on crowdin and I am currently reviewing current translations :wink:

jtraulle:
Is it possible to slightly increase the width of the sidebar from 23rem to 26rem ? "Canned response" is translated into "Réponses pré-enregistrées" or, shorter by "Réponses standardisées". But 23rem is not enough. As well as "Account Settings" -> "Paramètres du compte" and "Inboxes" -> "Boîtes de réception"

|                  | 23rem | 26rem |
|------------------|---------|--------|
| **Main** sidebar | ![main-sidebar-23rem](https://user-images.githubusercontent.com/613615/79697564-c83d2b00-8283-11ea-87ee-ae798fbaf058.png) | ![main-sidebar-26rem](https://user-images.githubusercontent.com/613615/79697570-cd9a7580-8283-11ea-8fd1-5f1cf2206604.png) |
| **Settings** sidebar | ![settings-sidebar-23rem](https://user-images.githubusercontent.com/613615/79697577-ddb25500-8283-11ea-993e-6b41c8640d46.png) | ![settings-sidebar-26rem](https://user-images.githubusercontent.com/613615/79697583-e440cc80-8283-11ea-8673-b1b33d569e90.png) |

jtraulle:
I just finished the review :tada:

PR #768, opened 2020-04-26
Slightly increase sidebar width

jtraulle:
## Description

This PR, propose to slightly increase the width of the sidebar to allow translated strings in French to be in one line only.

See https://github.com/chatwoot/chatwoot/issues/688#issuecomment-616211237 for more info and visual screenshots of change.


pranavrajs:
@jtraulle Sorry that we missed the comment earlier. I would like to approach this problem differently. I don't think it is a scalable solution to the problem we have, let us say there is a long inbox name which does not fit to the width. Increasing width would not solve that problem. 

I would prefer to have the text truncated like the one given below and show the full text as a popup when we hover on the item.

Screenshot 2020-04-27 13 21 20

Let me know what you think.

jtraulle:
@pranavrajs Thanks for your reply :slightly_smiling_face:  While I understand that this should be scalable for as many languages as possible, I tend to find that it is more readable when we can read the complete string.

So maybe a combination of both solution (increasing size of the sidebar : this would fit 80% of languages I suppose and a CSS ellipsis for the rare translations that would not fit).

pranavrajs:
@jtraulle Sounds good, could you make that change to truncate the text as well?

Please test this in tablet view as well.

jtraulle:
OK, I'll do that. Was stuck yesterday trying to install Yarn dependencies but with the ongoing lock-down situation worldwide, network congestion is more frequent. Finally got it installed and successfully build the docker image.  I'll update the PR ASAP :slightly_smiling_face:

pranavrajs:
> @pranavrajs OK, I'll do that. Was stuck yesterday trying to install Yarn dependencies but with the ongoing lock-down situation worldwide, network congestion is more frequent. Finally got it installed and successfully build the docker image. I'll update the PR ASAP 🙂

Sounds good. Yeah, the network congestion is real.  Hope you are staying safe.

jtraulle:
@pranavrajs I have added the ellipsis. I search for the tooltips definition and found them in /app/javascript/dashboard/i18n/default-sidebar.js but noticed that the **toolTip** keys does not get translated. For example, tooltip _Conversation from all subscribed inboxes_ (for the **Conversations** menu item) does not get translated. Is it possible to change that and allow the **toolTip** to be translated as well ? I search what to modify but without luck ... Thanks for your help :slightly\_smiling\_face:

pranavrajs:
@jtraulle I'll test this out tomorrow. 

> I search for the tooltips definition and found them in /app/javascript/dashboard/i18n/default-sidebar.js but noticed that the toolTip keys does not get translated. For example, tooltip Conversation from all subscribed inboxes (for the Conversations menu item) does not get translated. 

This tooltip is only used in Conversations. I think we can get rid of that attribute all together, it doesn't add much value now.

pranavrajs:
@jtraulle Nice work!

I found a couple of minor issues.

Screenshot 2020-06-09 12 19 05

1. Letters appear to be cut due to some reason, please see `g` in Agentes
2. I would like to have a tooltip on the last one. Right now, the user cannot see the entire text

caseyjhol:
Setting

```css
.wrap {
    padding: $space-small 0;
    margin: -#{$space-small} 0;
}
```
should fix the issue with the bottoms of letters being clipped.

sojan-official:
In our current design iteration, we are trying to expand the conversation workspace area and to that, we are evaluating shrinking the sidebar even further 😅 . I guess the changes here will become irrelevant then.

PR #899, opened 2020-05-29
fix bug within #toggleAgentTypingStatus test

RonWalker22:
The test was providing a false positive as the state of `isAgentTyping` was not changed

I just changed the initial state of isAgentTyping from `false` to `true`

## Type of change
expected)


pranavrajs:
@RonWalker22 Thanks for the PR. 👍

PR #1036, opened 2020-07-14
fixed db:migrate RoundRobin dup

tianlu1677:
Because RoundRobin had defined in file, so when db:create and db:migrate will raise error.
so change the migration file name


sojan-official:
@tianlu1677  thanks a lot for picking thing

PR #1084, opened 2020-07-23
fix: Add a check for 24 hour window before sending a message

### LINKED ISSUE #1009: WhatsApp Improvements due to timeout (templates?)
No

WhatApp (via Twilio) has a limit on that you can only send another outbound message if the last inbound message was within 24 hours. It would be helpful to have a timer or CW simply not allowing another message which you can't tell the user didn't get.

Adding support to send a template from a drop down where it could initiate a simple message like: "We want to give you an update, please reply so we can do so" etc.

Then the conversation can continue.

Additionally, adding "Status Callback URL" support would actually help see if messages have been read/delivered etc from such data provided by Twilio (https://www.twilio.com/docs/whatsapp/api#monitor-the-status-of-your-whatsapp-outbound-message). Currently there is no way to tell this at all.

pranavrajs:
Fixes #1009 

- Add a check on the frontend to check 24 hour window
Screenshot 2020-07-23 16 27 13

- Introduce CSS variables for colors
Screenshot 2020-07-23 at 4 28 31 PM


app/javascript/dashboard/components/widgets/conversation/MessagesView.vue
```diff
@@ -5,6 +5,18 @@
       :is-contact-panel-open="isContactPanelOpen"
       @contactPanelToggle="onToggleContactPanel"
     />
+    <div v-if="!currentChat.can_reply" class="messenger-policy--banner">
```
sojan-official, line 8: should be translatable ?

pranavrajs: Done.

PR #1095, opened 2020-07-26
feat: Customisable Email Templates

### LINKED ISSUE #1008: Outbound Emails - Customise Sender and Content
No

Currently CW sends emails based on a hard-set from address of the install as well as a hard-set name in a template. When using multiple inboxes, or even accounts, this causes confusion as to who the mail is from.

In reality, the From: name should be able to be set against the Inbox, as well as the From: address. 

The subject of the message as well as the message template including URL and domain should be easily customisable again on a per inbox basis so the content is exactly matching the site the person has visited. 

For now, we're going to have to turn off the emails because it seems to follow the one-hard set template/from data for all and in the world where people are extra vigilant about emails purporting to be from a specific sender, this would be a red flag for some.

sojan-official:
# Pull Request Template

## Description

* Ability to set custom email template layouts
* Ability to set custom email template content
* Ability to set layouts and contents at installation/account levels
* Add liquid support to Chatwoot
* Ability to use liquid in email templates

Fixes #1008 

## Type of change

## How Has This Been Tested?

## Checklist:


app/mailers/application_mailer.rb
```diff
@@ -1,9 +1,15 @@
 class ApplicationMailer < ActionMailer::Base
+  include ActionView::Helpers::SanitizeHelper
+
   default from: ENV.fetch('MAILER_SENDER_EMAIL', 'accounts@chatwoot.com')
-  layout 'mailer'
+  before_action { Current.account = params[:account] }
```
sony-mathew, line 5: What if mail is triggered by a background job (Sidekiq Worker)? It may not have the current account then. Or else we might want to set that somewhere to use this.

app/mailers/agent_notifications/conversation_notifications_mailer.rb
```diff
@@ -1,21 +1,27 @@
...
+    send_mail_with_liquid(to: @agent.email, subject: subject) and return
   end
 
   def conversation_assignment(conversation, agent)
     return unless smtp_config_set_or_development?
 
     @agent = agent
     @conversation = conversation
-    mail(to: @agent.email, subject: "#{@agent.name}, A new conversation [ID - #{@conversation.display_id}] has been assigned to you.")
+    subject = "#{@agent.name}, A new conversation [ID - #{@conversation.display_id}] has been assigned to you."
```
sony-mathew, line 16: Can/Shall we fetch this from a translation file? Does it make sense?

app/services/email_templates/db_resolver_service.rb
```diff
@@ -0,0 +1,69 @@
...
+  def find_installation_template
+    @@model.find_by(name: @template_name, template_type: @template_type, account: nil)
+  end
+
+  # Build path with eventual prefix
+  def build_path(prefix)
+    prefix.present? ? "#{prefix}/#{@template_name}" : @template_name
+  end
+
+  # returns a path depending if its a partial or template
```
sony-mathew, line 59: A more detailed comment here with an example case would be really helpful. It is very difficult to understand what is expected and what is been thrown out of this function here.

app/views/mailers/agent_notifications/conversation_notifications_mailer/conversation_creation.html.erb
```diff
@@ -1,4 +1,5 @@
-<p>Hi <%= @agent.name %>,</p>
+<p>Hi {{agent.name}}</p>
```
sony-mathew, line 1: Do all these files always rendered using Liquid templating engine and is yes would the extension still be `.html.erb` or something like `.liquid.erb`. make more sense?

app/views/mailers/agent_notifications/conversation_notifications_mailer/conversation_assignment.html.erb
```diff
@@ -1,4 +1,4 @@
-<p>Hi <%= @agent.name %>,</p>
+<p>Hi {{agent.name}},</p>
```
sony-mathew, line 1: Do all these files always rendered using Liquid templating engine and is yes would the extension still be `.html.erb` or something like `.liquid.erb`. make more sense?

app/mailers/application_mailer.rb
```diff
@@ -14,4 +20,36 @@ def global_config
   def smtp_config_set_or_development?
     ENV.fetch('SMTP_ADDRESS', nil).present? || Rails.env.development?
   end
+
+  def ensure_current_account(account)
```
sony-mathew, line 24: Are we using this function anywhere? I could not find a reference to this? Sorry if I missed it out somewhere.

app/mailers/agent_notifications/conversation_notifications_mailer.rb
pranavrajs: translation or a custom db value makes sense.

app/mailers/application_mailer.rb
sojan-official: the idea is to use https://api.rubyonrails.org/classes/ActionMailer/Parameterized.html . So we will be passing in the account at the mailer trigger point.

app/mailers/agent_notifications/conversation_notifications_mailer.rb
sojan-official: Will take this up a part of making emails translatable ticket.

app/views/mailers/agent_notifications/conversation_notifications_mailer/conversation_creation.html.erb
sojan-official: added a view handler for liquid.

app/mailers/application_mailer.rb
sojan-official: it was leftover initially but have repurposed the function.

app/mailers/application_mailer.rb
```diff
@@ -1,17 +1,48 @@
...
+  def liquid_locals
+    # expose variables you want to be exposed in liquid
+    {
+      global_config: GlobalConfig.get('INSTALLATION_NAME', 'BRAND_URL'),
+      action_url: @action_url
+    }
+  end
+
+  def ensure_current_account(account)
+    Current.account = account if params.present?
```
sony-mathew, line 46: Do we have a global `account` available here?

sojan-official: I guess we don't need an if check here. account param gets passed down from the before filter

PR #1163, opened 2020-08-23
chore: Improve developer experience

pranavrajs:
> There are only two hard things in Computer Science: cache invalidation and naming things. -- Phil Karlton

This PR is a migration from the existing Foundation 6.x to TailwindCSS.

### Reasons behind this PR

- Foundation is a great framework, but sometimes it is pretty hard to upgrade even the minor versions.
- Lack of utility classes forces us to write custom classes and it is incredibly painful to name them.
- Lack of standardization in the naming pattern we have across the codebase. Even though we have BEM in some places, there is no standard guideline on how to the classes. 
- Foundation class overrides make it painful to build an SFC. For example, we have a component defined for each message type in the widget. To use the same component in the dashboard, we have to update the style to match dashboard styles which makes the reusable component useless.
- The customization of colors is somewhat irregular as some of the colors use methods like `lighten(10%)` or `darken(20%)`. This is generating more number of colors than what is defined in the palette. 
- Usage of CSS variables is painful. 
- Non-standard use of base font-size, we have base font size set at 10px while the browser's default is 16px.


pranavrajs:
Closing in favour of #1206

PR #1192, opened 2020-09-03
chore: Add message_echo listener for Facebook

### LINKED ISSUE #1186: Echo messages from Facebook are not saved to Chatwoot
If a person replies from the Facebook messenger, that message is not displayed in Chatwoot

pranavrajs:
Fixes #1186


app/builders/messages/facebook/message_builder.rb
```diff
@@ -120,7 +121,7 @@ def message_params
       message_type: @message_type,
       content: response.content,
       source_id: response.identifier,
-      sender: contact
+      sender: @outgoing_echo ? nil : contact
```
sojan-official, line 124: should we need the instance variable ? or can we just use the `outgoing_echo` function?

app/models/inbox.rb
```diff
@@ -66,6 +66,10 @@ def web_widget?
     channel.class.name.to_s == 'Channel::WebWidget'
   end
 
+  def inbox_type
+    channel.class.name.to_s.downcase.remove('channel::', 'page', 'profile', 'sms').strip
```
pranavrajs, line 70: I would suggest to add a method `name` on Channel to resolve the name instead of this. This is just to ensure that the Twilio Whatsapp channel returns `Whatsapp` and not `Twilio`

```rb
def name
  medium === :whatsapp ? 'Whatsapp' : 'Twilio'
end
```

PR #1232, opened 2020-09-13
chore: Remove `emojione`, cleanup styles

pranavrajs:
- Remove emojione dependency  
- Use Emojis from unicode standard 13.0
- Use px instead of rem in shared emoji component.


nithindavid (approved):
🔥

PR #1324, opened 2020-10-08
Upgrade actionpack to >= 6.0.3.4

pranavrajs:
Upgrade bundle


sony-mathew:
This is already done as part of #1329

PR #1335, opened 2020-10-11
feat: Add Contacts page

### LINKED ISSUE #44: Add Customer view
- [ ] Create/Read/Update customer profiles

nithindavid:
**Screenshot**

image

Fixes #44


app/javascript/dashboard/routes/dashboard/dashboard.routes.js
```diff
@@ -1,14 +1,19 @@
 import AppContainer from './Dashboard';
 import settings from './settings/settings.routes';
 import conversation from './conversation/conversation.routes';
+import { routes as conversationRoutes } from './contacts/routes';
```
sojan-official, line 4: contactRoutes ?

pranavrajs:
1. Contacts are not listed when page is loaded.
2. Search is not working properly
3. Pagination button is not working properly.

Screenshot 2020-10-23 at 3 04 45 PM

4. Page size is wrong.

For page 1, it shows 1-8. For page 2, it shows 21-28

Screenshot 2020-10-23 at 3 06 09 PM

5. Backend page size is 25, frontend is using 20.

6. Current query in contacts_controller uses an AND condition (email and phone present). It should have been 
```
Current.account.contacts.where.not(email: [nil, '']).or(Current.account.contacts.where.not(phone_number: [nil, '']))
```
I guess we can add an additional attribute something like `identified`. If the user has email, phone or twitter user or another identifier we can mark it as `identified`. This way query would be simpler.

docker-compose.yaml
```diff
@@ -40,7 +40,7 @@ services:
       - NODE_ENV=development
```
pranavrajs, line 40: Please revert the changes from this file.

app/controllers/api/v1/accounts/contacts_controller.rb
```diff
@@ -1,17 +1,40 @@
...
+    render json: { error: 'Specify search string with parameter q' }, status: :unprocessable_entity if params[:q].blank? && return
+
+    contacts = Current.account.contacts.where.not(email: [nil, '']).or(Current.account.contacts.where.not(phone_number: [nil, '']))
+                      .where('name LIKE :search OR email LIKE :search', search: "%#{params[:q]}%")
+    @contacts_count = contacts.count
+    @contacts = contacts.left_outer_joins(:conversations)
+                        .select('contacts.*, COUNT(conversations.id) as conversations_count, MAX(conversations.contact_last_seen_at) as last_seen_at')
+                        .group('contacts.id')
+                        .includes([{ avatar_attachment: [:blob] }, { contact_inboxes: [:inbox] }])
+                        .page(@current_page).per(RESULTS_PER_PAGE)
```
pranavrajs, line 29: Duplicate method. Please create a method.

PR #1340, opened 2020-10-13
chore: Reduce the click area on Branding

pranavrajs:
- Reduce the click are on branding footer


app/javascript/widget/components/Branding.vue
```diff
@@ -34,21 +34,25 @@ export default {
...
-  filter: grayscale(1);
-  font-size: $font-size-small;
   justify-content: center;
-  text-align: center;
-  text-decoration: none;
   padding: $space-normal 0 $space-slab;
-  cursor: pointer;
+  text-align: center;
+
+  a {
```
nithindavid, line 42: Use a classname ?

nithindavid (changes requested):
Add a classname to tag, Looks good otherwise.

PR #1342, opened 2020-10-14
Update Spanish translation

RicardoRamirezR:
Fine tune Spanish translation


app/javascript/dashboard/i18n/locale/es/cannedMgmt.json
```diff
@@ -67,7 +67,7 @@
       },
       "CONFIRM": {
         "TITLE": "Confirmar eliminación",
-        "MESSAGE": "¿Está seguro de eliminar ",
+        "MESSAGE": "¿Estás seguro de eliminar ",
```
pranavrajs, line 70: I believe Estás is used in informal context. I don't really know spanish, but just checking whether the usage is correct.

pranavrajs (approved):
@RicardoRamirezR Thanks for the PR. I believe the previous version was generated using Google Translate. This is helpful. 👍

app/javascript/dashboard/i18n/locale/es/cannedMgmt.json
RicardoRamirezR: @pranavrajs mhmm, yes and not. I chose that one to be consistent across the whole translation. If you prefer the "formal" one, I can change all of them. i.e. Introduzca is the "formal" one, Introduce is the other one, etc.

pranavrajs: @RicardoRamirezR Thanks for the details. I would suggest to use the formal one if that is possible. :)

pranavrajs (changes requested):
As discussed, we would go ahead with the formal language :)

RicardoRamirezR:
@pranavrajs done the formal one!

app/javascript/dashboard/i18n/locale/es/cannedMgmt.json
RicardoRamirezR: Done

PR #1351, opened 2020-10-19
DRY up check_authorization method

RonWalker22:
While adding more model authorization policies to my fork, I noticed the `check_authorization` method was not DRY. So this is my attempt to clean up. 

I replaced the six `check_authorization` methods with one method inside of the `base controller`. 

One edge case that is accounted for is when the model you want to authroize is not the same as the controller name. For example, for `AgentsController` we want to check authorization for the `User` model. In this case, we just pass in the correct model to `super`.


sojan-official (approved):
👍

PR #1352, opened 2020-10-20
fix: Add compact type, fix spacing in contact panel

pranavrajs:
Fix spacing in sidepanel

### DIFF
--- a/app/javascript/dashboard/assets/scss/widgets/_buttons.scss
+++ b/app/javascript/dashboard/assets/scss/widgets/_buttons.scss
@@ -40,7 +40,8 @@
   }
 
   &.compact {
-    padding: 0;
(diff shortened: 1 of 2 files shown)


app/javascript/dashboard/assets/scss/widgets/_buttons.scss
```diff
@@ -38,6 +38,10 @@
   &.round {
     border-radius: $space-larger;
   }
+
+  &.compact {
+    padding: 0;
```
nithindavid, line 43: Not sure if putting 0 padding to left and right is a good idea, Can we at least put a small padding for left and right side ?

nithindavid (changes requested):

nithindavid (approved):

PR #1408, opened 2020-11-13
feat: Ignore tweets from blocked users

### LINKED ISSUE #1154: Twitter generating Conversations for blocked users
**Describe the bug**
If a blocked user tweets your @username, this is converted to a new conversation. Usually on Twitter, you would not expect to see this at all. You would not know it happened.

Even if you reply, they will not see the message either.

**To Reproduce**
Block @username
Tweet from that username to a username in your CW installation
New conversation appears from that username.

**Expected behavior**
I would imagine the expected behaviour would be more like twitter.com or your twitter client in that no message appears from a blocked user at all.

It is possible this is due to how the results are implemented from search. In which case, it may be pertinent to show a users blocked status in the CW interface, i.e at the top of the message or in the user card... "the user is blocked" etc.

Adding scope, ability to block a user from CW could be useful in situations that warrant it and likewise, unblocking them could be useful too.

pranavrajs:
- Ignore blocked_users

Fixes #1154

### DIFF
--- a/spec/lib/webhooks/twitter_spec.rb
+++ b/spec/lib/webhooks/twitter_spec.rb
@@ -10,6 +10,7 @@ describe Webhooks::Twitter do
   let!(:twitter_inbox) { create(:inbox, channel: twitter_channel, account: account, greeting_enabled: false) }
   let!(:dm_params) { build(:twitter_message_create_event).with_indifferent_access }
   let!(:tweet_params) { build(:tweet_create_event).with_indifferent_access }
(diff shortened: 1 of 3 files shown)


spec/lib/webhooks/twitter_spec.rb
```diff
@@ -22,6 +23,13 @@
     end
 
     context 'with tweet_params params' do
+      it 'does not create incoming message in the twitter inbox if it is a blocked user' do
```
sojan-official, line 26: 'if it is a blocked contact'  ?

sojan-official (approved):

PR #1412, opened 2020-11-15
chore: Automate conversation display_id generation with db triggers

### LINKED ISSUE #1406: Load testing using Locust on Self hosted Chatwoot
**Describe the bug**

DB started throwing errors when doing load testing using locust for 500 users at spawn rate of 6 users sec. Could you please tell us what is the purpose of the display_id and index below and why concurrent inserts to db are failing at higher loads ?

**Server logs**

The DB logs were showing this:

2020-11-13 04:41:07.991 UTC [4251] ERROR:  duplicate key value violates unique constraint "index_conversations_on_account_id_and_display_id"

2020-11-13 04:41:07.991 UTC [4251] DETAIL:  Key (account_id, display_id)=(4, 2548) already exists.
  | 2020-11-13 04:41:07.991 UTC [4251] STATEMENT:  INSERT INTO "conversations" ("account_id", "inbox_id", "created_at", "updated_at", "contact_id", "display_id", "contact_inbox_id") VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING "id"

2020-11-13 04:41:07.991 UTC [4253] ERROR:  duplicate key value violates unique constraint "index_conversations_on_account_id_and_display_id"
  | 2020-11-13 04:41:07.991 UTC [4253] DETAIL:  Key (account_id, display_id)=(4, 2548) already exists.
  | 2020-11-13 04:41:07.991 UTC [4253] STATEMENT:  INSERT INTO "conversations" ("account_id", "inbox_id", "created_at", "updated_at", "contact_id", "display_id", "contact_inbox_id") VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING "id"

 2020-11-13 04:41:08.329 UTC [4252] ERROR:  duplicate key value violates unique constraint "index_conversations_on_account_id_and_display_id"
 2020-11-13 04:41:08.329 UTC [4252] DETAIL:  Key (account_id, display_id)=(4, 2550) already exists.

**Environment**

The above chatwoot setup was on self hosted openshift servers using docker strategy.

**Additional context**

Help us to diagnose the issue we have indicated above. Also, please point us to any load testing use-cases and real life scenario studies on chatwoot in large scale deployments. It will be useful. Also statistics on the requests chatwoot can handle in production deployments will be useful and helpful.

sorumehta:
## Description

To allow for more new conversation requests without creating race conditions during database insert, I have created triggers on database level to create display_id automatically in a new migrations file (although it won't be included in schema.rb, because rails won't understand it) and I removed callback (set_display_id) from the model file conversations.rb. I have verified the display ids in database table by creating new conversations with multiple accounts.

But there is a new issue now with API response of new conversation. The id of new conversation in the JSON response is coming as null, although its a valid integer in the database. 

Since I am new to Rails development, I am finding it difficult to fix this new issue with my changes. I hope someone can help me figure it out

Fixes #1406 (app not allowing concurrent conversations request)

## Type of change
Database level (Postgres specific)

## How Has This Been Tested?

Tested manually by creating multiple new conversations from multiple accounts and making sure that the display_id for each account_id are incrementing sequentially.


sorumehta:
@pranavrajs  created the db level change which we discussed on issue #1406 . However, need your help with fixing new conversation API response. 
Its my first Rails project and first time contribution to open source, so if I am missing something important, please let me know. I will be happy to fix it.

pranavrajs:
@sony-mathew @sojan-official Can you please take a look at this?

sorumehta:
@pranavrajs Since removing set_display_id callback is giving me null, how can I get the display_id from this callback in the new trigger based create without doing "next_display_id = account.conversations.maximum('display_id').to_i + 1" ?

sorumehta:
@pranavrajs Fixed the null id issue. Please review.

sorumehta:
ci/circleci Backed Test phase says "No examples found."
How can I pass this test?

sojan-official:
@sorumehta most of the chatwoot installation options assume schema.rb as the source of truth for database structure.  
so they use `rake db:schema:load` instead of ` rake db:migrate`. 

the particular test failure you are seeing is because we set up the test database using `db:schema:load` . But because of the pending trigger migration, it's failing without getting into running tests.   

Let me look into what we can do here and get back to you on this.  Thanks a lot for taking this on.

sorumehta:
Thank you @sojan-official for taking a look. And yes, that's true, you do have to run 'rails db:migrate' (and I am not sure how well it would work on an existing production database, I just tested on a 'fresh' database). If you want to integrate this into schema, this link might be relevant: https://stackoverflow.com/questions/27507094/rails-schema-rb-does-not-include-new-custom-postgres-function

Also, I would be grateful if you could provide your feedback on my solution (and implementation) to allow concurrent conversations request. It's my first rails code so your feedback will be really helpful for me in learning.

sojan-official:
@sorumehta the proposed solution in the stackoverflow , would also require including `rake db:structure:load` . So we will need to think through about existing production systems.  

we can also look into https://github.com/jenseng/hair_trigger . the advantage here is that we will have a readable version of the trigger added to the model, So someone looking into the code will understand how the display_id is generated.  Otherwise, a person new to the codebase would find it hard to understand how the display id is generated.   

The code looks good to me. The usage of active record callbacks is done as they are intended 👍 . 

We usually opt for the rails way/readability when there isn't a significant tradeoff in performance. Hence you would see the reliance on active record callbacks vs utilizing database-specific features. In this particular case, I think the utilization of triggers should give us an advantage. 

Let's figure out the `schema.rb`  before getting this merged, But otherwise, it looks good to me.

app/models/conversation.rb
```diff
@@ -158,10 +159,7 @@ def self_assign?(assignee_id)
   end
 
   def set_display_id
-    self.display_id = loop do
-      next_display_id = account.conversations.maximum('display_id').to_i + 1
-      break next_display_id unless account.conversations.exists?(display_id: next_display_id)
-    end
+    self.display_id = account.conversations.find(id).display_id
```
sojan-official, line 162: you could do `self.reload` i believe.  
that should update the object with the values from DB 

https://apidock.com/rails/ActiveRecord/Persistence/reload

sorumehta:
@sojan-official Cool! Made the change.

sorumehta:
@sojan-official @pranavrajs So I did a load test (users=500, spawn rate=10) myself with and without my changes. Here are the results (12 GB RAM, 4 cores):
**without my changes:**
I got these logs in postgres:

ERROR:  duplicate key value violates unique constraint "index_conversations_on_account_id_and_display_id"
--
  | 2020-11-17 14:23:39.589 UTC [24147] DETAIL:  Key (account_id, display_id)=(1, 5878) already exists.
  | 2020-11-17 14:23:39.589 UTC [24147] STATEMENT:  INSERT INTO "conversations" ("account_id", "inbox_id", "created_at", "updated_at", "contact_id", "display_id", "contact_inbox_id") VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING "id"

lots of 500 status codes and successful ones had really long response times (in order of 10 seconds)

**with my changes:**
all 500 requests were successful and response times were less than a second.

So I think these changes would help in a scenario (don't know how likely though) where you have millions of customers and get more than half a million new conversation requests in a day.

sorumehta:
> @sorumehta the proposed solution in the stackoverflow , would also require including `rake db:structure:load` . So we will need to think through about existing production systems.
> 
> we can also look into https://github.com/jenseng/hair_trigger . the advantage here is that we will have a readable version of the trigger added to the model, So someone looking into the code will understand how the display_id is generated. Otherwise, a person new to the codebase would find it hard to understand how the display id is generated.
> 
> The code looks good to me. The usage of active record callbacks is done as they are intended 👍 .
> 
> We usually opt for the rails way/readability when there isn't a significant tradeoff in performance. Hence you would see the reliance on active record callbacks vs utilizing database-specific features. In this particular case, I think the utilization of triggers should give us an advantage.
> 
> Let's figure out the `schema.rb` before getting this merged, But otherwise, it looks good to me.

@sojan-official  I understand your point, it wouldn't be so easy to understand and install from a new user's perspective. Will look into the hair_trigger project you have mentioned.

sorumehta:
@sojan-official used hair_trigger this time to generate migration and it included trigger in schema this time.

sojan-official:
@sorumehta I think you missed out committing `gemfile.lock` hence the build failure. 
https://bundler.io/rationale.html#checking-your-code-into-version-control

sorumehta:
@sojan-official Can you help me make sense of the build failure logs? For example one of the error logs says:

5) Conversation Messages API POST /api/v1/accounts/{account.id}/conversations/<id>/messages when it is an authenticated user creates a new outgoing message
     Failure/Error: expect(response).to have_http_status(:success)
       expected the response to have a success status code (2xx) but it was 422
     # ./spec/controllers/api/v1/accounts/conversations/messages_controller_spec.rb:29:in `block (4 levels) in <top (required)>'

but when I tried this API on my machine, it returned with success (200). I have attached the screenshot in the comment.
test_api_response_messg

sorumehta:
@sojan-official let me know if you can help with this

sojan-official:
@sorumehta will take a look into this.

sorumehta:
@sojan-official Any updates on this?

sojan-official:
> @sojan-official Any updates on this?

@sorumehta  lost track of this one 🤦‍♂️ . will look into the specs today.

db/schema.rb
```diff
@@ -536,4 +536,18 @@
   add_foreign_key "contact_inboxes", "contacts"
   add_foreign_key "contact_inboxes", "inboxes"
   add_foreign_key "conversations", "contact_inboxes"
+  create_trigger("accounts_after_insert_row_tr", :generated => true, :compatibility => 1).
```
sojan-official, line 539: we will also have to iterate through all the existing accounts and create triggers for them, with the starting value to be max value of display id.

app/models/conversation.rb
```diff
@@ -293,4 +291,9 @@ def mute_key
   def mute_period
     6.hours
   end
+
+  # creating db triggers
+  trigger.before(:insert).for_each(:row) do
+    "NEW.display_id := nextval('account_seq_' || NEW.account_id);"
```
sojan-official, line 297: probably name the sequence as. `conv_dpid_seq_`  ?   
if we end up adding more display id sequences on like . `campaign_dp_id`

db/schema.rb
sojan-official: can probably add a new migration for this case

sorumehta:
@sojan-official I have added the sequence creation for existing accounts in migration 20201125123131 and changed the sequence name as you suggested. Please take a look when possible.

sojan-official:
> @sojan-official @pranavrajs So I did a load test (users=500, spawn rate=10) myself with and without my changes. Here are the results (12 GB RAM, 4 cores):
> **without my changes:**
> I got these logs in postgres:
> 
> ## ERROR: duplicate key value violates unique constraint "index_conversations_on_account_id_and_display_id"
>   | 2020-11-17 14:23:39.589 UTC [24147] DETAIL: Key (account_id, display_id)=(1, 5878) already exists.
>   | 2020-11-17 14:23:39.589 UTC [24147] STATEMENT: INSERT INTO "conversations" ("account_id", "inbox_id", "created_at", "updated_at", "contact_id", "display_id", "contact_inbox_id") VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING "id"
> 
> lots of 500 status codes and successful ones had really long response times (in order of 10 seconds)
> 
> **with my changes:**
> all 500 requests were successful and response times were less than a second.
> 
> So I think these changes would help in a scenario (don't know how likely though) where you have millions of customers and get more than half a million new conversation requests in a day.


@sorumehta  will do some testing and work on this to get it merged. 

Meanwhile, could you share the details of how you performed the load test? did you write any scripts or use any specific tool?

sorumehta:
@sojan-official I wrote a script using [locust](https://www.locust.io/) framework in python to load test. In the script, I generated 10 (spawn rate) new contact and conversation requests per second upto 500 users

sojan-official:
> @sojan-official I wrote a script using [locust](https://www.locust.io/) framework in python to load test. In the script, I generated 10 (spawn rate) new contact and conversation requests per second upto 500 users

If you could share the script. it would be nice.  Either way, thanks for the info. I am planning to add some documentation around testing.  Think we can add locust as an option.

sorumehta:
@sojan-official  Sure, I have shared the script, but its not the best script in the world. Also, it makes request to a different server with a random phone number each time, which in turn creates a new contact and conversation and returns a websocket url. You can replace the current POST request with new contact and new conversation API requests of Chatwoot.

Here's the code
`from locust import HttpUser, between, task,HttpLocust,TaskSet
import json
import time
from locust import events
import random
import functools

class WebsiteUser(HttpUser):

    def random_phone_number(self):
        a = functools.partial(random.randint, 0, 9)
        gen = lambda: "91{}{}{}{}{}{}{}{}{}{}".format(a(), a(), a(), a(), a(), a(), a(), a(), a(), a(), a())
        return gen()


    def on_start(self):
        print("starting new user")
        self.endpoint = "<your_endpoint_here>"
        self.lob = "prepaid"
        self.ph_number = self.random_phone_number()
        # self.ws = websocket.WebSocket()

    @task
    def chat_task(self):
        print("user task fired")
        # print(phone_number)
        payload = {'lob':self.lob, 'number': self.ph_number}
        headers = {'client_token': 'z3&pza3$V'}
        start_post_time = time.time()
        print("payload: ",payload)
        print("post url: http://"+self.endpoint+"/api/upstream/conversation/new")
        response = self.client.post("http://"+self.endpoint+"/api/upstream/conversation/new", 
                        data=json.dumps(payload),
                        headers=headers)

        if response.status_code != 200:
            try:
                print("response status code is: ", response.status_code)
                print("response body is: ", response.json())
                events.request_failure.fire(
                    request_type='Recv',
                    name='livechat',
                    response_time=time.time()-start_post_time,
                    response_length=0,
                    exception=None
                )
            except Exception:
                print("Exception happened for: ", response.text)
        else:
            json_response_dict = response.json()
            ws_url = json_response_dict["socket_path"]
            print("ws_url:", ws_url)

            print("passed")
            events.request_success.fire(
                request_type='Recv',
                name='livechat',
                response_time=time.time() - start_post_time,
                response_length=len(json_response_dict)
            )
                
    wait_time = between(100, 120)
`
> Think we can add locust as an option
Yes, I would highly recommend locust.

sorumehta:
@sojan-official I hope you are seeing the same results in your load test as me

sojan-official:
@sorumehta  its look good in my local, But we would need to test the changes against some production data dumps to ensure that there won't be any breakages.  will keep you updated

sorumehta:
@sojan-official Sure, thanks

sorumehta:
@sojan-official Did you get a chance to check it on production data?

sorumehta:
@pranavrajs are we still considering the merge?

pranavrajs:
@sorumehta Yes, we have not got a chance to test it yet. There are 2 things we want to test out. 

1. Testing the change against production data
2. See if there is an opportunity to remove raw SQL for active record queries in DB migration.

Thanks for the patience.

sorumehta:
@sojan-official just checking if there are any updates on this?

sojan-official:
@sorumehta  There isn't anything pending from your side. Since this has some critical data migrations which would be hard to reverse on any breakages. We didn't want to rush it in the 1.11.0 release.  We will get this merged for the 1.12.0.

sorumehta:
@sojan-official Thanks for following up on this

db/migrate/20201125123131_conv_dpid_seq_for_existing_accnts.rb
```diff
@@ -0,0 +1,46 @@
...
+            loop
+                exit when curr_accnt_id > max_accnt_id-1 ;
+                curr_accnt_id := curr_accnt_id + 1 ;
+                execute 'drop sequence IF EXISTS conv_dpid_seq_' || curr_accnt_id ;
+            end loop;
+        END IF;
+        raise notice '%', curr_accnt_id;
+    end; $$;
+    ")
+  end
```
sojan-official, line 45: @sorumehta we could probably abstract this logic from raw SQL into ruby code so that it's more readable.  Let me know your thoughts. 

```suggestion
  def up
    ::Acccount.find_in_batches do |accounts_batch|
      Rails.logger.info "migrated till #{accounts_batch.first.id}\n"
      accounts_batch.each do |account|
        display_id = Conversation.where(account_id: account_id).maximum('display_id')
        display_id ||= 0 # for accounts with out conversations
        ActiveRecord::Base.connection.exec_query("create sequence IF NOT EXISTS conv_dpid_seq_#{account.id} START #{display_id + 1}")
      end
    end
  end

  def down
    ::Acccount.find_in_batches do |accounts_batch|
      Rails.logger.info "migrated till #{accounts_batch.first.id}\n"
      accounts_batch.each do |account|
        ActiveRecord::Base.connection.exec_query("drop sequence IF EXISTS conv_dpid_seq_#{account.id}")
      end
    end
  end
```

sojan-official:
@sorumehta  I did some database migration tests and it seems things are looking good. Have a suggestion for one of the migrations. But let's get this merged in the next couple of days. Thanks again for the patience.

db/migrate/20201125123131_conv_dpid_seq_for_existing_accnts.rb
sorumehta: @sojan-official  This looks great, and much more readable. Shall I click on the 'Commit suggestion' button?

sojan-official: @sorumehta please go ahead. I will one more quick round of testing over that migrations.

sorumehta:
@sojan-official committed your suggestion

sojan-official:
Added https://github.com/chatwoot/chatwoot-locust-tests for running the load tests over these changes. 
@sorumehta  if you have some bandwidth, please take a look.

sojan-official (approved):
@sorumehta  Thanks a ton for the work over this PR. Also thanks a lot for the patience. 😅

sorumehta:
@sojan-official Thanks to you too for considering my changes and improving them. It was a great experience, and in the process, I got to know the awesomeness of Rails framework!

db/schema.rb
```diff
@@ -219,8 +219,8 @@
     t.integer "inbox_id", null: false
     t.integer "status", default: 0, null: false
     t.integer "assignee_id"
-    t.datetime "created_at", null: false
```
pranavrajs, line 222: @sojan-official Why was this change done?

sojan-official: I think this got crept in unintentionally. let me raise a PR fixing this.

PR #1418, opened 2020-11-16
fix: Avoid keys getting translated in availability

pranavrajs:
- Update STATUSES_LIST to array instead of object array


app/javascript/dashboard/components/layout/AvailabilityStatus.vue
```diff
@@ -55,21 +55,32 @@ export default {
...
   computed: {
     ...mapGetters({
       currentUser: 'getCurrentUser',
     }),
+    availabilityDisplayLabel() {
+      const availabilityIndex = this.statusKeys.findIndex(
+        key => key === this.currentUserAvailabilityStatus
+      );
+      return this.$t('PROFILE_SETTINGS.FORM.AVAILABILITY.STATUSES_LIST')[
+        availabilityIndex
```
nithindavid, line 71: Should we rely array index position from translation files, since its compared against order in our constant variable?

nithindavid (changes requested):

sojan-official (approved):

PR #1433, opened 2020-11-20
feat: Default locale

### LINKED ISSUE #830: set default locale
I can not specify locale for login screen, other than english

define default locale as enviroment variable for example DEFAULT_LOCALE .

using system variable for example LC_ALL.

**Additional context**
Also it seems that agent locale is allways english, even if account locale defined as greek (for example)

#### 1. Ability to define a default language on the installation. 

Define an environment variable `DEFAULT_LOCALE`.
1. If the variable is set, all non-authenticated pages would fallback to the default locale. 
2. Whenever a new account is created, the default language will be DEFAULT_LOCALE instead of en

#### 2. Ability to select different locale in non-authenticated pages
There is will be a dropdown at the bottom of the page to change the locale. If the user changes the locale, save it in the local storage and use this locale in his next visit.

azyzio:
# Pull Request

## Description

The issue was that unauthenticated pages were always using english locale, so all the work done by translators was essentially wasted - their translations were never displayed.

The PR introduces the DEFAULT_LOCALE env variable that is a fallback for params-based & account-specific locale.

When creating new account, locale set on the registration page will be used.

In addition, some refactoring has been done so there is only one source of truth for dashboard-related locale.

(note: I'm not a Vue.js developer, so I might not have used best practices on the front-end)

Fixes #830

## Type of change
- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?
Manual testing


app/javascript/dashboard/App.vue
```diff
@@ -28,18 +28,21 @@ export default {
 
   mounted() {
     this.$store.dispatch('setUser');
+    this.setLocale();
     this.initializeAccount();
   },
 
   methods: {
+    async setLocale() {
```
pranavrajs, line 36: This seems like a smart move. ❤️

pranavrajs:
@azyzio Thanks for the PR, will test it out and let you know if anything is required. @sojan-official Can you please take a look at the backend changes?

sojan-official (approved):
❤️

pranavrajs (approved):

pranavrajs:
@azyzio Made a small change around the locale as it was not setting the locale based on the account when logged in.

azyzio:
@pranavrajs Cool thanks 👍 

That's a little weird though as `I18n.locale` set by ruby should be set to account locale (unless there are request-specific params to overwrite it). Maybe the variable in `window.chatwootConfig` didn't get updated properly between logging in and navigating the dashboard. Sorry for that - I'm not not used to one-pager sites like those in vue.js.

pranavrajs:
@azyzio No worries, this PR was helpful. Setting I18n.locale wouldn't have any effect on the frontend as it is a separate app.  You have to manually set the language on the frontend.

PR #1435, opened 2020-11-21
fix: Add missing close button if bubble is hidden

### LINKED ISSUE #1434: Unable to close chat window when hideMessageBubble: true
**Describe the bug**

When using:

chatwootSettings = { hideMessageBubble: true };

And toggling the chat, at least on desktop the chat window does not have an X to close it. I believe it would be preferrable to have it there always (like it is when opened up on mobile).

Also, it should be possible to set the distance from the bottom of the browser window as now it has a weird huge gap (see screenshot).

Screenshot 2020-11-20 at 15 16 00

pranavrajs:
Fixes #1434 

- Add missing close button
- Fix the position of the widget holder if there is no bubble present

Screenshot 2020-11-21 at 6 19 43 PM

Screenshot 2020-11-21 at 6 19 30 PM

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -325,7 +325,7 @@ GEM
     pry-rails (0.3.9)
       pry (>= 0.10.4)
     public_suffix (4.0.5)
-    puma (4.3.5)
+    puma (4.3.6)
       nio4r (~> 2.0)
     pundit (2.1.0)
       activesupport (>= 3.0.0)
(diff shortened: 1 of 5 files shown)


Gemfile.lock
```diff
@@ -325,7 +325,7 @@ GEM
     pry-rails (0.3.9)
       pry (>= 0.10.4)
     public_suffix (4.0.5)
-    puma (4.3.5)
+    puma (4.3.6)
```
pranavrajs, line 328: https://github.com/puma/puma/issues/2342

sojan-official (approved):

PR #1442, opened 2020-11-23
enh: Create superadmin in seeds

azyzio:
# Pull Request

## Description

While testing anything related to superadmin panel or jobs, the developer had to manually create a superadmin account via console.
The PR modifies the seeds to create the superadmin account

## Type of change

- Developer experience enhancement

## How Has This Been Tested?

Manual testing


sojan-official:
@azyzio I would lean towards having a rake task for this or seed this based on a condition[environment variable] ensuring that its a development environment.  

Since `rake db:seed` is run by community installs. A person not aware of the super admin feature will leave the default account intact and might be susceptible towards attacks.  

thoughts? 
cc: @pranavrajs  @sony-mathew

azyzio:
@sojan-official ok, assumed seeds were never run on production - but I agree it might be possible with self-hosted distribution systems. What do you think about just running it `unless Rails.env.production?`

sojan-official:
> @sojan-official ok, assumed seeds were never run on production - but I agree it might be possible with self-hosted distribution systems. What do you think about just running it `unless Rails.env.production?`

`unless Rails.env.production?`  sounds good to me

pranavrajs:
unless Rails.env.production? sounds good to me too!

azyzio:
Thanks for the observation!
I added the condition :)

pranavrajs (approved):

PR #1467, opened 2020-12-01
Update local_resource.rb

SmartManoj:
# Pull Request Template

## Description

No need to split file type as it is not an URL and defaults to nil

### DIFF
--- a/lib/local_resource.rb
+++ b/lib/local_resource.rb
@@ -25,7 +25,7 @@ class LocalResource
   end
 
   def find_file_type
-    @file_type ? @file_type.split('/').last : Pathname.new(uri.path).extname
+    @file_type || Pathname.new(uri.path).extname
   end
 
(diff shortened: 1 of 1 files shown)


pranavrajs:
@SmartManoj I don't understand the change here. @sojan-official / @sony-mathew Can you please take a look?

SmartManoj:
here file_type will not have '/'.

rawan322 (approved):

lib/local_resource.rb
```diff
@@ -25,7 +25,7 @@ def encoding
   end
 
   def find_file_type
-    @file_type ? @file_type.split('/').last : Pathname.new(uri.path).extname
```
sojan-official, line 28: There are places where we pass encoding as the file type. hence this is intended.  
cc: @pranavrajs

PR #1547, opened 2020-12-23
Enhancement 731- Add pending message view for text messages

### LINKED ISSUE #731: Chore: Add pending message in dashboard
Now the messages sent from the dashboard are displayed only after the API call is a success. Change this to have a temporary message in frontend and update the status when the API call is a success.

nithindavid:
Fixes #731


app/javascript/dashboard/helper/commons.js
```diff
@@ -26,3 +30,19 @@ export const getTypingUsersText = (users = []) => {
...
 };
+
+export const createPendingMessage = data => {
+  const timestamp = new Date().getTime() / 1000;
+  const tempMessageId = getUuid();
+  const pendingMessage = {
+    ...data,
+    content: data.message,
+    id: tempMessageId,
+    echo_id: tempMessageId,
```
pranavrajs, line 41: I think it is better to use echoId instead of echo_id here.

app/javascript/dashboard/store/modules/conversations/index.js
```diff
@@ -85,17 +85,19 @@ export const mutations = {
       selectedChatId: conversationId,
     });
     if (!chat) return;
-    const previousMessageIndex = chat.messages.findIndex(
-      m => m.id === message.id
+    const { echo_id: tempMessageId } = message;
+    const pendingMessageIndex = chat.messages.findIndex(
+      m => m.id === message.id || m.id === tempMessageId
```
pranavrajs, line 90: `m.echo_id === tempMessageId`?

app/javascript/shared/constants/messages.js
```diff
@@ -0,0 +1,12 @@
+export const MESSAGE_STATUS = {
+  FAILED: 'failed',
+  SUCCESS: 'sent',
```
pranavrajs, line 3: SENT?

app/views/api/v1/accounts/conversations/messages/create.json.jbuilder
```diff
@@ -1,6 +1,7 @@
 json.id @message.id
 json.content @message.content
 json.inbox_id @message.inbox_id
+json.echo_id @message.echo_id
```
pranavrajs, line 4: Please add this in index.json.jbuilder as well. @sojan-official Any thoughts on moving this entire thing to a partial?

app/javascript/dashboard/helper/commons.js
```diff
@@ -1,5 +1,9 @@
 /* eslint no-console: 0 */
```
pranavrajs, line 1: I know this may not be the right place, I guess we can remove no-console flag

app/views/api/v1/accounts/conversations/messages/create.json.jbuilder
sojan-official: let move this to a partial.

pranavrajs (commented):
Looks good overall, nice work!

app/javascript/shared/constants/messages.js
nithindavid: Matching response from action cable

pranavrajs: I meant the label of sent, it is still 'success'.

WevrLabs:
Right on time!
I was composing a long reply last night, and just as the moment I hit Enter, boom! Connection dropped briefly coincidentally and then I had to re-write it all over again.

app/javascript/dashboard/store/modules/conversations/index.js
nithindavid: `echo_id` is just o get the echo, we are keeping the temp value in `id` itself. Else we will have clear echo_id and check each time as there is response from both `action-cable` and `api`

app/javascript/dashboard/components/widgets/conversation/Message.vue
```diff
@@ -139,17 +151,27 @@ export default {
         'is-image': this.hasImageAttachment,
       };
     },
+    isPending() {
+      const isPending = this.data.status === MESSAGE_STATUS.PROGRESS;
```
pranavrajs, line 155: This variable is useless, please return directly.

pranavrajs:
@nithindavid The entire messages are moving up and down after a message is sent.

https://www.loom.com/share/52bff3e21bcb4b8e91cfb7f586d4e4b8

pranavrajs (approved):
👍 LGTM

PR #1563, opened 2020-12-29
feat: API to add label to contacts

pranavrajs:
- Add labels to contact
- List label of a contact


sojan-official (dismissed):

app/controllers/concerns/label_helper.rb
```diff
@@ -0,0 +1,10 @@
+module LabelHelper
```
sojan-official: `XYCHelper` naming convention is (mostly) used in Rails views to share reusable code. We could evaluate a different naming convention for controller concerns.

sojan-official (approved):

PR #1569, opened 2020-12-31
fix: Update `listen` gem

pranavrajs:
- Update `listen` gem to latest

This helps in reducing the CPU usage in Mac OS Big Sur

Related: https://github.com/guard/listen/issues/478


pranavrajs:
@sojan-official Merging this as it is blocking my development workflow.

PR #1623, opened 2021-01-07
Updates reply box UI 👀

### LINKED ISSUE #1622: Update reply box UI to accommodate formatting
**Why?**
With upcoming formatting toolbar and attachment preview, The reply box needs to revamp it's design.

nithindavid:
Fixes #1622 

**New design preview**
- Default state
image
image

- @sojan-official @pranavrajs  Need feedback on the position if emoji widget, placed it outside rather placing over he input box  box so that when I append an emoji I expect to see it in the input box.
image

- Private note
image

- With attachment
image


app/javascript/dashboard/components/widgets/WootWriter/ReplyBottomPanel.vue
```diff
@@ -0,0 +1,164 @@
...
+  display: flex;
+  justify-content: space-between;
+  padding: var(--space-slab) var(--space-normal);
+
+  &.is-note-mode {
+    background: var(--y-50);
+  }
+}
+
+.button {
```
nithindavid, line 101: These styles are close to tailwind button styles, Didn't move it to a generic file since polluting it might cause breakages. Also we might be going with tailwind soon.

pranavrajs (approved):

Issue #900, opened 2020-05-29
Feature: Show a setup installation page when no superadmin/accounts are available


pranavrajs:
**Problem**

The current installation process in Heroku/any other provider is hard. After the installation is up and running user has to go to DB to create an admin account.

**Solution**

To make this process easier, create a setup page where the user can provide an email address and password.

pranavrajs:
@sojan-official Can't we close this issue?

sojan-official:
Closing this since admin accounts are created during installation flow. We can revisit creating super admin accounts at a later point.

ChrisvanChip:
> Closing this since admin accounts are created during installation flow. We can revisit creating super admin accounts at a later point.

Isn't it an idea to make the first admin automatically super admin? If a company requires more super admin accounts, they can do so through the console.

sojan-official:
@ChrisvanChip Starting from the upcoming release, the first user account on chatwoot will automatically be a super admin as well.

PR #1640, opened 2021-01-11
feat: Installation On-boarding flow

sojan-official:
- Have an onboarding flow for fresh installations

Changes: 

- Allow users to set up the account on the first load.
- Account setup would be required to use the application.
- During the setup, Chatwoot collects name, company name, email, password (minimum fields required to set up an account)
- If the user is willing to receive release notes, newsletter & feedback surveys, he/she would be able to check the checkbox to subscribe to these during the setup. 
  - If this is checked the information would be shared with `hub.chatwoot.com`. 
  - If the user does not check this field, none of the above fields would be shared with `hub.chatwoot.com`

Ref: #900


app/views/super_admin/onboarding/index.html.erb
```diff
@@ -0,0 +1,59 @@
...
+        <div class="row align-center">
+          <div class="small-12 medium-4 column">
+            <%= form_tag('/super_admin/onboarding', class: 'login-box column align-self-top') do %>
+              <div class="column log-in-form">
+                <% if flash[:error].present? %>
+                <div data-alert class="alert-box warning"><%= flash[:error] %></div>
+                <% end %>
+
+                <label>
+                  Name
```
pranavrajs, line 25: Should we use i18n here?

config/routes.rb
```diff
@@ -212,6 +212,11 @@
     end
   end
 
+  namespace :super_admin do
```
pranavrajs, line 215: IMO Superadmin namespace is misleading at the moment. We are not creating a super admin, we just create an account.

Issue #866, opened 2020-05-15
Feature: Organize Agents into teams 


sojan-official:
It's difficult to manage agents in chatwoot system if the number of agents is huge. Let's introduce teams/permission management at a higher level to mitigate this.  

todo: explore the idea further.

AwakenTheJaken:
+1 for this, we were planning on using chatwoot 🚀 as a replacement for Livezilla 🤮 but without this feature and the ability to route a user to a specified team it makes it difficult for us to make the jump.

pranavrajs:
API for this is merged, will be continuing the work in 1.13.0

PR #1654, opened 2021-01-14
feat: Team APIs

sojan-official:
Teams feature would allow users to group agents into different teams and it would allow assigning to a team.

**Attributes:**

- **id** - Auto generated id
- **name** - The name of the team
- **description** - The description about the team
- **allow_auto_assign** - If this is true, assigning to the team would auto-assign the conversation to an available agent in the team
- **account_id** - The account to which the team belongs

**Relations:**

- Team has many users
- Users has many teams

**APIs:** 

- API for CRUD on teams
- API to add/remote user from a team
- API to list all the users of a team
- API to list all the teams a user belongs to

Ref: #866


app/controllers/api/v1/accounts/team_members_controller.rb
```diff
@@ -0,0 +1,24 @@
+class Api::V1::Accounts::TeamMembersController < Api::V1::Accounts::BaseController
+  before_action :fetch_team
+  before_action :check_authorization
+
+  def index
+    render json: @team.team_members.map(&:user)
```
pranavrajs, line 6: Should we use agent partial here?

app/controllers/api/v1/accounts/team_members_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def create
+    @team.team_members.find_or_create_by(user_id: params[:user_id])
+    head :ok
```
pranavrajs, line 11: Let us return team_member here

pranavrajs (approved):

PR #1659, opened 2021-01-15
fix: Add correct thread message_id to the email message

pranavrajs:
The intention is to allow public to beta test email channel
- Fixed thread issues with outgoing email, now it would select the last incoming email message_id
- Fixed outgoing message subject name, set it to `re: original message name`
- Send email in email channel without 2 minute delay
- Display generated email address in the UI


sony-mathew (commented):
LGTM. (Needs some spec changes and code-climate failure fixes)

sojan-official (approved):

Issue #1480, opened 2020-12-02
Issue with conversations


KonstantinReido:
**Describe the bug**

In the list of conversation on the counter, I see 8, but the list contains only 1. Also, the system sends email notifications to admin, but without a visible message.
<img width="399" alt="Screen Shot 2020-12-02 at 16 42 55" src="https://user-images.githubusercontent.com/172571/100901860-fb5bc400-34cc-11eb-985a-d276e20258cc.png">

I think it can be related. On "Retries" page of Sidekiq admin panel, I periodically see jobs. Killing them.

<img width="1239" alt="Screen Shot 2020-12-02 at 18 36 24" src="https://user-images.githubusercontent.com/172571/100902390-82a93780-34cd-11eb-9212-abc3b93202e7.png">

I don't see in DB contacts with these ids

**Environment**

Self hosted installtion on Heroku. 

Chatwood v1.10.0

pranavrajs:
@KonstantinReido Can you please share the browser logs?

KonstantinReido:
@pranavrajs like this?
```
XHRGEThttps://support.xxx.com/api/v1/accounts/1/labels
[HTTP/1.1 200 OK 129ms]

XHRGEThttps://support.xxx.com/api/v1/accounts/1/inboxes
[HTTP/1.1 200 OK 379ms]

XHRGEThttps://support.xxx.com/api/v1/accounts/1/conversations?status=open&assignee_type=me&page=1
[HTTP/1.1 200 OK 537ms]

XHRGEThttps://support.xxx.com/api/v1/accounts/1/agents
[HTTP/1.1 304 Not Modified 453ms]

XHRGEThttps://support.xxx.com/auth/validate_token
[HTTP/1.1 304 Not Modified 429ms]

XHRGEThttps://support.xxx.com/api/v1/accounts/1/
[HTTP/1.1 304 Not Modified 204ms]

XHRGEThttps://support.xxx.com/api/v1/accounts/1/conversations?status=open&assignee_type=all&page=1
[HTTP/1.1 200 OK 553ms]

GEThttps://www.gravatar.com/avatar/eebe522533ae9c9056f144036dbc63d3?d=404
[HTTP/2 404 Not Found 403ms]

TypeError: t is null
    unreadMessagesCount conversations.js:12
    unreadMessagesCount conversations.js:11
    getUnreadCount ConversationCard.vue:112
    VueJS 3
    hasUnread ConversationCard.vue:116
    VueJS 2
[...]

PR #1683, opened 2021-01-22
Chore: Fix Deserialisation Error in Jobs

sojan-official:
# Pull Request Template

## Description

- Ignore Jobs on Deserialisation error
- Show empty conversations in chat UI

Fixes: #1480

## Type of change

## How Has This Been Tested?

## Checklist:


app/jobs/application_job.rb
```diff
@@ -1,2 +1,6 @@
 class ApplicationJob < ActiveJob::Base
+  # https://api.rubyonrails.org/v5.2.1/classes/ActiveJob/Exceptions/ClassMethods.html
+  discard_on ActiveJob::DeserializationError do |_job, error|
```
pranavrajs, line 3: Isn't it suppressing the error? Should we try to fix the reason why deserialization error occurs? I guess this is to remove the errors like "Conversation with id not found" and "contact with id not found" errors in job.

pranavrajs (approved):
I have a comment around deserialization error, otherwise looks fine.

app/views/api/v1/conversations/partials/_conversation.json.jbuilder
```diff
@@ -11,7 +11,9 @@ json.meta do
 end
 
 json.id conversation.display_id
-if conversation.unread_incoming_messages.count.zero?
+if conversation.messages.count.zero?
```
pranavrajs, line 14: If you can cover this with a test, that would be nice.

app/jobs/application_job.rb
sojan-official: The error occurs because you are passing in an object as an argument and that object doesn't exist in the database.  [Either the object is deleted / its a race condition where the job is picked up before the item is committed to the database.]

After looking into logs the primary cases I have seen are 
- Contact id not found ( Result of merge action where original contact was deleted ) 
- Contact Avatar job ( same ) 
- Conversation Id ( conversation wasn't committed cause of some error. but event dispatch job was called on update ) 

There isn't much we can do to recover/reprocess the job in case of these deserialization errors.

pranavrajs (approved):

PR #1720, opened 2021-02-01
Add name error validation for account setting Profile name #1714

### LINKED ISSUE #1714: Agent Name cannot be shorter than 4 characters if edited by Admin
**Describe the bug**

Admins cannot set Agent Name to anything shorter than 4 characters on users while the users themselves do not have this restriction when editing Full Name.

**To Reproduce**

1. Go to **Settings**
2. Add an agent or edit an existing agent
3. Enter an **Agent Name** shorter than 4 characters
4. See Agent Name field light up in red and the Edit/Add Agent button become disabled

**Expected behavior**

I expected the admin to be able to set an Agent Name shorter than 4 characters. 

**Additional context**

We were running a self hosted v1.12.0.

In our case we wanted to only show agent's given name due to privacy. Two agents, of which I was one of them, apparently had too short names. Originally we wanted to set the proper Full Name and set the given name as Display Name, but soon realised we - as admins - could not set the Display Name for agents. This in turn led us to discover the minimum name limit, at least for admins. We were also a bit bummed that we could not restrict agents from editing Full Name and Display Name, but that is borderline offtopic for this bug report.

Manually removing `disabled="disabled"` from the submit-button HTML in the browser dev tools seem to allow the form submit properly though. =)

karthik-sivadas:
# Pull Request Template

## Description

Add name error validation for account settings Profile name. Have set minimum string length for the name as 4.
Screenshot 2021-02-01 at 11 24 55 PM

Fixes #1714

## Type of change

Bugfix (a non-breaking change that fixes an issue)

## How Has This Been Tested?

Setting the profile name in the account settings with a string value less than 4 throws an error as shown in the above image.

`bundle exec rspec spec/models/user_spec.rb`

### DIFF
--- a/app/javascript/dashboard/routes/dashboard/settings/agents/AddAgent.vue
+++ b/app/javascript/dashboard/routes/dashboard/settings/agents/AddAgent.vue
@@ -102,7 +102,7 @@ export default {
   validations: {
     agentName: {
       required,
-      minLength: minLength(4),
+      minLength: minLength(1),
     },
     agentEmail: {
       required,
--- a/app/javascript/dashboard/routes/dashboard/settings/agents/EditAgent.vue
+++ b/app/javascript/dashboard/routes/dashboard/settings/agents/EditAgent.vue
@@ -112,7 +112,7 @@ export default {
   validations: {
     agentName: {
       required,
-      minLength: minLength(4),
+      minLength: minLength(1),
     },
     agentType: {
       required,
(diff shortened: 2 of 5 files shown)


pranavrajs (approved):
@karthik-sivadas Thanks for the PR, the changes look good to me. Should we add the validation at the backend as well?

karthik-sivadas:
> @karthik-sivadas Thanks for the PR, the changes look good to me. Should we add the validation at the backend as well?

Also adding `validates_length_of :name, minimum: 4` to the `User` model.

pranavrajs:
@karthik-sivadas Changes looks good, can you please update this at 

`app/javascript/dashboard/routes/dashoard/settings/agents/AddAgent.vue` and
`app/javascript/dashboard/routes/dashoard/settings/agents/EditAgent.vue`

pranavrajs (approved):
LGTM

Issue #1345, opened 2020-10-15
Enhancement: Add sounds to staff view.


iammattmartin:
**Is your enhancement request related to a problem? Please describe.**
Yes, when staff have the admin area open and dealing with a chat, unless they keen a concisious view of that, conversation in a running chat may be forgotten.

The visitor gets a discreet noise on a new message, extending this to staff would be good as an optional setting per staff member.

This will help that when people are live on site, chats are not forgotten as active.

**Describe the solution you'd like**
The same noise visitors get when a new message is received.

mohmmd256:
Might also bring this to attention:

![Screen Shot 2020-10-16 at 5 54 10 PM](https://user-images.githubusercontent.com/7905931/96280881-d1ac1380-0fd8-11eb-910f-3d5bbaeccc01.png)

pranavrajs:
We have looked at the different ways to enable this. Most of it has been hacky and that's the reason why we suggested to go ahead with the push notification. I can see that this has growing interest. We will take a look at it again. 

@mohmmd256 Are you talking about the widget side? We don't have one on dashboard.

iammattmartin:
Push is good but if you have a massive monitor (I use a 49" ultra wide screen), these are often missed due to being top right corner.

Also, we use Fluid (https://fluidapp.com/) to have CW in an app box so we can keep on the dock. Push notifications don't work for it.

When there is a native CW app/wrapper that may fix that but then the notification view and being able to see them still remains a problem. With bigger monitors appearing, that may happen more often. 

Finally, with as many notifications as some people may get in a day, it may be lost in all the noise too...

mohmmd256:
@pranavrajs Yes correct, I am referring to the widget.

tabrza:
@pranavrajs - what are your thoughts on moving forwards on this?

pranavrajs:
@tabrza #989 changes only the sound file, it doesn't solve the problem.

pranavrajs:
This is tracked via #1738, the changes are working, but we need to add an ability to mute the notifications, otherwise, it will be an annoying feature. Moving it to the new milestone as the development would be continued.

PR #1738, opened 2021-02-07
fix: Add alert for incoming messages on dashboard

nithindavid:
Fixes - #1345

**Note**
- Does not work on safari, Seems to work on Firefox/Chrome. (User needs to make some interaction with page, scroll, touch )
- Need to add more cases to when to play the sound.
- Need an initial review on sound part.


app/views/layouts/vueapp.html.erb
```diff
@@ -50,5 +50,8 @@
     <div id="app"></div>
     <noscript id="noscript">This app works best with JavaScript enabled.</noscript>
     <%= yield %>
+    <audio autoplay loop muted id="notif-audio">
+      <source src="<%= audio_path 'ding.mp3' %>" type="audio/mp3">
```
pranavrajs, line 54: Why can't we use the existing audio file?

pranavrajs (commented):
Except for the audio path, PR looks good. Can we add a setting to disable the sound notifications on UI, otherwise this would create a mess.

app/javascript/shared/helpers/AudioNotificationHelper.js
```diff
@@ -7,3 +7,31 @@ export const playNotificationAudio = () => {
...
+
+export const newMessageNotification = data => {
+  if (document.hidden) {
+    playAudioFromIframe();
+  } else {
+    const { conversation_id: currentPageId } = window.WOOT.$route.params;
+    const { conversation_id: incomingConvId } = data;
+    if (currentPageId !== incomingConvId) {
+      playAudioFromIframe();
+    }
```
pranavrajs, line 35: Do we need to do this? I guess we should only alert the user with the sound only if the window is out of focus? 

@sojan-official thoughts?

whande1992:
I also felt the need for sound notification, in addition to the visual one, as the agents cannot identify when there is a customer in line, a new conversation or message. I look forward to the appeal.

app/javascript/shared/helpers/AudioNotificationHelper.js
sojan-official: The ideal case is if there is no user interaction on the page for x minutes. I could be away from my desk to get my coffee and definitely the sound would help me not to miss out the conversation.

If we have the UI for disabling audio alert I would go for sound on all notifications even when the page is on focus

- Accounts with high volume conversations can have the sounds disabled.
- Accounts where volume is low, the notification would help me not to miss out on messages. 

We could take it as a second iteration to disable the notifications if the user is active on the page

app/views/layouts/vueapp.html.erb
nithindavid: Needs to be in assets folder to include in `.erb` files. @sojan-official Any idea to use the same file in `app/javascript/shared/assets/audio/ding.mp3` in `.erb` files?

sojan-official:
Let's get this merged with a setting to disable audio alerts in the profile.
cc: @pranavrajs @nithindavid

app/javascript/dashboard/routes/dashboard/settings/profile/NotificationSettings.vue
```diff
@@ -1,6 +1,32 @@
...
+        <p>
+          {{ $t('PROFILE_SETTINGS.FORM.AUDIO_NOTIFICATIONS_SECTION.NOTE') }}
+        </p>
+      </div>
+      <div class="columns small-9">
+        <div>
+          <input
+            id="audio_enable_alert"
+            v-model="enableAudioNotifications"
+            class="notification--checkbox"
```
pranavrajs, line 17: unused classname

app/views/layouts/vueapp.html.erb
pranavrajs: Please move it to the public folder in this case.

pranavrajs: Use an .ogg file not mp3

config/environments/production.rb
```diff
@@ -76,7 +76,7 @@
   # Use a different logger for distributed setups.
   # require 'syslog/logger'
   # config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new 'app-name')
-  
+
```
pranavrajs, line 79: Unnecessary change, please revert.

app/javascript/shared/helpers/AudioNotificationHelper.js
```diff
@@ -4,6 +4,44 @@ export const playNotificationAudio = () => {
...
+  } else {
+    const { conversation_id: currentPageId } = window.WOOT.$route.params;
+    const currentUserId = window.WOOT.$store.getters.getCurrentUserID;
+    const {
+      enable_audio_alerts: enableAudioNotifications = false,
+    } = window.WOOT.$store.getters.getUISettings;
+    const { conversation_id: incomingConvId, sender_id: senderId } = data;
+    const isFromCurrentUser = currentUserId === senderId;
+    const playAudio =
+      currentPageId !== incomingConvId &&
```
pranavrajs, line 39: `currentPageId` is confusing, please rename it to `currentConversationId`

app/javascript/shared/helpers/AudioNotificationHelper.js
```diff
@@ -4,6 +4,44 @@ export const playNotificationAudio = () => {
...
+};
+
+export const newMessageNotification = data => {
+  if (document.hidden) {
+    playAudioFromIframe();
+  } else {
+    const { conversation_id: currentPageId } = window.WOOT.$route.params;
+    const currentUserId = window.WOOT.$store.getters.getCurrentUserID;
+    const {
+      enable_audio_alerts: enableAudioNotifications = false,
```
pranavrajs, line 34: `enableAudioAlerts`, please be consistent with the names

app/javascript/dashboard/routes/dashboard/settings/profile/NotificationSettings.vue
```diff
@@ -277,6 +310,12 @@ export default {
 
       this.updateNotificationSettings();
     },
+    handleAudioInput(e) {
+      this.enableAudioNotifications = e.target.checked;
```
pranavrajs, line 314: Please use either `enableAudioAlert` or `enableAudioNotification`, don't mix both.

pranavrajs (changes requested):
Please see the inline comments.

app/javascript/shared/helpers/AudioNotificationHelper.js
```diff
@@ -4,6 +4,44 @@ export const playNotificationAudio = () => {
...
+    } = window.WOOT.$store.getters.getUISettings;
+    const { conversation_id: incomingConvId, sender_id: senderId } = data;
+    const isFromCurrentUser = currentUserId === senderId;
+    const playAudio =
+      currentPageId !== incomingConvId &&
+      !isFromCurrentUser &&
+      enableAudioNotifications;
```
pranavrajs, line 41: I think the audio alert should only be sent for incoming messages, if a teammate replies to the message why should we alert another user?

@sojan-official Thoughts

pranavrajs:
A couple of other issues: 

1. There is no separation between the sections

Screenshot 2021-02-28 at 8 14 00 PM

2. Clicking on "Play audio notification" does not give me feedback. While other checkboxes show an alert that the settings have been updated.

3. Sound does not work in Firefox.

app/javascript/dashboard/routes/dashboard/settings/profile/NotificationSettings.vue
nithindavid: Already in use classname

nithindavid:
> While other checkboxes show an alert that the settings have been updated, clicking on "Play audio notification" does not give me feedback.

The Update UI doesn't have promise support, Which is out of scope of this issue. Will fix in a new PR since couple of other parts also needs to be updated.

sojan-official:
- Go to settings page and enable audio alert
- go to conversations page and refresh the page 
- Go back to the settings page and the audio alert is unchecked ( refreshing the pages brings back the checkbox as checked )

Screenshot 2021-03-02 at 8 20 52 PM

nithindavid:
> * Go to settings page and enable audio alert
> * go to conversations page and refresh the page
> * Go back to the settings page and the audio alert is unchecked ( refreshing the pages brings back the checkbox as checked )
> 
> Screenshot 2021-03-02 at 8 20 52 PM

Fixed

sojan-official (approved):

PR #1743, opened 2021-02-08
fix: Fix bot message rendering in dashboard

pranavrajs:
Fix input_email, input_select, form message rendering in dashboard.

### DIFF
--- a/app/javascript/dashboard/components/widgets/conversation/Message.vue
+++ b/app/javascript/dashboard/components/widgets/conversation/Message.vue
@@ -72,7 +72,7 @@ import Spinner from 'shared/components/Spinner';
 import contentTypeMixin from 'shared/mixins/contentTypeMixin';
 import BubbleActions from './bubble/Actions';
 import { MESSAGE_TYPE, MESSAGE_STATUS } from 'shared/constants/messages';
-
+import { generateBotMessageContent } from './helpers/botMessageContentHelper';
 export default {
   components: {
     BubbleActions,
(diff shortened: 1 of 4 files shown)


app/javascript/dashboard/components/widgets/conversation/Message.vue
```diff
@@ -99,7 +99,45 @@ export default {
   },
   computed: {
     message() {
-      return this.formatMessage(this.data.content, this.isATweet);
+      let messageContent = this.formatMessage(this.data.content, this.isATweet);
```
nithindavid, line 102: Move to a helper maybe since it's long and testable if separated?

app/javascript/dashboard/components/widgets/conversation/Message.vue
```diff
@@ -99,7 +99,45 @@ export default {
...
+      if (this.contentType === 'input_select') {
+        const {
+          submitted_values: submittedValues = [],
+        } = this.contentAttributes;
+        const [selectedOption] = submittedValues;
+
+        if (selectedOption && selectedOption.title) {
+          messageContent += `<strong>${selectedOption.title}</strong>`;
```
nithindavid, line 110: If this parsing is before, can we use markdown instead of HTML ?

app/javascript/dashboard/components/widgets/conversation/Message.vue
```diff
@@ -99,7 +99,45 @@ export default {
...
+        }
+      } else if (this.contentType === 'form') {
+        const {
+          items,
+          submitted_values: submittedValues = [],
+        } = this.contentAttributes;
+        if (submittedValues.length) {
+          const submittedObject = submittedValues.reduce(
+            (acc, keyValuePair) => {
+              acc = { ...acc, [keyValuePair.name]: keyValuePair.value };
```
nithindavid, line 125: Just acc[x] = computed_value ? Is enough right ?

nithindavid (changes requested):

app/javascript/dashboard/components/widgets/conversation/Message.vue
pranavrajs: The content becomes unreadable with all the \n and spaces. I would prefer to go with this.

nithindavid (approved):

PR #1749, opened 2021-02-10
feat: Use audio tag instead of download file button

### LINKED ISSUE #1458: Audio player for audio message
If the conversation is happening in Whatsapp, people tend to use audio messages along with text messages. The current interface shows a download button for all file types except the image files. This makes it harder for the support agents to hear the content as they have to download it first and play it locally. 

### Feature

Identify a good library to support inline audio player and play audio inside the app itself.

pranavrajs:
Screenshot_2021-02-11 Chatwoot(1)

Looks a little bit ugly. Since this would get the product to be useful, I'm inclined to this approach.

@sojan-official @nithindavid Any thoughts?

Fixes #1458

### DIFF
--- a/app/helpers/file_type_helper.rb
+++ b/app/helpers/file_type_helper.rb
@@ -9,6 +9,8 @@ module FileTypeHelper
       'image/bmp'
     ].include?(content_type)
 
+    return :audio if content_type.include?('audio/')
+
     :file
   end
 end
--- a/app/javascript/dashboard/components/widgets/conversation/Message.vue
+++ b/app/javascript/dashboard/components/widgets/conversation/Message.vue
@@ -21,8 +21,11 @@
               :url="attachment.data_url"
               :readable-time="readableTime"
             />
+            <audio v-else-if="attachment.file_type === 'audio'" controls>
+              <source :src="attachment.data_url" />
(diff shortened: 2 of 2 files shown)


sojan-official (approved):
We can go with this and do another iteration to improve the audio UI.

nithindavid (approved):

nithindavid:
👍

WevrLabs:
Doesn't it recognize `.ogg` files?

Screen Shot 2021-02-11 at 5 38 18 PM

pranavrajs:
@WevrLabs It should where is this coming from? Whatsapp?

WevrLabs:
> @WevrLabs It should where is this coming from? Whatsapp?

Correct, a voice note from WhatsApp (tested sending an `mp3` from agent to the user, and it plays as should, but I guess tho voice notes from WhatsApp are received in `.ogg` format).

pranavrajs:
Can you please inspect what is the type of the file returned in the API?

WevrLabs:
> Can you please inspect what is the type of the file returned in the API?

I looked inside twilio console logs, and saw the media file, I guess it's `.ogg` extension.

PR #1759, opened 2021-02-13
fix: Update the message id generation logic

### LINKED ISSUE #1450: Message-ID potentially invalid, per RFC2822
**Describe the bug**

Message-ID in mail on self-hosted sent from CW seems invalid. i.e

```Message-ID: <conversation/xxx-xxxx-xxxx-xxx/messages/xxx@>```

There should be hostname or domain after the @

Both SMTP_ADDRESS and SMTP_DOMAIN are set in the .env file so not sure where else this would need to be set or if this is just an oversight?

I would however say that using the same domain as the sender address (if set per Inbox) would be best.

### PR DESCRIPTION
**Pull request by pranavrajs**
Fixes #1450

### CODE REVIEW

**Comment by pranavrajs** [comment]
@sojan-official Can you please take a look at this fix?

**Review by sony-mathew — APPROVED** [review]
LGTM.

**Comment by sojan-official** [comment]
Our current message IDs should be valid 
ref: https://stackoverflow.com/a/594393/939299

allowed characters as per RFC2822.
```
::alpha::|::digit::|[!#$%&'*+-/=?^_`{}|~.]
```

PR #1760, opened 2021-02-13
feat: Assign conversations to team

nithindavid:
Fixes part of #866 

## Todo
- [ ] append assignee type on option value, since team and agent can have same id


nithindavid:
Duplicate #1849

PR #1765, opened 2021-02-14
Include website_token key in the inboxes response

### LINKED ISSUE #1764: Send website_token in the /inboxes response
There is a general requirement of wanting to have the website_token when an inbox for a website is created. The current API sends it as a part of script string of the JSON response.

### **The key is part of the swagger doc but not in the actual response.**

website_token key in the JSON response

N/A

**Additional context**
It's a real simple PR to submit

srameshr:
Many self-hosted instances require this key as a part of QA & maintenance. Since this key is not sensitive its better to send it as a part of the JSON response instead of having to use a regex to get it from the widget_script_tag

# Pull Request Template

## Description

Many self-hosted instances require this key as a part of QA & maintenance. Since this key is not sensitive its better to send it as a part of the JSON response instead of having to use a regex to get it from the widget_script_tag

Fixes #1764
https://github.com/chatwoot/chatwoot/issues/1764

## Type of change

## How Has This Been Tested?

Ran the relevant tests.

## Checklist:

### DIFF
--- a/app/views/api/v1/models/_inbox.json.jbuilder
+++ b/app/views/api/v1/models/_inbox.json.jbuilder
@@ -12,6 +12,7 @@ json.welcome_title resource.channel.try(:welcome_title)
 json.welcome_tagline resource.channel.try(:welcome_tagline)
 json.enable_auto_assignment resource.enable_auto_assignment
 json.web_widget_script resource.channel.try(:web_widget_script)
+json.website_token resource.channel.try(:website_token)
 json.forward_to_address resource.channel.try(:forward_to_address)
 json.phone_number resource.channel.try(:phone_number)
 json.selected_feature_flags resource.channel.try(:selected_feature_flags)


pranavrajs (approved):

pranavrajs:
@srameshr Thanks for the PR. Can you please share the exact use case you are solving by adding this key?

srameshr:
@pranavrajs Our self hosted Chatwoot instance for our org needs the website_token key for some upkeep and maintenance. We create an inbox from the API and sometimes from the dash and in either case the key is missing.

srameshr:
@pranavrajs Also, is that key used to identify the website? Cant we just use window.host in the widget_script instead of this key?

pranavrajs:
@srameshr Yes, website token is the identifier. It cannot be done with the URL since we are not enforcing a unique URL on website inboxes.

srameshr:
@pranavrajs But that should be the right behavior on DB right? You can't or should not have two chat widgets on one domain + subdomain combination.

PR #1766, opened 2021-02-14
fix: Remove teams on account user destroy

pranavrajs:
- Remove user from teams on removing from accounts.

### DIFF
--- a/app/services/agents/destroy_service.rb
+++ b/app/services/agents/destroy_service.rb
@@ -1,0 +1,28 @@
+class Agents::DestroyService
+  pattr_initialize [:account!, :user!]
+
+  def perform
+    destroy_notification_setting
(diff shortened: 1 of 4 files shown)


sojan-official (approved):

app/services/agents/destroy_service.rb
```diff
@@ -0,0 +1,28 @@
+class Agents::DestroyService
+  pattr_initialize [:account!, :user!]
+
+  def perform
```
sojan-official, line 4: should we wrap this up in a transaction to ensure we don't half deleted states ?

pranavrajs: Makes sense, will change it tomorrow.

sojan-official (approved):

PR #1806, opened 2021-02-19
Feat: Create contact from contacts page

nithindavid:
- Adds create contact form in a model

image

### DIFF
--- a/app/javascript/dashboard/routes/dashboard/conversation/contact/EditContact.vue
+++ b/app/javascript/dashboard/routes/dashboard/conversation/contact/EditContact.vue
@@ -11,7 +11,7 @@
         :contact="contact"
         :in-progress="uiFlags.isUpdating"
         :on-submit="onSubmit"
-        @success="onCancel"
+        @success="onSuccess"
         @cancel="onCancel"
       />
     </div>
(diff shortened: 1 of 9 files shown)


app/javascript/dashboard/routes/dashboard/conversation/contact/EditContact.vue
```diff
@@ -11,6 +11,8 @@
         :contact="contact"
         :in-progress="uiFlags.isUpdating"
         :on-submit="onSubmit"
+        @success="onCancel"
```
pranavrajs, line 14: `onSuccess`

pranavrajs (changes requested):
Please see the inline comments.

app/javascript/dashboard/routes/dashboard/conversation/contact/EditContact.vue
nithindavid: I prefer not to use `on` before emitted events. as `@success` is already short for `v-on:success`. @pranavrajs What do you think ?

pranavrajs: I meant that the callback is wrong. `@success` is fine.

nithindavid: The idea was to call modal closing function after success, I'll create another function

nithindavid: Made the changes

pranavrajs (approved):

PR #1823, opened 2021-02-23
Feat: Show working hours on widget

### LINKED ISSUE #234: As a business owner, I would like to set business hours on the widget
- [ ] Create UI to accept business active hours from the UI
- [ ] Create a DB field to store the business hours
- [ ] Create an API to update the business hours.
- [ ] Provide an indication in widget if current time is outside of the business hours.

nithindavid:
Fixes #234 
image


db/schema.rb
```diff
@@ -49,6 +49,7 @@
     t.string "support_email", limit: 100
     t.integer "settings_flags", default: 0, null: false
     t.integer "feature_flags", default: 0, null: false
+    t.string "timezone", default: "UTC"
```
sojan-official, line 52: remove this change
```suggestion
```

lib/current.rb
```diff
@@ -3,6 +3,11 @@ module Current
   thread_mattr_accessor :account
   thread_mattr_accessor :account_user
 
+  def account=(account)
+    super
+    Time.zone = account.timezone
+  end
```
sojan-official, line 9: remove this changes

```suggestion

```

db/migrate/20201207062430_change_working_hours_to_zero.rb
```diff
@@ -0,0 +1,11 @@
+class ChangeWorkingHoursToZero < ActiveRecord::Migration[6.0]
```
sojan-official, line 1: remove this file. we already have this migration.

sojan-official (changes requested):
some comments

PR #1856, opened 2021-03-05
Chore: Responsive fixes for common screen widths

nithindavid:
Fixes congestion because #1835 

Here's how it'll look on different screen sizes (title as pixels)
**1040**

**1280**

**1368**

**1440**

**1740**


nithindavid (commented):
👍

pranavrajs (approved):

PR #1859, opened 2021-03-05
Fix: apply alert conditions while document is hidden

nithindavid:
**Why**
- Notifications were playing while I was away from chatwoot tab, irrespective of the condition.
- Play for private notes from others


pranavrajs (dismissed):
Looks good! Please write a test case for `shouldPlayAudio`

Issue #1647, opened 2021-01-13
SMTP related environment variables not used in production.rb


milansimek:
Several variables present in the `.env` file are not being used in `config/environments/production.rb`

See `.env.example`:
https://github.com/chatwoot/chatwoot/blob/develop/.env.example#L51

`SMTP_AUTHENTICATION` and `SMTP_ENABLE_STARTTLS_AUTO` are configurable in this file.

However, in `production.rb` these values are hard-coded:
https://github.com/chatwoot/chatwoot/blob/develop/config/environments/production.rb#L108

Besides these two variables, it would be good to also have a `SSL_VERIFY_MODE` env variable.

Shall I create a pull request for this?

sojan-official:
@milansimek  thanks for spotting that. feel free to go ahead.

PR #1868, opened 2021-03-07
chore: Make SMTP environment variables configurable

sojan-official:
fixes: #1647

### DIFF
--- a/app/mailers/conversation_reply_mailer.rb
+++ b/app/mailers/conversation_reply_mailer.rb
@@ -1,5 +1,5 @@
 class ConversationReplyMailer < ApplicationMailer
-  default from: ENV.fetch('MAILER_SENDER_EMAIL', 'accounts@chatwoot.com')
+  default from: ENV.fetch('MAILER_SENDER_EMAIL', 'Chatwoot <accounts@chatwoot.com>')
   layout :choose_layout
 
   def reply_with_summary(conversation, message_queued_time)
@@ -116,7 +116,7 @@ class ConversationReplyMailer < ApplicationMailer
   def from_email_address
     return @inbox.email_address if @inbox.email_address
 
-    ENV.fetch('MAILER_SENDER_EMAIL', 'accounts@chatwoot.com')
+    account_support_email
   end
 
   def custom_message_id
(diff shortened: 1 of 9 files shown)


config/initializers/mailer.rb
```diff
@@ -0,0 +1,48 @@
...
+  config.action_mailer.raise_delivery_errors = true
+
+  # Config related to smtp
+  smtp_settings = {
+    domain: ENV.fetch('SMTP_DOMAIN', ''),
+    address: ENV.fetch('SMTP_ADDRESS', 'localhost'),
+    port: ENV.fetch('SMTP_PORT', 587)
+  }
+
+  if ENV.fetch('SMTP_AUTHENTICATION', '').present?
```
pranavrajs, line 21: Would this break the exisiting apps where they have not set SMTP_AUTHENTICATION? Also why is this under a condition?

sojan-official: it only makes sense to pass authentication-related parameters to hash only when an authentication type is specified via `SMTP_AUTHENTICATION`.

in the current documentation even though we ask people to set this environment variable, the value for this is hardcoded in the production config. That's the reason why the SMTP server which has a different authentication mode than `login` isn't working

app/mailers/conversation_reply_mailer.rb
```diff
@@ -105,15 +105,15 @@ def reply_email
     end
   end
 
-  def from_email_with_name
+  def from_email
     if should_use_conversation_email_address?
-      "#{assignee_name} <#{account_support_email}>"
```
pranavrajs, line 110: Why is the assignee_name removed?

app/mailers/conversation_reply_mailer.rb
```diff
@@ -146,11 +146,7 @@ def current_domain
   end
 
   def account_support_email
```
pranavrajs, line 148: do we need this method?

config/initializers/mailer.rb
```diff
@@ -0,0 +1,48 @@
...
+    smtp_settings[:user_name] = ENV.fetch('SMTP_USERNAME', '')
+    smtp_settings[:password] = ENV.fetch('SMTP_PASSWORD', '')
+    smtp_settings[:authentication] = ENV.fetch('SMTP_AUTHENTICATION', 'login')
```
pranavrajs, line 24: I guess this should be a symbol.

PR #1873, opened 2021-03-08
Fix: emoji box not showing up

nithindavid:
**Why**
Emoji box getting overlayed by recent changes to move header up the conversation box


app/javascript/dashboard/components/widgets/conversation/ConversationBox.vue
```diff
@@ -80,9 +80,7 @@ export default {
   display: flex;
   background: var(--color-background-light);
   margin: 0;
-  height: auto;
-  flex: 1 1;
-  overflow: hidden;
+  height: calc(100vh - 63px);
```
nithindavid, line 83: 63px is Fixed Height of header

PR #1914, opened 2021-03-15
Fix: Hide prechat for sessions inititated with setUser

### LINKED ISSUE #1894: Pre-chat form asks for name and email even when these are set in SDK
Hi there,
It appears that the pre-chat form still asks the user for their name and email address, even though these may have already been set via the SDK.

Name and Email already set in SDK:
Screen Shot 2021-03-10 at 8 43 26 AM

Pre-chat form should not ask for name and email if set in SDK:
Screen Shot 2021-03-10 at 8 53 07 AM

nithindavid:
Fixes #1894


app/javascript/widget/store/modules/contacts.js
```diff
@@ -1,15 +1,36 @@
 import ContactsAPI from '../../api/contacts';
 import { refreshActionCableConnector } from '../../helpers/actionCable';
 
+const state = {
+  currentUser: {},
```
pranavrajs, line 5: What happens if the user reloads the page, `setUser` would be called only once if there is no change in the data.

app/controllers/api/v1/widget/contacts_controller.rb
```diff
@@ -1,4 +1,9 @@
 class Api::V1::Widget::ContactsController < Api::V1::Widget::BaseController
+  def index
+    process_hmac
+    render json: @contact
```
pranavrajs, line 4: please pass only email and name

app/controllers/api/v1/widget/contacts_controller.rb
```diff
@@ -1,4 +1,9 @@
 class Api::V1::Widget::ContactsController < Api::V1::Widget::BaseController
+  def index
+    process_hmac
```
pranavrajs, line 3: please move this as 

```
process_hmac before_action: [:index, :update]
```

config/routes.rb
```diff
@@ -150,7 +150,7 @@
             post :transcript
           end
         end
-        resource :contact, only: [:update]
+        resource :contact, only: [:update, :index]
```
pranavrajs, line 153: Please order maintain the order index, update

app/javascript/widget/store/modules/contacts.js
```diff
@@ -1,27 +1,58 @@
...
+};
+
+const SET_CURRENT_USER = 'SET_CURRENT_USER';
+const SET_CONTACTS_UI_FLAG = 'SET_CONTACTS_UI_FLAG';
+
+export const getters = {
+  getCurrentUser(_state) {
+    return _state.currentUser;
+  },
+  getUIFlags(_state) {
```
pranavrajs, line 18: UI flags is unused, please remove it.

app/javascript/widget/store/modules/contacts.js
nithindavid: Added a case to fetch from API

PR #1987, opened 2021-03-25
fix: Handle rendering HTML only email in UI

pranavrajs:
Render HTML only emails in UI


app/javascript/dashboard/components/widgets/conversation/Message.vue
```diff
@@ -107,11 +107,22 @@ export default {
         this.contentAttributes,
         this.$t('CONVERSATION.NO_RESPONSE')
       );
-      let messageContent =
-        this.formatMessage(this.data.content, this.isATweet) +
-        botMessageContent;
+      if (
+        this.data?.content_attributes?.email?.html_content?.full &&
```
nithindavid, line 111: Too big to understand. Can we separate into a variable ?

nithindavid: The entire statement inside if into a variable, so it's readable

app/javascript/dashboard/components/widgets/conversation/Message.vue
```diff
@@ -107,11 +107,22 @@ export default {
...
+        this.data.message_type === 0
```
nithindavid, line 112: Use already existing constant for `status` ?

app/javascript/dashboard/components/widgets/conversation/Message.vue
```diff
@@ -107,11 +107,22 @@ export default {
...
+      ) {
+        let parsedContent = new DOMParser().parseFromString(
+          this.data?.content_attributes?.email?.html_content?.full || '',
```
nithindavid, line 115: We can move this into a variable since it's reused

nithindavid (changes requested):
Please see the individual comments

app/javascript/dashboard/components/widgets/conversation/Message.vue
nithindavid: Please try to avoid the `snake_key` usage

nithindavid (approved):

PR #2012, opened 2021-03-29
chore: Fix Current.user nil issue when using access tokens

sojan-official:
Fix Current.user nil issue when using access tokens


pranavrajs (approved):
LGTM

PR #2025, opened 2021-03-31
fix: Use reply content if available

pranavrajs:
Instead of rendering the entire thread, display only the replies in email channel


pranavrajs:
@sojan-official Let us get this merged, it wouldn't break anything. There is still an issue with the quoted text, we can take that separately.

sojan-official (approved):

PR #2047, opened 2021-04-03
Fixes: Agent name overflow in dropdown menu 

### LINKED ISSUE #2039: Bug: Text overflow in assign agent dropdown.
image

iamsivin:
Fixes #2039 

**After**

Screenshot 2021-04-08 at 1 52 03 PM


app/javascript/dashboard/assets/scss/plugins/_multiselect.scss
```diff
@@ -37,6 +37,14 @@
     font-size: $font-size-small;
     font-weight: $font-weight-normal;
 
+    span {
+      display: inline-block;
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+      width: 12rem;
```
nithindavid, line 45: Can we try without fixed width?

nithindavid (changes requested):
We need to merge #2027 first, then we will see if we have a new workaround.

nithindavid (approved):

PR #2060, opened 2021-04-06
Fix: Populate only agents part of the inbox in conversations

nithindavid:
Why?
agents who doesn't belong inbox also gets listed in assign.


muhsin-k:
@nithindavid The code looks good to me. The only small UI suggestion is loader is not vertically centred.
Screenshot 2021-04-06 at 7 53 29 PM

nithindavid:
> @nithindavid The code looks good to me. The only small UI suggestion is loader is not vertically centred.
> Screenshot 2021-04-06 at 7 53 29 PM

This is fixed now as the dropdown height has changed.

muhsin-k (approved):

PR #2061, opened 2021-04-06
Fix: clear search input

### LINKED ISSUE #1955: Bug: Clear search input when page changes
image

niklashaug:
# Pull Request Template

## Description

Search input is now cleared when the user either toggles the assignee tab or when he applies filters (inbox or labels).

Fixes #1955

## Type of change

## How Has This Been Tested?

Tested on local machine.


nithindavid (approved):
Thanks for the PR @niklashaug  🎉 I'll merge it after testing on my local.

app/javascript/dashboard/routes/dashboard/conversation/search/PopOverSearch.vue
```diff
@@ -124,6 +133,9 @@ export default {
     closeSearch() {
       this.showSearchBox = false;
     },
+    clearSearchTerm() {
+      this.searchTerm = null;
```
pranavrajs, line 137: I would suggest making it an empty string rather than null.

app/javascript/dashboard/routes/dashboard/conversation/search/PopOverSearch.vue
```diff
@@ -83,6 +83,9 @@ export default {
       conversations: 'conversationSearch/getConversations',
       uiFlags: 'conversationSearch/getUIFlags',
     }),
+    currentPage() {
+      return this.$store.state.conversationPage.currentPage;
```
pranavrajs, line 87: I guess this should be a getter defined in the state.

pranavrajs (changes requested):
Nice work! Thanks for the PR. Please take a look at the individual comments.

cc: @nithindavid

app/javascript/dashboard/routes/dashboard/conversation/search/PopOverSearch.vue
niklashaug: There is already a getter 'conversationPage/getCurrentPage' but it requires an argument (the current assignee tab). Should I add another getter that doesn't require the argument to conversationPage.js?

nithindavid: Yes, Can you add another getter that doesn't require the argument to conversationPage.js?

niklashaug: @nithindavid sure. How should I name this getter without an argument then? Or should I just enhance the existing getter and check whether the argument has been provided or not?

nithindavid: If you could refactor the current `getCurrentPage` function to a new more sensible name would be the best, as there is only one usage so far.

niklashaug: So I should give `getCurrentPage` a new name such as `getCurrentPageFilter` or `getCurrentFilteredPage` and then add `getCurrentPage` that just returns that object without any filter applied?

nithindavid: 'getCurrentPageFilter' and 'getCurrentPageWithFilter' sounds good

niklashaug: To make sure I got it right:
I'm going to rename `getCurrentPage` to either `getCurrentPageFilter` or `getCurrentPageWithFilter` and create a function `getCurrentPage` that returns the object without having to provide any filter as arguments?

nithindavid: Yes 👍

muhsin-k:
**QA**

- Search text gets cleared when the user toggles mine, unassigned and all tabs
- Search text gets cleared  when the user clicks on labels
- Search text gets cleared  when the user clicks on the Inbox

https://user-images.githubusercontent.com/12408980/115671795-bc090500-a2ff-11eb-8438-1faec0de5dd7.mov

@pranavrajs Everything is working as expected.

pranavrajs:
@muhsin-k Thanks! Would you be please test one more case? 

- Open the search with some text
- Create a new conversation from the widget

Can you please verify whether the search input text stays as it is?

muhsin-k:
@pranavrajs Search input text doesn't get cleared on a new conversation
https://user-images.githubusercontent.com/12408980/115676194-33409800-a304-11eb-9cdc-3bb94c5ec1bb.mov

pranavrajs (approved):

pranavrajs:
Aw! Sorry, I messed up the commit message. :(

PR #2086, opened 2021-04-12
Responsive issue in activity message

iamsivin:
Fixes responsive issue in activity message.


app/javascript/dashboard/components/widgets/conversation/Message.vue
```diff
@@ -143,6 +143,9 @@ export default {
       return `https://twitter.com/${screenName}`;
     },
     alignBubble() {
+      if (this.data.message_type === 2) {
```
nithindavid, line 146: Instead of using `2`, we have constants at `import { MESSAGE_TYPE } from 'shared/constants/messages';`. Please use them.

nithindavid (changes requested):
Please see the comments

PR #2101, opened 2021-04-14
feat: Contactable Inboxes API

### LINKED ISSUE #1678: Cannot save additional attributes via conversations API
**Describe the bug**

https://github.com/chatwoot/chatwoot/blob/develop/app/controllers/api/v1/accounts/conversations_controller.rb#L90 sets `additional_attributes` to `params[:additional_attributes]`, however `params[:additional_attributes]` is a `ActionController::Parameters` and not a `Hash`, hence it cannot pass the validation in https://github.com/chatwoot/chatwoot/blob/develop/app/models/conversation.rb#L145.

Screenshot 2021-01-21 at 12 27 58 PM

Hence `additional_attributes` will always be saved as an empty hash.

**To Reproduce**

1. Make a post request to create a conversation with following json body:
```json
{
  "source_id": "<Source id>",
  "additional_attributes": {
    "comment_id": 1
  }
}
```
2. Created conversation's `additional_attributes` is an empty object

**Expected behavior**

Conversation's `additional_attributes` should be saved and returned successfully.

**Additional context**

https://github.com/chatwoot/chatwoot/blob/develop/spec/controllers/api/v1/accounts/conversations_controller_spec.rb#L138-L148 should add a test case where `params` contains `additional_attributes` hash and make sure that it's returned in the API response.

```ruby
    it 'creates a new conversation with additional attributes' do
        allow(Rails.configuration.dispatcher).to receive(:dispatch)
        post "/api/v1/accounts/#{account.id}/conversations",
             headers: agent.create_new_auth_token,
             params: { source_id: contact_inbox.source_id, additional_attributes: { comment_id: 1 } },
             as: :json

        response_data = JSON.parse(response.body, symbolize_names: true)
        expect(response_data[:additional_attributes]).to eq({ comment_id: 1 })
      end
```

sojan-official:
- endpoint which lists inboxes through which a contact can be contacted
- conversation creation API auto-creates contact_inbox for specific channels [ Twilio, email, api]
- Ability to send the initial message payload along with the conversation creation
- Fixes #1678 ( issue saving additional attributes for conversation )


app/views/api/v1/accounts/contacts/contactable_inboxes.json.jbuilder
```diff
@@ -0,0 +1,6 @@
+json.payload do
+  json.array! @contactable_inboxes do |contactable_inbox|
+    json.inbox contactable_inbox[:inbox]
```
pranavrajs, line 3: Please use inbox.json.jbuilder

pranavrajs (changes requested):
Looks good overall! Please use jbuilder for inbox

nithindavid (approved):

PR #2153, opened 2021-04-23
Enhancement: Dropdown component for labels.

iamsivin:
**Describe the changes**

Created a dropdown component with a search input for labels.

Screenshot 2021-04-23 at 5 34 30 PM

https://user-images.githubusercontent.com/64252451/115868552-436a8b80-a45a-11eb-9a76-58c0e9ac9125.mov

### DIFF
--- a/app/javascript/dashboard/routes/dashboard/conversation/ContactPanel.vue
+++ b/app/javascript/dashboard/routes/dashboard/conversation/ContactPanel.vue
@@ -53,6 +53,16 @@
         </multiselect>
       </div>
     </div>
+    <contact-custom-attributes
+      v-if="hasContactAttributes"
+      :custom-attributes="contact.custom_attributes"
+    />
+    <conversation-labels :conversation-id="conversationId" />
+    <contact-conversations
+      v-if="contact.id"
+      :contact-id="contact.id"
+      :conversation-id="conversationId"
+    />
     <div v-if="browser.browser_name" class="conversation--details">
       <contact-details-item
         v-if="location"
@@ -101,16 +111,6 @@
         emoji="🕰"
       />
     </div>
-    <contact-custom-attributes
-      v-if="hasContactAttributes"
-      :custom-attributes="contact.custom_attributes"
-    />
-    <conversation-labels :conversation-id="conversationId" />
-    <contact-conversations
-      v-if="contact.id"
-      :contact-id="contact.id"
-      :conversation-id="conversationId"
-    />
   </div>
 </template>
 
@@ -296,14 +296,14 @@ export default {
 .contact--panel {
   background: white;
   border-left: 1px solid var(--color-border);
-  font-size: $font-size-small;
+  font-size: var(--font-size-small);
   overflow-y: auto;
   overflow: auto;
   position: relative;
-  padding: $space-one;
+  padding: var(--space-one);
 
   i {
-    margin-right: $space-smaller;
+    margin-right: var(--space-smaller);
   }
 }
 
@@ -317,22 +317,22 @@ export default {
 
 .close-button {
   position: absolute;
-  right: $space-normal;
-  top: $space-slab;
-  font-size: $font-size-default;
-  color: $color-heading;
+  right: var(--space-normal);
+  top: var(--space-slab);
+  font-size: var(--font-size-default);
+  color: var(--color-heading);
 }
 
 .conversation--details {
-  padding: 0 var(--space-slab);
+  padding: var(--space-two) var(--space-slab) 0 var(--space-slab);
 }
 
 .conversation--labels {
-  padding: $space-medium;
+  padding: var(--space-medium);
 
   .icon {
-    margin-right: $space-micro;
-    font-size: $font-size-micro;
+    margin-right: var(--space-micro);
+    font-size: var(--font-size-micro);
     color: #fff;
   }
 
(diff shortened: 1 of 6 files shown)


app/javascript/dashboard/routes/dashboard/conversation/labels/LabelBox.vue
```diff
@@ -2,93 +2,117 @@
...
     <spinner v-else></spinner>
   </div>
 </template>
 
 <script>
 import { mapGetters } from 'vuex';
-import AddLabelToConversation from './AddLabelToConversation';
 import ContactDetailsItem from '../ContactDetailsItem';
 import Spinner from 'shared/components/Spinner';
+import LabelDropdown from 'shared/components/ui/LabelDropdown.vue';
```
nithindavid, line 55: We can avoid `.vue` in imports. Please do this for other changes also.

app/javascript/dashboard/routes/dashboard/conversation/labels/LabelBox.vue
```diff
@@ -100,19 +124,32 @@ export default {
...
+
+    toggleLabels() {
+      this.showSearchDropdownLabel = !this.showSearchDropdownLabel;
     },
-    closeEditModal() {
-      bus.$emit('fetch_conversation_stats');
-      this.isEditing = false;
+
+    closeDropdownLabel() {
+      this.showSearchDropdownLabel = false;
```
nithindavid, line 133: Can we rename this to `showSearchDropdownLabel` to `showLabelSelector` or something like that ?

nithindavid (changes requested):
Great PR 🎉
I have left some minor suggestions and also there are couple of places where we can use css variables like for font-weight and border-radius. Please change them.

app/javascript/dashboard/routes/dashboard/conversation/ContactPanel.vue
```diff
@@ -53,6 +53,16 @@
         </multiselect>
       </div>
     </div>
+    <contact-custom-attributes
+      v-if="hasContactAttributes"
+      :custom-attributes="contact.custom_attributes"
+    />
+    <conversation-labels :conversation-id="conversationId" />
+    <contact-conversations
```
pranavrajs, line 61: This can grow to a long list, please keep it at the end.

app/javascript/shared/components/ui/LabelDropdown.vue
```diff
@@ -0,0 +1,212 @@
...
+                    class="label-color--display"
+                    :style="{ backgroundColor: label.color }"
+                  />
+                  <span>{{ label.title }}</span>
+                </div>
+                <i
+                  v-if="selectedLabels.includes(label.title)"
+                  class="icon ion-checkmark-round"
+                />
+              </button>
```
pranavrajs, line 37: I guess this can be a label component with 2 variants.

1. Label which shows the text.
2. Label which shows the text and the close button.

app/javascript/shared/components/ui/LabelDropdown.vue
```diff
@@ -0,0 +1,212 @@
...
+      {{ $t('CONTACT_PANEL.LABELS.LABEL_SELECT.TITLE') }}
+    </h4>
+    <div class="search-wrap">
+      <input
+        ref="searchbar"
+        v-model="search"
+        type="text"
+        class="search-input"
+        autofocus="true"
+        placeholder="Filter"
```
pranavrajs, line 13: Please use i18n.

app/javascript/shared/components/ui/LabelDropdown.vue
```diff
@@ -0,0 +1,212 @@
...
+      type: Function,
+      default: () => {},
+    },
+    onAdd: {
+      type: Function,
+      default: () => {},
+    },
+    onRemove: {
+      type: Function,
+      default: () => {},
```
pranavrajs, line 81: The best here is to use emit events from the component, instead of passing down the method.

pranavrajs:
@iamsivin Great work on the PR. This is amazing. ❤️ 

One observation I have is that there is a component that shows the modal to select labels (previous UX). I don't see that getting removed in this PR. There will be i18n text associated with the previous UX that also needs to be removed.

app/javascript/shared/components/ui/LabelDropdown.vue
pranavrajs: Now that the storybook is merged, it will be a great place to start building components.

app/javascript/dashboard/routes/dashboard/conversation/ContactPanel.vue
nithindavid: I asked @iamsivin to move this up because 
1. I think tag is more of a ticket property which is often looked/scanned unlike the details like Timezone/browser etc.
2. The agent might add Labels on the go. So it should stay with Agent/Team/Labels which makes more sense because of similarity and ease off access. Also this won't grow more than 3 lines usually. 

Let me know if this makes sense.

app/javascript/shared/components/ui/LabelDropdown.vue
iamsivin: Yeah sure.

app/javascript/dashboard/routes/dashboard/conversation/ContactPanel.vue
pranavrajs: This makes sense. I agree with this. 

My point was around contact-conversations, not labels. The number of conversations can grow. I believe the metadata should be shown above conversations.

PR #2163, opened 2021-04-27
chore: Upgrade yarn packages

nithindavid:
- Resolves package versions


pranavrajs:
@nithindavid Shall we close this PR? This has a lot of conflicts and I don't think this is required anymore as we merged the storybook PR.

PR #2175, opened 2021-04-28
feat:  Campaign APIs

sojan-official:
## Changes

- Add models for campaigns 
- Add API end points for campaigns ( index, create, show, update)
- Add API end points for inbox campaigns ( index )


app/models/campaign.rb
```diff
@@ -0,0 +1,49 @@
...
+#  description   :text
+#  enabled       :boolean          default(TRUE)
+#  title         :string           not null
+#  trigger_rules :jsonb
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#  account_id    :bigint           not null
+#  display_id    :integer          not null
+#  inbox_id      :bigint           not null
+#  sender_id     :integer
```
pranavrajs, line 16: Should this be owner or something else? sender seems to be a confusing name.

pranavrajs (approved):
Nice work! This looks good to me except a small comment around sender variable naming.

app/models/campaign.rb
sojan-official: This column isnt to store the person who created the campaign. But the agent who would appear as the sender for the campaign message. 

If the value is left blank, we will show the sender to be bot. 


I think we can design the ui in a way that this behaviour is clearly denoted.

PR #2183, opened 2021-04-29
Fix Timezone and update open_at logic

### LINKED ISSUE #2182: Out of office message is not working properly even after setting business hours and setting timezone
**Describe the bug**

Out of office message is not working properly even after setting business hours and setting timezone

**To Reproduce**

Steps to reproduce the behaviour:
1. Create website inbox
2. Set business hours
3. Start conversation in web widget
4. Out of office message is not sent even when outside outside business hours.

vishal-pandey:

Fixes #2182 

## Type of change


pranavrajs:
@vishal-pandey Thanks for the PR.

@sojan-official Can you please take a look at this? I'm concerned about the part where `Time.zone = inbox.timezone` is set. Would it have issues as it is not reset in the next call?

PR #2192, opened 2021-04-30
feat: Campaign conversation builder

sojan-official:
This PR implements APIs to trigger a campaign

- the widget should hit the /widget/events controller with the following payload when conditions to trigger a campaign is satisfied 

```
{
  name: "campaign.triggered"
  event_info: {
      campaign_id: campaign_id
   }
}
```


app/builders/campaigns/campaign_conversation_builder.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def conversation_params
+    {
+      account_id: @campaign.account_id,
+      inbox_id: @contact_inbox.inbox_id,
+      contact_id: @contact_inbox.contact_id,
+      contact_inbox_id: @contact_inbox.id,
+      campaign_id: @campaign.id
+    }
```
pranavrajs, line 33: Should we add browser params here also? That might be relevant right?

pranavrajs: Product question: Should we create the conversation in resolve status by default? Because this would be shown in the UI and might be a noise right?

pranavrajs (dismissed):
Looks good overall! Nice work! 

I have 2 comments about the implementation. Please see the individual comments.

app/builders/campaigns/campaign_conversation_builder.rb
sojan-official: @pranavrajs  browser params are now passed into the builder,
let's take the starting conversation in resolved status as a setting in our next iteration.

Issue #1710, opened 2021-01-30
Bug: Text overlaps in conversation area


amitpatelx:
**Describe the bug**

Time of the message overlaps in conversation area with channel name
 
**To Reproduce**

1. Send a conversation through any channel
2. Visit Dashboard immediately
3. Check the conversation area

**Expected behavior**

Text should not be overlap. Instead of `less than a minute ago` may be we can use `just now`

**Screenshots**

![Chatwoot](https://user-images.githubusercontent.com/1289894/106356057-a00a4e80-6322-11eb-8460-bcc66419254c.png)

**Environment**

Self hosted development branch 

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04 LTS
 - Browser: Chrome
 - Version:88.0.4324.96 (Official Build) (64-bit)

pranjal0819:
## I found some issue with contact name like this

1. Time of the message overlaps with the name
![image](https://user-images.githubusercontent.com/40784971/111986652-40673d00-8b34-11eb-94c3-850575009139.png)


2. Email overlaps with name
![image](https://user-images.githubusercontent.com/40784971/111986820-6e4c8180-8b34-11eb-96b3-5daf781d9948.png)


3. Message gone outside of the window
![image](https://user-images.githubusercontent.com/40784971/111986950-92a85e00-8b34-11eb-965a-902bca351b25.png)


4.Time of the message overlaps with the name
![image](https://user-images.githubusercontent.com/40784971/111987035-af449600-8b34-11eb-9130-8b7d8507bd23.png)

## Suggestions
Truncate name based on pixel width.


cc: @pranavrajs @sladyn98

PR #2208, opened 2021-05-04
fix: Truncates overlapping name in contact table

nithindavid:
Fixes part of https://github.com/chatwoot/chatwoot/issues/1710#issuecomment-804015603

- Truncates and refactors to use better markup

**Before** 
image

**After**
image


pranavrajs:
@nithindavid Can you please share the issue as well here? I do not have the context on what is being fixed. A before / after screencast is helpful.

pranavrajs (approved):

PR #2224, opened 2021-05-05
chore: Ensure privilege validations for API endpoints

sojan-official:
- Add necessary authorizations for agents and admins for various API endpoints

# Testing 

- Permission differences between agents and  administrators 
- Agents shouldn't be able to access conversations outside inboxes they have access to ( via direct URLs)
- Agents should be able to interact with crud APIs for inbox member additions 
- Contactable Inboxes on the contacts page should only list inboxes to which the agent has access to


app/controllers/api/v1/accounts/contacts/conversations_controller.rb
```diff
@@ -8,9 +8,7 @@ def index
   private
 
   def inbox_ids
-    if Current.user.administrator?
-      Current.account.inboxes.pluck(:id)
-    elsif Current.user.agent?
```
pranavrajs, line 13: If the user is an admin, I guess we could show every inbox available.

sojan-official: `assigned_inboxes` handle this condition. it returns all inboxes for admin users and only the specific inboxes for agents

pranavrajs:
@sojan-official Looks like the new APIs have an issue with access. Otherwise, it looks fine, let us get it merged soon.

sojan-official:
@pranavrajs  back to you

pranavrajs (approved):

PR #2226, opened 2021-05-05
chore: Add color variants to label component

nithindavid:
- Adds support for primary, secondary, alert, success, warning

image


app/javascript/dashboard/components/widgets/Label.vue
```diff
@@ -1,12 +1,9 @@
...
-  <div
-    :class="labelClass"
-    :style="{ background: bgColor, color: textColor }"
-    :title="description"
-  >
+  <div :class="labelClass" :style="labelStyle" :title="description">
+    <i v-if="icon" class="label--icon" :class="icon" @click="onClick" />
     <span v-if="!href">{{ title }}</span>
-    <a v-else :href="href" :style="{ color: textColor }">{{ title }}</a>
-    <i v-if="showIcon" class="label--icon" :class="icon" @click="onClick" />
```
pranavrajs, line 9: I guess this should be kept, if we want to use `+` instead of `X`

nithindavid: I have covered that with a new `<i` tag above.

pranavrajs (approved):

muhsin-k (approved):

PR #2231, opened 2021-05-06
chore: Refactor sidebar button with button component

iamsivin:
Screenshot 2021-05-06 at 11 43 38 AM


app/javascript/dashboard/components/layout/AvailabilityStatus.vue
```diff
@@ -20,22 +20,28 @@
...
             </woot-dropdown-item>
           </woot-dropdown-menu>
         </div>
       </transition>
 
-      <button class="status-change--change-button" @click="openStatusMenu">
+      <woot-button
+        variant="clear"
+        color-scheme="secondary"
+        class-names="status-change--change-button link"
```
nithindavid, line 40: Can check if we need these classnames?

nithindavid (changes requested):

app/javascript/dashboard/components/layout/AvailabilityStatus.vue
iamsivin: Yes, "status-change--change-button" is used in spec.js  file.

PR #2234, opened 2021-05-06
chore: Refactor contact page button with button component

iamsivin:
Uses `woot-button` instead of deprecated components.


app/javascript/dashboard/components/widgets/TableFooter.vue
```diff
@@ -12,39 +12,41 @@
         v-if="totalCount"
         class="primary button-group pagination-button-group"
       >
-        <button
-          class="button small goto-first"
+        <woot-button
+          size="small"
+          class-names="goto-first"
           :class="firstPageButtonClass"
```
nithindavid, line 18: This `firstPageButtonClass` computed method is to get `disabled` class. But we have `is-disabled` prop for button. Please use the props.

nithindavid (changes requested):
Please `is-disabled` prop directly in `woot-button`

nithindavid:
> Please `is-disabled` prop directly in `woot-button`

Please do this for other woot-buttons also

PR #2235, opened 2021-05-06
feat: Increase the width of the message bubble

pranavrajs:
- Increase the width of the bubble to accommodate more content on smaller screens.

| Before | After |
| -- | -- |
| Screenshot 2021-05-06 at 2 04 31 PM | Screenshot 2021-05-06 at 2 03 58 PM |

### DIFF
--- a/app/javascript/dashboard/assets/scss/widgets/_conversation-view.scss
+++ b/app/javascript/dashboard/assets/scss/widgets/_conversation-view.scss
@@ -236,7 +236,7 @@
 
   .wrap {
     @include margin($zero $space-normal);
-    max-width: 69%;
+    max-width: 85%;
 
     .sender--name {
       font-size: $font-size-mini;


app/javascript/dashboard/assets/scss/widgets/_conversation-view.scss
```diff
@@ -236,7 +236,7 @@
 
   .wrap {
     @include margin($zero $space-normal);
-    max-width: 69%;
+    max-width: 85%;
```
nithindavid, line 239: As a refactor we have a max-width set for the immediate children `.bubble` also. Can you remove that from there and bring it here and `max-width: min(50rem, 85%)` ?

nithindavid (changes requested):

app/javascript/dashboard/assets/scss/widgets/_conversation-view.scss
nithindavid: 👆 Please check this if works for all attachment types.

nithindavid (approved):

PR #2236, opened 2021-05-06
chore: Fix emails being sent with the wrong translations

sojan-official:
Fix for emails being sent with the wrong translations

## QA Instructions 
------------------

This PR introduces config changes to enable account creation and switching accounts from the dashboard for the default test account.  This could be used to test the changes easily in local environments

Screenshot 2021-05-07 at 4 19 39 PM

- Create multiple accounts for a user through the super admin panel or via the dashboard 
- Assign different languages to each of these accounts 
- Switch over into various accounts via  dashboard account switcher and trigger cases where emails are sent out 
- Ensure each email is sent according to the correct account language 

notes:  only some of the chatwoot emails are translated
- titles of auth related emails ( account confirmation. password reset, agent confirmation )
- titles of conversation continuity and transcript emails


pranavrajs (dismissed):
Changes look good! @sojan-official Can you please share the steps to test this out? I can do a QA as well.

sojan-official:
@pranavrajs  updated

app/controllers/api/v1/accounts/conversations_controller.rb
```diff
@@ -41,7 +41,9 @@ def unmute
   end
 
   def transcript
-    ConversationReplyMailer.conversation_transcript(@conversation, params[:email])&.deliver_later if params[:email].present?
+    if params[:email].present?
```
pranavrajs, line 44: Should we send an unprocessable entity response if the email param is not present?

app/mailers/application_mailer.rb
```diff
@@ -65,11 +65,11 @@ def locale_from_account(account)
   end
```
pranavrajs, line 65: There is a `Current.account` reference in line number 46. I think that also needs to be updated.

pranavrajs (commented):
I noticed one issue with the current account in liquid template variable. Please see the individual comment.

sojan-official:
@pranavrajs  we will have to test this for [custom email templates](https://www.chatwoot.com/docs/self-hosted/configuration/email-notifications/).

pranavrajs (approved):

PR #2241, opened 2021-05-07
feat: Add inline editing support for crm attributes

nithindavid:
Adds inline editing feature for contact attributes.
https://user-images.githubusercontent.com/1277421/117456063-6bcbad00-af65-11eb-8cda-8654f6ceaaa4.mov


app/javascript/dashboard/modules/contact/components/ContactAttribute.vue
```diff
@@ -8,7 +8,26 @@
         {{ label }}
       </h4>
     </div>
-    <div class="value-wrap">
+    <div v-show="isEditing">
```
pranavrajs, line 11: A comment on the implementation. Should we use content-editable instead of input box? Something like this https://www.npmjs.com/package/vue-contenteditable

nithindavid: This looks like good improvement over the current one. I'll pickup this an improvement in a new PR.

pranavrajs (approved):

PR #2250, opened 2021-05-11
feat: APIs for Integration Hooks

sojan-official:
- Introduces JSON Schema validations via JSONSchemer
- Add CRUD APIs for integration hooks


pranavrajs (approved):
Looks good to me! 👍  @sojan-official Do we need to test out the existing integrations?

sojan-official:
@pranavrajs  There shouldn't be a significant impact on existing integrations apart from the additional validations added on the model. The specs should cover this case.

PR #2254, opened 2021-05-12
chore: Execute campaigns based on matching url

muhsin-k:
- Start campaigns based on the URL specified in trigger rules  

## How to Test
1. Enable inbox campaigns

Open `app/javascript/dashboard/routes/dashboard/settings/inbox/Settings.vue`
Add the code on line number 337

```
{
 key: 'campaign',
 name: this.$t('INBOX_MGMT.TABS.CAMPAIGN'),
}
```
2. Go to any website inbox

3. Click on the settings

4. Go to the campaigns tab

5. Create a new campaign by entering details title, messages, sender, URL and time on page

6. Open the URL which configured website inbox

7. Wait for the time specified in the campaign

8. Check messages are coming over the widget


app/javascript/dashboard/routes/dashboard/settings/inbox/Settings.vue
```diff
@@ -438,7 +438,7 @@ export default {
       }
     },
     async updateAgents() {
-      const agentList = this.selectedAgents.map((el) => el.id);
+      const agentList = this.selectedAgents.map(el => el.id);
```
pranavrajs, line 441: I'm not sure whether this actually conforms to the eslint rules we have. This seems there is either a problem with the prettier config we have or there is some issue with the editor plugins you use. Could you please remove these?

app/javascript/sdk/IFrameHelper.js
```diff
@@ -140,6 +146,9 @@ export const IFrameHelper = {
         IFrameHelper.pushEvent('webwidget.triggered');
       }
     },
+    onLocationChange: url => {
+      IFrameHelper.sendMessage('change-url', { url });
```
pranavrajs, line 150: There is an event `set-current-url`. Could you please check if we can reuse that?

app/javascript/widget/helpers/campaignHelper.js
```diff
@@ -0,0 +1,41 @@
+import { triggerCampaign } from 'widget/api/campaign';
+window.campaignTimers = [];
```
pranavrajs, line 2: Is there a way to avoid window object. I understand that we need a global object. Can we do it in a different way?

app/javascript/widget/helpers/campaignHelper.js
```diff
@@ -0,0 +1,41 @@
...
+      );
+    });
+  // Execute campaigns
+  filteredCampaigns.forEach(campaign => {
+    const {
+      trigger_rules: { time_on_page: timeOnPage },
+      id: campaignId,
+    } = campaign;
+    const timeoutID = setTimeout(async () => {
+      await triggerCampaign({ campaignId });
```
pranavrajs, line 35: I don't think we need an await here. It doesn't seem to have any effect.

app/javascript/widget/helpers/campaignHelper.js
```diff
@@ -0,0 +1,41 @@
...
+        timeOnpage: item?.trigger_rules?.time_on_page,
+        url: item?.trigger_rules?.url,
+      };
+    })
+    .filter(item => {
+      return (
+        stripTrailingSlash({ URL: item.url }) ===
+        stripTrailingSlash({ URL: currentURL })
+      );
+    });
```
pranavrajs, line 27: Can you please refactor this code? Ideally, move it to a method and have a test case for the same.

app/javascript/widget/helpers/campaignHelper.js
```diff
@@ -0,0 +1,41 @@
...
+      id: campaignId,
+    } = campaign;
+    const timeoutID = setTimeout(async () => {
+      await triggerCampaign({ campaignId });
+    }, timeOnPage * 1000);
+    window.campaignTimers.push(timeoutID);
+  });
+};
+
+export { startCampaigns };
```
pranavrajs, line 41: Please use `export const startCampaigns`

app/javascript/widget/helpers/campaignHelper.js
```diff
@@ -0,0 +1,41 @@
+import { triggerCampaign } from 'widget/api/campaign';
```
pranavrajs, line 1: Please add a test cases for this file.

app/javascript/widget/App.vue
```diff
@@ -184,6 +188,11 @@ export default {
           this.showUnreadView = true;
         } else if (message.event === 'unset-unread-view') {
           this.showUnreadView = false;
+        } else if (message.event === 'change-url') {
+          startCampaigns({
+            allCampaigns: this.campaigns,
+            currentURL: message.url,
+          });
```
pranavrajs, line 195: I would suggest this also to be an action in the campaign module. Otherwise, changes in campaigns object would re-render the entire tree.

pranavrajs (changes requested):
Nice work! I have some comments about the implementation. Please see the individual comments.

app/javascript/dashboard/routes/dashboard/settings/inbox/Settings.vue
muhsin-k: Some issue with `arrowParens` rule. I will add the rule `"arrowParens": "avoid"` in **.prettierrc** file

app/javascript/sdk/IFrameHelper.js
muhsin-k: The event `set-current-url` fires only when url hash changes. It doesn't work on non-hash URL changes.

app/javascript/widget/helpers/campaignHelper.js
muhsin-k: Added class for campaign helper

muhsin-k:
@pranavrajs Please check the lastest changes

app/javascript/widget/helpers/campaignTimer.js
```diff
@@ -1,14 +1,33 @@
...
+  };
+
+  clearTimers = () => {
+    this.campaignTimers.forEach(timerId => {
+      clearTimeout(timerId);
+      this.campaignTimers[timerId] = null;
+    });
+  };
+}
+
```
pranavrajs, line 27: export default new CampaignTimer();

app/javascript/widget/store/modules/campaign.js
```diff
@@ -29,10 +28,7 @@ export const actions = {
       campagins: formattedCampaigns,
       currentURL,
     });
-    campaignTimer.init({
-      allCampaigns: filteredCampaigns,
-      currentURL,
-    });
+    campaignTimer.initTimers({ allCampaigns: filteredCampaigns });
```
pranavrajs, line 31: One small feedback on the naming. Shouldn't it be enough to have the variable named as `campaigns` rather than `allCampaigns`

PR #2300, opened 2021-05-18
docs: Adds documentation for Integration APIs

sojan-official:
docs: Adds documentation for Integration APIs


app/views/api/v1/models/_hook.json.jbuilder
```diff
@@ -1,4 +1,7 @@
 json.id resource.id
-json.app resource.app.params.to_h
-json.enabled resource.enabled?
+json.app_id resource.app_id
+json.status resource.enabled?
```
pranavrajs, line 3: Is this API being using in the frontend anywhere?

sojan-official: nope

Issue #2304, opened 2021-05-19
Trick to use Slack Integration?


wflanagan:
I'm attempting to deploy the Chatwoot app using CapRover. I have configured the Slack app in my account. Further, I've configured as the documentation states in the Chatwoot help.  

When I attempt to hit "connect", I receive an error page from Slack stating: 

```
Something went wrong when authorizing this app.
Try going back and authorizing again. If problems persist, contact support for help.

Error details
Please specify `client_id`
```

When I manually add the client_id param to the URL, and hit it.. I get the OAuth dialog. If I complete this, ti returns to Chatwoot, but does not provide any sort of notification. Further, the "Slack" connect button still says "connect." 

Caprover is hosted on DigitalOcean. Latest Chrome. MacOs Big Sur.

The documentation states to use the base url as the redirect URL. I'm wondering if that's not right? Also, it seems the URL that's produced by what I'm guessing is Omniauth doesn't have the client_id in it. 

So, net, is there a trick to get the Slack integration working?

wflanagan:
I'll add that I have a few services that integrate with Slack. And, most have this problem. I think that Slack made a change to their OAuth requirements.

pranavrajs:
@wflanagan I can reproduce this in the staging environment. I'm not sure whether something changed recently.

wflanagan:
@sojan-official Hey, thanks for the docker reference. FYI.. the docker image boots basically stating the following in the logs: 

2021-05-21T20:11:38.849972239Z Switch to inspect mode.
2021-05-21T20:11:38.851176267Z
2021-05-21T20:11:45.696337306Z Switch to inspect mode.
2021-05-21T20:11:45.697945289Z
2021-05-21T20:11:52.549161984Z Switch to inspect mode.
2021-05-21T20:11:52.550676445Z
2021-05-21T20:11:59.385619918Z Switch to inspect mode.
2021-05-21T20:11:59.386869082Z

and is non-operational.

sojan-official:
@wflanagan  the images which we provide on the hub are the base images.  Using the base image you will have to build images for the server and worker with the required service

ref: https://github.com/chatwoot/chatwoot/blob/develop/docker-compose.production.yaml
https://www.chatwoot.com/docs/self-hosted/deployment/docker

ilearnio:
This issue still happens for me after I deployed to heroku yesterday using your "One click Heroku deploy" button

pranavrajs:
[...]

PR #2309, opened 2021-05-20
fix: Slack integration redirect url missing client ID

sojan-official:
Fixes: #2304

### DIFF
--- a/app/views/api/v1/models/_app.json.jbuilder
+++ b/app/views/api/v1/models/_app.json.jbuilder
@@ -2,6 +2,7 @@ json.call(resource.params, *resource.params.keys)
 json.name resource.name
 json.description resource.description
 json.enabled resource.enabled?(@current_account)
+json.action resource.action
 json.button resource.action
 json.hooks do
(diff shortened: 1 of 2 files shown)


nithindavid (approved):

wflanagan:
@sojan-official Is it possible to get a new docker image?

sojan-official:
@wflanagan  you can use the `develop` branch image until we release the next version. 

https://hub.docker.com/layers/94291362/chatwoot/chatwoot/develop/images/sha256-cf87f8eebb47435a4389a6435489cee385862347e7e866167da55308263c4832?context=explore

wflanagan:
@sojan-official Thanks, I appreciate that. But, I think that's not a public image. I can't access it via that URL anyway.

sojan-official:
> @sojan-official Thanks, I appreciate that. But, I think that's not a public image. I can't access it via that URL anyway.

https://hub.docker.com/r/chatwoot/chatwoot/tags please use the develop tag

Issue #2077, opened 2021-04-09
Allow adding regex validation on form message type inputs.


nimitwalia89:
Currently adding validations is not easy. We basically have to check the users submitted values and send message again with same inputs. It would be easier to developer and better user experience if we allow adding regex for input validations.

sojan-official:
The form text inputs now support the following attributes 

`pattern` : standard html attribute that supports regex 
`pattern_error` : a custom error message for the regex validation
`required` :  whether field is required


example 

```
 { "name": 'email', "placeholder": 'Please enter your email', "type": 'email', "label": 'Email', "required": 'required',          "pattern_error": 'Please fill this field', "pattern": '^[^\s@]+@[^\s@]+\.[^\s@]+$' },
 { "name": 'text_area', "placeholder": 'Please enter text', "type": 'text_area', "label": 'Large Text', "required": 'required',          "pattern_error": 'Please fill this field' },      
 { "name": 'text', "placeholder": 'Please enter text', "type": 'text', "label": 'text', "default": 'defaut value', "required": 'required',          "pattern": '^[a-zA-Z ]*$', "pattern_error": 'Only alphabets are allowed' },      
```

PR #2317, opened 2021-05-21
feat: Add regex validation on form message input

iamsivin:
**After**

https://user-images.githubusercontent.com/64252451/120766477-e9201880-c537-11eb-8de7-ee84cbb8c9ed.mov

**Before**

https://user-images.githubusercontent.com/64252451/119114663-fa4f2c80-ba43-11eb-8598-dcc66690b4e3.mov

fixes: #2077


sojan-official:
I think we should add an error indication on the form field, or the end-user who is not aware of the validation could be confused.

app/javascript/shared/components/ChatForm.vue
```diff
@@ -90,6 +91,20 @@ export default {
...
+    fieldValidation(name, regex = '') {
+      const formValue = this.formValues[name];
+      if (!regex) {
+        return true;
+      }
+      const [, pattern, flags] = regex.match(/\/(.*)\/([a-z]*)/);
+      const reg = new RegExp(pattern, flags);
+      if (!reg.test(formValue)) {
+        return false;
+      }
```
nithindavid, line 104:
```suggestion
reg.test(formValue);
```

nithindavid: We can simplify it like this.

nithindavid (changes requested):

app/javascript/shared/components/ChatForm.vue
```diff
@@ -90,6 +91,20 @@ export default {
       }
       this.$emit('submit', this.formValues);
     },
+
+    fieldValidation(name, regex = '') {
```
pranavrajs, line 95: Please consider a better name for the method.

app/javascript/shared/components/ChatForm.vue
```diff
@@ -90,6 +91,20 @@ export default {
...
+      const formValue = this.formValues[name];
+      if (!regex) {
+        return true;
+      }
+      const [, pattern, flags] = regex.match(/\/(.*)\/([a-z]*)/);
+      const reg = new RegExp(pattern, flags);
```
pranavrajs, line 101: I didn't understand this part. Why do we need this? Can directly use new RegExp(regex)

pranavrajs (commented):
Nice work! I have some comments about the implementation, please see the individual comments.

app/javascript/shared/components/ChatForm.vue
```diff
@@ -3,14 +3,28 @@
...
-          :name="item.name"
-          :placeholder="item.placeholder"
-          :disabled="!!submittedValues.length"
-        />
+        <div v-if="item.type === 'email' || item.type === 'text'">
+          <div v-if="item.type === 'email'">
+            <input
+              v-model="formValues[item.name]"
+              :type="item.type"
+              :class="{ error: hasFetch }"
```
nithindavid, line 11: We can have more meaningful variable name like `hasTouched`.

nithindavid (changes requested):

XuanThuLe:
lephamkhoinguyen@gmail.cơm

app/javascript/shared/components/ChatForm.vue
XuanThuLe: Ok

XuanThuLe:
Ok

XuanThuLe:
Ok

app/javascript/shared/components/ChatForm.vue
```diff
@@ -90,6 +106,32 @@ export default {
...
+      const [, pattern, flags] = regex.match(/\/(.*)\/([a-z]*)/);
+      const reg = new RegExp(pattern, flags);
+      return reg.test(formValue);
+    },
+
+    isTouched() {
+      this.hasTouched = true;
+    },
+
+    emailValidation(name, type) {
```
sojan-official, line 124: w can rename the function to be something more generic. The regex doesn't need to be `email`. It could be a specific code, date, anything

lib/woot_message_seeder.rb
```diff
@@ -76,11 +76,9 @@ def self.create_sample_form_message(conversation)
       content: 'form',
       content_attributes: {
         "items": [
-          { "name": 'email', "placeholder": 'Please enter your email', "type": 'email', "label": 'Email' },
-          { "name": 'text_area', "placeholder": 'Please enter text', "type": 'text_area', "label": 'Large Text' },
-          { "name": 'text', "placeholder": 'Please enter text', "type": 'text', "label": 'text', "default": 'defaut value' },
-          { "name": 'select', "label": 'Select Option', "type": 'select', "options": [{ "label": '🌯 Burito', "value": 'Burito' },
-                                                                                      { "label": '🍝 Pasta', "value": 'Pasta' }] }
+          { "name": 'email', "placeholder": 'Please enter your email', "type": 'email', "label": 'Email', "regex": '/^[^\s@]+@[^\s@]+\.[^\s@]+$/' },
+          { "name": 'text_aread', "placeholder": 'Please enter text', "type": 'text_area', "label": 'Large Text' },
```
sojan-official, line 80:
```suggestion
          { "name": 'text_area', "placeholder": 'Please enter text', "type": 'text_area', "label": 'Large Text' },
```

app/javascript/shared/components/ChatForm.vue
```diff
@@ -3,14 +3,28 @@
...
         <label>{{ item.label }}</label>
-        <input
-          v-if="item.type === 'email' || item.type === 'text'"
-          v-model="formValues[item.name]"
-          :type="item.type"
-          :name="item.name"
-          :placeholder="item.placeholder"
-          :disabled="!!submittedValues.length"
-        />
+        <div v-if="item.type === 'email'">
```
sojan-official, line 6: We would need to do the validation on all input text elements available in chatwoot
`text`, `email`

So let's remove the specific condition added.

sojan-official (commented):
@nithindavid  @muhsin-k  thoughts on using https://www.w3schools.com/tags/att_input_pattern.asp 
pattern attribute ?

app/javascript/shared/components/ChatForm.vue
iamsivin: Changed this to the input pattern attribute method.

app/javascript/shared/components/ChatForm.vue
iamsivin: Changed this method to the input pattern attribute method.

lib/woot_message_seeder.rb
```diff
@@ -76,11 +76,10 @@ def self.create_sample_form_message(conversation)
       content: 'form',
       content_attributes: {
         "items": [
-          { "name": 'email', "placeholder": 'Please enter your email', "type": 'email', "label": 'Email' },
-          { "name": 'text_area', "placeholder": 'Please enter text', "type": 'text_area', "label": 'Large Text' },
-          { "name": 'text', "placeholder": 'Please enter text', "type": 'text', "label": 'text', "default": 'defaut value' },
-          { "name": 'select', "label": 'Select Option', "type": 'select', "options": [{ "label": '🌯 Burito', "value": 'Burito' },
```
sojan-official, line 82: I think the select example got removed. let's add this back

iamsivin: Yeah sure.

app/models/concerns/content_attribute_validator.rb
```diff
@@ -2,7 +2,7 @@ class ContentAttributeValidator < ActiveModel::Validator
   ALLOWED_SELECT_ITEM_KEYS = [:title, :value].freeze
   ALLOWED_CARD_ITEM_KEYS = [:title, :description, :media_url, :actions].freeze
   ALLOWED_CARD_ITEM_ACTION_KEYS = [:text, :type, :payload, :uri].freeze
-  ALLOWED_FORM_ITEM_KEYS = [:type, :placeholder, :label, :name, :options, :default].freeze
+  ALLOWED_FORM_ITEM_KEYS = [:type, :placeholder, :label, :name, :options, :default, :required, :regex].freeze
```
sojan-official, line 5: let rename `:regex` to `:pattern`  since that confer with the standards.
Let's also introduce `:title` attribute so that the validation condition can be described by the API user

> Tip: Use the global title attribute to describe the pattern to help the user.
ref: https://www.w3schools.com/tags/att_pattern.asp

app/javascript/shared/components/ChatForm.vue
```diff
@@ -1,23 +1,48 @@
...
+            :pattern="item.regex"
+            :required="item.required ? 'required' : false"
+            :name="item.name"
+            :placeholder="item.placeholder"
+            :disabled="!!submittedValues.length"
+          />
+          <span class="error-message">{{ $t('CHAT_FORM.INVALID.EMAIL') }}</span>
+        </div>
+        <div v-if="item.type === 'text'">
+          <input
```
sojan-official, line 26: let's add the pattern and title attribute to item.type 'text' as well

sojan-official:
@nithindavid  @muhsin-k  @pranavrajs  Could one you take a final pass?

muhsin-k (approved):

Issue #2276, opened 2021-05-17
feat: APIs to create and update Agent Bots


sojan-official:
- APIs to create and update installation wide agent bots via platform apps
- APIs to create and update account-specific agent bots via admin user tokens

PR #2323, opened 2021-05-21
feat: Add AgentBot APIs

sojan-official:
fixes: #2276


app/controllers/api/v1/accounts/agent_bots_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def destroy
+    @agent_bot.destroy
+    head :ok
+  end
+
+  private
+
+  def fetch_agent_bot
+    @agent_bot = Current.account.agent_bots.find(params[:id])
```
pranavrajs, line 28: Should we use `permitted_params` here?

db/schema.rb
```diff
@@ -97,7 +97,8 @@
     t.string "outgoing_url"
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
-    t.boolean "hide_input_for_bot_conversations", default: false
```
pranavrajs, line 100: This is used in the frontend, we might need to evaluate the impact.

spec/controllers/api/v1/accounts/agent_bots_controller_spec.rb
```diff
@@ -0,0 +1,179 @@
...
+      it 'shows the agent bot' do
+        get "/api/v1/accounts/#{account.id}/agent_bots/#{agent_bot.id}",
+            headers: agent.create_new_auth_token,
+            as: :json
+
+        expect(response).to have_http_status(:success)
+        expect(response.body).to include(agent_bot.name)
+      end
+
+      it 'would not show a global agent bot' do
```
pranavrajs, line 51: Why is this logic added here? I guess if we display all the bots with account nil in index API, then we should be consistent with show API.

pranavrajs: Editing is different case, you might not want to allow editing global bots, individual display should be allowed.

app/views/api/v1/models/_inbox.json.jbuilder
```diff
@@ -22,6 +22,9 @@ json.phone_number resource.channel.try(:phone_number)
 json.selected_feature_flags resource.channel.try(:selected_feature_flags)
 json.reply_time resource.channel.try(:reply_time)
 json.reauthorization_required resource.channel.try(:reauthorization_required?) if resource.facebook?
+json.agent_bot do
```
pranavrajs, line 25: I would advise against this. If we need to fetch whether there is an agent bot associated with the inbox, we can fetch it separately. 

Adding it here would create unnecessary DB queries in the dashboard while fetching inboxes or a particular inbox.

pranavrajs: Also, this might be an N+1 query.

pranavrajs: Ideally inbox/{inbox_id}/agent_bots is a good way to do it.

pranavrajs (changes requested):
Nice work! Looks good overall.

db/schema.rb
sojan-official: remove the frontend code base on this.

pranavrajs (approved):

Issue #2334, opened 2021-05-25
Allow to set bot name in campaigns and/or disable creating new chats until user answers campaign


lucargir:
**Is your enhancement request related to a problem? Please describe.**
It is an enhancement of the campaigns feature

**Describe the solution you'd like**
When creating a campaign, one can choose to send the campaign on behalf of any of the agents or a bot.

- if the campaign is sent on behalf of an agent, a new chat is created and assigned to them even if the user does not reply to the campaign. This becomes bothersome as many users won't reply to the campaign and new chats are created and assigned to the agent, taking space in their inbox and getting notifications by e-mail.
This could be solved by sending campaigns without creating chats until the user replies. Maybe not assigning the chat until the user replies? Or not creating a chat at all until the user replies, and then assigning the campaign to the chat. 

- If a campaign is sent by the bot, it literally says "Bot" on the chat bubble... This is a bit crappy for the user, because even though everyone knows these messages are automated, the system is literally saying "this was sent by a bot, not a real person".
Being able to change the bot name would be enough.

**Describe alternatives you've considered**
Using the bot agent for now works...

muhsin-k:
> **Is your enhancement request related to a problem? Please describe.**
> It is an enhancement of the campaigns feature
> 
> **Describe the solution you'd like**
> When creating a campaign, one can choose to send the campaign on behalf of any of the agents or a bot.
> 
> * if the campaign is sent on behalf of an agent, a new chat is created and assigned to them even if the user does not reply to the campaign. This becomes bothersome as many users won't reply to the campaign and new chats are created and assigned to the agent, taking space in their inbox and getting notifications by e-mail.
>   This could be solved by sending campaigns without creating chats until the user replies. Maybe not assigning the chat until the user replies? Or not creating a chat at all until the user replies, and then assigning the campaign to the chat.

We will fix this issue in the next release.

> * If a campaign is sent by the bot, it literally says "Bot" on the chat bubble... This is a bit crappy for the user, because even though everyone knows these messages are automated, the system is literally saying "this was sent by a bot, not a real person".
>   Being able to change the bot name would be enough.
> 

It is a good suggestion. We will consider this in the next release

> **Describe alternatives you've considered**
> Using the bot agent for now works...

iammattmartin:
Part of this is dupe of #2288. Just to link the two.

stonecharioteer:
[...]

PR #2335, opened 2021-05-25
feat: Create campaign conversation only if user interacts with the bubble 

### LINKED ISSUE #2288: Don't create conversation unless clicked with campaign
Yes

Currently, it looks like a new conversation is created when a rule is matched but this might be non-ideal as there will be a high percentage of chats simply with no content (replies) (or where the X (closure) is pressed). It also triggers a new conversation email which isn't ideal.

A conversation should really only be generated when a visitor interacts with the campaign and provides some details (#2287) or responds.

Likewise, if a user clicks the X, it should learn, close and not keep popping up with the message on each new page as that will get **really** annoying.

muhsin-k:
Fixes #2288 and #2334


app/javascript/widget/App.vue
```diff
@@ -109,10 +119,28 @@ export default {
...
+        this.setUserLastSeen();
+        this.executeCampaign({ campaignId, websiteToken });
+      });
     },
     setPopoutDisplay(showPopoutButton) {
       this.showPopoutButton = showPopoutButton;
     },
+    setCampaignView() {
+      const { messageCount, activeCampaign } = this;
+      if (this.isIFrame && !isEmptyObject(activeCampaign) && !messageCount) {
```
nithindavid, line 137: Would be more readable if we have constant with name suggesting what it does inside `if`.

app/javascript/widget/components/UnreadMessage.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <div class="chat-bubble-wrap">
-    <div class="chat-bubble agent">
+    <div
+      :class="['chat-bubble agent', checkMessageIsCampaign && 'campaign']"
+      v-on="checkMessageIsCampaign ? { click: onClickCampaign } : {}"
```
nithindavid, line 5: Can we use `@click` here and moving conditions as a new method?

app/javascript/widget/components/UnreadMessage.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <div class="chat-bubble-wrap">
-    <div class="chat-bubble agent">
+    <div
+      :class="['chat-bubble agent', checkMessageIsCampaign && 'campaign']"
```
nithindavid, line 4: Can we follow the `{ 'x': true }` inside classes? Also the `chat-bubble agent` can be moved outside as a separate `class="chat-bubble agent"`

nithindavid (changes requested):

app/javascript/widget/components/UnreadMessage.vue
```diff
@@ -38,8 +46,20 @@ export default {
...
+      type: Number,
+      default: null,
+    },
   },
   computed: {
+    campaignClass() {
+      return {
+        campaign: this.messageType === 'campaign',
+      };
+    },
```
pranavrajs, line 59: This is not being used anywhere

app/javascript/widget/components/UnreadMessage.vue
```diff
@@ -1,6 +1,10 @@
 <template>
   <div class="chat-bubble-wrap">
-    <div class="chat-bubble agent">
+    <div
+      :class="{ campaign: checkMessageIsCampaign }"
```
pranavrajs, line 4: I don't think we need to have a conditional logic here, we can enable click on both campaign messages and normal messages.

Also please consider using <button> instead of <div />

tabrza:
@pranavrajs - is it possible to only create the conversation when the user sends a message or keep it as an option on the campaign for in the settings section? I've noticed that users in attempts to click away the campaign, click on the bubble which result in a message being sent to the agent whereas the user has no interest in interacting with the agent

PR #2337, opened 2021-05-25
fix: Uses label component for add label button

nithindavid:
**Why**
The current add label uses button which doesn't match the label size.
**After**

image

### DIFF
--- a/app/javascript/shared/components/ui/dropdown/AddLabel.vue
+++ b/app/javascript/shared/components/ui/dropdown/AddLabel.vue
@@ -1,11 +1,11 @@
 <template>
-  <span role="button" class="label--add" @click="addLabel">
+  <woot-button variant="link" class="label--add" @click="addLabel">
     <woot-label
       color-scheme="secondary"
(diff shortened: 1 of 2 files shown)


app/javascript/shared/components/ui/dropdown/AddLabel.vue
```diff
@@ -1,34 +1,34 @@
...
-    variant="hollow"
-    size="tiny"
-    icon="ion-plus-round"
-    color-scheme="secondary"
-    class-names="button-wrap"
-    @click="toggleLabels"
-  >
-    {{ $t('CONTACT_PANEL.LABELS.MODAL.ADD_BUTTON') }}
-  </woot-button>
+  <span role="button" class="label--add" @click="addLabel">
```
pranavrajs, line 2: Please refrain from @click listeners on non-button attributes. Otherwise, we will have to handle keyboard listeners, tabIndex etc here.

nithindavid: @pranavrajs Done.

PR #2342, opened 2021-05-26
chore: Update Facebook Messenger gem

sojan-official:
- update Facebook messenger gem
- clean up the dead code


pranavrajs (approved):
Looks good! Nice work on cleaning up unused routes. 

Tested it on staging, created a new inbox, tested both incoming and outgoing messages

Issue #1542, opened 2020-12-23
Canned responses should be available to paste at any point in the chat


pranavrajs:
- Canned responses should be available at any point in the chat. Right now, it works only at the start of the conversation
- Allow canned responses in private notes.

sojan-official:
the possible use case for canned responses in private notes :  

When transferring a conversion across team ( support to engineering ), There is a template which the agent has to fill out for the other team.

PR #2356, opened 2021-06-01
enhancement: Adds support for canned command key everywhere on rich text editor

nithindavid:
This makes it easy to insert a canned response everywhere in the rich text editor.
Note: The simple text editor doesn't get this functionality. I'm creating a new pr for default to use rich-text-editor.

Fixes: #1542


pranavrajs (changes requested):
Looks really nice! I don't have any comments on the PR, it looks good.

Although I noticed a bug in the implementation. If the "Enter to send" is enabled, then it wouldn't allow the select the canned response, instead, it would send the message directly. You might need to handle that as well.

pranavrajs:
https://user-images.githubusercontent.com/2246121/120425565-0e702380-c38c-11eb-9b44-2f02db2ab4ce.mp4

pranavrajs (approved):

PR #2363, opened 2021-06-02
chore: Adds support for prose-mirror on editor

nithindavid:
**This PR will**
- Use common dependencies from woot-prosemirror package
- Makes the default editor work on top of prose-mirror.


app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
```diff
@@ -18,22 +18,10 @@
         v-on-clickaway="hideEmojiPicker"
         :on-click="emojiOnClick"
       />
-      <resizable-text-area
```
pranavrajs, line 21: We cannot use the rich content editor for Facebook/Twitter/Whatsapp as they don't support markdown format we use.

pranavrajs (changes requested):
I have a comment about removing the normal editor. Please see the comment above.

nithindavid:
@pranavrajs This is addressed in #2393

PR #2372, opened 2021-06-03
fix: Retain focus after sending message in rich content editor

### LINKED ISSUE #2200: New reply box loses focus
Since the most recent change (addition of the "Enter to send" checkbox), when you enter a reply with that selected and press enter, on sending, the focus on the input is lost so you have to manually click within that to send another message.

This behaviour has changed as it always used to retain focus on that input.

v1.15.1

pranavrajs:
Fixes #2200

### DIFF
--- a/app/javascript/dashboard/components/widgets/WootWriter/Editor.vue
+++ b/app/javascript/dashboard/components/widgets/WootWriter/Editor.vue
@@ -136,9 +136,11 @@ export default {
     showCannedMenu(updatedValue) {
       this.$emit('toggle-canned-menu', !this.isPrivate && updatedValue);
     },
-    value(newValue) {
+    value(newValue = '') {
       if (newValue !== this.lastValue) {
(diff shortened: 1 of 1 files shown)


nithindavid (dismissed):

app/javascript/dashboard/components/widgets/WootWriter/Editor.vue
```diff
@@ -136,9 +136,11 @@ export default {
     showCannedMenu(updatedValue) {
       this.$emit('toggle-canned-menu', !this.isPrivate && updatedValue);
     },
-    value(newValue) {
+    value(newValue = '') {
       if (newValue !== this.lastValue) {
-        this.state = createState(newValue, this.placeholder, this.plugins);
+        const tr = this.state.tr;
```
nithindavid, line 141:
```suggestion
        const { tr } = this.state;
```

nithindavid (approved):

PR #2374, opened 2021-06-03
 chore: Add different color for messages from bot

nithindavid:
Users will be able to easily distinguish agent messages with ones sent by bots.

**Preview**
image


app/javascript/dashboard/components/widgets/conversation/Message.vue
```diff
@@ -197,11 +197,15 @@ export default {
...
         'is-image': this.hasImageAttachment,
         'is-text': this.hasText,
+        'is-from-bot': this.isSentByBot,
       };
     },
     isPending() {
       return this.data.status === MESSAGE_STATUS.PROGRESS;
     },
+    isSentByBot() {
+      return !this.sender.type || this.sender.type === 'bot';
```
pranavrajs, line 207: sender can be null

pranavrajs (approved):

sojan-official (approved):

muhsin-k (approved):

app/javascript/dashboard/components/widgets/conversation/Message.vue
pranavrajs: @nithindavid Sorry about this, this was not needed. This was handled already in the computed property.

pranavrajs (changes requested):
On QA, I found that the message turns violet first and then becomes blue. Please see the attached video.


https://user-images.githubusercontent.com/2246121/120775981-5e441b80-c541-11eb-8e16-33bd6e6343b9.mp4

nithindavid:
@pranavrajs Thank for pointing out! I have added the fix.

pranavrajs (approved):

Issue #2343, opened 2021-05-27
feat: Public APIs for API channels


sojan-official:
- Enable a public API interface for API channels which can be toggled on
- The API interface will be similar to that of current website channel
- The APIs should be documented so that custom interfaces can be built on top
- There should be APIs to connect to the chatwoot websockets
- This will enable building custom chat widget / mobile chat interfaces over chatwoot

PR #2375, opened 2021-06-03
feat: Add Public APIs for API Channel

sojan-official:
fixes: #2343


sojan-official:
ref : https://github.com/chatwoot/api_channel_demo

PR #2400, opened 2021-06-08
chore: Rails Upgrade

sojan-official:
Apart from the basic regression testing 
-------

- [x] Sentry Integration
- [x] Test Facebook Messenger 
- [x] Test Azure Storage 

Addresses: #2737


.nvmrc
```diff
@@ -1 +1 @@
-12.16.1
+16.1.0
```
pranavrajs, line 1: We might need to add this to the doc as well.

app/controllers/api/v1/accounts/base_controller.rb
```diff
@@ -14,7 +14,7 @@ def ensure_current_account
     account = Account.find(params[:account_id])
     if current_user
       account_accessible_for_user?(account)
-    elsif @resource&.is_a?(AgentBot)
```
pranavrajs, line 17: Was this an intended change?

app/controllers/api/v1/accounts/inbox_members_controller.rb
```diff
@@ -9,7 +9,7 @@ def create
       update_agents_list
       head :ok
     rescue StandardError => e
-      Rails.logger.debug "Rescued: #{e.inspect}"
+      Rails.logger.debug { "Rescued: #{e.inspect}" }
```
pranavrajs, line 12: Instead of this can we use sentry.capture.

app/models/channel/facebook_page.rb
```diff
@@ -47,14 +47,14 @@ def subscribe
       ]
     )
   rescue => e
-    Rails.logger.debug "Rescued: #{e.inspect}"
+    Rails.logger.debug { "Rescued: #{e.inspect}" }
```
pranavrajs, line 50: Same comment here as above, do we need to report this to Sentry rather than debug logging? This wouldn't be available in the production logs, right?

pranavrajs (commented):
Nice work! ❤️ There is a lot of changes, let us have a proper test plan.

.nvmrc
sojan-official: At the moment we aren't using nvm to manage node.js.  It's scattered across the platforms.
Let's take this as a subsequent change to ensure we use a consistent version of the node everywhere. 

https://github.com/chatwoot/docs/issues/66

app/controllers/api/v1/accounts/base_controller.rb
sojan-official: Yep, the safe operator here seemed like we are suppressing the chances of an error.

app/controllers/api/v1/accounts/inbox_members_controller.rb
sojan-official: These are changes from robocop auto-correct. 
Let's take it as a separate task to audit the tracing: https://github.com/chatwoot/chatwoot/issues/2736

app/models/channel/facebook_page.rb
sojan-official: https://github.com/chatwoot/chatwoot/issues/2736

muhsin-k (approved):

PR #2415, opened 2021-06-11
feat: Add settings for audio alert notifications

### LINKED ISSUE #1864: Sounds trigger on events
**Describe the bug**

Following the newly added sound notifications, these appear to trigger at odd times.

When closing a ticket the sound triggers as does when a ticket is re-opened, assigned and un-assigned.

I would assume the sound should only trigger when a message is updated (from a chat message) instead of any type of update?

iamsivin:
Screenshot 2021-06-24 at 11 33 54 AM

Fixes #1864


app/javascript/shared/helpers/AudioNotificationHelper.js
```diff
@@ -60,20 +72,32 @@ export const shouldPlayAudio = (
...
+  const { conversation_id: incomingConvId } = data;
+  const currentConv =
+    window.WOOT.$store.getters.getConversationById(incomingConvId) || {};
+  const assignee = currentConv.meta.assignee;
+  let id;
+  let userId;
+  if (assignee) {
+    id = assignee.id;
+  }
+  userId = assignee.id;
```
nithindavid, line 84: This is not needed I think as `userId` and `id` will always have same value.

app/javascript/shared/helpers/AudioNotificationHelper.js
```diff
@@ -60,20 +72,32 @@ export const shouldPlayAudio = (
...
     data,
     currentConvId,
     currentUserId,
     isDocHiddden
   );
-
-  if (enableAudioAlerts && playAudio) {
+  const playAudioByUserSettings = shouldPlayByUserSettings(
+    enableAudioAlerts,
+    userId,
```
nithindavid, line 97: We should be using `currentUserId` instead of `userId` . See my above comment.

nithindavid (changes requested):

iamsivin:
@pranavrajs  I have fixed the reviews comments.

app/javascript/shared/helpers/AudioNotificationHelper.js
```diff
@@ -51,29 +64,43 @@ export const shouldPlayAudio = (
...
-
+export const getAssigneeFromNotification = currentConv => {
+  let id;
+  if (currentConv.meta) {
+    const assignee = currentConv.meta.assignee;
+    if (assignee) {
+      id = assignee.id;
+    }
+  }
+  return id;
```
nithindavid, line 79:
```suggestion
  if (currentConv.meta) {
    const { assignee = {} }= currentConv.meta;
    const { id } = assignee;
    return id;
  }
  return undefined;
```

nithindavid: Please write a spec for this function.

nithindavid (changes requested):

pranavrajs (approved):

PR #2491, opened 2021-06-22
feat: Adds inbox info to previous conversations

### LINKED ISSUE #2166: Add inbox info to previous conversations
Lets say I have several inbox, website live chat and e-mail support. If some customer contacts me on different channels, I can see the previous conversations from that customer in `Contacts` >> `View details` but it does not show which conversation is from which inbox. 

chrome_7uKM0UZ1sD

Is it possible to add a tag like this: (to show which conversation belong to which channel)
chrome_CP8ekXkoFY

iamsivin:
Fixes #2166 

Screenshot 2021-06-22 at 9 43 45 PM

### DIFF
--- a/app/javascript/dashboard/routes/dashboard/conversation/ContactConversations.vue
+++ b/app/javascript/dashboard/routes/dashboard/conversation/ContactConversations.vue
@@ -16,7 +16,7 @@
           v-for="conversation in previousConversations"
           :key="conversation.id"
           :chat="conversation"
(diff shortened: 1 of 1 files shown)


muhsin-k:
@iamsivin It would be great if we can show avatar too

pranavrajs:
@muhsin-k Avatar would of the same person, right? This is on the contact sidebar. I don't think we need to add that.

muhsin-k (approved):

PR #2494, opened 2021-06-22
feat: Add CSAT Message Template Hooks 

sojan-official:
Triggers CSAT Survey Messages when a conversation is resolved for inboxes where CSAT is enabled.

fixes: #2470

### DIFF
--- a/app/services/message_templates/hook_execution_service.rb
+++ b/app/services/message_templates/hook_execution_service.rb
@@ -50,7 +50,7 @@ class MessageTemplates::HookExecutionService
     # for now csat only available in web widget channel
     return unless inbox.web_widget?
     return unless inbox.csat_survey_enabled?
-    
+
     true
   end
 
--- a/app/services/message_templates/hook_execution_service.rb
+++ b/app/services/message_templates/hook_execution_service.rb
@@ -48,6 +48,8 @@ class MessageTemplates::HookExecutionService
 
   def should_send_csat_survey?
     return unless conversation.resolved?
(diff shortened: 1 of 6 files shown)


app/services/message_templates/hook_execution_service.rb
```diff
@@ -5,18 +5,22 @@ def perform
...
 
   delegate :inbox, :conversation, to: :message
   delegate :contact, to: :conversation
 
+  def trigger_templates
+    # TODO: let's see whether this is needed and remove this and related logic if not
+    # ::MessageTemplates::Template::OutOfOffice.new(conversation: conversation).perform if should_send_out_of_office_message?
+    ::MessageTemplates::Template::Greeting.new(conversation: conversation).perform if should_send_greeting?
+    ::MessageTemplates::Template::EmailCollect.new(conversation: conversation).perform if inbox.enable_email_collect && should_send_email_collect?
+    ::MessageTemplates::Template::CsatSurvey.new(conversation: conversation).perform if should_send_csat_survey?
```
pranavrajs, line 21: In my opinion, this shouldn't run on every message. It should be an event on the conversation after_commit hook.

sojan-official: Our conversation model is already pretty bloated. Hence the thought process to keep all the message templates in a single place for maintainability. Performance impact should be minimal since we already have the conversation object in memory. 

Let's refactor the execution service in a future iteration to run on conversation commit hooks as well.

pranavrajs (approved):

PR #2498, opened 2021-06-23
feat: Add APIs to create custom views on the dashboard

pranavrajs:
- Add a model to save custom filters
- Add APIs to save/list/update/destroy custom filter
- Add specs for the controller APIs
- Add documentation for the controller APIs.

### DIFF
--- a/db/migrate/20201123195011_create_platform_apps.rb
+++ b/db/migrate/20201123195011_create_platform_apps.rb
@@ -2,7 +2,6 @@ class CreatePlatformApps < ActiveRecord::Migration[6.0]
   def change
     create_table :platform_apps do |t|
       t.string :name, null: false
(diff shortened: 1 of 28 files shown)


db/migrate/20201123195011_create_platform_apps.rb
```diff
@@ -2,7 +2,9 @@ class CreatePlatformApps < ActiveRecord::Migration[6.0]
   def change
     create_table :platform_apps do |t|
       t.string :name, null: false
-      t.timestamps
+      t.integer :type, null: false, default: 0
```
sojan-official, line 5: i think there is a minor breakage here, probably resolving merge conflicts.

sojan-official (approved):

Issue #2472, opened 2021-06-21
feat: CSAT Responses


pranavrajs:
- [x] Add a `csat_response` table to capture the response 
      [point, feedback text, converastion_id, assigned_id]
- [x] Add an API to fetch the list of responses 
      (paginated - default page size: 25)
      (Date range filter)
- [x] Generate CSAT Records when contact interact with the component

sojan-official:
Raises https://github.com/chatwoot/chatwoot/pull/2503 for the first 3 items.
Let's take up the remaining reporting APIs in another PR

PR #2503, opened 2021-06-23
feat: CSAT APIs

sojan-official:
references: #2472

-  CSAT Responses Model
-  CSAT Responses API ( pagination ) ( date range filter )
-  CSAT Responses generation on component interaction


db/schema.rb
```diff
@@ -267,6 +267,23 @@
     t.index ["team_id"], name: "index_conversations_on_team_id"
   end
 
+  create_table "csat_survey_responses", force: :cascade do |t|
+    t.bigint "account_id", null: false
+    t.bigint "conversation_id", null: false
+    t.bigint "message_id", null: false
+    t.integer "rating", null: false
+    t.text "feedback_text"
```
pranavrajs, line 275: feedback is enough right?

sojan-official: At a glance, `csat_survey_responses.feedback` might be confusing, whether it's numeric feedback or text feedback.
We could also go with `additional_feedback` or  `feedback_message`.

pranavrajs (approved):
@sojan-official This looks good! Please create a separate PR for includes query to remove n+1 queries.

PR #2512, opened 2021-06-25
feat: Ability to create a new conversation if the previous conversation is resolved

### LINKED ISSUE #2466: Allow the user to create a new conversation if the previous one is resolved
- If CSAT survey is enabled
  - Disallow user to reply to a resolved conversation, show new conversation button
 
- If CSAT survey is disabled
  - The behaviour of the input box remains the same.

muhsin-k:
Fixes #2466


nithindavid (approved):

app/javascript/widget/views/Home.vue
```diff
@@ -163,6 +170,10 @@ export default {
         this.showAttachmentError = false;
       }, 3000);
     });
+    bus.$on(BUS_EVENTS.START_NEW_CONVERSATION, () => {
+      this.isOnCollapsedView = true;
```
pranavrajs, line 174: Do we need to set this?

app/javascript/widget/views/Home.vue
```diff
@@ -163,6 +170,10 @@ export default {
...
+      this.startNewConversation = true;
```
pranavrajs, line 175: Can we have a better naming for this variable? `startNewConversation` seems like an action rather than a state which holds the value of whether the pre-chat form is enabled or not.

app/views/api/v1/widget/messages/create.json.jbuilder
```diff
@@ -2,6 +2,7 @@ json.id @message.id
 json.content @message.content
 json.inbox_id @message.inbox_id
 json.conversation_id @message.conversation.display_id
+json.conversation_status @message.conversation.status
```
pranavrajs, line 5: I couldn't see this attribute being used anywhere. Do we need this change?

app/javascript/widget/components/ChatFooter.vue
```diff
@@ -1,36 +1,83 @@
 <template>
-  <footer class="footer">
+  <footer v-if="!hideReplyBox" class="footer">
     <ChatInputWrap
       :on-send-message="handleSendMessage"
       :on-send-attachment="handleSendAttachment"
     />
   </footer>
+  <custom-button
```
pranavrajs, line 8: `woot-button`?

pranavrajs:
@muhsin-k On QA, I noticed that the form display email and name field again. If it can be hidden by passing a prop from a parent, that would be nice. We would only need the message field there.

app/javascript/widget/views/Home.vue
pranavrajs: Never mind, we do.

muhsin-k: Yes.  It is used for skipping the team availability page on a new conversation. https://github.com/chatwoot/chatwoot/blob/186c3e225b4e50035df2a7e7494f1e3c8efac138/app/javascript/widget/views/Home.vue#L133

app/javascript/widget/views/Home.vue
muhsin-k: How about `ON_NEW_CONVERSATION`?

app/javascript/widget/components/ChatFooter.vue
muhsin-k: We are not using ```woot-button``` on the widget side anywhere. Will raise a separate PR for this

app/javascript/widget/views/Home.vue
pranavrajs: `isOnNewConversation` sounds good.

PR #2532, opened 2021-06-28
feat: Adds delete option for messages in conversation

nithindavid:
image


iamsivin (dismissed):

app/javascript/dashboard/modules/conversations/components/MessageContextMenu.vue
```diff
@@ -0,0 +1,75 @@
+<template>
+  <div class="context-menu">
+    <woot-button
+      icon="ion-more"
```
pranavrajs, line 4: Can we use 20px for this icon? It seems really small.

app/javascript/dashboard/components/widgets/conversation/Message.vue
```diff
@@ -211,6 +234,29 @@ export default {
...
     },
   },
+  methods: {
+    handleContextMenuClick() {
+      this.showContextMenu = !this.showContextMenu;
+    },
+    async handleDelete() {
+      const { conversation_id: conversationId, id: messageId } = this.data;
+      try {
+        await this.$store.dispatch('deleteMessage', {
```
pranavrajs, line 244: [vuex] unknown action type: deleteMessage

app/javascript/dashboard/i18n/locale/en/conversation.json
```diff
@@ -27,6 +27,8 @@
     "REMOVE_SELECTION": "Remove Selection",
     "DOWNLOAD": "Download",
     "UPLOADING_ATTACHMENTS": "Uploading attachments...",
+    "SUCCESS_DELETE_MESSSAGE": "Message deleted successfully",
```
pranavrajs, line 30: Please correct the spelling.

pranavrajs (changes requested):
I could get the option to work. When I click on delete message, it throws an error.

app/javascript/dashboard/components/widgets/conversation/Message.vue
pranavrajs: Never mind, I merged the store, it worked.

pranavrajs:
On QA, I figured that the dropdown goes behind the side panel.

Screenshot 2021-06-29 at 11 20 19 AM

nithindavid:
> On QA, I figured that the dropdown goes behind the side panel.
> 
> Screenshot 2021-06-29 at 11 20 19 AM

@pranavrajs Our current menu doesn't change position by context. Might need another PR to fix this.

pranavrajs:
Can't we add `position: fixed` to the dropdown pane so that it would show something like the one below? That way we don't need to put effort to fix the positioning based on context.

Screenshot 2021-06-29 at 2 05 05 PM

Screenshot 2021-06-29 at 2 04 59 PM

nithindavid:
@pranavrajs It won't move on scroll if we put fixed

nithindavid:
@pranavrajs here's a solution without fixed, Making the menu position by message type (left or right)

image

pranavrajs:
@nithindavid Works like a charm! Nice work!

PR #2534, opened 2021-06-29
chore: Change the notification text for mention notification type

### LINKED ISSUE #1711: Change the notification text for mention notification type.
Right now, the notification text for mention is `You have been mentioned in the #[conversation-id] by [user.name]`. Change this to the actual message.

Format: `[#<conversation-id>] <actual-message-content>`

eg: `[#34] This seems like a bug. @sojan-official can you please take a look?`

muhsin-k:
Fixes #1711

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -55,6 +55,7 @@ en:
       conversation_creation: "[New conversation] - #%{display_id} has been created in %{inbox_name}"
       conversation_assignment: "[Assigned to you] - #%{display_id} has been assigned to you"
       assigned_conversation_new_message: "[New message] - #%{display_id} %{content}"
+      conversation_mention: "You have been mentioned in conversation [ID - %{display_id}] by %{name}"
   conversations:
     messages:
       deleted: This message was deleted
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -55,7 +55,6 @@ en:
       conversation_creation: "[New conversation] - #%{display_id} has been created in %{inbox_name}"
       conversation_assignment: "[Assigned to you] - #%{display_id} has been assigned to you"
       assigned_conversation_new_message: "[New message] - #%{display_id} %{content}"
-      conversation_mention: "You have been mentioned in conversation [ID - %{display_id}] by %{name}"
```
pranavrajs, line 58: Isn't this text used in the email notifications?

pranavrajs (commented):
Nice work!

config/locales/en.yml
muhsin-k: No. I have verified by searching the entire project using the code ```notifications.notification_title. conversation_mention```

pranavrajs: I looked at the `conversation_reply_mailer`, it is not translated. 

I would suggest keeping this text for now instead of deleting it as the translation would also get deleted. It is an unnecessary effort from the  I will send a follow-up PR moving this text.

pranavrajs (approved):
Looks good to me, @sojan-official any thoughts of the helper methods?

sojan-official (approved):
LGTM

PR #2579, opened 2021-07-06
feat: Support messages along with the actions in Dialogflow response

### LINKED ISSUE #2570: Accept Multiple Responses From Dialogflow
When using handoff with Dialogflow, you want to inform the user what is happening. If the user is just put in handoff mode and sit there for minutes screaming into the void until an agent responds, that's a bad experience. Even on the Chatwoot website there is a demo showing that you can have the bot reply with a message when it hands the chat off to an agent, but this is not the case.

Currently, only one response is accepted by Chatwoot from Dialogflow. You can put multiple responses in Dialogflow, a custom payload and a text response. If a custom payload AND a text response is sent, both should be shown in chat. The current behavior is that the text will override the custom payload and the chat is never handed off to an agent like expected.

I've tried to play around with Dialogflow and setup multiple intents for the same input to try to get it to receive the 2 responses, but Sojan says that there is currently not a way to do so and that Chatwoot only accepts one response.

**Additional context**
image
image
image

Both responses are set up, but the chat never moves onto an agent like it should.

pranavrajs:
- Process every fulfilment message instead of the first one.
- Create multiple messages/actions based on the response

Fixes #2570

### DIFF
--- a/lib/integrations/dialogflow/processor_service.rb
+++ b/lib/integrations/dialogflow/processor_service.rb
@@ -39,14 +39,17 @@ class Integrations::Dialogflow::ProcessorService
   end
 
   def process_response(message, response)
-    text_response = response.query_result['fulfillment_text']
-
-    content_params = { content: text_response } if text_response.present?
-    content_params ||= response.query_result['fulfillment_messages'].first['payload'].to_h
-
-    process_action(message, content_params['action']) and return if content_params['action'].present?
-
-    create_conversation(message, content_params)
+    responses = response.query_result['fulfillment_messages']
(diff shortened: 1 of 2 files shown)


pranavrajs:
Tested it locally by creating a Dialogflow bot.

**DialogFlow Config**

Screenshot 2021-07-06 at 11 45 46 PM 

**Bot Behaviour**

https://user-images.githubusercontent.com/2246121/124647913-31728380-deb4-11eb-8378-59a629a40fa1.mp4

sojan-official (approved):

PR #2581, opened 2021-07-07
chore: Update availability status everywhere if the user changes the status from the account menu

### LINKED ISSUE #2580: bug: Availability status is not reflecting in the agent multi-select until we refresh the page
https://user-images.githubusercontent.com/12408980/124705780-cd7fa780-df13-11eb-83db-ae244686d44f.mov

muhsin-k:
Fixes #2580


app/javascript/dashboard/mixins/agentMixin.js
```diff
@@ -0,0 +1,35 @@
+import { mapGetters } from 'vuex';
+
+export default {
+  computed: {
+    getAgents() {
```
pranavrajs, line 5: As this is a computed property let us not try to use a variable name that resembles the method name. I guess the `get` or `check` prefixes is suited for methods rather than a computed property. A suggestion here is to name this property `assignableAgents`

app/javascript/dashboard/mixins/agentMixin.js
```diff
@@ -0,0 +1,35 @@
...
+          email: 'None',
+        },
+        ...agents,
+      ].map(item =>
+        item.id === this.currentUser.id
+          ? {
+              ...item,
+              availability_status: this.currentUser.availability_status,
+            }
+          : item
```
pranavrajs, line 31: Is it better to update the state when the action is performed instead of merging it here? Are there any other conditions which I have missed?

pranavrajs (commented):
Nice work on moving the logic to a mixin. I have 2 minor comments on the implementation.

app/javascript/dashboard/mixins/agentMixin.js
muhsin-k: Mainly this action is performing after the user changes the status from the account menu. To update the assignable agent's list, we need to pass the currently selected inbox id. Unfortunately, we are not keeping the `inboxId` in auth module. That's why I moved this logic to agent mixin.

pranavrajs: Sounds good.

pranavrajs (approved):

iamsivin (approved):

PR #2582, opened 2021-07-07
chore: Add fix for type error in message view

### LINKED ISSUE #2386: TypeError: cannot read property 'dataFetched' of undefined
https://sentry.io/share/issue/64ad988dbf5d4fcdaf76fb24f7fce059/

muhsin-k:
Fixes #2386


app/javascript/dashboard/components/widgets/conversation/MessagesView.vue
```diff
@@ -292,7 +292,9 @@ export default {
       this.setScrollParams();
 
       const dataFetchCheck =
-        this.getMessages.dataFetched === true && this.shouldLoadMoreChats;
+        this.getMessages &&
+        this.getMessages.dataFetched === true &&
+        this.shouldLoadMoreChats;
```
pranavrajs, line 297: I wouldn't go with this fix as this just suppresses the error and the actual problem is not solved.

The problem here I guess is that this view is getting render even when a conversation is not available.

pranavrajs:
@muhsin-k Closing this as the solution doesn't fix the issue. Let us discuss the possible solution.

PR #2600, opened 2021-07-09
chore: Disable delete option for deleted messages

### LINKED ISSUE #2583: bug: Showing delete option for deleted messages
Screenshot 2021-07-07 at 1 15 30 PM

muhsin-k:
Fixes #2583


app/javascript/dashboard/components/widgets/conversation/Message.vue
```diff
@@ -66,6 +65,7 @@
       v-if="isBubble"
       :is-open="showContextMenu"
       :show-copy="hasText"
+      :show-delete="showDelete"
```
pranavrajs, line 68: I guess we can disable the entire option. Copy doesn't make sense if the message is already deleted.

pranavrajs (approved):

Issue #2586, opened 2021-07-07
"This message was deleted"


iammattmartin:
When a message is deleted showing "This message was deleted", there should be an option to *not* show this on the widget.

If you're correcting a spelling, to have that show looks very bad.

The same applies to emails, it's looks incredibly unprofessional.

- This _really_ needs to be documented as to what the behaviour is going to be as I would not expect the widget to show the deleted message or this being sent by email, but would expect an admin to still be able to see this.

pranavrajs:
@iammattmartin Makes sense, thanks for the feedback. 

It creates confusion among the users when they send a message and it is deleted by the agent and disappears from the widget. The intention was to avoid this. So that when a user sends a message and the agent deletes it, the user would still be able to see the message with a deleted content. 

For agent side messages, it makes sense to avoid showing it on the widget side and email.

iammattmartin:
I agree with that. If the message deleted is from a user, that is fine - so they know what happened but I would say that staff doing so is to correct an error in a message that they've sent (and before delivery via email) etc.

pranavrajs:
Sure, that makes sense. We will update the behaviour.

PR #2614, opened 2021-07-13
fix: Hide deleted messages on widget side

### LINKED ISSUE #2584: bug: Attachments are showing after deleting a message
https://user-images.githubusercontent.com/12408980/124720966-50116280-df26-11eb-8869-918618dfb84f.mov

muhsin-k:
Fixes #2584 and #2586


pranavrajs:
As per the conversation we had at https://github.com/chatwoot/chatwoot/issues/2586#issuecomment-875594747, we could completely hide the bubble if the message is deleted.

muhsin-k:
> As per the conversation we had at [#2586 (comment)](https://github.com/chatwoot/chatwoot/issues/2586#issuecomment-875594747), we could completely hide the bubble if the message is deleted.

@pranavrajs That makes sense. I will update the PR accordingly

app/javascript/widget/mixins/messageMixin.js
```diff
@@ -0,0 +1,13 @@
+export default {
```
nithindavid, line 1: Nice work 🎉

nithindavid (approved):

Issue #2553, opened 2021-07-02
NoMethodError: undefined method contact


muhsin-k:
https://sentry.io/share/issue/14ccf0894aa646e6810436cd13991a76/

HariniKrishnan:
Hi I would like to contribute for this issue, Shall I go ahead and take it?

sojan-official:
@HariniKrishnan  Please go ahead. Let us know if you need any help.

HariniKrishnan:
Thanks a lot @sojan-official. Sure will reach out for help..

HariniKrishnan:
Hi can you please help me in reproducing this issue?

sojan-official:
@HariniKrishnan The issue occurs when chatwoot processes incoming emails from a contact who already exists in the database and the from address has a different character casing. 

ie `john@acme.inc` exists in the database and In the following line `processed_mail.from.first` returns '`John@acme.inc`
https://github.com/chatwoot/chatwoot/blob/a5bc81b304ac9d0ada34ee360b7642bc53c07793/app/mailboxes/support_mailbox.rb#L59

This would in turn leads to the `create_contact` method call, which would throw the `Email already taken` error, which was suppressed by the line 
https://github.com/chatwoot/chatwoot/blob/a5bc81b304ac9d0ada34ee360b7642bc53c07793/app/builders/contact_builder.rb#L59


-----

Fixes 

- Let's downcase `processed_mail.from.first` before performing the checks and passing it into other methods as an argument.

- Also After logging the error message in the following line, let's also raise the error so that we can catch the issue early.
https://github.com/chatwoot/chatwoot/blob/a5bc81b304ac9d0ada34ee360b7642bc53c07793/app/builders/contact_builder.rb#L59

ps: Setting up an actual chatwoot test environment with email could be tricky. So you could just approach it by writing an RSpec test case.  Let us know if you have questions.

HariniKrishnan:
Thanks a ton for detailed explanation @sojan-official I tried to setup but in vain. Sure will fix and test in rspec.

HariniKrishnan:
hi @sojan-official created PR. can you please review?

PR #2678, opened 2021-07-21
2553: fix for contact find exception

HariniKrishnan:
## Description

send a email of already exisiting contact with different case say "Care@example.com" , before this fix, it will throw error because of case mismatch it goes to contact create. Now it will retrieve existing contact

Fix for (https://github.com/chatwoot/chatwoot/issues/2553)

## Type of change

## How Has This Been Tested?

Spec file has been written and it passed locally.

## Checklist:


app/mailboxes/support_mailbox.rb
```diff
@@ -58,12 +58,15 @@ def create_conversation
...
   def find_or_create_contact
-    @contact = @inbox.contacts.find_by(email: processed_mail.from.first)
+    # changing to downcase to avoid case mismatch while finding
+    @contact = @inbox.contacts.find_by(email: processed_mail.from.first.downcase)
     if @contact.present?
       @contact_inbox = ContactInbox.find_by(inbox: @inbox, contact: @contact)
     else
       create_contact
     end
+  rescue StandardError => e
```
sojan-official, line 68: We won't need this rescue as we intend for it to error out in unexpected cases. We capture it in sentry and fix it. Rescuing here, In a way, suppresses that intention. so let's remove this.

We could also add a `raise` after the logging statement in this rescue block where the original error was suppressed.

https://github.com/chatwoot/chatwoot/blob/a7ca55c0801561bbd0c4a9e21a775569914a7d6c/app/builders/contact_builder.rb#L58

app/mailboxes/support_mailbox.rb
```diff
@@ -58,12 +58,15 @@ def create_conversation
   end
 
   def find_or_create_contact
-    @contact = @inbox.contacts.find_by(email: processed_mail.from.first)
+    # changing to downcase to avoid case mismatch while finding
+    @contact = @inbox.contacts.find_by(email: processed_mail.from.first.downcase)
```
sojan-official, line 62: We would also need to downcase the email passed into the contact builder in 
https://github.com/chatwoot/chatwoot/blob/a7ca55c0801561bbd0c4a9e21a775569914a7d6c/app/mailboxes/support_mailbox.rb#L75

Since we are repeating the `.downcase` method, another approach could be implementing a method inside  `chatwoot/app/presenters/mail_presenter.rb ` which downcases the from addresses by default so that we wouldn't need to repeat calling downcase 

example
```
def from
   @mail.from.map(&:downcase)
end
```
ref: https://ruby-doc.org/stdlib-2.5.1/libdoc/delegate/rdoc/SimpleDelegator.html

----

let's also downcase the email used for querying here
https://github.com/chatwoot/chatwoot/blob/a7ca55c0801561bbd0c4a9e21a775569914a7d6c/app/builders/contact_builder.rb#L45

sojan-official (changes requested):
Thanks a lot for the changes. I have added some review comments. 
Please feel free to let me know if you have questions or suggestions.

app/mailboxes/support_mailbox.rb
HariniKrishnan:
```suggestion
  rescue StandardError => e
     Rails.logger.info "#{e.message} in find/create_contact :: backtrace: #{e.backtrace}"
     raise e
   end
```
Shall I change like the above one?!

sojan-official: @HariniKrishnan Let's  remove the rescue block from `app/mailboxes/support_mailbox.rb` all-together and place it in `chatwoot/app/builders/contact_builder.rb` instead with the `raise e` as you suggested. 

Since `create_contact` calls `ContactBuilder`, let's log the error at the root itself.

HariniKrishnan: sure will do that

sojan-official:
@HariniKrishnamoorthy Checking in to see if you need any help with this. Also, let us know if your bandwidth is limited. 
We will take up the PR from here and get this merged.

HariniKrishnan:
@sojan-official Hi yes on it, past two weeks I didn't had time to work, that's why there is a delay, sorry for this... we will push ASAP

spec/fixtures/files/welcome.eml
```diff
@@ -1,4 +1,4 @@
-From: Sony Mathew <sony@chatwoot.com>
+From: Sony Mathew <Sony@chatwoot.com>
```
HariniKrishnan, line 1: changed here to ensure it shouldn't break anything for existing cases.

HariniKrishnan:
@sojan-official so sorry for late fix, seriously caught up and so sorry for not updating prior, it was easy fix, could've done earlier. Thanks for detailed explanation in review comments helped understand the issues quickly. made changes please verify

sojan-official (approved):
@HariniKrishnan Thanks a lot for the changes.  

I have made a couple of minor changes to this PR. 
- Removed the additional test case for support_mailbox, primarily because we have tests for mail decorator that ensures from address to be always downcased. 
- minor fixes to make rubocop happy 

Will QA and get this merged.

HariniKrishnan:
@sojan-official thanks a lot, this my very first contribution to open source, there were lot of learnings, thank you for your patience and guidance means a lot.

PR #2689, opened 2021-07-22
feat: Add APIs for custom attribute definitions

### LINKED ISSUE #2655: feat: Custom attribute schema
The custom attributes are rendered as text right now. This is not ideal. We need to allow rich content rendering for the attributes. To support this, add an option to save a custom attribute schema of conversation/contact. The custom attribute schema would define the type of the object which would help the frontend to render the value.

- Create a model to save the custom attribute schema with the following attributes
    - attribute_key - The key which will be used.
    - attribute_type - Let us start with `number`, `text`, `currency`, `percent`, `link`.
    - attribute_display_name - How will it be displayed in the UI.
    - default_value - Provide a default value to the attribute.
    - type_of_object - Contact / Conversation
- CRUD APIs for custom attributes.
- Migration to add the existing custom attributes to the newly constructed schema objects.
    - Create an in-memory array to save the available custom attribute.

pranavrajs:
- Add schema to store custom attribute definition
- Add APIs for CRUD operations on Custom Attribute Definitions

Fixes #2655

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_07_14_110714) do
+ActiveRecord::Schema.define(version: 2021_07_22_095814) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_stat_statements"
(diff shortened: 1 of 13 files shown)


db/schema.rb
```diff
@@ -262,7 +262,7 @@
     t.datetime "agent_last_seen_at"
     t.jsonb "additional_attributes", default: {}
     t.bigint "contact_inbox_id"
-    t.uuid "uuid", default: -> { "gen_random_uuid()" }, null: false
+    t.uuid "uuid", default: -> { "public.gen_random_uuid()" }, null: false
```
sojan-official, line 265: intended ?

sojan-official (commented):
Should we limit the access to these APIs to administrators alone?

db/schema.rb
pranavrajs: No. Updated it.

pranavrajs:
@sojan-official Please take a look at this.

sojan-official (approved):

PR #2709, opened 2021-07-26
fix: Update tweet character count logic

### LINKED ISSUE #2695: Messages missing in Twitter
Long messages sent from Chatwoot are missing on Twitter. There are no logs to debug this. We would have to figure out whether it is something with the character encoding/escaping.

TODO:

- [x] Fix sending long messages.
- [x] Show an indication to the agent if the message is not delivered to the end-user.

pranavrajs:
- Consider the tweet author's username in the character count.
- Display Tweet source link

Screenshot 2021-07-26 at 11 05 13 PM

Fixes #2695


sojan-official (approved):

app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
```diff
@@ -198,7 +205,7 @@ export default {
       }
       if (this.isATwitterInbox) {
         if (this.conversationType === 'tweet') {
-          return MESSAGE_MAX_LENGTH.TWEET;
+          return MESSAGE_MAX_LENGTH.TWEET - this.replyToUserLength - 2;
```
nithindavid, line 208: Can we make this more readable ?

app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
```diff
@@ -235,9 +242,22 @@ export default {
...
+        sender: {
+          additional_attributes: { screen_name: screenName = '' } = {},
+        } = {},
+      } = selectedTweet;
+      return screenName ? screenName.length : 0;
+    },
     isMessageEmpty() {
-      if(!this.message) {
-        this.message = '';
+      if (!this.message) {
```
nithindavid, line 259:
```suggestion
      return !this.message
```

app/javascript/dashboard/components/widgets/conversation/bubble/Actions.vue
```diff
@@ -113,6 +127,13 @@ export default {
   }
 }
 
+.right {
+  .ion-reply,
+  .ion-android-open {
```
nithindavid, line 132: Can we target with a generic class name like, `action-icons` or something ? So we don't have to rely on ion classes.

nithindavid (changes requested):
Nice PR 👍 Have left some minor comments

app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
pranavrajs: Oops! Sorry I didn't notice this review message. I will fix it in a separate one.

app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
pranavrajs: Doesn't make much difference IMHO.

app/javascript/dashboard/components/widgets/conversation/bubble/Actions.vue
pranavrajs: This makes sense, I will create a separate PR for the same, missed this review.

PR #2749, opened 2021-08-03
feat: CSAT for all channels

### LINKED ISSUE #2653: CSAT for every channel (backend)
Right now, CSAT surveys are supported only in the web-widget channel. To support CSAT surveys in all channels, we would start by sending a link once the conversation is resolved.

Once the conversation is resolved, the system should create a message with input_csat type. In each send_on_service listener, there should be an additional check to alter the content to send the link for the CSAT survey.

URL format: `<installation-url>/app/responses/<response-uuid>`

**Todo**:

- [x] #2671
- [ ] Update API to provide rating and response. Save the response in both the message object as well as the CSAT response object.
- [ ] Behaviour: 
   - [ ] Lock interaction based on the day: If the CSAT is a message sent more than 14 days before, then throw 404.
   - [ ] If already submitted, add an additional flag on the response to show that the message is submitted.

**Notes**

- Shouldn't work for twitter

muhsin-k:
Fixes #2653


app/models/message.rb
```diff
@@ -130,6 +130,13 @@ def webhook_data
     }
   end
 
+  def content
+    # move this to a presenter
+    return self[:content] if !input_csat? || !inbox.web_widget?
+
+    "Please rate this conversation , #{ENV['FRONTEND_URL']}/survey/responses/#{conversation.uuid}"
```
sojan-official, line 137: Lets add translations for this

sojan-official (commented):
The code looks good to me. Let's do a round of testing for the behavior on all channels and get this merged. 

- shouldn't get triggered for tweet conversations

sojan-official:
@pranavrajs  @muhsin-k  we might need to revisit whether we should start new conversations for Facebook and Twitter on resolution.

sojan-official:
## email 

Screenshot 2021-08-21 at 12 08 47 AM

## response page

Screenshot 2021-08-21 at 12 08 38 AM

sojan-official (dismissed):

sojan-official:
@nithindavid  could you take a look at the front end code, if it's all good let's get this merged?

app/javascript/survey/views/Response.vue
```diff
@@ -33,26 +47,138 @@
...
+    isRatingSubmitted() {
+      return this.surveyDetails && this.surveyDetails.rating;
+    },
+    isFeedbackSubmitted() {
+      return this.surveyDetails && this.surveyDetails.feedback_message;
+    },
+    isButtonDisabled() {
+      return !(this.selectedRating && this.feedback);
+    },
+    enableBanner() {
```
nithindavid, line 101: Would be nice to see `isSomething` pattern being followed like other functions to keep it consisent.

app/javascript/survey/components/Feedback.vue
```diff
@@ -0,0 +1,55 @@
...
+    <label class="text-base font-medium text-black-800">
+      {{ $t('SURVEY.FEEDBACK.LABEL') }}
+    </label>
+    <text-area
+      v-model="feedback"
+      class="my-5"
+      :placeholder="$t('SURVEY.FEEDBACK.PLACEHOLDER')"
+    />
+    <custom-button
+      class="flex items-center font-medium float-right"
```
nithindavid, line 12: We could avoid float classes as we have flex styles. We could wrap it in a div to float it right.

nithindavid (changes requested):
@muhsin-k Nice work :D I have left some minor comments

nithindavid (approved):

PR #2757, opened 2021-08-05
feat: Use relay server for self-hosted mobile push notifications

pranavrajs:
- Use relay server


Brokray:
If I understand this MR correctly we will be able to have push notifications using the official mobile app even if we are pointing to a self-hosted instance of Chatwoot?
Will there be any config to do in the self-hosted instance?

pranavrajs:
@Brokray Yes, that is correct. You don't have to do anything on your end. You will be able to use the official mobile app and receive notifications without any additional step.

PR #2796, opened 2021-08-12
fix: Fixes parsing issue with custom attributes

nithindavid:
Fix for - https://sentry.io/share/issue/dcecadd36fbb4bbab9a11199bb0e8147/

- Checks for type of value before parsing 
- Now supports object/array and number along with string
- Displays undefined and null as undefined and null

image


muhsin-k:
**QA** has been done in local environment

https://user-images.githubusercontent.com/12408980/129313802-21ceffeb-88a0-465d-aa79-b7cb5ba8035d.mov

muhsin-k (approved):

PR #2810, opened 2021-08-13
chore: parallel runs in circle

sojan-official:
- todo 
- [ ] Fix coverage reporting


tejaswinichile:
@pranavrajs  @sojan-official I fixed the parallelism and code-coverage issue in the last commit.

sojan-official:
@tejaswinichile  Great work 🔥 

- Just noted that this PR shows an increase of 8% change in coverage. We might need to look into how
Screenshot 2021-08-20 at 7 37 00 PM

- Should we also look into splitting the frontend tests ? At the moment i think they are run in all 4 containers

tejaswinichile:
> Should we also look into splitting the frontend tests? At the moment I think they are run in all 4 containers
Yes, I can do it when I'm done with other tasks. 
cc @sojan-official  Let's keep this open so that we will fix it in one go.

PR #2841, opened 2021-08-19
chore: consider X-original-sender to create contact in case of group mail

### LINKED ISSUE #2715: chore: Email parser should consider `x-original-sender` header
**Problem:**

For emails forwarded from google groups, Google rewrites the FROM address to the group email and the original email will be available under `X-Original-Sender`. This is not handled in Chatwoot at the moment. So instead of displaying the original email as the sender, Chatwoot would show group email as the sender. 

**Fix:**

- Handle the case for 'X-Original-Sender' while creating a contact from the email inbox.

**More context:**

We are using Action Mailbox for parsing emails. In the support mailbox, we are not handling the case mentioned above. https://github.com/chatwoot/chatwoot/blob/develop/app/mailboxes/support_mailbox.rb#L61

tejaswinichile:
## Description

Find or Create contact with X-Original-Sender if present in case of mail replied by the member of the google groups.

Fixes #2715 

## Type of change

## How Has This Been Tested?

Tested through local specs by adding the new header field `X-Original-Sender`

## Checklist:


app/mailboxes/support_mailbox.rb
```diff
@@ -41,6 +41,10 @@ def decorate_mail
     @processed_mail = MailPresenter.new(mail, @account)
   end
 
+  def original_sender
```
sojan-official, line 44: let's have this method in the `MailPresenter` itself so that all custom logic over the mail object remains in the same place.

sojan-official (changes requested):
added a minor comment. let me know your thoughts.

app/mailboxes/support_mailbox.rb
tejaswinichile: Cool 💯

sojan-official (approved):

PR #2948, opened 2021-09-02
enhancement: Improves design of conversation sidebar

nithindavid:
image


nithindavid:
@iamsivin You could use this design on the contact page side panel also.

pranavrajs (approved):

PR #2975, opened 2021-09-07
feat: Configure Chatwoot & Analytics for SaaS app

pranavrajs:
This PR introduces Chatwoot inside Chatwoot and PostHog as an analytics provider.

- This would not affect any Self-hosted installations, the SDK gets initialized only if the token and set in InstallationConfig.

### DIFF
--- a/app/views/devise/token.json.jbuilder
+++ b/app/views/devise/token.json.jbuilder
@@ -1,4 +1,7 @@
 json.payload do
   json.success true
   json.partial! 'auth.json.jbuilder', resource: @resource
+  json.data do
+    json.created_at @resource.created_at
+  end
 end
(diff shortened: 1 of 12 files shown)


app/views/devise/token.json.jbuilder
```diff
@@ -1,4 +1,7 @@
 json.payload do
   json.success true
   json.partial! 'auth.json.jbuilder', resource: @resource
+  json.data do
+    json.created_at @resource.created_at
+  end
```
pranavrajs, line 6: @sojan-official @tejaswinichile I didn't want to update this attribute for all other APIs, that is why this is written like this. Partial has a `data` attribute, it is getting merged without issues.

PR #2988, opened 2021-09-09
chore: Move Facebook event processing to worker

sojan-official:
- Move Facebook event processing to worker


tejaswinichile (approved):
LGTM 👍🏼

PR #3021, opened 2021-09-15
chore: Use the name of the sender from the mail object

### LINKED ISSUE #2911: Use the name of the sender from the mail object
### Context

Right now, the name of the contact in an email channel is taken from the first part of the email. We need to extract the name of the user from the mail object.

For eg: in the email below, the name of the contact would be saved as Sony (from Sony@chatwoot.com) instead of Sony Mathew.

https://github.com/chatwoot/chatwoot/blob/65f3e83afd9cdf1d5f94a818ab89d5ad50a31d49/spec/fixtures/files/welcome.eml#L1-L3

### Todo

- [x] Parse the name from the email object.
- [x] Update support/reply mailbox to use the name.

tejaswinichile:
# Pull Request Template

## Description

if the sender email is `Sony Mathew <Sony@chatwoot.com>` Contact should be built with `Sony Mathew`

Fixes #2911 

## Type of change

## How Has This Been Tested?

- Added email channel
- Sent an email to the registered email
- checked the sender email and name in the contact sidebar

## Checklist:


app/mailboxes/support_mailbox.rb
```diff
@@ -82,6 +82,6 @@ def create_contact
   end
 
   def identify_contact_name
-    processed_mail.from.first.split('@').first
+    processed_mail.sender_name
```
pranavrajs, line 85: If there is no name in the email, then the system auto-generates a random name for the user. To avoid that, please provide the first line as a fallback. 

Screenshot 2021-09-15 at 6 25 34 PM

pranavrajs (changes requested):

app/mailboxes/support_mailbox.rb
tejaswinichile: oh o! Yeah 👍🏼

PR #3060, opened 2021-09-21
fix: set default content type as text for message

### LINKED ISSUE #2974: Bug: content_type is nil for new message in webhook events
ref: https://discord.com/channels/647412545203994635/650257251856285707/884475687170342962 

> Hi, we have connected with a bridge server to send message from chatwoot to whatsapp by configuring webhook url and all that. So, when we try to get the content_type in req.body, because only then different types of messages can be sent (as for now, we have done sending just text messages), the value is null. How to get the content_type as it is showing null? I have attached the app logs for reference

Chatwoot_content_type

## Acceptence 
- [ ] Ensure content_type can't be null
- [ ] Backfill old data with content_type : 'text' for old messages through a migration

murtaza-bagwala:
## Description

Set content_type as text if its null.
Update content_type as text for all the messages where it was null.

Fixes #2974 

## Type of change

## How Has This Been Tested?

Tried passing payload without content_type and it set  content_type as text by default.

## Checklist:


sojan-official:
Hey @murtaza-bagwala  Thanks a lot for the changes.

But one more thing we have to be conscious about is that we have mesages being created from places other than `app/builders/messages/message_builder.rb` , So rather than fixing it in each of these places. Let do it right at the model itself, we can also add a validation on the message model for presence validation as well. 

example :

chatwoot/app/models/message.rb
```
before_validation :ensure_content_type
validates :content_type, presence: true


 private

  def ensure_content_type
    self.content_type ||= 'text'
  end

```

------
or we could just set a default at the database level itself through a migration and skip the logic in code all together 
ref: https://blog.arkency.com/how-to-add-a-default-value-to-an-existing-column-in-a-rails-migration/

```
change_column_null( :messages, :content_type, false, '0')
```

I feel more lenient towards doing it at the database level.

murtaza-webonise:
Yes, makes more sense

I am making these changes

murtaza-bagwala:
Yes, makes more sense

I am making these changes

spec/controllers/api/v1/accounts/conversations/messages_controller_spec.rb
```diff
@@ -50,6 +50,7 @@
         expect(response_data['content_attributes']['external_created_at']).to eq time_stamp
         expect(conversation.messages.count).to eq(1)
         expect(conversation.messages.last.sender_id).to eq(agent_bot.id)
+        expect(conversation.messages.last.content_type).to eq('text')
```
tejaswinichile, line 53: Can we add one more test with content blank and then check content type.

sojan-official: ^^ I guess we could just have the test in model spec since the code changes are done there.

PR #3066, opened 2021-09-22
feat: API to Filter reports by teams

### LINKED ISSUE #2916: Feat: API to  Filter reports by teams
- [ ] API to Filter reports by teams
- [ ] Update API to accept team information
- [ ] API to download team report
- [ ] Swagger documentation updates

aswindevps:
## Description

This feature allows the user to filter reports by teams along with account labels, agents and inboxes.

Fixes #2916

## Type of change

## How Has This Been Tested?

1. Existing reports API's should function properly after the API changes.
2. Verify the team reports/summary API's via Postman.
3. Verify the CSV downloading API for teams.

## Checklist:


app/builders/v2/report_builder.rb
```diff
@@ -56,6 +58,10 @@ def label
     @label ||= account.labels.where(id: params[:id]).first
   end
 
+  def team
+    @team ||= account.teams.where(id: params[:id]).first
```
pranavrajs, line 62: `.find` is enough right?

app/models/team.rb
```diff
@@ -40,4 +40,12 @@ def add_member(user_id)
   def remove_member(user_id)
     team_members.find_by(user_id: user_id)&.destroy
   end
+
+  def messages
+    account.messages.where(conversation_id: conversations.pluck(:id))
```
tejaswinichile, line 45: To improve query: we can have index on `index_messages_on_account_id_and_conversation_id unique: true`
I ran explain on the above query on my local and might be expensive in the future with a lot of records:
```Sort  (cost=45.05..46.00 rows=381 width=245)
   Sort Key: created_at
   ->  Seq Scan on messages  (cost=0.00..28.72 rows=381 width=245)
         Filter: ((account_id = 1) AND (conversation_id = ANY ('{33,16,18,1,9,2,19,32,8,10,11,5,6,17,4,13,34,30,15,12,3,7,14}'::integer[])))
(4 rows)```

We can take it later in the tech debt also just have it mentioned somewhere we should unique: true index on the combination of the conversation_id and account_id in the message table.

tejaswinichile (changes requested):

app/builders/v2/report_builder.rb
aswindevps: `.find` would throw an exception if record not found but `.where` return an empty array.

tejaswinichile: In that case use `find_by(id: params[:id])`.

aswindevps: Is there any specific reason for `where` condition? Because i can see the same for inbox, user and label above.

@sojan-official

tejaswinichile: `find_by` uses `where(id: params[:id]).take` internally which is faster than `where(id: params[:id]).first` so as we need first matching record always go with find or find_by according to if you want to handle the exception or not.

app/models/team.rb
sojan-official: Let's take the improvements as separate tech debt.

app/builders/v2/report_builder.rb
sojan-official: @aswindevps  Using .find isn't a bad choice here.  The reason why .find would throw and exception is because the param supplied to the method is an invalid team id.  ( wrong id format or a team id that doesn't belong to the account )
In such cases we would want the error to be thrown, which will be handled by https://github.com/chatwoot/chatwoot/blob/0c24df96a8444b7e85b0a8395deb9c34371767aa/app/controllers/concerns/request_exception_handler.rb#L12 and we would throw a 404 .

we could switch to find for labels and inbox as well.

app/models/team.rb
sojan-official: as a part of the tech debt we can also revisit the specs around reports and see how we can improve the overall execution time there.

sojan-official:
Except for the suggestion on switching to find, the PR looks good to me. Lets create a tech debt for the other discussion and get this merged.

app/builders/v2/report_builder.rb
aswindevps: Switched to `.find` method for inboxes, agents, labels and teams.

PR #3074, opened 2021-09-23
chore: Move agent availability to Account level

sojan-official:
## Changes

- [ ] Move agent availability to account level 
- [ ] API for admins to manage other agents availability
- [ ] API to disable the agents going offline automatically

### DIFF
--- a/app/controllers/api/v1/accounts/agents_controller.rb
+++ b/app/controllers/api/v1/accounts/agents_controller.rb
@@ -16,7 +16,7 @@ class Api::V1::Accounts::AgentsController < Api::V1::Accounts::BaseController
 
   def update
     @agent.update!(agent_params.except(:role))
-    @agent.current_account_user.update!(role: agent_params[:role]) if agent_params[:role]
+    @agent.current_account_user.update!(role: agent_params[:role], availability: new_agent_params[:availability], auto_offline: new_agent_params[:auto_offline],) if agent_params[:role]
     render partial: 'api/v1/models/agent.json.jbuilder', locals: { resource: @agent }
   end
 
@@ -51,18 +51,20 @@ class Api::V1::Accounts::AgentsController < Api::V1::Accounts::BaseController
       account_id: Current.account.id,
       user_id: @user.id,
       role: new_agent_params[:role],
-      inviter_id: current_user.id
+      inviter_id: current_user.id,
+      availability: new_agent_params[:availability],
+      auto_offline: new_agent_params[:auto_offline],
     )
   end
 
   def agent_params
-    params.require(:agent).permit(:email, :name, :role)
+    params.require(:agent).permit(:email, :name, :role, :availability, :auto_offline)
   end
 
   def new_agent_params
     # intial string ensures the password requirements are met
     temp_password = "1!aA#{SecureRandom.alphanumeric(12)}"
-    params.require(:agent).permit(:email, :name, :role)
+    params.require(:agent).permit(:email, :name, :role, :availability, :auto_offline)
           .merge!(password: temp_password, password_confirmation: temp_password, inviter: current_user)
   end
 
--- a/app/controllers/api/v1/profiles_controller.rb
+++ b/app/controllers/api/v1/profiles_controller.rb
@@ -2,7 +2,6 @@ class Api::V1::ProfilesController < Api::BaseController
   before_action :set_user
 
   def show
-    render partial: 'api/v1/models/user.json.jbuilder', locals: { resource: @user }
   end
 
   def update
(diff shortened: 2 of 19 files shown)


app/controllers/api/v1/accounts/agents_controller.rb
```diff
@@ -9,19 +9,16 @@ def index
     @agents = agents
   end
 
-  def destroy
```
sojan-official, line 12: cleaned up agents controller slightly while enabling it to accept the additional params `availability` and `auto_offline`

app/controllers/api/v1/profiles_controller.rb
```diff
@@ -15,19 +13,26 @@ def update
     @user.update!(profile_params)
   end
 
+  def availability
```
sojan-official, line 16: new routes in the profile for the agent to update their availability. 
new route because now updating availability will requires the specific account id

app/controllers/api/v1/profiles_controller.rb
```diff
@@ -15,19 +13,26 @@ def update
...
+    @user.account_users.find_by!(account_id: availability_params[:account_id]).update!(availability: availability_params[:availability])
+  end
+
   private
 
   def set_user
     @user = current_user
   end
 
+  def availability_params
```
sojan-official, line 26: at the moment configuring `auto_offline` is not available at the agent level. admin determines it.

app/javascript/dashboard/api/auth.js
```diff
@@ -161,9 +161,9 @@ export default {
     });
   },
 
-  updateAvailability({ availability }) {
-    return axios.put(endPoints('profileUpdate').url, {
-      profile: { availability },
+  updateAvailability(availabilityData) {
```
sojan-official, line 164: change this to work for the new backend endpoint

app/javascript/dashboard/components/layout/AvailabilityStatus.vue
```diff
@@ -26,7 +26,9 @@
                 color-scheme="secondary"
                 class-names="status-change--dropdown-button"
                 :is-disabled="status.disabled"
-                @click="changeAvailabilityStatus(status.value)"
+                @click="
```
sojan-official, line 29: the component doesn't depend on the current user anymore. But on current account id

app/javascript/dashboard/store/modules/auth.js
```diff
@@ -40,7 +40,11 @@ export const getters = {
   },
 
   getCurrentUserAvailabilityStatus(_state) {
-    return _state.currentUser.availability_status;
+    const { accounts = [] } = _state.currentUser;
```
sojan-official, line 43: getting online status from account users instead.

app/javascript/shared/helpers/BaseActionCableConnector.js
```diff
@@ -1,6 +1,6 @@
 import { createConsumer } from '@rails/actioncable';
 
-const PRESENCE_INTERVAL = 60000;
```
sojan-official, line 3: reduced the time frame to make the presence feel more real time

app/models/account_user.rb
```diff
@@ -24,15 +26,20 @@
 #
 
 class AccountUser < ApplicationRecord
+  include AvailabilityStatusable
```
sojan-official, line 29: moving availability to account user

app/models/concerns/availability_statusable.rb
```diff
@@ -2,29 +2,29 @@ module AvailabilityStatusable
   extend ActiveSupport::Concern
 
   def online_presence?
-    return if user_profile_page_context?
```
sojan-official, line 5: cleaned up the logic to make it more readable

lib/online_status_tracker.rb
```diff
@@ -1,5 +1,5 @@
 module OnlineStatusTracker
-  PRESENCE_DURATION = 60.seconds
+  PRESENCE_DURATION = 20.seconds
```
sojan-official, line 2: to keep up with the frontend change

sojan-official:
@muhsin-k  Please see the impact of this on mobile apps

app/controllers/api/v1/accounts/agents_controller.rb
```diff
@@ -9,19 +9,16 @@ def index
...
-    @agent.current_account_user.destroy
-    head :ok
-  end
+  def create; end
 
   def update
-    @agent.update!(agent_params.except(:role))
-    @agent.current_account_user.update!(role: agent_params[:role]) if agent_params[:role]
-    render partial: 'api/v1/models/agent.json.jbuilder', locals: { resource: @agent }
+    @agent.update!(agent_params.slice(:name).compact)
```
tejaswinichile, line 15: Are the roles check already available, who can modify the agent's data and he/she is not destroying his own record?

sojan-official: @tejaswinichile We have a policy check which only allows admins to access `create`, `edit`, and `destroy` endpoints. `index` is available for all users ( both admins and agents ) in the system.  

- An admin can update the name of the user in the system
- An admin can remove himself from the account 
- An admin can remove other users from the account ( both agents and admins )

sojan-official:
@muhsin-k  We have the following API changes 
- endpoint to switch the availability has been changed to  post api/v1/profile/availability which accepts account_id and availability for params 
- for the `user` object in auth calls, `availability` has been moved to the nested accounts objects instead of the previous parent object

PR #3128, opened 2021-10-03
feat: Autogenerate vapid keys for push notifications

### LINKED ISSUE #1589: Autogenerate push notification keys
Right now, our users have to generate a VAPID key and update in the ENV for the browser notification to work. This is an unnecessary additional step.

- Autogenerate VAPID keys on installation or initial boot. (https://github.com/zaru/webpush#generating-vapid-keys)
- Save the generated VAPID keys to the database and cache it in Redis along with global cache.
- Use it for registering push notification on the web.

This reduces a step to install Chatwoot. 🔥

santhosh-chinnasamy:
# Pull Request Template
Fixes #1589

## Type of change

## How Has This Been Tested?

- ran rspec locally
- in browser, confirmed keys are accessbile from window.chatwootConfig object via debugger console.

## Checklist:


sojan-official (changes requested):
@santhosh-chinnasamy, thanks a lot for the changes. It looks good. But we might want to take a slightly different approach here. 

- [ ]  Creating new data models and columns just for these two keys is overkill. So instead, Chatwoot has a data model called InstallationConfig, which you can rely upon. 
ref: https://github.com/chatwoot/chatwoot/blob/develop/app/models/installation_config.rb

- [ ] We should factor in existing installations where they have these keys configured in environment variables. So we should add migrations or write code in a way that relies on that as a fallback. 

- [ ] need a hook in seeding to populate the values for new installations ( already covered )
---

But Considering the latter two, let's follow this approach instead. 

- [ ] Create a new vapid service class that tries to fetch existing keys from the database. If unavailable, it will try fetching the ENV variable / generate value using Webpush and then creating a new record with either. 
ref: https://github.com/chatwoot/chatwoot/blob/40d0b2faf32dd7f66a65469f6de24144af4a2717/lib/chatwoot_hub.rb#L8
 - Saves us from writing the migration for older clients
 - Still, support ENV variables until we completely deprecate them


Let me know if you have any thoughts, questions, or comments.

santhosh-chinnasamy:
@sojan-official Okay.

vishnu-narayanan:
@santhosh-chinnasamy Could you please pull the latest changes from `develop` and resolve the conflicts ?

santhosh-chinnasamy:
Okay I'll do it.

vishnu-narayanan (approved):
LGTM

PR #3139, opened 2021-10-05
Fix/1925 delete user while removing him from the account

### LINKED ISSUE #1925: Delete user account while removing from the account
If the user is associated with only one account, delete the user record also while removing it from the account.

#### Problem

When a user is removed from an account, we are removing the base user account considering that the user can be present in more than one account. This creates orphan user records in the system if the user is associated with only one account.

#### Solution

Delete user account also if the user has only one account associated with them.

#### Acceptance criteria

- Remove user from the account
  - If the user has only one account associated with them, delete the user record also. 
  - If the user has more than one account, remove the association only.

#### Special considerations

- Removing a user from an account should remove them from the teams, inboxes
- Removing a user from an account should not remove messages sent by the user, instead, it should point to a null value and should be handled in the UI.

murtaza-bagwala:
# Pull Request Template

## Description

Delete user(with no accounts) from the DB while removing from the account.

Fixes #1925

## Type of change

## How Has This Been Tested?

Tried creating users with different roles, added them to different accounts, and then removed them.

## Checklist:

### DIFF
--- a/app/controllers/super_admin/account_users_controller.rb
+++ b/app/controllers/super_admin/account_users_controller.rb
@@ -13,6 +13,7 @@ class SuperAdmin::AccountUsersController < SuperAdmin::ApplicationController
   def destroy
     if requested_resource.destroy
       flash[:notice] = translate_with_resource('destroy.success')
+      return redirect_to super_admin_account_url(requested_resource.account.id) if requested_resource.user_deleted?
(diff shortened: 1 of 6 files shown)


murtaza-bagwala:
@sojan-official not sure why this testcase is getting failed.

PR #3151, opened 2021-10-06
feat: Delete user while removing him from the account

### LINKED ISSUE #1925: Delete user account while removing from the account

murtaza-bagwala:
# Pull Request Template

## Description

Delete user (with no accounts) while removing from the account

Fixes #1925 

## Type of change

## How Has This Been Tested?

Tried creating different users with different roles across different accounts and deleting them.

## Checklist:


spec/controllers/super_admin/account_users_controller_spec.rb
```diff
@@ -19,4 +19,17 @@
...
+
+  describe 'DELETE /super_admin/account_users/:id' do
+    context 'when user has only one account' do
+      let(:account_user) { create(:account_user) }
+
+      it 'redirects to the account details' do
+        sign_in super_admin
+        delete "/super_admin/account_users/#{account_user.id}"
+        expect(response).to have_http_status(:redirect)
+        expect(response).to redirect_to("/super_admin/accounts/#{account_user.account.id}")
```
tejaswinichile, line 31: Can you check if the user is not present or the user is destroyed after the delete method?

tejaswinichile (changes requested):

spec/controllers/super_admin/account_users_controller_spec.rb
tejaswinichile: @murtaza-bagwala Please notify once you update the comment.

murtaza-bagwala: sure

sojan-official:
closing this as we haven't been actively working on this for a while. Let's reopen if we prioritise this in future.

PR #3177, opened 2021-10-09
feat: add healthcheck

### LINKED ISSUE #2238: Health-check / status endpoint

When deploying behind a load balancer or other frontend, a relatively cheap health-check is desirable that indicates that the application is ready to receive requests but doesn't do heavy things like template rendering.

An endpoint that simply returns 200 OK if Chatwoot is healthy and ready to receive requests.

Ideally the health check would also take into account PG/Redis/etc connectivity, but it ideally would not couple the rate of that check to the rate that the health check endpoint is called at, because in some cloud environments the health check endpoint can be called at quite a high rate.

If it's practical, checking the health of db/redis/etc every N seconds and storing a health status, and then returning that stored health status from the health check endpoint, would be ideal. Alternatively some short-term caching mechanism could be used on the health check endpoint's response.

I've searched the repository and documentation to see if there is an existing health endpoint but could not find one. At present I'm simply using `/`, but the template rendering, request processing, etc, is quite heavy so the health checks cause considerable load.

santhosh-chinnasamy:
Fixes #2238

## Type of change

## How Has This Been Tested?
- by checking urls exposed by the gem in browser and curl

## Checklist:


sojan-official:
Closing this PR since there wasn't much activity over this for a while. Let's reopen this once this becomes a priority again.

PR #3292, opened 2021-10-27
feat: Add Null values at the last while sorting

tejaswinichile:
## Description

Consider null values at the last while applying sort


tejaswinichile:
@sojan-official Added the spec for checking the null values remains at last.

pranavrajs:
@tejaswinichile Made a small change to change the sort on county using the name instead of code. The reason is that the country for some countries is different from its name (for eg: United Arab Emirates is AE). This would come before Argentina and would create confusion that the sort is not working.

app/javascript/dashboard/routes/dashboard/contacts/components/ContactsTable.vue
```diff
@@ -178,12 +178,14 @@ export default {
         {
           field: 'company',
           key: 'company',
+          sortBy: this.sortConfig.company_name || '',
```
nithindavid, line 181: Instead of `''` we could use `undefined`

nithindavid (changes requested):

tejaswinichile:
@pranavrajs @sojan-official Please have a look at this one.

pranavrajs (approved):

sojan-official (approved):

app/javascript/dashboard/routes/dashboard/contacts/components/ContactsTable.vue
tejaswinichile: @nithindavid I have used it as we used it for other filters, e.g email, and phone_number. Should I change from `''` to `undefined` for the previous filters too?

nithindavid (approved):

PR #3293, opened 2021-10-27
feat: Adds support for paste attachment with keys.

iamsivin:
Fixes # (issue)

## Type of change

## How Has This Been Tested?

1. Copy a file or image.
2. Come back to the app and open a chat.
3. And press ⌘ + V.

## Checklist:


app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
```diff
@@ -314,11 +314,20 @@ export default {
...
+    document.removeEventListener('paste', this.onPaste);
   },
   methods: {
+    onPaste(e) {
+      const data = e.clipboardData.files;
+      if (!data) {
+        return;
+      }
+      this.onFileUpload(data[0]);
+    },
```
nithindavid, line 330:
```suggestion
    onPaste(e) {
      const data = e.clipboardData.files;
      if (!data.length || !data[0]) {
        return;
      }

      const file = data[0];
      const { name, type, size } = file;
      this.onFileUpload({ name, type, size, file });
    },
```

nithindavid: We can keep like this to get rid of other changes. All the other changes can be reverted I think.

nithindavid (changes requested):

nithindavid (approved):

PR #3298, opened 2021-10-28
feat: IMAP Email Channel

### LINKED ISSUE #2520: IMAP ingest
As a head of customer success, for a business with a single email support ingress, I would like to support imap email ingest, so that omni-channel support is supported (inbound). So that, the adminstrative overhead of configuring each agent their own email is avoided, a single record of customer communication (ideally per-issue) can be maintained.

When I added an inbox with email, I sort of expected to need to provide IMAP details, so that like Zendesk, all emails to support could then be further redirected to the appropriate areas using rules and filters.

Currently, I can setup SES with an S3 bucket, and use a lambda to classify emails, which can be made available to bespoke apps or an API.

**Additional context**
Seemed like a missing feature. I'm just evaluating the self-hosted edition. Really impressive, but this did seem like a critical missing feature for any business using something like this.

aswindevps:
This change allows the user to configure both IMAP and SMTP for an email inbox. IMAP enables the user to see emails in Chatwoot. And user can use SMTP to reply a email conversation.

User can use the default settings to send and receive emails for email inbox if both IMAP and SMTP are disabled.

Fixes #2520

## Type of change

## How Has This Been Tested?

1. Configure both IMAP and SMTP for outlook email.
2. Send a mail from gmail.
3. Check new emails as conversations in Chatwoot after configuring the IMAP.
4. Reply to the email conversation from Chatwoot.
5. Check for reply mail in gmail.
6. Verify the outlook email ID in the reply mail in gmail.

## Checklist:


robguthrie:
I'm so excited about this. This is what we require to adopt Chatwoot. Happy to test when it's at that stage.

config/schedule.yml
```diff
@@ -3,13 +3,12 @@
...
-  class: "Internal::CheckNewVersionsJob"
+  cron: '0 12 */1 * *'
+  class: 'Internal::CheckNewVersionsJob'
   queue: scheduled_jobs
 
 # executed At every 5th minute..
 trigger_scheduled_items_job:
-  cron: "*/5 * * * *"
-  class: "TriggerScheduledItemsJob"
+  cron: '*/1 * * * *'
```
sojan-official, line 12: should we change this back to 5 minutes?  or we could choose something in between and update the comment above.

spec/mailboxes/imap/imap_mailbox_spec.rb
```diff
@@ -0,0 +1 @@
+
```
sojan-official, line 1: intentional?

sojan-official (commented):
Code Looks good to me. 
We could add more specs for `app/mailboxes/imap/imap_mailbox.rb` to ensure coverage.

spec/mailboxes/imap/imap_mailbox_spec.rb
aswindevps: Working on this spec.

config/schedule.yml
aswindevps: We can change it back to 5 min.

app/mailboxes/imap/imap_mailbox.rb
```diff
@@ -0,0 +1,76 @@
+class Imap::ImapMailbox
+  include MailboxHelper
+  attr_accessor :channel, :account, :inbox, :conversation, :processed_mail
+
+  def process(mail, channel)
+    @inbound_mail = mail
```
sojan-official, line 6: lets add a check here to prevent the email loop for chatwoot notification emails: 

ref: https://github.com/chatwoot/chatwoot/blob/e5a80195ab5516e4c3f4a55b8fb72c32338d2e0d/app/mailboxes/support_mailbox.rb#L10

sojan-official:
Minor comment about adding a check to prevent notification email loop. The code looks good to be otherwise, let's QA and get this merged.  

+ create a doc issue to track documentation.

app/mailboxes/imap/imap_mailbox.rb
aswindevps: Done

sojan-official (approved):
LGTM

aswindevps:
Below scenarios are tested and verified.

1. The user sends an email, the agent replies to the email, user replies to the email.
    **The replies to the conversation should not create a separate conversation.**
2. User sends an email, the agent sends a reply, the agent sends a reply, user replies to the first mail.
    **The replies to the conversation should not create a separate conversation.**
3. User sends an email, the user sends a follow-up, the agent sends a reply, user replies mail.
    **The replies to the conversation should not create a separate conversation.**
4. The user sends an email, after that Chatwoot inbox is created. Send a follow-up to the first mail. Send another to the first mail. 
    **Even though the parent is not available in Chatwoot, it should show as the same thread.**
5. If I have more than one inbox connect in Chatwoot and the user sends the same message to both inboxes.
    **It should create conversation in both inboxes with the same message.**

sojan-official:
> I'm so excited about this. This is what we require to adopt Chatwoot. Happy to test when it's at that stage.

@robguthrie This change is now available on the chatwoot develop branch and you should be able to start using it.

robguthrie:
Whoooo hoooo! Congratulations. It's been a pleasure to watch the development process you have.  I'll try to spin it up this weekend.

Lewiscowles1986:
Thanks for working on this. I'd forgotten about chatwoot until this point; but it is great to see features rolled out in this way. I'll forward to the person I was looking at support management solutions for.

entttom:
What mail service are you guys using? every case from above creates a new conversation (gmail free as service provider).

> Below scenarios are tested and verified.
> 
> 1. The user sends an email, the agent replies to the email, user replies to the email.
>    **The replies to the conversation should not create a separate conversation.**
> 2. User sends an email, the agent sends a reply, the agent sends a reply, user replies to the first mail.
>    **The replies to the conversation should not create a separate conversation.**
> 3. User sends an email, the user sends a follow-up, the agent sends a reply, user replies mail.
>    **The replies to the conversation should not create a separate conversation.**
> 4. The user sends an email, after that Chatwoot inbox is created. Send a follow-up to the first mail. Send another to the first mail.
>    **Even though the parent is not available in Chatwoot, it should show as the same thread.**
> 5. If I have more than one inbox connect in Chatwoot and the user sends the same message to both inboxes.
>    **It should create conversation in both inboxes with the same message.**

PR #3313, opened 2021-11-02
Bug: Instagram render template message twice

### LINKED ISSUE #3196: Instagram account conversation triggers welcome message twice
If a new conversation is created in Instagram account inboxes, then the welcome message is triggered twice. However, it is sent only once to the sender.

| Dashboard view | Instagram sender view |
| -- | -- |
| Screenshot 2021-10-11 at 1 51 33 PM |  translate |

tejaswinichile:
# Pull Request Template

## Description

When we send a welcome message to the Instagram user that message comes back again to save it in the Chatwoot inbox, we should consider a template message as a repeated message and neglect that.

Fixes #3196 

## Type of change

## How Has This Been Tested?

Tested on staging

## Checklist:

### DIFF
--- a/app/builders/messages/instagram/message_builder.rb
+++ b/app/builders/messages/instagram/message_builder.rb
@@ -38,6 +38,10 @@ class Messages::Instagram::MessageBuilder < Messages::Messenger::MessageBuilder
     @outgoing_echo ? :outgoing : :incoming
   end
 
+  def message_identifier
+    message[:mid]
+  end
+
   def message_source_id
     @outgoing_echo ? recipient_id : sender_id
   end
@@ -64,10 +68,6 @@ class Messages::Instagram::MessageBuilder < Messages::Messenger::MessageBuilder
 
   def message_content
     @messaging[:message][:text]
-  end
-
-  def content_attributes
-    { message_id: @messaging[:message][:mid] }
   end
 
   def build_message
@@ -103,22 +103,17 @@ class Messages::Instagram::MessageBuilder < Messages::Messenger::MessageBuilder
       account_id: conversation.account_id,
       inbox_id: conversation.inbox_id,
       message_type: message_type,
-      source_id: message_source_id,
+      source_id: message_identifier,
       content: message_content,
-      content_attributes: content_attributes,
       sender: @outgoing_echo ? nil : contact
     }
   end
 
   def already_sent_from_chatwoot?
     cw_message = conversation.messages.where(
-      source_id: nil,
-      message_type: %w[template outgoing],
-      content: message_content,
-      private: false,
-      status: :sent
+      source_id: @messaging[:message][:mid]
     ).first
-    cw_message.update(content_attributes: content_attributes) if cw_message.present?
+
     cw_message.present?
   end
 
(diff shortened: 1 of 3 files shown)


tejaswinichile:
This is how it was looking before
Screenshot 2021-11-02 at 4 11 06 PM
>

sojan-official:
@tejaswinichile  While looking into the code, I just noticed that we are handling the way messages are being sent to Facebook and Instagram differently. 

For facebook -> send the message and update the message-id to the source id 
https://github.com/chatwoot/chatwoot/blob/03b1a3d045bbfc221bcf07c8155eccd2d0d88905/app/services/facebook/send_on_facebook_service.rb#L19

for Instagram, -> sending the message without storing the source id. 
https://github.com/chatwoot/chatwoot/blob/03b1a3d045bbfc221bcf07c8155eccd2d0d88905/app/services/instagram/send_on_instagram_service.rb#L60

I think if possible, we should have the message updated with the Instagram message Id as well. This will let us handle the echo message better. We can check whether we already have a matching message with the particular source id before creating a new message.  

The content scan logic we have at the moment might break in cases where an agent replied same message body from chatwoot dashboard as well directly from facebook or Instagram.  Chatwoot would only create a single message with this case instead of creating two ?

Also the source id will ensure that we can show a sent status for that particular message more reliably

tejaswinichile:
@sojan-official When I checked it last time because I wanted to handle it in the same way in Facebook update, but in Instagram it wasn't returning any message-id let me confirm it again.

tejaswinichile:
@sojan-official I just got it, I didn't do it because the message id was not available for message reaction. As we are not handling message reaction right now I have added message-id changes to check the duplicate message.

app/builders/messages/instagram/message_builder.rb
```diff
@@ -103,7 +107,7 @@ def message_params
       account_id: conversation.account_id,
       inbox_id: conversation.inbox_id,
       message_type: message_type,
-      source_id: message_source_id,
+      source_id: message_indentifier,
```
pranavrajs, line 110: identifier?

pranavrajs (commented):
Looks like there is a typo in the method name.

app/builders/messages/instagram/message_builder.rb
tejaswinichile: Took some time to realize that.

app/builders/messages/instagram/message_builder.rb
```diff
@@ -112,12 +116,9 @@ def message_params
...
     cw_message = conversation.messages.where(
-      source_id: nil,
-      message_type: 'outgoing',
-      content: message_content,
-      private: false,
-      status: :sent
+      source_id: @messaging[:message][:mid]
     ).first
+
     cw_message.update(content_attributes: content_attributes) if cw_message.present?
```
sojan-official, line 122: Is updating the content_attributes of the message intentional? Could we add a comment here on why we need to do this?

tejaswinichile: It's not necessary now since we are storing the same in the message source id.
Removing the same.

sojan-official (approved):

Issue #2879, opened 2021-08-25
feat: Refactor actions on widget side to support multiple conversations


nithindavid:
**Estimation** - 2
- Currently the store doesn't support multiple conversations. We have started the efforts to refactor the store here -> #2851
- This is a follow up to #2851 to make the actions compatible with the mentioned pull request
- While refactoring make sure to move existing business logic in actions to helpers to keep the store logic clean and simple
- With current reactor we have made the mutation to do a single state key update rather than doing a task. So create a message in state we have to call `addConversationEntry` and `addConversationId`  to complete the task.  So when it comes upto actions we need to follow task based approach which in turn calls multiple mutations to achieve that task.
- Adds specs for the action and helpers

More context can be found here https://github.com/chatwoot/chatwoot/issues/2355#issuecomment-898370642

PR #3344, opened 2021-11-09
feat: Updates widget conversation state to support multiple conversations

nithindavid:
`widget/v2` is the feature branch where we need to merge the new state.

More info about state can be found here #2879

### DIFF
--- a/app/javascript/widget/store/modules/conversation/actions.js
+++ b/app/javascript/widget/store/modules/conversation/actions.js
@@ -22,7 +22,7 @@ export const actions = {
           conversationId,
         });
         commit(
-          'messageV2/addMessagesEntry',
+          'message/addMessagesEntry',
           { conversationId, messages: [lastMessage] },
           { root: true }
         );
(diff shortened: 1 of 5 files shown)


app/javascript/widget/store/modules/conversation/actions.js
```diff
@@ -1,112 +1,171 @@
...
+          { root: true }
+        );
+        commit('addMessageIdsToConversation', {
+          conversationId,
+          messages: [lastMessage],
+        });
+      });
     } catch (error) {
-      // Ignore error
+      throw new Error(error);
```
fayazara, line 35: I think this throw inside a catch will be ignored, since it is not inside a try block.

fayazara (commented):
Looks, good. Have left a smal question/doubt.

app/javascript/widget/store/modules/conversation/actions.js
nithindavid: This will be thrown again, which an be caught when the action is called in UI. So we will know it has failed or not to show errors in UI.

fayazara (approved):
Looks good on the codefront

PR #3354, opened 2021-11-10
Limit redis and postgres access to localhost

Jnig:
## Description

The default docker-compose configuration will expose redis, postgres and rails directly to the internet when the service is started on a virtual machine.

In most cases that is not what you want, and especially for redis and postgres exposing the services could be a potential security risk. By adding 127.0.0.1 access is limited to localhost and access is only possible after nginx oder another web server is configured as reverse proxy.

```
0.0.0.0:6379->6379/tcp         chatwood_redis_1
0.0.0.0:5432->5432/tcp         chatwood_postgres_1
```

## Type of change

## How Has This Been Tested?

Modified docker-compose file was started with docker-compose up

## Checklist:


tejaswinichile:
@vishnu-narayanan Can you check this?

vishnu-narayanan (approved):
@Jnig Thank you for the PR. 😄

docker-compose.production.yaml
```diff
@@ -13,7 +13,7 @@ services:
       - postgres
       - redis
     ports:
-      - 3000:3000
+      - '127.0.0.1:3000:3000'
```
sojan-official, line 16: @vishnu-narayanan  should we enable the binding for 3000 alone to be open. It should be relatively harmless considering it's just the webserver.  
+ it will make it easy for someone to verify the installation on a VPS before setting up a frontend proxy like Nginx?

vishnu-narayanan: @sojan-official Since this is the production file, I prefer we go with the secure default. To verify the installation, one can do something like, `curl -I localhost:3000` or change the production compose file to drop the `127.0.0.1` binding.

sojan-official: sounds good 👍

vishnu-narayanan:
Note: Moving forward, anyone using `docker-compose.production.yaml` need to have something like `Nginx`to proxy the requests to the container. 

If you want to verify whether the installation is working, try `curl  -I localhost:3000` to see if it returns `200`. Also, you could temporarily drop the `127:0.0.1:3000:3000` for rails to `3000:3000` to access your instance at http://<external-ip>:3000. It's recommended to revert this change back and use `Nginx` in front.

PR #3386, opened 2021-11-15
chore: Prevent notification email loop

sojan-official:
Configuring an agent email also as a support email inbox leads to conversations getting created in a loop if notifications were also configured to the same email.


app/mailboxes/support_mailbox.rb
```diff
@@ -7,6 +7,9 @@ class SupportMailbox < ApplicationMailbox
                     :decorate_mail
 
   def process
+    # prevent loop from chatwoot notification emails
+    return if notification_email_from_chatwoot?
```
tejaswinichile, line 11: We can just add it in `before_processing`, as it will stop the cycle of finding the channel and loading the account.

tejaswinichile (commented):
One small comment everything else LGTM 👍🏼

app/mailboxes/support_mailbox.rb
sojan-official: @tejaswinichile I am using the decorated mail object inside the helper method to identify the `original_sender` of the email. Since the callbacks are interdependent. let's take it as refactoring for later.

tejaswinichile (approved):
💯

PR #3402, opened 2021-11-17
feat: Remove ionicons v2 dependency on widget

pranavrajs:
- Remove IonIcons font from the widget(66kb less during load).
- Use FluentIcons on the widget
- Create a component for fluent icons


fayazara (dismissed):
The new icons look really modern. Did a QA locally.

nithindavid (dismissed):
Code looks good, I'll do QA and will update here.

app/javascript/widget/assets/scss/_utilities.scss
```diff
@@ -0,0 +1,3 @@
+.icon-button {
+  @include button-size;
```
nithindavid, line 2: Since tailwind is already there, we could stick to `w-8` and `h-8` for the same size.

PR #3405, opened 2021-11-17
Chore: Improve conversation auto resolution

### LINKED ISSUE #1583: chore: Improve Conversation Auto resolution job logic
When we implement the conversation auto resolution job, chatwoot didn't have a mechanism to implement cronjobs, Now that we have `sidekiq-cron` . let's revisit this implementation.  

## Current Logic 
A single job is enqueued for each conversation to keep track of resolution. It also gets re-enqueued when the conversation is reopened 
ref: https://github.com/chatwoot/chatwoot/blob/a2764e5c1dbd65abbf6e26d50d974d39d4bb1ffc/app/models/conversation.rb#L174

## Proposed Logic
Add a hook into scheduled jobs and follow a Model similar to how we execute the campaign jobs.  
ref: https://github.com/chatwoot/chatwoot/blob/a2764e5c1dbd65abbf6e26d50d974d39d4bb1ffc/app/jobs/trigger_scheduled_items_job.rb#L6

- [ ] scheduled jobs will iterate over accounts with auto resolution configured and scheduled a new account conversation resolution job 
- [ ] Account conversation resolution job will fetch conversations that has passed auto resolution time period and will resolve them 
- [ ] Clean up the existing logic that schedules a single job per conversation.

shivamsinghchahar:
# Pull Request Template

## Description

Add a cron job that checks all of the accounts that have auto resolution configured. The cron job then triggers another job that marks all the auto resolvable conversations as `resolved` for an account.

The old background job has also been removed.

Fixes #1583 

## Type of change

New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Tested manually and also added tests for the new background jobs that were added.

## Checklist:


app/jobs/conversations/resolution_job.rb
```diff
@@ -0,0 +1,8 @@
+class Conversations::ResolutionJob < ApplicationJob
+  queue_as :medium
+
+  def perform(account:)
+    resolvable_conversations = account.conversations.open.select(&:auto_resolve?)
```
sojan-official, line 5: This will load up all the open conversation objects and does computation in memory which might not be necessary in this case.  Instead, let's query from the database only the conversations that have passed the resolution period. 

feel free to clean up if required, move it to a scope etc.

```suggestion
    resolvable_time = Time.now.utc - account.auto_resolve_duration.days
    resolvable_conversations = account.conversations.open.where('last_activity_at < ? ', resolvable_time )
```

app/models/conversation.rb
```diff
@@ -151,6 +151,15 @@ def recent_messages
     messages.chat.last(5)
   end
 
+  def auto_resolve?
```
sojan-official, line 154: this method won't be required, since are querying only the required objects in the job itself.

config/schedule.yml
```diff
@@ -12,4 +12,9 @@ trigger_scheduled_items_job:
   cron: "*/5 * * * *"
   class: "TriggerScheduledItemsJob"
   queue: scheduled_jobs
-  
\ No newline at end of file
+
+# executed At 00:00 every night
+account_conversations_resolution_scheduler_job:
```
sojan-official, line 17: rather than having a separate schedule for this job. let's run it inside the "TriggerScheduledItemsJob" 

- We will have one job that triggers all the routine maintenance jobs
- Also conversations will be resolved through out the day by the system, rather than as one batch at start of every day.

sojan-official (commented):
@shivamsinghchahar  Thanks a lot for the changes. Looks good for the most part. I have added a couple of review comments. Let's QA this post that.

shivamsinghchahar:
@sojan-official Thanks a lot for the review! I have made the changes. Please let me know if more changes are needed.

PR #3406, opened 2021-11-17
fix: Set default value of sortBy to empty string

pranavrajs:
`sortBy` needs the default value to be an empty string not undefined.


pranavrajs:
@tejaswinichile Contact sort is not working in prod, this PR fixes it.

tejaswinichile (approved):
LGTM 💯

PR #3415, opened 2021-11-18
feat: Use vue-router on widget route management

pranavrajs:
- Use Vue router for widget route management
- Use dynamic import for minimizing initial loadsize


muhsin-k:
@pranavrajs I have found a small issue while testing the flow. The read messages appear again when the user refreshes the widget window

https://user-images.githubusercontent.com/12408980/148332959-ae0865d8-1056-497f-857c-3416a71a54cc.mov

PR #3432, opened 2021-11-22
fix: Pass label title instead of id to filter API

pranavrajs:
- Use label title instead of id in the Filter API as per the changes made in #3418


fayazara (approved):
Looks good.

PR #3438, opened 2021-11-22
Update schema files

tejaswinichile:
Update schema files

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_11_22_061012) do
+ActiveRecord::Schema.define(version: 2021_11_22_112607) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_stat_statements"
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -281,6 +273,28 @@
     t.index ["phone_number"], name: "index_channel_whatsapp_on_phone_number", unique: true
   end
 
+  create_table "companies", force: :cascade do |t|
```
pranavrajs, line 276: We don't have companies table yet, we might need another PR to clean this up  😆

tejaswinichile: @pranavrajs Ohh is it my changes? Let me check this. :D

tejaswinichile:
@sojan-official This https://github.com/chatwoot/chatwoot/pull/3438#issuecomment-975396972 is a good approach, I am gonna do it `actions` table.
Any idea why Am I getting companies table changes?

sojan-official:
@tejaswinichile  you might have the companies table in your local database from an older migration. Whenever you run `rake db:migrate` rails will create a new schema file with your local database copy.  Two options to fix this is

- if you have data in your local you don't want to lose, delete the table using raw SQL
- or  checkout develop and do a rake db:reset

tejaswinichile:
@sojan-official Oh yes! now I remember that I was working on custom attributes to save in companies table like 2-3 months back 😁  Yes, we don't need this change in development yet.

db/schema.rb
```diff
@@ -326,7 +318,7 @@
     t.datetime "agent_last_seen_at"
     t.jsonb "additional_attributes", default: {}
     t.bigint "contact_inbox_id"
-    t.uuid "uuid", default: -> { "public.gen_random_uuid()" }, null: false
+    t.uuid "uuid", default: -> { "gen_random_uuid()" }, null: false
```
sojan-official, line 321: this change doesn't seem intended.

db/schema.rb
```diff
@@ -326,7 +318,6 @@
     t.datetime "agent_last_seen_at"
     t.jsonb "additional_attributes", default: {}
     t.bigint "contact_inbox_id"
-    t.uuid "uuid", default: -> { "public.gen_random_uuid()" }, null: false
```
sojan-official, line 329: @tejaswinichile 😅 this is needed though

tejaswinichile: I just did reset :D

Issue #2557, opened 2021-07-02
Strange permission issue for a user having agent and admin access to different Accounts


hackedunit:
**Describe the bug**

When a user has an agent access to an Account and also an Admin access to a different Account, the user is unable to click on Inboxes or create an Inbox under the account with Admin access. Looks like the permissions are not being applied to the user after he switches to the Account with admin access. Strange thing is that if the permission (in super admin) is applied in the order of 1. Admin access 2. Agent access, this would work fine. But if its the other way around, the issue exists.

**To Reproduce**

1. Add a user, say Paul, as an agent
2. Add Paul again to a new account as an Administrator
3. Login as Paul and switch to the account with Admin access
4. Try to create an Inbox, you will see its not clickable.

**Expected behavior**

Paul should be able to create an Inbox as he has Admin access under that Account.

**Screenshots**

In this order, Paul won't be able to create Inbox:

<img width="1472" alt="CleanShot 2021-07-02 at 15 17 11@2x" src="https://user-images.githubusercontent.com/13460/124256140-96427c80-db48-11eb-9fd7-9cbf9d627ecc.png">

This works fine:

<img width="1462" alt="CleanShot 2021-07-02 at 15 18 11@2x" src="https://user-images.githubusercontent.com/13460/124256264-ba9e5900-db48-11eb-8307-ef6847c01560.png">

**Environment**

Tested on both self hosted and Chatwoot.com - was able to reproduce on both.

sojan-official:
@hackedunit  Thanks for reporting this, we will take a look.

pranavrajs:
I can confirm this bug. 

**Todo:**

- [ ] Remove the active account user from the user model.
- [ ] Deprecate `role` from the user payload.
- [ ] Use roles from accounts array in the response payload.
- [ ] Make changes in the UI to use account role instead of role attribute.

PR #3445, opened 2021-11-23
fix: Use role permissions from accounts instead of role attribute

pranavrajs:
Fixes #https://github.com/chatwoot/chatwoot/issues/2557


sojan-official:
Test Cases 

with multiple accounts. 
- [ ] route to login if not logged in
- [ ] route to dashboard from the login page if logged in
- [ ] general usability with multiple accounts

PR #3451, opened 2021-11-23
Fix: fixing mail to and in_reply_to issues

tejaswinichile:
Fix: 
1. https://sentry.io/organizations/thoughtwoot-04/issues/2730446421/?project=5410824&query=is%3Aunresolved&statsPeriod=14d
2. https://sentry.io/organizations/thoughtwoot-04/issues/2742345718/?project=5410824&query=is%3Aunresolved&statsPeriod=14d


tejaswinichile:
I am going to add test cases tomorrow.

app/mailboxes/reply_mailbox.rb
```diff
@@ -26,6 +26,8 @@ def find_relative_conversation
   end
 
   def conversation_uuid_from_to_address
+    return if mail.to.nil?
```
sojan-official, line 29: if `mail.to.nil?` we might need to look into the `X-forwarded-for` header I believe. Lets add a method to parse the `to` address in our mail decorator as well. 

cc: @pranavrajs

pranavrajs: Yes, please check if mail gem already handles this, otherwise let us add a fallback.

tejaswinichile: Okay, let me make this change and I'll add the specs for this new decorator.

tejaswinichile:
@sojan-official Can you check this.

app/mailboxes/reply_mailbox.rb
```diff
@@ -26,6 +26,8 @@ def find_relative_conversation
   end
 
   def conversation_uuid_from_to_address
+    return check_forwarded_for_email if mail.to.blank?
```
sojan-official, line 29: rather than placing the logic in this file, I think we could do some clean-up and implement a method in https://github.com/chatwoot/chatwoot/blob/develop/app/presenters/mail_presenter.rb#L89  . 

We could add a method similar to `original_sender`, called `original_reciever` with a comment on how the `to` could be null in specific cases.
https://github.com/chatwoot/chatwoot/blob/c23e2c23d4151a2879b33e5e74c808565246730f/app/presenters/mail_presenter.rb#L89

replace the mail.to with  `decorated_mail_instance.original_reciever` 

A person new to the codebase could totally miss out on the reasons why we implemented this check.  + all the mail-related logic in one place would help us to handle it for all mailboxes without logic duplication.

sojan-official (commented):
@tejaswinichile  The changes look good to me. But I think it's worth investing some time into this and rewriting the logic a bit for the sake of future maintainability.  If you think it's a larger effort,  we can add a fix me block and create a tech debt for the same. 

Notes for QA 

- we will need to recreate this specific scenario with an email with empty .to and ensure that there won't be any breakages

app/mailboxes/reply_mailbox.rb
tejaswinichile: @sojan-official 
I was thinking that but we have this sequence in the reply mailbox,
```rb
before_processing :conversation_uuid_from_to_address,
                    :find_relative_conversation,
                    :verify_decoded_params,
                    :decorate_mail
```
So didn't get much time to check why we have decorated email after `conversation_uuid_from_to_address` and we need `x-forwarded-for` email in the `conversation_uuid_from_to_address`

If we can spend a little more time on this I can figure out and refactor this. I will convert this to draft for now, I will pick this up after the automation rule changes.

sojan-official: @tejaswinichile  sounds good 👍 . I had a quick look and seems like we could do the following

 - We could decouple the account related logic in MailPresenter
 - decorate the mail at the start of the sequence using mail presenter 
 - replace the `mail.x` calls with `decorated_object.x` calls 
 - and initialize the final mail object combined with the account logic at the end of the sequence.  
 
But it seems like a larger effort. so either we could park this issue till we get to it, or we could have a separate tech debt for the same.  Either sounds good to me.

tejaswinichile:
@sojan-official Let's park this, I will get to it by coming Friday.

tejaswinichile:
@sojan-official Could you take a look at this?

sojan-official (approved):
Code LGTM. 
let's QA and get this merged

tejaswinichile:
@sojan-official I have tested it with the production example, I have added one more spec after that as the original mail do contain `in-reply-to` address, so that case is being handled by the current change. Please have a look.

tejaswinichile:
New finding: email in the issue has `in_reply_to` email with <reply+uuid@something.com> pattern

tejaswinichile:
RCA of the issue: 

This particular mail: `globalid: gid://chatwoot/ActionMailbox::InboundEmail/367238`
Routed to reply mailbox using Message source_id which is present in the system.

1. When it gets to the reply mailbox it checks if the to email address array has any matching conversation patterns, but as `mail.to` is nil and here we got the exception.
2. This particular email doesn't have a forwarded-for email address but we are handling that going forward.
3. Added test cases to make sure: To email address, X-forwarded-for address, and in-reply-to with conversation uuid is not present.

sojan-official (approved):

PR #3453, opened 2021-11-24
fix: Render checkbox if string value passed 

muhsin-k:
Before adding rich types in custom attributes, all the boolean values are stored in string format ("true"/"false"). This fix will render all these values in boolean format.

### How Has This Been Tested?

1. Create a new contact attribute with type **checkbox**
2. Set attribute via SDK method

Ex: 

```
 window.$chatwoot.setCustomAttributes({
  cloudCustomer: "false" // "true"
});
```
3. Check this attribute is rendering properly in the contact sidebar

### DIFF
--- a/app/javascript/dashboard/components/CustomAttribute.vue
+++ b/app/javascript/dashboard/components/CustomAttribute.vue
@@ -154,10 +154,10 @@ export default {
 
   computed: {
     formattedValue() {
-      if (this.attributeType === 'date') {
+      if (this.isAttributeTypeDate) {
         return format(new Date(this.value || new Date()), DATE_FORMAT);
       }
-      if (this.attributeType === 'checkbox') {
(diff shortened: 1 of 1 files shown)


app/javascript/dashboard/components/CustomAttribute.vue
```diff
@@ -156,6 +153,15 @@ export default {
   },
 
   computed: {
+    formattedValue() {
+      if (this.attributeType === 'date') {
+        return format(new Date(this.value || new Date()), DATE_FORMAT);
+      }
+      if (this.attributeType === 'checkbox') {
```
iamsivin, line 160:
```suggestion
      if (this.isAttributeTypeCheckbox) {
```

iamsivin:
Please see the comment. Looks good 👍🏻

iamsivin (approved):

Issue #3503, opened 2021-12-02
bug: CCing people from other conversation


Robinjoeh:
**Describe the bug**

Chatwoot is CC:ing contacts based on the previous message.

**To Reproduce**

1. Open a conversation
2. Reply to a conversation and add CC recipient, email will be sent correctly
3. Reply to another conversation (from a different person) and the person you CC:ed in the previous conversation will be included in this one as well. The "CC" field in the conversation will still be shown empty, and you will not see that you have mistakenly CC:ed the person until the message is sent.

**Environment**

Self-hosted, v1.22.1.

Robinjoeh:
Have you been able to reproduce this? It has happened several times now, and if you need more info I'll be happy go gather more details.

sojan-official:
@Robinjoeh We have been able to reproduce the issue and we are working on fix for this

Robinjoeh:
@sojan-official Perfect, keep up the good job!

PR #3515, opened 2021-12-07
fix: Fixes cc bcc being sent when mail head is empty  

nithindavid:
Fixes: #3503

### DIFF
--- a/app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
+++ b/app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
@@ -310,9 +310,6 @@ export default {
         this.showMentions = false;
       }
     },
-    ccEmails(newVal) {
-      console.log(newVal);
-    },
   },
 
   mounted() {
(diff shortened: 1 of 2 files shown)


app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
```diff
@@ -309,6 +310,9 @@ export default {
         this.showMentions = false;
       }
     },
+    ccEmails(newVal) {
```
fayazara, line 313: Can remove this block.

fayazara (changes requested):
Looks good. Minor correction

app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
```diff
@@ -280,7 +281,7 @@ export default {
       return !this.message.trim().replace(/\n/g, '').length;
     },
     showReplyHead() {
-      return !this.isOnPrivateNote && this.isAnEmailChannel;
+      return true || (!this.isOnPrivateNote && this.isAnEmailChannel);
```
muhsin-k, line 284: Revert this change since reply email head only for email inboxes

nithindavid, line 284:
```suggestion
      return !this.isOnPrivateNote && this.isAnEmailChannel;
```

PR #3525, opened 2021-12-08
chore: Fix pre-commit hooks

muhsin-k:
- Upgraded husky from 6 to 7
- Git hooks are configured using individual files for hooks in a `.husky/` directory.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -28,6 +28,8 @@ module.exports = {
     }],
     'vue/html-self-closing': 'off',
     "vue/no-v-html": 'off',
+    'vue/singleline-html-element-content-newline': 'off',
+    'vue/require-default-prop': 'off',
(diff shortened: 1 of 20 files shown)


.eslintrc.js
```diff
@@ -28,6 +28,9 @@ module.exports = {
     }],
     'vue/html-self-closing': 'off',
     "vue/no-v-html": 'off',
+    'vue/singleline-html-element-content-newline': 'off',
+    'vue/require-default-prop': 'off',
+    'vue/require-prop-types': 'off',
```
muhsin-k, line 33: This change was intentional in order to make small PR. I will create separate PR for fixing all the Eslint issues

pranavrajs: Made it a warning instead of error.

PR #3526, opened 2021-12-08
chore: Add the ability to use same key for conversation and contact

muhsin-k:
- Added attribute model in custom attribute index

### DIFF
--- a/app/models/custom_attribute_definition.rb
+++ b/app/models/custom_attribute_definition.rb
@@ -16,7 +16,7 @@
 #
 # Indexes
 #
-#  attribute_key_index                               (attribute_key,account_id) UNIQUE
+#  attribute_key_model_index                         (attribute_key,attribute_model,account_id) UNIQUE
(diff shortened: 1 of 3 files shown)


sojan-official (approved):

app/models/custom_attribute_definition.rb
```diff
@@ -16,16 +16,15 @@
...
+#  attribute_key_model_index                         (attribute_key,attribute_model,account_id) UNIQUE
 #  index_custom_attribute_definitions_on_account_id  (account_id)
 #
 class CustomAttributeDefinition < ApplicationRecord
   scope :with_attribute_model, ->(attribute_model) { attribute_model.presence && where(attribute_model: attribute_model) }
   validates :attribute_display_name, presence: true
 
   validates :attribute_key,
-            presence: true,
-            uniqueness: { scope: :account_id }
```
sojan-official, line 28: we don't need to remove the uniqueness validation, but just extended the scope to both account_id and attribute model

sojan-official (approved):

PR #3527, opened 2021-12-08
Fix: Add foreign key with delete cascade for destroy_async

tejaswinichile:
refer: https://www.bigbinary.com/blog/rails-6-1-allows-associations-to-support-destroy_async-option-with-dependent-key

`destroy_async option should not be used when the association is backed by foreign key constraints in the database`
So to resolve this I have added foreign key with `cascade: delete`


aswindevps (approved):

sojan-official:
@tejaswinichile should we do the same for other relations for which we have foreign keys as well? which we moved to destory_async

tejaswinichile:
@sojan-official We are tracking this issue here 
Add foreign key with delete cascade for destroy_async #3530

PR #3581, opened 2021-12-15
Fix user email reconfirmation flow

aswindevps:
## Description

Users can change their email from profile settings. They will be logged out immediately. Users can log in again with the updated email without verifying the same. This is a security problem.

So this change enforce the user to reconfirm the email after changing it. Users can log in with the updated email only after the confirmation.

Fixes: https://huntr.dev/bounties/7afd04b4-232e-4907-8a3c-acf8bd4b5b22/

## Type of change

## How Has This Been Tested?

1. Go to user profile settings.
2. Update the email. 
  a. User should be logged out. 
  b. User should receive a confirmation link in the updated email.
3. Confirm the updated email.
4. Login with the updated email.
5. Go to user profile settings and verify the updated email.

## Checklist:

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -177,4 +177,10 @@ class User < ApplicationRecord
       type: 'user'
     }
   end
+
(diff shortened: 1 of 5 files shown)


tejaswinichile (approved):
LGTM 💯

PR #3582, opened 2021-12-15
Fix: Add max retries for all jobs

tejaswinichile:


app/jobs/inboxes/fetch_imap_emails_job.rb
```diff
@@ -2,6 +2,8 @@ class Inboxes::FetchImapEmailsJob < ApplicationJob
   queue_as :low
 
   def perform(channel)
+    return unless channel.imap_enabled?
```
aswindevps, line 5: please remove '?'

`return unless channel.imap_enabled`

tejaswinichile: @aswindevps `channel.imap_enabled?` also works as `channel.imap_enabled` it's a ruby convention to write it with a question mark at the end if it's returning a boolean value.

sojan-official: this should be okay. ref: https://stackoverflow.com/a/1345854/939299

sojan-official:
@tejaswinichile  seems like we will have to bump up to rails 6.1.4.2

 https://weblog.rubyonrails.org/2021/12/14/Rails-6-0-4-2-6-1-4-2-and-7-0-0-rc2-have-been-released/

tejaswinichile:
@sojan-official I added it in the different PR, let me add it here.

sojan-official (approved):

PR #3585, opened 2021-12-15
feat: Show errors for message bubbles

nithindavid:
image


fayazara (approved):
Code Looks Good!

sojan-official:
let's try to improve the error message rendering. 
We could take it up in subsequent Pr. until then let's not show this or use a hardcoded error message in the tooltip.

PR #3640, opened 2021-12-21
chore: Update puma config as per recommendations

sojan-official:
ref:
https://devcenter.heroku.com/articles/concurrency-and-database-connections#threaded-servers
https://www.speedshop.co/2017/10/12/appserver.html


sojan-official:
We will need to iterate and choose the apt values for various installation set ups that chatwoot provides.

PR #3703, opened 2022-01-06
chore: Notification subscription delete API

muhsin-k:
- Added delete notification subscription API for clearing all the push tokens when logged out from the mobile app (https://github.com/chatwoot/chatwoot-mobile-app/issues/430)


sojan-official (approved):
LGTM

PR #3747, opened 2022-01-12
Fix: Pagination bug with filters payload

### LINKED ISSUE #3739: API call is failing on selecting load more conversations after applying filter
**Describe the bug**

The API call is failing when we click to load more conversations after applying the filter.

**To Reproduce**

Steps to reproduce the behaviour:
1. Go to Dashboard and open the filter modal
2. Click on status -> open and apply filter and now you can see the filtered conversation.
3. Open dev tool and go to Network tab and verify the filters API and go back to dashboard
4. Scroll down to and select Load more conversations after that you can see the API call is getting an error.
5. You can see the steps in the below screen recording.

https://user-images.githubusercontent.com/64252451/149061127-5f48c359-9816-49db-b854-4a46e4dc6325.mov

 - OS: Mac OS
 - Browser chrome
 - Version 98

fayazara:
# Pull Request Template

## Description

Fixes the pagination error which was causing filter api to return a 500 error.

Fixes #3739 

## Type of change

## Checklist:


app/javascript/dashboard/components/ChatList.vue
```diff
@@ -346,6 +346,9 @@ export default {
       if (!this.hasAppliedFilters) {
         this.fetchConversations();
       } else {
+        this.appliedFilters[
+          this.appliedFilters.length - 1
+        ].query_operator = null;
```
nithindavid, line 351: Is there a way to remove the query operator when it's applied?

fayazara: You mean when making the api request?

nithindavid: No, Like the applied filter object shouldn't coming with the non empty `query_operator` in the first place I guess. That would be the root cause of this problem. 

@iamsivin  also came across this issue in another place, So fixing the root would be the best case scenario.

nithindavid: Possibly in `app/javascript/dashboard/components/widgets/conversation/ConversationAdvancedFilter.vue`

fayazara: @nithindavid We need to remove this manually because there is no way of knowing if a filter appended in the list would be the last unless we click submit.

But I have moved removing this operator to the payload generator helper method in https://github.com/chatwoot/chatwoot/pull/3747/commits/18e10fc6a8e824ab6e4e3256c037bee9a7642509, so this has to be never done manually in our logic. Is be this okay?

app/javascript/dashboard/helper/filterQueryGenerator.js
```diff
@@ -11,7 +11,7 @@ const generatePayload = data => {
     }
     return item;
   });
-
+  payload[payload.length - 1].query_operator = null;
```
nithindavid, line 14:
```suggestion
// For every query added, the query_operator is set default to and so the
// last query will have an extra query_operator, this would break the api. 
// Setting this to null for all query payload

  payload[payload.length - 1].query_operator = null;
```

nithindavid (approved):

Issue #3684, opened 2022-01-03
whatsapp360 dialog press replay button but i didn't receive the message in chatwoot


voko3456:
![image](https://user-images.githubusercontent.com/20530550/147931260-8a645681-cb3e-4409-a4cc-06ada643a75e.png)

voko3456:
#3795 still not fixed 😥

PR #3751, opened 2022-01-13
chore: Support Whatsapp Interactive messages (360Dialog)

sojan-official:
Render button and list reply clicks as text messages in agent dashboard

fixes: #3684


voko3456:
The issue still not fixed 😥 if I send template message via 360dialog api then press on reply button I didn't receive message on chatwoot

Issue #3626, opened 2021-12-20
bug: Background inconsistencies with campaigns table 


nithindavid:
![image](https://user-images.githubusercontent.com/1277421/146713225-f386cb2b-b28d-40a4-be9f-2d79103c83a6.png)

anoymouscoder:
@pranavrajs @nithindavid Could you provide more information about this issue? And what is expect behaviours?

nithindavid:
@anoymouscoder As you can see the screenshot, the white background is getting cut even before the table ends. (See the color change behind the delete button)
The expectation is to have white bg till end of the row/table.

amaljosea:
Taking a look at this issue

PR #3774, opened 2022-01-15
fix: Background inconsistencies with campaigns table

amaljosea:
Might not be the best way to fix it, but looks like working.

Fixes https://github.com/chatwoot/chatwoot/issues/3626


nithindavid:
@amaljosea Thanks for the PR :)

nithindavid (approved):

PR #3874, opened 2022-01-31
feat: Email signature API support

### LINKED ISSUE #3799: feat: API for saving Email Signature with profile settings
**Context**
This is a follow up/breakdown of #3378. 
With every agent in chatwoot we need to support email signatures. The email signature would come under profile settings. This would need changes to  `PUT /api/v1/profile` API. With support for `email_signature` and `email_signature_enabled` keys.

`email_signature` - Would need to support HTML text also. Would also need to take care sanitation here

`email_signature_enabled` - If this is set to true, we would append the html value in `email_signature` to every outgoing email of the agent.

tejaswinichile:
# Pull Request Template

## Description

Fixes #3799 

## Type of change

## How Has This Been Tested?

Added test cases to save and accept email signatures.

## Checklist:


tejaswinichile:
email_signature_api

@nithindavid Let's have a call when you are back, I have added basic API changes to save email_signature and email_signature_enabled.

nithindavid:
@tejaswinichile I noticed the `email_signature_enabled` doesn't get toggled to `false` if the `email_signature` is not empty. Can we make it independent of each other? So that email signature will remain even if it is disabled ?

db/schema.rb
```diff
@@ -747,6 +747,8 @@
     t.jsonb "ui_settings", default: {}
     t.jsonb "custom_attributes", default: {}
     t.string "type"
+    t.boolean "email_signature_enabled", default: false, null: false
```
pranavrajs, line 750: Can we use a generic name here? We can use the same signature across channels.

tejaswinichile: @pranavrajs Just `signature_enabled` will work?

tejaswinichile:
> I noticed the email_signature_enabled doesn't get toggled to false if the email_signature is not empty. Can we make it independent of each other?

Oh okay, that's because of the validation rule, let me change that.

@nithindavid

db/schema.rb
pranavrajs: let us put it as message_signature and message_signature_enabled

nithindavid (approved):

tejaswinichile:
Merging this to save the email signature at the profile endpoint.

Issue #292, opened 2019-11-26
The header should collapse after a conversation from widget starts


nithindavid:

sojan-official:
@nithindavid  @pranavrajs  this is fixed in #293 right ?

pranavrajs:
Closed via #293

PR #3890, opened 2022-02-02
fix: Ongoing campaign URL validation

muhsin-k:
Fixes [#292](https://github.com/chatwoot/product/issues/295)


tejaswinichile (approved):
LGTM 💯

app/models/campaign.rb
```diff
@@ -86,6 +88,21 @@ def ensure_correct_campaign_attributes
     end
   end
 
+  def validate_url
+    return unless trigger_rules['url']
+
+    errors.add(:url, 'invalid') if inbox.inbox_type == 'Website' && !url_valid?(trigger_rules['url'])
```
aswindevps, line 94: Better to have a proper error msg instead of just `invalid`.

muhsin-k: @aswindevps The final response should be like this
Screenshot 2022-02-04 at 2 10 28 PM

PR #3921, opened 2022-02-07
fix: Omit invalid urls in custom attributes

muhsin-k:
Fixes https://github.com/chatwoot/product/issues/293


iamsivin (approved):
Code looks good to me👍🏻

fayazara (approved):
Tested, works as expected

PR #3934, opened 2022-02-08
feat: bulk actions to update status

### LINKED ISSUE #3845: Add the API to change the status of multiple conversations at once
- [ ] Accept conversation ids, status as the parameters
- [ ] Verify that the user has permission to resolve the conversation (should have access to the inbox / should be an admin). Return a 401 if conversation permissions are not available.
- [ ] Namespace under bulk_actions/status_change

**Additional thoughts:**
- Can be done as a job instead of doing it while calling the API.

tejaswinichile:
# Pull Request Template

## Description

Added the endpoints for bulk update conversation status.

Fixes #3845 

## Type of change

## How Has This Been Tested?

Tested on local with postman

## Checklist:


tejaswinichile:
Before update:
open_conv

Postman request to update the conversation status:
erquest


After update:
snoozed

app/jobs/conversations/status_change_job.rb
```diff
@@ -0,0 +1,8 @@
+class Conversations::StatusChangeJob < ApplicationJob
+  queue_as :medium
+
+  def perform(account:, conversation_ids:, status:)
+    conversations = account.conversations.where(display_id: conversation_ids)
+    conversations.update(status: status)
```
pranavrajs, line 6: @tejaswinichile Would it show an activity message that the "conversation was resolved by X" if we do it like this?

tejaswinichile: Yes, to create the activity we have `after_commit` assert actions set on the conversation model.

tejaswinichile:
QA checklist

1. Accept conversation ids and status and update the respective conversations with status
2. Accept conversation ids and agent_id and update the respective conversations with agent_id
3. Accept conversation ids and team_id and update the respective conversations with team_id
4. Accept conversation ids and labels and update the respective conversations with labels
5. Accept conversation ids and labels to remove and remove the labels from conversations

pranavrajs:
@tejaswinichile I'd like to add the following to QA checklist.

- [ ] Changing the status creates an activity message.
- [ ] Changing the assignee team/agent creates an activity message.
- [ ] Changing the labels creates an activity message.

app/controllers/api/v1/accounts/bulk_actions_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+      method_type: :delete,
+      user: current_user,
+    )
+    head :ok
+  end
+
+  private
+
+  def permitted_params
+    params.permit(:status, :assignee_id, :team_id, conversation_ids: [], labels: [])
```
sojan-official, line 25: Let's make the param structure more open towards the future changes than it is deeply tied to the scope of the story.  This would make it easier for the client implementations to accommodate future changes with minimal impact.  

an example param structure will be 
```
{
 ids: [ ] #rather than making it rigid towards conversations
 type: 'Conversation' #would be helpful for the future to extend support for contacts
 
 # wrapping the `attributes to be updated` in a key will let us permit the params based on object type
 fields: {
    status: 'open',
    assignee_id: '2'
 }
 # since labels are a separate object, lets have a separate key for that
 labels: {
    add: [],
    remove: [],
 }
}


```

app/jobs/bulk_actions_job.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def bulk_conversation_update(params)
+    params = available_params(params)
+    conversations.each do |conversation|
+      add_labels(conversation)
+      conversation.update(params)
+    end
+  end
+
+  def bulk_delete_conversation(_params)
```
sojan-official, line 30: let's consider renaming this method since it doesn't reflect the action it performs. 
also, the parameters seem to be unused in this method.

app/jobs/bulk_actions_job.rb
```diff
@@ -0,0 +1,48 @@
...
+    @conversation_ids = params.delete(:conversation_ids)
+    @labels = params.delete(:labels)
+    @conversations = account.conversations.where(display_id: @conversation_ids)
+    @method_type = method_type
+    Current.user = user
+    bulk_update(params)
+  end
+
+  def bulk_update(params)
+    if @method_type == :delete
```
sojan-official, line 15: method type :delete shouldn't be required based on the previous comments since everything related to the action will be wrapped under create

sojan-official (commented):
Add some review comments through which we can improve the implementation and make it more resilient towards the future without further rewrites. 

Please let me know your thoughts.

spec/controllers/api/v1/accounts/bulk_actions_controller_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+        expect(Conversation.last.status).to eq('open')
+        expect(Conversation.first.assignee_id).to eq(nil)
+
+        post "/api/v1/accounts/#{account.id}/bulk_actions",
+             headers: agent.create_new_auth_token,
+             params: { type: 'Conversation', fields: { status: 'snoozed' }, ids: %w[1 2 3] }
+
+        expect(response).to have_http_status(:success)
+
+        perform_enqueued_jobs do
```
sojan-official, line 47: Since we are calling `perform` directly here, it beats the purpose of the `perform_enqueued_jobs` jobs block. It doesn't test the condition of whether the HTTP request was responsible for triggering the Bulkactions job.  

We should do either of the following 
1.  wrap the post request inside `perform_enqueued_jobs`
```
perform_enqueued_jobs do
  post "/api/v1/accounts/#{account.id}/bulk_actions",
             headers: agent.create_new_auth_token,
             params: { type: 'Conversation', fields: { status: 'snoozed' }, ids: %w[1 2 3] }
end

# this ensures that if code is correct bulkactionsjob will be triggered internally on the request
# then do expectations that the conversation objects are updated

```
2. rewrite the controller test to check whether the job is enqueued correctly and have a separate test for the bulk actions job  (  feels more right ) 


```
# psuedo code

# bulk actions controller spec 
bulk_actions_job = double
allow(BulkActionsJob).to receive(:new).and_return(bulk_actions_job)
allow(bulk_actions_job).to receive(:perform)

 post "/api/v1/accounts/#{account.id}/bulk_actions",
             headers: agent.create_new_auth_token,
             params: { type: 'Conversation', fields: { status: 'snoozed' }, ids: %w[1 2 3] }

expect(bulk_actions_job).to have_received(:perform).with(the parameters)
## tests that HTTP request enqueued correct job 


## Create a separate test file for bulk actions job where we test the various conditions


```

sojan-official (commented):
Looks good mostly.  few comments around the tests.

sojan-official (dismissed):

app/jobs/bulk_actions_job.rb
```diff
@@ -0,0 +1,57 @@
+class BulkActionsJob < ApplicationJob
+  queue_as :medium
+  attr_accessor :records
+
+  MODEL_TYPE = ['Conversation'].freeze
+
+  def perform(account:, params:, user:)
+    @account = account
+    Current.user = user
```
sojan-official, line 9: @tejaswinichile  just noticed this. If we set current.user here. it will remain in the thread. So we need to ensure that it's unset at the end of the execution. We also need to vary of cases where the job could fail and the unsettling of Current.user doesn't happen. 

Probably we can have an initializer in the application job which unsets the Current. user as the start of the execution of each job to sure.

tejaswinichile: @sojan-official  Yes, this is for the activity message, let me unset it after the execution.

app/jobs/bulk_actions_job.rb
```diff
@@ -0,0 +1,65 @@
...
+    @params = params
+    @records = records_to_updated(params[:ids])
+    bulk_update
+  rescue StandardError => e
+    Current.reset
+    Sentry.capture_exception(e)
```
sojan-official, line 15:
```suggestion
  ensure
    Current.reset
```

app/jobs/bulk_actions_job.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def bulk_update
+    bulk_remove_labels
+    bulk_conversation_update
+    unset_current_user
```
sojan-official, line 21:
```suggestion
```

app/jobs/bulk_actions_job.rb
```diff
@@ -0,0 +1,65 @@
...
+  def records_to_updated(ids)
+    current_model = @params[:type].camelcase
+    return unless MODEL_TYPE.include?(current_model)
+
+    current_model.constantize&.where(account_id: @account.id, display_id: ids)
+  end
+
+  def unset_current_user
+    Current.reset
+  end
```
sojan-official, line 64:
```suggestion
```

sojan-official (commented):
minor cosmetic changes

config/routes.rb
```diff
@@ -40,6 +40,7 @@
             resource :contact_merge, only: [:create]
           end
 
+          resource :bulk_actions, only: [:create, :destroy]
```
sojan-official, line 43:
```suggestion
          resource :bulk_actions, only: [:create]
```

sojan-official (approved):

PR #3959, opened 2022-02-10
feat: add specs for reply-to 

### LINKED ISSUE #3834: Support reply-to header in emails? Can make PR
I recently tried to integrate chatwoot into our systems. We don't actually want real time chat (we don't have capacity to deliver on realtime support, we're a tiny team in an uncommon timezone). But I really love chatwoot as a issue tracking system. It beats all the other open-source products out there.

Our main communication channel is email. I tried integrating the live chat widget, but realized that we can't respond via email later on if the conversation is started in a chat channel, also the customers get disappointed when we can't respond in real time. Then i tried integrating your API, it required a lot of logic and network requests to create or search a contact, conversation, add message etc, and has the same problems as the chat channel.

By far the simplest way to integrate chatwoot, for us, is to have a web-form which sends an email. The FROM header is fixed (we can't spoof a from header from our server), but I can put the customer name and email into the REPLY-TO header. I have made a 2 line change to chatwoot to prefer the reply-to header when parsing an email in the IMAP processor. This way, our systems can send emails in from web-forms, and we can reply via email from within chatwoot. It works really well!

2 changes are on my mind. Firstly, if you would like, I can make a pull request with my changes to prefer a REPLY-TO header before the FROM header in your email processor.

The other idea, is the ability to reply via email when a conversation is started in a chat or api channel, and I'm wondering what you think about that.

Thanks so much for this great product!

tejaswinichile:
Fixes #3834 

Specs added for community PR.


aswindevps (approved):
LGTM

pranavrajs (commented):
The spec actually doesn't cover the cases which the PR addressed.

1. When a new conversation is created: If FROM and REPLY-TO are present, then the contact of the contact should be created using REPLY-TO.
2. When a new message is created in an existing conversation, then REPLY-TO would be given preference to select the contact.

The spec written is just testing the email object which doesn't ensure that the contact is created with the correct email address.

spec/mailboxes/reply_mailbox_spec.rb
```diff
@@ -59,12 +60,37 @@
...
+      before do
+        conversation_1.update!(uuid: '6bdc3f4d-0bec-4515-a284-5d916fdde489')
+      end
+
+      it 'prefer reply-to over from address' do
+        described_class.receive reply_to_mail
+        expect(conversation_1.messages.last.content).to eq("Let's talk about these images:")
+
+        email = conversation_1.messages.last.content_attributes['email']
+        expect(reply_to_mail.mail['Reply-To'].value).to include(email['from'][0])
```
tejaswinichile, line 90: > 2. When a new message is created in an existing conversation, then REPLY-TO would be given preference to select the contact. 

When a new message is created in the already present conversation with the message-id/in-reply-to/to address we just append the new message we don't append the contact anywhere.

Should I make that change?

The above test checks, the mail body is saved in content attributes with a reply-to address as from address.

pranavrajs: Let us do that as a separate PR. I guess we are not creating a new contact when a new message comes in.

pranavrajs (approved):

Issue #1372, opened 2020-10-29
Bug: Server responds with a broken image url on file upload


robinWongM:
**Describe the bug**

It seems that server will respond with the image url before it succeed the uploading when an agent is uploading an image. The browser will fetch the image when the image isn't ready and user will get a "broken" image.

**To Reproduce**

1. Click on attachment icon in a chatbox
2. Select an image to upload
3. Wait until an broken image appears

**Expected behavior**

The agent should get a perfect image rather than a broken image.

**Screenshots**

![image](https://user-images.githubusercontent.com/2606137/97541760-a675eb00-1a00-11eb-96f5-cc9f270e1b9f.png)

**Browser logs**

![image](https://user-images.githubusercontent.com/2606137/97542065-17b59e00-1a01-11eb-9bef-fd5aec61d7f9.png)

![image](https://user-images.githubusercontent.com/2606137/97542011-040a3780-1a01-11eb-99a0-96c334fa6295.png)

**Server logs**

![image](https://user-images.githubusercontent.com/2606137/97541651-7595b600-1a00-11eb-9d02-23559d0a6334.png)

**Environment**

self hosted: Kubernetes cluster (on premise) with docker image `chatwoot/chatwoot:develop`

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Windows
 - Browser [e.g. chrome, safari] Edge
 - Version [e.g. 22] 87.0.664.12

**Additional context**

If I click "Reload the image" from the context menu, the image will be displayed as expected since the uploading process is done then.

pranavrajs:
@robinWongM Are you using Cloud storage like S3 or GCS? If not, could you please try it with any cloud storage. Our current docker images doesnot support disk based file uploads. This is a known issue as seen in #1166

robinWongM:
> @robinWongM Are you using Cloud storage like S3 or GCS? If not, could you please try it with any cloud storage. Our current docker images doesnot support disk based file uploads. This is a known issue as seen in #1166

I haved tried self-hosted minio which is S3-compatible, and chatwoot behaved the same. Unfortunately I couldn't use other cloud storage since the Kubernetes host is unable to access the Internet directly.

When I use local storage, chatwoot container is configured to persist its `/app/storage` on a Persistent Volume (while the docker-compose.production.yml doesn't). So when chatwoot is restarted (the Kubernetes Pod is deleted and recreated), previously uploaded file won't gone. It did work.

[...]

PR #3966, opened 2022-02-11
Chore/delayed deploy of direct uploads

tejaswinichile:
Conditional: Direct upload


tejaswinichile:
@pranavrajs I don't why in indirect-upload/server-upload, sometimes it doesn't load image preview in the widget, it shows the image after the refresh, till then it shows the download link. 

I couldn't take the screenshot of the same because that issue is not appearing anymore.
Could you test it once from your side too?

sojan-official:
@tejaswinichile its intended behaviour. The reason why we wanted to do direct uploads was because of #1372. As a temporary fall back the image preview component renders a link when the image URL returns 404

pranavrajs (approved):

PR #3973, opened 2022-02-14
feat: Group by filter in reports

### LINKED ISSUE #3662: Add option for group by week / month in reports
Got a customer request to include "group by week/month" option for reports.

> Would you all consider adding a "group by" dropdown to the left of the date range dropdown in the reports section? We would love to be able to look at trends by week and by month within the app rather than just by day (the by day trend is awesome, but we have virtually no conversations on weekends and it can be tough to spot a trend over time due to random one day spikes that can occur). 

It seems like a nice feature to include. 

@sojan-official @tejaswinichile Please take a look at this request and assess the changes required.

aswindevps:
## Description

This change allows the user to group the reports based on Day, Week, Month, and Year. This filter depends on the date range chosen. Please note the date range to group by mapping.

Last 7 days - Defaults to Day
Last 30 days - Day, Week
Last 3 months - Day, Week, Month
Last 6 months - Day, Week, Month
Last year - Day, Week, Month
Custom date range - Day, Week, Month, Year

Fixes #3662

## Type of change

## How Has This Been Tested?

https://user-images.githubusercontent.com/5848740/153804813-0effc0a5-57f1-4e9d-be36-d8167d58a810.mp4

## Checklist:


app/javascript/dashboard/i18n/locale/en/report.json
```diff
@@ -59,6 +59,29 @@
     "CUSTOM_DATE_RANGE": {
       "CONFIRM": "Apply",
       "PLACEHOLDER": "Select date range"
+    },
+    "GROUP_BY_FILTER_DROPDOWN_LABEL": "Group By",
+    "GROUP_BY_DAY": "day",
+    "GROUP_BY_WEEK": "week",
+    "GROUP_BY_MONTH": "month",
+    "GROUP_BY_YEAR": "year",
+    "GROUP_BY_OPTIONS": {
```
fayazara, line 68: We need to move this object to a separate constants file and use use locale files for labels and text. Since we are using this object in `app/javascript/dashboard/routes/dashboard/settings/reports/Index.vue:165`, it might break in later if other locale files are not matching.

fayazara (changes requested):

fayazara:
@aswindevps Metrics like Avg First response time & Avg Resolution time change if we change chart type from Day, Week, Month.

https://user-images.githubusercontent.com/15716057/153808145-c891a541-84c4-4e6a-9f73-bca771e7b37d.mp4

Is this supposed to happen? Since only the chart representation is getting updated, shouldn't this be linked to the date-range?

aswindevps:
> @aswindevps Metrics like Avg First response time & Avg Resolution time change if we change chart type from Day, Week, Month.
> 
>  CleanShot.2022-02-14.at.11.26.29.mp4 
> Is this supposed to happen? Since only the chart representation is getting updated, shouldn't this be linked to the date-range?

@pranavrajs, please share your thoughts on this.

app/javascript/dashboard/i18n/locale/en/report.json
aswindevps: @pranavrajs can we take this as a separate task?

pranavrajs:
@aswindevps This looks like a bug. As @fayazara mentioned the representation is only changing right? The time period is constant, it should not change the avg/total values shown.

aswindevps:
@pranavrajs we can disable the group by filter for Avg First response time & Avg Resolution.
For the above 2 metrics, the grouping will be day-wise and take the average. So changing the group by filter won't change the value. These 2 metrics depend only on the date range. 

Other metrics will change based on the date range as well as the group by filter.

fayazara (dismissed):
Tested locally, working as expected

app/builders/v2/report_builder.rb
```diff
@@ -64,26 +64,30 @@ def team
...
-         .group_by_day(:created_at, range: range, default_value: 0)
+         .group_by_period(params[:group_by] || I18n.t('reports.default_group_by'),
+                          :created_at, range: range, default_value: 0, permit: %w[day week month year])
          .count
   end
 
   def incoming_messages_count
     scope.messages.incoming.unscope(:order)
-         .group_by_day(:created_at, range: range, default_value: 0)
+         .group_by_period(params[:group_by] || I18n.t('reports.default_group_by'),
```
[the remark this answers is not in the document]
aswindevps: Replaced with constant

spec/builders/v2/report_builder_spec.rb
```diff
@@ -247,6 +259,19 @@
...
+        params = {
+          type: :label,
+          id: label_2.id,
+          since: (Time.zone.today - 3.days).to_time.to_i.to_s,
+          until: Time.zone.today.to_time.to_i.to_s,
+          group_by: 'test'.to_s
+        }
+
+        builder = V2::ReportBuilder.new(account, params)
+        expect { builder.summary }.to raise_error(ArgumentError)
```
tejaswinichile, line 273: Can we have the tests against the correct group, with a metrics check?

tejaswinichile (changes requested):
The backend code looks good, just a small comment about the positive test case.

spec/builders/v2/report_builder_spec.rb
aswindevps: Done

fayazara (approved):
Looks good!

PR #3983, opened 2022-02-15
feat: Show cc from last email on reply editor

### LINKED ISSUE #3815: feat: Keep CC:ed emails in agent replies
The list of CC:ed emails is always empty and the agent has to add the CC:ed contacts to every reply.

Make sure to always include the latest email CC:ed emails in the field by default to make sure the agent does not forget to include them.

image

nithindavid:
Fixes #3815


Robinjoeh:
Won't this cause an issue when agents are making a private note? Looking at the code, it seems the CC:ed contacts will be added also to private notes?

nithindavid:
@Robinjoeh Thanks for the suggestion. I'll verify it 👍

fayazara (approved):
LGTM

PR #3987, opened 2022-02-15
Fix: Html mail and text content

tejaswinichile:
Related issue https://github.com/chatwoot/chatwoot/pull/3964/files


tejaswinichile:
Adding this commit to the previous PR #3964 . Closing this one

PR #3999, opened 2022-02-16
feat: Show banner message if Chatwoot update available

muhsin-k:
Fixes https://github.com/chatwoot/product/issues/216

#### How Has This Been Tested?

- Open the file `app/javascript/dashboard/App.vue`
- Update the line number 110
```
this.latestChatwootVersion = '5.0.1';
```

- Reload the dashboard

- Check the update banner are displaying on top of the page

Screenshot 2022-02-28 at 3 40 41 PM

- When click on dismiss icon, the banner should disappear


pranavrajs:
@nithindavid Please take the next steps on this PR.

app/javascript/dashboard/components/ui/UpdateBanner.vue
```diff
@@ -0,0 +1,88 @@
+<template>
+  <div class="banner">
```
pranavrajs, line 2: Can we use the banner component @iamsivin created?

app/javascript/dashboard/App.vue
```diff
@@ -72,10 +96,27 @@ export default {
...
+    },
+    dismissUpdateBanner() {
+      const updatedDismissedItems = this.getDismissedUpdates;
+      updatedDismissedItems.push(this.latestChatwootVersion);
+      localStorage.setItem(
+        'dismissedUpdates',
+        JSON.stringify(updatedDismissedItems)
+      );
+      this.showUpdateBanner = false;
+      this.latestChatwootVersion = this.globalConfig.appVersion;
```
pranavrajs, line 118: I would suggest creating a LocalStorage library to save and extract JSON data. Something like this https://gist.github.com/mritzco/72ac6ae497097b1657546b061a13706a

app/javascript/dashboard/components/layout/sidebarComponents/Secondary.vue
```diff
@@ -225,7 +225,7 @@ export default {
 .secondary-menu {
   background: var(--white);
   border-right: 1px solid var(--s-50);
-  height: 100vh;
+  height: 100%;
   width: 19rem;
```
pranavrajs, line 229: Please test it out on mobile as well as other screens.

pranavrajs (changes requested):

pranavrajs:
@muhsin-k Is there anything pending in this PR? Can we ship this in the current release?

app/javascript/dashboard/assets/scss/_layout.scss
```diff
@@ -10,17 +10,32 @@ body {
...
+  flex-grow: 0;
+  min-height: 0;
   width: 100%;
 }
 
-.app-root {
+.banner + .app-wrapper {
+  .button--fixed-right-top {
+    top: 5.6 * $space-one;
+  }
```
iamsivin, line 21:
```suggestion
.update-banner + .app-wrapper {
  .button--fixed-right-top {
    position: absolute;
    top: 5.6 * $space-one;
  }
```

Adding `position: absolute;` will solve this issue mentioned below in the screencast


https://user-images.githubusercontent.com/64252451/158370848-b830ebc9-2e30-43d6-b45b-7320338c9a50.mov

iamsivin:
Action buttons below the input field are hidden if the banner is shown. It is overflowing you can see in the below screencast.


https://user-images.githubusercontent.com/64252451/158371682-6fab7e76-0db8-4974-998d-edd2c9c8e6e5.mov

app/javascript/dashboard/assets/scss/_layout.scss
fayazara: > Adding `position: absolute;` will solve this issue mentioned below in the screencast. And `.banner` is the class name of the banner component so it will affect the button when the other banner is active.


This only works if we are having the button inside the header component (eg - inbox settings).

iamsivin: Yes that's right, but I have tried in this case and it works.

pranavrajs:
@iamsivin Let us fix this in a separate PR. It seems like the happy path is ready and clean, so let us go ahead with that..

PR #4018, opened 2022-02-18
Fix: add option to choose 24 hour working slot

### LINKED ISSUE #3491: bug: can't configure 24 hour working slots in business hours
When I try to configure 24-hour working slots 12:00 am - 12:00 am in business hours, even though the UI responds with a success message, the slots get reset to the older values. 

https://user-images.githubusercontent.com/73185/144212016-7b48d100-10b8-4b11-9acf-c364b6736864.mov

aswindevps:
## Description

This change allows the user to choose a 24-hour working slot.

Fixes #3491

## Type of change

## How Has This Been Tested?

https://user-images.githubusercontent.com/5848740/154708382-af828c73-858a-429c-84ef-73bc1f349849.mp4

## Checklist:


app/models/working_hour.rb
```diff
@@ -69,4 +71,11 @@ def close_after_open
 
     errors.add(:close_hour, 'Closing time cannot be before opening time')
   end
+
+  def update_open_all_day_hours
+    return unless open_all_day?
+
+    WorkingHour.find_by(day_of_week: day_of_week, open_all_day: false)&.update(closed_all_day: false, open_hour: 0, open_minutes: 0, close_hour: 23,
```
sojan-official, line 78: This will iterate over all the WorkingHour records in the DB. 

better to do 

```
before_validation :ensure_open_all_day_hours

def ensure_open_all_day_hours
   return unless open_all_day? 
  
   self.closed_all_day = false
   self.open_hour = 0
   self.open_minutes = 0
   self.close_hour = 23
   sefl.close_minutes = 59
end


```

We could also add a xor validation for closed_all_day & open_all_day to ensure both doesn't turn to be true.

pranavrajs:
@aswindevps Please make the necessary changes in the widget as well. (eg: Team Availability component, `mixins/availability.js`)

app/models/working_hour.rb
```diff
@@ -69,4 +71,14 @@ def close_after_open
 
     errors.add(:close_hour, 'Closing time cannot be before opening time')
   end
+
```
sojan-official, line 74: let's also add. a validation to ensure both the open_all_day and closed_all_day parameters can be saved as true at the same time .

example:

```
 validate :open_all_day_and_closed_all_day

 def open_all_day_and_closed_all_day
    return unless open_all_day? && closed_all_day?
    errors.add(:base, "open_all_day and closed_all_day cannot be true at the same time")
 end
```

aswindevps: Just a thought, the `ensure_open_all_day_hours` validation always updates `closed_all_day` to `false` if `open_all_day` is `true` right? So do we need this check? Or am I missing any scenarios?

aswindevps: okay

sojan-official (approved):

PR #4024, opened 2022-02-21
fix: Message signature going with other channels message

### LINKED ISSUE #4025: Email signature being sent out with all channels
With the new email signature feature. If you enable it when writing an e-mail and leave it on after sending it, every chat message gets the signature as well. Even the one's where email signature is not supposed to work.

nithindavid:
Fixes #4025

The message signature was getting attached to every inbox irrespective of the settings.


fayazara:
@nithindavid The signature doesn't seem to work, no errors in console too.

https://user-images.githubusercontent.com/15716057/155665098-0e249bb6-37f4-4fb0-8307-2d25102d267a.mp4

pranavrajs:
@fayazara This is a good find. Made some changes to fix it. Can you please test it again?

**Case 1: If email channel and Message Signature is not enabled in profile settings (value is empty)**

This might be the case you saw. In this case, the signature option should not be shown or should show "no signature available" message. I've updated this to the following.

Screenshot 2022-02-25 at 1 45 20 PM

**Case 2: if email channel, Message Signature is configured in profile settings (value is available)**

Clicking on the button should show the configured signature and while sending a message, it should append the configured message.


**Case 3: if not email channel, Message Signature is configured in profile settings, previously enabled in email channel (value is available)**

- Button should not be shown
- Message should not have signature added to it.

fayazara (approved):
Tested all three cases, works correctly.

CleanShot 2022-02-25 at 16 32 17

PR #4011, opened 2022-02-17
Fix: html mail fix with html_body content

tejaswinichile:
RCA:

Mail had content-type : text/Html
Mail.html_part and mail.parts was empty

In that case, we considered the raw text body but from this fix, if only HTML content is present then we are considering raw HTML mail body too.

This fix fixes all the HTML type email

The previous PR that we closed, I had introduced refactoring of code there which apparently didn't work for some mail types which I removed from here.

PR #4030, opened 2022-02-21
Fix: show empty emails in chat screen

### LINKED ISSUE #4020: "No content available" for e-mail messages without body
**Describe the bug**
If a user sends an email to an inbox monitored by chatwoot, but only puts his question in the subject, with an empty message body, the email will not be processed properly. Instead, an empty chat history appears, in which it is not possible to read the message.

**To Reproduce**
Send a blank email to Chatwoot and only specify the subject line

**Expected behavior**
The message is displayed, you can see what the subject says

**Screenshots**
current behavior:
image

Expected behavior
screenshot 2022-02-19 13-24-03

**Environment**
self hosted installation

nithindavid:
Fixes #4020


sojan-official (commented):
Looks good to me. lets test it out with an email with just the subject

sagararyal:
This partially fixes #4020

`No Content available` is also randomly showing up for emails that have a body, and a subject.  However, chatwoot somehow fails to render them.

Maybe there's an issue with the way cron fetches the emails from SMTP?

sojan-official:
@sagararyal  the latter is a separate issue which we are looking at https://github.com/chatwoot/chatwoot/pull/4011/files
cc: @tejaswinichile

sagararyal:
@sojan-official great, but i think these are all different issues that behave similarly.

Here's what i manage to replicate, the issue is caused by emails that contain reply-to address.
Please put this where it belongs.


Message content:
  
Screen Shot 2022-02-22 at 7 50 18 PM


Gmail has full message:

Screen Shot 2022-02-22 at 7 56 28 PM



in Chatwoot.
user is created for the reply-to, however the original message context is lost.
Screen Shot 2022-02-22 at 7 51 09 PM

aswindevps (approved):
LGTM

Issue #4070, opened 2022-02-25
bug: Issue with automation event loop


sojan-official:
Conversation updated event keeps triggering on empty update transactions

PR #4071, opened 2022-02-25
fix: Automation loop on conversation_update

sojan-official:
fixes: #4070


pranavrajs (approved):
Test locally with the conditions that triggered infinite loop earlier, it seems to be working

Issue #3918, opened 2022-02-06
How to reset the user on logout for multi-page site


lilsweetkara:
[DISCLAIMER: First of all, I am a terrible coder. I learned a few things from my husband, who is a software engineer, but I am generally speaking terrible at this. Sorry.)

I am trying to have the chat bubble appear only for logged in users. I have a simple (crappy) node app using EJS.
I am using app.chatwoot.com

I placed the element in the header, and basically run a reset/hide it if the user is not logged in.
You mentioned in the docs that I should run `reset()` when the user logs out. The trouble I have is that the logout page is just an URL in my server that zaps the session and redirect to `/` So, no page is actually displayed.

I put thall of the code in the header, with every page, and will call reset() if the user is not logged in (or, `setUser()` if the user IS logged in). However, it feels very wasteful (or, my husband's words, "It's the worst idea since the invention of the atomic bomb")

Also, I have to have that `alreadyReset()` variable there, since 1) I need to wait for `chatwood:ready` 2 After reset, the event `chatwoot:ready` is emitted again, and reset() is run again. And again. And again. and...

Am I doing it all wrong? (Or, is it really worse than the atomic bomb?)

````

<%if(session.loggedIn) { %>

<script>
    (function(d,t) {
      var BASE_URL="https://app.chatwoot.com";
      var g=d.createElement(t),s=d.getElementsByTagName(t)[0];
      g.src=BASE_URL+"/packs/js/sdk.js";
      g.defer = true;
      g.async = true;
      s.parentNode.insertBefore(g,s);
      g.onload=function(){
        window.chatwootSDK.run({
          websiteToken: 'XXXXXXXXXXXXXXXX,
          baseUrl: BASE_URL
        })
      }
    })(document,"script");

    window.addEventListener('chatwoot:ready', function () {
        debugger
        window.$chatwoot.setUser('<%=session.loggedInEmail%>', {
            email: '<%=session.loggedInEmail%>',
            name: '<%=session.loggedInName%> <%=session.subscriber ? 'PAID' : 'FREE'%>' 
            identifier_hash: '<%=hash%>',
        });
    });        

  </script>
  
  <% } else { // This happens if user is NOT logged in  %>    
    <script>
      let alreadyReset = false
      window.chatwootSettings = {
        showPopoutButton: false
      };      
        (function(d,t) {
          var BASE_URL="https://app.chatwoot.com";
          var g=d.createElement(t),s=d.getElementsByTagName(t)[0];
[...]

PR #4079, opened 2022-02-28
fix: Add resetTriggered flag to fix reset loop

### LINKED ISSUE #3963: bug: Calling sdk event reset results in infinite loop with ready event
https://github.com/chatwoot/chatwoot/issues/3918#issuecomment-1033037680

```
<script>
  (function(d,t) {
    var BASE_URL="https://my-base-url.com";
    var g=d.createElement(t),s=d.getElementsByTagName(t)[0];
    g.src=BASE_URL+"/packs/js/sdk.js";
    g.defer = true;
    g.async = true;
    s.parentNode.insertBefore(g,s);
    g.onload=function(){
      window.chatwootSDK.run({
        websiteToken: 'my-website-token',
        baseUrl: BASE_URL
      })
    }
  })(document,"script");

  window.addEventListener('chatwoot:ready', function () {
    console.log('chatwoot:ready');
    window.$chatwoot.reset();
  });
</script>
```

As you can see above, that's just copy and paste from the Chatwoot docs, nothing else, but the log message console.log('chatwoot:ready');, will keep firing again and again until I remove it from the code.

So please, help us out, how can we reset the session? Please provide a step-by-step explanation, or a snippet of code with an example, because so far, the only solution that I see is to:

Load an HTML page and run the window.$chatwoot.reset();
Redirect the user to another page where there's no window.$chatwoot.reset();
But as you can imagine, that's not an applicable solution for us.

@maxacarvalho Thanks for reporting this.

pranavrajs:
- Added a resetTriggered flag to check if the loaded event was triggered by a reset method. 
- If resetTriggered, the ready event will not be fired again.

Fixes #3963, #3918

### DIFF
--- a/app/javascript/packs/sdk.js
+++ b/app/javascript/packs/sdk.js
@@ -25,6 +25,7 @@ const runSDK = ({ baseUrl, websiteToken }) => {
     launcherTitle: chatwootSettings.launcherTitle || '',
     showPopoutButton: chatwootSettings.showPopoutButton || false,
     widgetStyle: chatwootSettings.widgetStyle || 'standard',
+    resetTriggered: false,
 
     toggle(state) {
       IFrameHelper.events.toggleBubble(state);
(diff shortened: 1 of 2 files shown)


muhsin-k (approved):
QA has been done on staging and everything is working as expected.

Issue #2828, opened 2021-08-16
End conversation from chat widget


pranavrajs:
**Context:**

Provide an option to the user to end the conversation once they are done with it. This would move the conversation from the queue.

**Changes:**
- [ ] Add a dropdown on the widget to select the end conversation button. Design attached. Move the existing open in a new window button to the dropdown.
- [ ] Add an API to resolve the conversation.
- [ ] Add an activity message "Conversation was resolved by %{user_name}".

**Design:**

<img width="432" alt="end-chat(2)" src="https://user-images.githubusercontent.com/2246121/129533601-5d7eb2dd-0840-4735-b191-33694c7ba961.png">

iammattmartin:
I'd just like to add that hiding the expand option probably isn't great and the end chat should really be more visible. 

Often the three dots is used for things like sound, email transcript and contact details. Not core functionality for the window itself.

pranavrajs:
@iammattmartin There is not enough space for the end chat button which is why we went with this design. This makes sense. Let us revisit the design.

PR #4102, opened 2022-03-02
fix: Show new message screen instead of input field

pranavrajs:
While testing #2828, I noticed that the new conversation flow if `allow_messages_after_resolved` is false, is not working properly. 

The earlier implementation: We were redirecting the user to a pre-chat form with contact fields disabled. The user would hit on the `create conversation` button which would eventually call the Conversation Create API. This flow got broken during Vue Router upgrade. 

The current behavior: Once we click on `Start New Conversation`, it would clear the messages locally, and show an input box. There is no logic to check whether a conversation should be created or not while sending a message. This resulted in the behavior where the messages were appended to the previous conversation. 

Fix: Updated the current implementation to use the redirect to pre-chat form logic.

Here is a loom that explains the behavior. https://www.loom.com/share/c864b3a7268d45768cf1abbe7da2915c

cc: @nithindavid @sojan-official


iamsivin (approved):
Code LGTM👍🏻

fayazara (approved):

muhsin-k:
@pranavrajs The codes looks good to me. I have found a small issue while doing QA.

**Steps to reproduce**

1. Create new conversation from the web widget
2. Resolve from the dashboard
3. Enable email and name option in pre-chat form
4. Reload the widget, click on "Start new conversation"
5. The fields email and name are missing in the pre-chat form

https://user-images.githubusercontent.com/12408980/156505257-57805d49-b27b-48eb-941a-4978f4974b0b.mov

pranavrajs:
@muhsin-k I knew this case. I took this decision because the contact information was not available always. There are some more improvements I'm planning as follow-ups, I will fix this case also in one of those PRs.

ChrisvanChip:
Hey @pranavrajs, 

I have the feeling that this PR causes no new chats to be made, but just cleans the history for the customer:

---

image
_The conversation history for the agent_

---

image
_The conversation bubbles for the customer, even though those messages are from an earlier chat that was resolved._

---

image
_The chat widget for the customer, only seeing the last chat._

---

I would like to have a new chat after the inquiry has been resolved (like the image below). It's really inconvenient right now.

image

PR #4103, opened 2022-03-02
chore: Pass sso_account_id to select the account during SSO Login

pranavrajs:
- Redirect the user to the correct account if sso_account_id is passed in the SSO URL.

Fixes https://github.com/chatwoot/product/issues/324

### DIFF
--- a/app/javascript/dashboard/api/auth.js
+++ b/app/javascript/dashboard/api/auth.js
@@ -13,7 +13,7 @@ export default {
         .post('auth/sign_in', creds)
         .then(response => {
           setAuthCredentials(response);
-          resolve();
+          resolve(response.data);
         })
         .catch(error => {
           reject(error.response);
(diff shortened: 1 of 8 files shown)


pranavrajs:
@sojan-official This is ready for review.

To test case 1:

1. Normal login - Login to the dashboard as usual as an admin and an agent.
2. Platform App login - 

2.1 SSO link with valid Account ID: Generate an SSO link for the user and append an sso_account_id to the URL. The login should redirect the user to the correct account.

2.2 SSO link with an invalid Account ID: Generate an SSO link for the user and append a random number as sso_account_id to the URL. After login, it should take you to the first account which the user belongs to.

sojan-official:
@pranavrajs the described scenarios worked in cases where the user was logged out. But the following scenarios failed while testing 

Case 1: 
- `John` is already logged in and is active on `account 1`
- SSO link was generated for John for `account 2 `
- John was redirected to `account 1` itself

Case 2:
- `John` is already logged in and is active on `account 1`
- SSO link was generated for `smith`
-  on redirect `John` wasn't logged out and `Smith` was logged in. the existing session wasn't reset

sojan-official (approved):
Tested all the described QA conditions and its working

Issue #3701, opened 2022-01-06
Unable to receive documents and files in WhatsApp via 360Dialog


sanjayws:
**Describe the bug**

- I setup 360Dialog and basic texting and images work fine. However, i cannot receiving PDF or allowable mimes like .docx (person to CW)
- Sending documents however is fine (from CW to person)

**To Reproduce**
1) Use 360dialog
2) Send hello, followed by an allowable attachment, e.g. pdf

**Expected behavior**
Documents should be rendered in CW inbox.

**Server logs**
[...]

PR #4173, opened 2022-03-14
fix: error receiving documents and files in WhatsApp via 360Dialog 

sojan-official:
fixes: https://github.com/chatwoot/chatwoot/issues/3701


sojan-official:
verified that document attachments are working in staging.

PR #4135, opened 2022-03-09
feat: Custom fields in pre-chat form

muhsin-k:
Fixes #2744

PR #4182, opened 2022-03-15
revert: Custom fields in pre-chat form #4135

muhsin-k:
This reverts commit e730804b4888fa0fc21fb1681f4e2a53d752e645.


pranavrajs:
@sojan-official Reverting this case for now as it seems like the migrations/config is missing in some cases. We will get this shipped in the next release.

pranavrajs (approved):

PR #4188, opened 2022-03-15
chore:  "Channel::TwilioSms" to be unique on account_sid+phone_number

### LINKED ISSUE #4187: bug: Prevent duplicate account_sid+phone_number for Channel::TwilioSms
Currently the app permits multiple `Channel::TwilioSms` records with duplicated `account_sid`+`phone_number` values.

However, `Twilio::IncomingMessageService` searches for the correct `Channel::TwilioSms` by `account_sid`+`phone_number`:
https://github.com/chatwoot/chatwoot/blob/796a7805dbc67ff7e0b27e95c6c2bc320b32a856/app/services/twilio/incoming_message_service.rb#L23-L26

If these values are duplicated in multiple records then the record that Twilio::IncomingMessageService find is indeterminate and may alternate between queries.

jordan-brough:
## Description

`Twilio::IncomingMessageService` searches for the correct `Channel::TwilioSms` by `account_sid`+`phone_number`:
https://github.com/chatwoot/chatwoot/blob/796a7805dbc67ff7e0b27e95c6c2bc320b32a856/app/services/twilio/incoming_message_service.rb#L23-L26

If these values are duplicated then the record it will find is indeterminate and may alternate between queries.

Fixes #4187

## Type of change


sojan-official:
@jordan-brough Thanks for the changes. Is there a chance for some installations to have duplicate records for this index? hence we might have to add data fix along in the migration where we try to clear out the duplicate values before trying to create the index.

tejaswinichile (approved):
LGTM 💯

sojan-official:
@tejaswinichile  before merging let's make sure that the migrations would show the appropriate error message as mentioned in the discussion.

tejaswinichile:
@sojan-official Yes 👍🏼 I will do the needfull.

tejaswinichile:
@jordan-brough Can you update the migration for the previous duplicate records with the error message and if possible add test cases?

jordan-brough:
@tejaswinichile 

> update the migration for the previous duplicate records with the error message

👍 done. How does that look?

> if possible add test cases?

I don't know if there are any applicable test cases here? I don't think y'all have specs covering migrations and the model change is just a standard `validates … uniqueness`.

jordan-brough:
Gotcha, that makes sense.

Even if phone numbers should be unique globally, it seems like it might be good to continue to look up the channel by `phone_number` + `account_sid` in `Twilio::IncomingMessageService`. If we only look up by phone number that seems like a possible security vulnerability since an attacker could guess or discover a phone number much more easily than an account sid, and this endpoint is open to the world.

And if that's the case maybe we do go ahead and have the unique index + validation be on `account_sid` + `phone_number`? And drop the old unique index on `account_id` + `phone_number`.  Thoughts?

sojan-official:
@jordan-brough agree with your point that we should continue to look up the channel by phone_number + account_sid in Twilio::IncomingMessageService.

`The phone numbers should be unique globally` comes from the recent pattern we are following for newer channels where the uniqueness of a channel's base id [phone numbers, channel ids, etc.] is guaranteed. i.e. while the server admin/super admin looks up into the data from the console, there won't be two channels with the same base id

- let's have a unique index for `account_sid` and `phone_number` for the case of lookups
- let's drop the old unique index on account_id + phone_number, which seems irrelevant. 
- let's have a unique constraint on `phone_number`  to satisfy the case mentioned above. (* when present : ref: #4242 )

Let me know your thoughts.

sojan-official:
Tested out the changes against Chatwoot SaaS and updated the migrations with logic to clear empty duplicate inboxes. 
Also, removed the account_id phone_number index and introduced a unique phone number index.

PR #4202, opened 2022-03-17
chore: Improve n+1 queries in agents endpoint

sojan-official:
Fixes the n+1 queries that get triggered while accessing agents endpoint


pranavrajs (approved):
Looks good to me.

PR #4237, opened 2022-03-22
chore: Use Round Robin service for team assignment

sojan-official:
Before this change, auto-assignment in teams wasn't using round-robin service.
Hence the agent's online, offline status wasn't respected by auto-assignment.

This change will make teams give priority to online agents for assignments.

fixes: https://github.com/chatwoot/product/issues/356

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,7 +110,7 @@ jobs:
       - run:
           name: Run backend tests
           command: |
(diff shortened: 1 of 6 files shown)


pranavrajs:
@sojan-official Can you please fix the failing spec?

sojan-official:
@pranavrajs  this should be good to go.

pranavrajs (approved):
LGTM

PR #4243, opened 2022-03-22
Use "contain_exactly" in some tests

jordan-brough:
## Description

The ordering of these is not specified and is not guaranteed.

On my laptop I'm getting failures for these tests due to ordering.


tejaswinichile:
@jordan-brough Order is not important here. Thank you so much for the PR and the effort.

tejaswinichile (approved):
LGTM 💯

PR #4244, opened 2022-03-22
add nice scrollbars to emoji picker

nosovk:
# Pull Request Template

## Description

image

should look like that: 
image

Refer to article to see options: https://css-tricks.com/almanac/properties/s/scrollbar/

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Actually I test changes in browser, but it would be nice to use correct colour variables.

## Checklist:


nosovk:
Yep, I already putt screenshot in first message

sojan-official:
@nosovk could you also mention the `OS` and `browser` in which you captured the screenshots?

nosovk:
It was Chrome Version 99.0.4844.82 on windows 10.
It would not work on firefox.
Refer to article to see options: https://css-tricks.com/almanac/properties/s/scrollbar/

nithindavid:
Hello @nosovk 
I think the root problem is the width of the container. We could ideally fix this by increasing  width of the emoji container. The scrollbar is only for webkit. Also this happens on chrome only when user has a physical mouse attached(Not fully sure). So I'm not feeling good about pushing a different experience only to one browser.

Let me know your thoughts.

nosovk:
There is another syntax for the same case in Firefox.
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scrollbars
I will add it tomorrow.
But I'm sure that having better UX in any browser is a good point to add few styles.
If you want we can add those styles to body, for fixing all scrollbars in Widget.

nithindavid:
@nosovk Thanks for the PR but IMHO targeting just webkit would not ideally fix the issue. We need to fully support all browsers for this. We will put this on hold and will revisit when the Firefox support is added.
Let me know your thoughts.

nosovk:
Actually there is a solution for Firefox also
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scrollbars

pranavrajs:
@nithindavid This PR was open for some time, please take this on priority and suggest the next steps here.

PR #4245, opened 2022-03-22
🏃‍♀️ Vite migration

pranavrajs:
- Webpacker is discontinued, new gem is shakapacker. I tried migrating to the new gem, it was painful, didn't work. (Probably due to the dependencies we have in the code). The error messages were cryptic.
- Webpack takes too much time to compile the assets which was affecting the productivity.
- Webpacker 5.x had a hard dependency on PostCSS 7 which blocked the migration of TailwindCSS. (latest is 3.x, we are on 1.9 - IE support is no longer required)

**Migration to Vite**

This PR migrates the repository to use Vite for development and production bundle. 

- I have ensured that the path `packs/js/sdk.js` remains the same so that the widget scripts are backward compatible.
- Add `.vue` to the resolve files as default support is only for `.js` extensions.
- Made sure that all bundles are working properly. (Have tested some of the happy paths in dashboard, widget and survey bundle)

**Risks**

- I'm getting a 500 error as described in https://vite-ruby.netlify.app/guide/troubleshooting.html#requests-to-vite-sporadically-return-a-500-error-response. I've not found a sustainable solution to fix this yet.
- The bundles have to be tested in production. I've test this in development only.
- Specs are running, but need to remove the babel dependencies from the codebase.
- Storybook has not been tested.

**How to test this PR?**

- Pull the changes, run `bundle` & `yarn`. You are good to test.


Procfile.dev
```diff
@@ -1,3 +1,4 @@
 backend: bin/rails s -p 3000
```
sojan-official, line 1: do we need the `backend` ? now that that we have `web`

fayazara:
@pranavrajs I ran `yarn` and `bundle` and getting this error when I try starting the dev environment with foreman. Should I be installing something?
Screenshot2022-03-23 at 19 11 24

pranavrajs:
Closing it now as it is a challenge to fix the conflicts, will create a new PR from develop.

PR #4246, opened 2022-03-22
Timezone fixes for ReportBuilder and for ReportController spec

jordan-brough:
I noticed this because I was getting test failures in `spec/controllers/api/v2/accounts/report_controller_spec.rb` running the tests locally on my laptop.

- ReportBuilder wasn't using the specified time zone for the timestamp in the
"build" method
- The ReportController spec was calling `Time.zone.today.to_time`, but
that sequence loses time zone information because `today` is a Date and
Dates don't have time zones in Ruby, so the time ended up in the system
time start-of-date instead of the `Time.zone` start-of-date.

### DIFF
--- a/app/builders/v2/report_builder.rb
+++ b/app/builders/v2/report_builder.rb
@@ -19,7 +19,7 @@ class V2::ReportBuilder
   # For backward compatible with old report
   def build
     timeseries.each_with_object([]) do |p, arr|
-      arr << { value: p[1], timestamp: p[0].to_time.to_i }
+      arr << { value: p[1], timestamp: p[0].in_time_zone(@timezone).to_i }
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/builders/v2/report_builder.rb
```diff
@@ -19,7 +19,7 @@ def timeseries
   # For backward compatible with old report
   def build
     timeseries.each_with_object([]) do |p, arr|
-      arr << { value: p[1], timestamp: p[0].to_time.to_i }
+      arr << { value: p[1], timestamp: p[0].in_time_zone(@timezone).to_i }
```
jordan-brough, line 22: Is there any chance this update will change/break existing behavior downstream? It seems like the intended result, but I don't have 100% context on how this is used.

aswindevps: It should not change/break the existing behavior. @pranavrajs any thoughts?

pranavrajs: @sojan-official Please take a lead on this.

sojan-official:
@jordan-brough the changes look good to me. Thanks a lot for the work

PR #4272, opened 2022-03-24
Don't send confirmation email when creating User via platform api

jordan-brough:
Call `skip_confirmation!` before saving the user (instead of `confirm` after
saving the user) so that the user is marked as confirmed already and the
confirmation email isn't sent.

The platform api automatically confirms users so we don't need to send this
email.

### DIFF
--- a/app/controllers/platform/api/v1/users_controller.rb
+++ b/app/controllers/platform/api/v1/users_controller.rb
@@ -7,8 +7,8 @@ class Platform::Api::V1::UsersController < PlatformController
 
   def create
     @resource = (User.find_by(email: user_params[:email]) || User.new(user_params))
+    @resource.skip_confirmation!
     @resource.save!
-    @resource.confirm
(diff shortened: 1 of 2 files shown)


app/controllers/platform/api/v1/users_controller.rb
```diff
@@ -7,8 +7,8 @@ class Platform::Api::V1::UsersController < PlatformController
 
   def create
     @resource = (User.find_by(email: user_params[:email]) || User.new(user_params))
+    @resource.skip_confirmation!
     @resource.save!
-    @resource.confirm
```
jordan-brough, line 11: Is it possible that some users of the platform api are relying on this email being sent for whatever reason?
If so maybe this would need a deprecation warning first or a configuration option or something?

sojan-official: @jordan-brough I think the earlier behaviour was unintentional. 
Let's go ahead with the merge. We will mention the change in the release notes.

Issue #4233, opened 2022-03-21
Providing a bad phone number should not break PATCH for other attributes (email, name, etc.)


Vadorequest:
I tried to provide the phone number, but it crashed my whole integration with "422 (Unprocessable Entity)" error.

PATCH https://app.chatwoot.com/api/v1/widget/contact?website_token=DJGn3WJ5dyMCUPh3rQHRJ8r6&cw_conversation=eyJhbGciOiJIUzI1NiJ9.eyJzb3VyY2VfaWQiOiJmMTJjYzhmYy02MmEwLTQ1MjMtYTdmMS1hYmM0MjQ3MzMyZmIiLCJpbmJveF9pZCI6MTI0NDZ9.rIiRaI9CJk69R7mZVdqT1pp63ruw9DZ9X_BU92sdhU4 422 (Unprocessable Entity)

Looking into the network logs, I saw that:
message: "Phone number should be in e164 format"

It's not mentioned in the doc.

Also, and that's the main point, providing a bad phone number should NOT cancel the whole PATCH query, it should just ignore the phone number.

Right now, because I provided a bad phone number (not E164), it didn't store the email/name/company either, because the whole PATCH query failed altogether. This is not the best behavior.

Can you:

- Update the documentation to mention the phone_number format that should be used
- Change your PATCH handler so that a bad phone number doesn't break the whole thing?

Also, converting a phone number to E164 is not a trivial task to do, especially when being limited by using no-code tools.

pranavrajs:
@muhsin-k Please take a look at this change.

pranavrajs:
@Vadorequest Just to make sure that we understand the issue properly.

1. Chatwoot requires you to send the phone number in the specified format. We don't allow duplicate number as well the number in a different format as it would fail the integration we have in place. Changing this constraint might be hard.

2. We will update the documentation for the same.

3. I understand that you would have issue with the data you would have collected from different tools. Would it be okay if you use a custom attribute to save phone number instead of the default phone number if you cannot convert it to e.164 format.

Vadorequest:
I'm not asking you to change the phone number format, I guess if you chose to use E164 it was for a good reason.

But what you could do, is change your PATCH implementation so that if any of the attribute sent is not valid, it wouldn't invalidate the whole payload, but only the concerned attribute.

For instance, if I send name + phone number, if the phone number is in a bad format, the name should still be updated.

[...]

PR #4309, opened 2022-03-28
fix: Referer URL validation

muhsin-k:
Fixes [#354](https://github.com/chatwoot/product/issues/354)


app/models/conversation.rb
```diff
@@ -241,6 +243,12 @@ def mute_period
     6.hours
   end
 
+  def validate_referer_url
+    return unless additional_attributes['referer']
+
+    errors.add(:referer_url, 'invalid') unless url_valid?(additional_attributes['referer'])
```
sojan-official, line 249: @muhsin-k  rather than throwing an error let's discard the value instead
ref: #https://github.com/chatwoot/chatwoot/issues/4233#issuecomment-1074846146

sojan-official (approved):

PR #4333, opened 2022-03-30
design: Changes conversation actions order

nithindavid:
**After**

image

**Why**
Based on usage, team is rarely used. After assigning the conversation to an agent, My next move will be to add labels. So i have to skip the team assignment block for next action. Also most times I end up clicking the team block for labels


pranavrajs:
@nithindavid This is a subjective choice, I'm not sure whether I'm aligned with this change. If this change is needed, I would suggest to have a draggable implementation so that the customers can choose the order they want.

fayazara (approved):
LGTM

iamsivin (approved):

Robinjoeh:
> @nithindavid This is a subjective choice, I'm not sure whether I'm aligned with this change. If this change is needed, I would suggest to have a draggable implementation so that the customers can choose the order they want.

I would agree with you, this highly depends on how much the particular user is using the 'teams'. For larger teams, definitely this change would not make sense.

From a design perspective, it also looks better with two dropdowns followed by a different styled section.

nithindavid:
@Robinjoeh @pranavrajs Makes sense. I'll close this now and will look at the draggable workaround suggested.

PR #4336, opened 2022-03-30
enhancement: persistent assign to me banner on conversation

nithindavid:
For #4418

## Description
While using chatwoot the current banner wasn't usable because it would vanish the moment I sent a message. This would make it more accessible for users. From now on the banner will stay there.

**Cases when the banner will be shown**

1. When conversation is unassigned 
2. When conversation is assigned to another agent but no agent has sent any message
3. When conversation is assigned to another agent but all outgoing messages sent by you(current user)
4. When conversation is unassigned but you have sent an outgoing message along with other outgoing replies from other agents.


fayazara (approved):
Working as expected.

Tested locally on the `widget_tests` page
1. Tested with no replies on a conversation
2. Tested with Unassigning a conversation.

app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
```diff
@@ -245,13 +271,11 @@ export default {
...
     },
     showSelfAssignBanner() {
-      if (this.message !== '' && !this.isOnPrivateNote) {
+      if (!this.isOnPrivateNote) {
         if (!this.assignedAgent) {
-          return true;
-        }
-        if (this.assignedAgent.id !== this.currentUser.id) {
-          return true;
+          return this.isOnlyAgentReplyMine || this.hasNoAgentReplied;
```
pranavrajs, line 276: @nithindavid Can you please share the logic to show the banner here?

If there is no assigned agent, then we show the banner if no agent has replied to it or if I'm the only one replied to it. So if I'm replied to a conversation that has been assigned to a different person, then this wouldn't be shown right?

In the current scenario, we would show the banner if I'm replying to a conversation which is not assigned to me or unassigned.

nithindavid: We could show banner if assigned to another agent but if the replies are only from you or no one has replied. Right now this shows always on all tickets.

pranavrajs:
Closing it here as we are not ready with product specs for this change. Let us reopen/raise a new PR once we decide on the logic and implementation.

PR #4373, opened 2022-04-04
fix:  Unread notification count in multiple accounts

### LINKED ISSUE #4367: The notification count is wrong in the realtime-notification update
The real-time notification update is returning the total number of notifications if the user has more than one accounts. It should only return the number of notifications in the current account.

**The problem:**

Screenshot 2022-04-04 at 11 34 46 AM

As seen here the notifications for the user in account 1 is 7, but when a new notification is created, the bell icon would show 14.

**Todo:**

- [ ] Send the only number of notifications in the account where notification is created.

muhsin-k:
Fixes #4367

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -188,10 +188,10 @@ class User < ApplicationRecord
     mutations_from_database.changed?('email')
   end
 
-  def notifications_meta
+  def notifications_meta(account_id)
     {
-      unread_count: notifications.where(read_at: nil).count,
-      count: notifications.count
+      unread_count: notifications.where(account_id: account_id, read_at: nil).count,
+      count: notifications.where(account_id: account_id).count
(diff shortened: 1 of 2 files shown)


sojan-official (approved):

app/models/user.rb
```diff
@@ -188,10 +188,10 @@ def will_save_change_to_email?
     mutations_from_database.changed?('email')
   end
 
-  def notifications_meta
+  def notifications_meta(account_id)
     {
-      unread_count: notifications.where(read_at: nil).count,
-      count: notifications.count
+      unread_count: notifications.where(account_id: account_id, read_at: nil).count,
```
pranavrajs, line 193: @muhsin-k @sojan-official Can we add a test case for the same?

PR #4414, opened 2022-04-07
enhancement: Remove clickaway for expanded reply box

### LINKED ISSUE #4409: enhancement: Expanded reply box should not close when clicked outside
While using the expanded reply box to compose a big message, The reply box will close if I accidentally click outside the reply box area or the backdrop. It's behaving similar to a modal now, would be better not to close it automatically.

@pranavrajs @sojan-official @fayazara Thoughts ?

nithindavid:
Fixes #4409

- Expanded reply box now closes only on click of close button or when esc key is pressed


iamsivin:
@nithindavid Can you remove the import line? I think we don't need it. 

https://github.com/chatwoot/chatwoot/blob/6266427dc0f56e273214a113c527b40cdfd847c0/app/javascript/dashboard/components/widgets/conversation/MessagesView.vue#L111

iamsivin (approved):
LGTM 👍🏻

PR #4421, opened 2022-04-08
feat: Add `phone_number` & `custom_attributes` in create conversation end point

muhsin-k:
- Added support for adding `phone_number` & `custom_attributes` in widget create conversation end point

### DIFF
--- a/app/controllers/api/v1/widget/conversations_controller.rb
+++ b/app/controllers/api/v1/widget/conversations_controller.rb
@@ -69,6 +69,7 @@ class Api::V1::Widget::ConversationsController < Api::V1::Widget::BaseController
   end
 
   def permitted_params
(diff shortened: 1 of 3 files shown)


sojan-official (approved):

app/controllers/api/v1/widget/conversations_controller.rb
```diff
@@ -7,12 +7,18 @@ def index
...
+      process_update_contact
       @conversation = create_conversation
       conversation.messages.create(message_params)
     end
   end
 
+  def process_update_contact
+    update_contact(contact_email) if @contact.email.blank? && contact_email.present?
+    update_contact_phone_number(contact_phone_number) if @contact.phone_number.blank? && contact_phone_number.present?
+    @contact.update!(name: contact_name) if contact_name.present?
```
pranavrajs, line 19: If a name exists already, this line would overwrite the existing name.

PR #4422, opened 2022-04-08
fix: GCP Redis managed don't support redis client setname command

mjattiot:
If using Redis on GCP, actioncable will attend to use the redis client setname command. However such command is not supported.
There is no way to change that behavior in cable.yml, so this alternative solution works as proposed here : https://github.com/rails/rails/issues/38244#issuecomment-575454444


pranavrajs:
@sojan-official Can you please check this PR and suggest the next steps?

sojan-official:
@mjattiot thanks a lot for the changes. let me learn about the impact of this change on existing installations and come back.
If there is no further concerns, will get this merged.

sojan-official:
ref: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/75173

mjattiot:
@sojan-official sorry I didnt get your message. We are using this patch in a chatwoot production installation and everything is working fine.

sojan-official:
@mjattiot since I didn't have a chance to evaluate the full impact of disabling id for action cable. I went with the route of adding an environment variable which can be used to  control this behaviour  `REDIS_DISABLE_CLIENT_COMMAND` 

Otherwise, I have tested the changes using the GCP memory store, and everything seems to be working.

mjattiot:
Cool thx !

PR #4442, opened 2022-04-11
Automation bug fix

tejaswinichile:
- [ ] 1. Webhook is working
- [ ] 2. Mute conversation - Directly marks as resolved - Should it be just mute or mute + resolved? - Shows the label as "Conversation was marked resolved by Automation     System".
- [ ] 3. Snooze conversation is not working.
- [ ] 4. Send Email to team - only sends email to the first person in the team index
- [ ] 5. Deleting automations are still running sometimes - not able to reproduce this.
- [ ]      Getting emails from Send Message to team - even if this was deleted.
- [ ] 6. Send message is working - But does not happen in realtime, need to refresh the page - Websocket is not getting updated on the automation trigger.
- [ ] 6. Deactivate automation has stopped working - Returns 422 unprocessable entity

### DIFF
--- a/app/controllers/api/v1/accounts/automation_rules_controller.rb
+++ b/app/controllers/api/v1/accounts/automation_rules_controller.rb
@@ -1,6 +1,6 @@
 class Api::V1::Accounts::AutomationRulesController < Api::V1::Accounts::BaseController
   before_action :check_authorization
-  before_action :fetch_automation_rule, only: [:show, :update, :destroy, :clone]
+  before_action :fetch_automation_rule, only: [:show, :update, :destroy, :clone, :staus_update]
 
   def index
     @automation_rules = Current.account.automation_rules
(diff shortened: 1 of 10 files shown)


app/controllers/api/v1/accounts/automation_rules_controller.rb
```diff
@@ -1,6 +1,6 @@
 class Api::V1::Accounts::AutomationRulesController < Api::V1::Accounts::BaseController
   before_action :check_authorization
-  before_action :fetch_automation_rule, only: [:show, :update, :destroy, :clone]
+  before_action :fetch_automation_rule, only: [:show, :update, :destroy, :clone, :staus_update]
```
pranavrajs, line 3: Spelling mistake

app/controllers/api/v1/accounts/automation_rules_controller.rb
```diff
@@ -19,6 +19,11 @@ def create
 
   def show; end
 
+  def staus_update
```
pranavrajs, line 22: Can you please add specs for this endpoint?

sojan-official (approved):

PR #4465, opened 2022-04-12
fix(widget): fetch old conversations when inside an iFrame

### LINKED ISSUE #4464: Fetch old conversations when inside an IFrame
My team is trying to embed the widget directly into our application without using the SDK using an IFrame.

However because of the line below the old conversations for the user with given `cw_conversation` token are not fetched.
https://github.com/chatwoot/chatwoot/blob/923b4637db3d255da9607b0b60206c23d4cee02c/app/javascript/widget/App.vue#L83

Is there a specific reason this fetch is not being executed on load?

Execute the following line when inside an IFrame and also fetch 
https://github.com/chatwoot/chatwoot/blob/923b4637db3d255da9607b0b60206c23d4cee02c/app/javascript/widget/App.vue#L89

You can trigger the fetching of messages by sending a `config-set` event. However, sending this message adds an extra loading indicator to the widget causing some bad UX.

https://github.com/chatwoot/chatwoot/blob/923b4637db3d255da9607b0b60206c23d4cee02c/app/javascript/widget/App.vue#L230

maartenvn:
## Description

Fetch old conversations when loading the widget inside an IFrame.

Fixes #4464 

## Type of change


pranavrajs:
@maartenvn Sorry for the delay, I'm not sure if I understand the change. Can you please share the use case you are trying to solve? 

We are fetching the old conversations even if it is outside of the Iframe. The sequence of calls is different in iFrame and non-iFrame use-cases. Please see the following code. 

https://github.com/chatwoot/chatwoot/blob/3d824855f0d8ceba733092f8048e809a92d5628a/app/javascript/widget/App.vue#L225-L232

pranavrajs (changes requested):
Please see the comment above.

pranavrajs:
Closing this as there is no activity on this. Also, we are refactoring the code in https://github.com/chatwoot/chatwoot/tree/refactor-widget

PR #4494, opened 2022-04-18
Fix: Slack update for file upload

tejaswinichile:
# Pull Request Template

## Description

When you upload a file from slack chat it's not appearing in the chatwoot inbox.
Slack updated the webhook event.


sojan-official (commented):
let's update the specs with the latest webhook format.

tejaswinichile:
@sojan-official I'm done with the spec changes, please have a look.

spec/lib/integrations/slack/incoming_message_builder_spec.rb
```diff
@@ -71,6 +71,18 @@
         expect(conversation.messages.last.content).to eql('this is test https://chatwoot.com Hey @Sojan Test again')
         expect(conversation.messages.last.attachments).to be_any
       end
+
+      it 'saves attachments if params file present and message blocks is not present' do
+        attachment_without_message_block = slack_attachment_stub.tap { |key| key[:event].delete(:blocks) }
```
tejaswinichile, line 76: This condition is failing according to the new changes https://github.com/chatwoot/chatwoot/pull/4494/files#diff-d27cad30e8a04dee27c18017fe61d997b066b15b3aa7b1188170f44be34cb3eaR57 where slack webhook is not sending the blocks for upload files event.

sojan-official: @tejaswinichile  if slack updated the webhooks format there isn't a point in retaining the old structure in our tests. So rather than adding the new spec, lets update `/Users/Sojan/Projects/Chatwoot/repos/chatwoot/spec/support/slack_stubs.rb` the attachment stub and ensure that the test case above this passes.

sojan-official (commented):
minor comments on updating the specs

sojan-official (approved):

PR #4496, opened 2022-04-18
fix: Update between business hour calculation

pranavrajs:
https://www.loom.com/share/5986f76066ec421ca549c4a7b746a385

Fixes https://github.com/chatwoot/product/issues/403


fayazara (approved):
Tested locally, LGTM!

PR #4498, opened 2022-04-18
chore: Sanitize html content 

### LINKED ISSUE #4318: feat:  HTML sanitizer component
**Problem**
There are many places we are using `v-html` to display HTML content, which may lead to **XSS** attacks. So we need to make sure that the HTML that's being rendered is sanitized and that it's safe to render in the browser.

**Solution**
- Create new component for displaying HTML content
- Pass element, content, class as props. Use `DomPurify` method to sanitize it.
- Whenever v-html has to be used, we will use the component instead.

muhsin-k:
Fixes #4318 and https://github.com/chatwoot/product/issues/355

- Added [vue-dompurify-html](https://github.com/LeSuisse/vue-dompurify-html) library
- Replaced `v-html` with `v-dompurify-html` everywhere in the app


fayazara (approved):
Tested all the places. this directive was used - working correctly.

iamsivin (approved):
Code LGTM👍🏻

PR #4515, opened 2022-04-20
chore: Update Help Center APIs to accomodate locale

pranavrajs:
- Add config for the locale in the portal
- Add locale in category


app/models/kbase/portal.rb
```diff
@@ -26,5 +27,5 @@ class Kbase::Portal < ApplicationRecord
 
   validates :account_id, presence: true
   validates :name, presence: true
-  validates :slug, presence: true
+  validates :slug, presence: true, uniqueness: true
```
pranavrajs, line 30: Added a unique constraint here.

app/controllers/api/v1/accounts/kbase/portals_controller.rb
```diff
@@ -21,7 +23,11 @@ def destroy
   private
 
   def fetch_portal
-    @portal = current_account.kbase_portals.find(params[:id])
+    @portal = Current.account.kbase_portals.find_by(slug: permitted_params[:id])
```
pranavrajs, line 26: instead of id used a slug for search.

PR #4516, opened 2022-04-20
fix: Dig params in widget contact end point

muhsin-k:
- Fixed contact update API failure if we don't pass email and phone number


pranavrajs:
@sojan-official @muhsin-k If this is happening in prod, let us create a hotfix branch and raise the PR against it instead of develop.

muhsin-k:
> @sojan-official @muhsin-k If this is happening in prod, let us create a hotfix branch and raise the PR against it instead of develop.

It is not happening on production

tejaswinichile (approved):
LGTM 💯

PR #4519, opened 2022-04-20
fix: Bypass invalid values in contact patch end point

### LINKED ISSUE #4233: Providing a bad phone number should not break PATCH for other attributes (email, name, etc.)

muhsin-k:
Fixes #4233 
- Added check for email and phone number validation
-  Prevent update If the values are invalid


tejaswinichile (approved):
LGTM 💯

PR #4550, opened 2022-04-25
chore: Ability to Remove password info from sentinel config

NusretOzates:
# Redis Sentinel Bugfix

## Description

We are using Chatwoot on our Openshift platform with Redis sentinels. The current code gives a "Resolving the “ERR Client sent AUTH, but no password is set” error. The reason for this error is we send password info for every sentinel but they don't have individual passwords. We only need to send the password in the base config.

Fixes # (issue)

## Type of change

## How Has This Been Tested?

With the current version, sidekiq or chatwoot cannot run with sentinels. After the change, we managed to run chatwoot and sidekiq on our server. You can test it by changing the env variables to use a Redis Sentinel.

## Checklist:

- [X] I have performed a self-review of my own code
- [X] Any dependent changes have been merged and published in downstream modules


NusretOzates:
CircleCI tests failed at the end because the failed test was expected to have a this:

```
 let(:expected_sentinels) do
      [
        { host: 'sentinel_1', port: '1234', password: 'some-strong-password' },
        { host: 'sentinel_2', port: '4321', password: 'some-strong-password' },
        { host: 'sentinel_3', port: '26379', password: 'some-strong-password' }
      ]
    end
```

which is directly about what I've changed so actually the test should change too

sojan-official:
@NusretOzates  thanks a lot for the changes. let me learn about the impact of this change on existing installations and come back. If there are no further concerns, will get this merged.

NusretOzates:
@sojan-official You are welcome :) I'm happy to contribute to the chatwoot as I'm working with it very closely!

tejaswinichile:
Check against the documentation https://github.com/redis-store/redis-rails
Set the config on local as given here https://github.com/redis-store/redis-rails#usage-with-redis-sentinel and it is working fine.

tejaswinichile (dismissed):
Tested on local with the given set of configurations in the documentation, password is not needed.
LGTM 💯

sojan-official:
@tejaswinichile I think sentinel also supports authentication with passwords for each instance as well. https://redis.io/docs/manual/sentinel/#configuring-sentinel-instances-with-authentication 

So we need to make sure we don't break any such configuration which was added earlier.

sojan-official:
@NusretOzates @tejaswinichile  What do think about introducing a new environment variable 
`REDIS_SENTINELS_AUTH` instead which can accept the value `false` and remove the password from the config based on the environment variable so that existing installations won't be affected.

NusretOzates:
@sojan-official I think it makes sense the only problem is. You will laugh at this but I'm a Python Developer/MLE so I can only read/understand Ruby but I cannot write Ruby so adding this is up to you guys 😄

sojan-official:
@NusretOzates No worries. We will take further steps to get this merged. Thanks again for the work

sojan-official:
Note: 

``` 
# By default, Chatwoot will pass REDIS_PASSWORD as the password value for sentinels
# Use the following environment variable to customize passwords for sentinels. 
# Use empty string if sentinels are configured without passwords
# REDIS_SENTINEL_PASSWORD=
```

PR #4607, opened 2022-05-03
Fix: Campaign triggers first_response reporting event

### LINKED ISSUE #4563: Campaign triggers first_response reporting event
Campaigns trigger the first_response reporting event even without agents responding to it. If an outgoing message is created from a campaign, it should not be marked as first-response.

**To reproduce:**

- Create a campaign
- Go to website where web widget is live
- Wait for the campaign message to appear. Click on the campaign.
- See the database/reports to see the first_response time of the conversation.

aswindevps:
## Description

This change prevents the campaign from triggering the first_response reporting event.

Fixes #4563

## Type of change

## How Has This Been Tested?

https://www.loom.com/share/99ec5092c7fe4e21a954515b774e42ce

## Checklist:


app/models/message.rb
```diff
@@ -6,6 +6,7 @@
 #  content             :text
 #  content_attributes  :json
 #  content_type        :integer          default("text"), not null
+#  created_by          :string
```
pranavrajs, line 9: I think it is better to use content_attributes here.

app/models/message.rb
```diff
@@ -168,7 +169,8 @@ def update_contact_activity
   def dispatch_create_events
     Rails.configuration.dispatcher.dispatch(MESSAGE_CREATED, Time.zone.now, message: self, performed_by: Current.executed_by)
 
-    if outgoing? && conversation.messages.outgoing.count == 1
+    campaign_outgoing_messages = conversation.messages.outgoing.where(created_by: 'Campaign')
```
pranavrajs, line 172: I would suggest saving campaign_id instead of "campaign" text, so it is easy for us to identify the campaign from which the message was triggered.

app/models/message.rb
```diff
@@ -168,7 +169,8 @@ def update_contact_activity
...
+    if outgoing? && (conversation.messages.outgoing.count - campaign_outgoing_messages.count) == 1
```
pranavrajs, line 173: Can't we just query messages where type = outgoing and campaign_id is nil?

app/builders/campaigns/campaign_conversation_builder.rb
```diff
@@ -21,7 +21,8 @@ def perform
 
   def message_params
     ActionController::Parameters.new({
-                                       content: @campaign.message
+                                       content: @campaign.message,
+                                       created_by: 'Campaign'
```
pranavrajs, line 25: I would suggest passing campaign_id here.

pranavrajs (changes requested):

db/migrate/20220503071925_add_created_by_to_message.rb
```diff
@@ -0,0 +1,5 @@
+class AddCreatedByToMessage < ActiveRecord::Migration[6.1]
+  def change
+    add_column :messages, :created_by, :string, default: nil
```
sojan-official, line 3: let's use campaign_id as suggested above for the attribute so that it stays consistent with the attribute we have in conversation. We will also be able to add belongs to => campaign association easily then.

pranavrajs: Do we need it as a separate attribute or can we add it in the content_attributes? For the large part, this field would be nil, right?

sojan-official: We can have it in content_attributes 👍

aswindevps: @pranavrajs At the moment the content_attributes column is JSON type. So it's difficult to query campaign_id from content_attributes. We need to take this as tech debt.

pranavrajs: @aswindevps I don't think we are querying campaign_id anywhere in this change, right? I guess what @sojan-official meant was that If at all required, we could query it from content_attributes.

sojan-official: After this change, we will have to figure out the number of outgoing messages without campaign_id. 
ref: https://github.com/chatwoot/chatwoot/blob/4f9d41962a3f98206297679ff65d72b9ab3af214/app/models/message.rb#L171

app/models/message.rb
aswindevps: Added new `jsonb` column `additional_attributes`

app/models/message.rb
aswindevps: Saved `campaign_id`  in `additional_attributes`.

app/models/message.rb
aswindevps: Updated to : 

outgoing? && conversation.messages.outgoing.where("(additional_attributes->'campaign_id') is null").count == 1

app/builders/campaigns/campaign_conversation_builder.rb
aswindevps: Done

db/migrate/20220503071925_add_created_by_to_message.rb
aswindevps: `if outgoing? && conversation.messages.outgoing.where("(additional_attributes->'campaign_id') is null").count == 1`

sojan-official (approved):

tejaswinichile (approved):
LGTM 💯

PR #4610, opened 2022-05-03
Fix: slack repeated callback event message for attached files

### LINKED ISSUE #4591: Message with Attachment triggers an unnecessary message if Slack is connected
**To Reproduce:**

- Connect Slack integration to the account
- Create a message with an attachment

**Dashboard View:** 

Screenshot 2022-05-02 at 8 05 08 PM

**Slack Thread View:**

Screenshot 2022-05-02 at 8 06 03 PM

tejaswinichile:
# Pull Request Template

## Description

Fixing "Attached File!" message from the file attachment message from slack.

Fixes #4591 

## Type of change

## How Has This Been Tested?

Tested on staging and checked all the tests are passing well.

## Checklist:


lib/integrations/slack/incoming_message_builder.rb
```diff
@@ -135,4 +135,8 @@ def file_type(attachment)
       :file
     end
   end
+
+  def attached_file_message?
```
sojan-official, line 139: can we add a link/documentation on when this event occurs / why?

lets also add a test case for this condition.

tejaswinichile: @sojan-official 
This is what we are adding from our side no slack changes, we add it as [initial comment for the attached file](https://github.com/chatwoot/chatwoot/blob/develop/lib/integrations/slack/send_on_slack_service.rb#L69) so when we fixed receiving the message from slack other than the supported message [here](https://github.com/chatwoot/chatwoot/pull/4494/commits/07a1c0da1e522d76e37b5f0cecdb4613389ab9b6) it is saving the 'Attached file!' message in CW dashboard when we get postback message for the same.

So the flow is:
1. You send a file from CW to the customer with the attached file which adds an 'Attached File!' message.
2. Which sends to slack and sends send it back to the CW with callback
3. Before, We had checked for the same; if it supported the message, but now we have one more condition which passes for the 'Attached File!' message.
4. Now I am adding this check-in 'Incoming message builder' if the message is 'Attached File!' ignore it and prevent it from creating in the CW.

Adding the test cases for the same.

sojan-official: @tejaswinichile makes sense.  can we add this as a comment in the code please :D. I am pretty sure someone will get confused with this condition down the lane when they look at the code.

tejaswinichile:
@sojan-official I have added the comment, please have a look.

sojan-official (approved):

PR #4649, opened 2022-05-09
Fix: Remove api from twitter oauth redirect URL

tejaswinichile:
Fixing: https://twittercommunity.com/t/error-when-authorize-twitter-nothing-to-see-here/170967


pranavrajs:
@sojan-official Agree! I don't think we need to rush this change.

tejaswinichile:
@pranavrajs @sojan-official 
Yeah, I will just have it as a draft.

pranavrajs:
Closing it here as the issue is resolved https://twittercommunity.com/t/error-when-authorize-twitter-nothing-to-see-here/170967/25

PR #4663, opened 2022-05-10
Fix: Agent Reports counts when they have access to multiple accounts

### LINKED ISSUE #4660: bug: Agent Reports counts are broken when they have access to multiple accounts
Image

Agent overview reports seems to be displaying data across all the accounts at the moment. We want to scope it to the specific account which administrator is viewing at the moment. 

## Fix

https://github.com/chatwoot/chatwoot/blob/f64cf85ab24e933e9469d0eab45c748f4f859f3e/app/helpers/report_helper.rb#L20

should be called with 
```
(get_grouped_values scope.conversations.where(account_id: @account).count
```

- should ensure that all the other scopes in this helper is also properly account scoped.

aswindevps:
## Description

This change restricts the agent report API to fetch agent metrics from the current account. 

Fixes #4660

## Type of change

## How Has This Been Tested?

https://www.loom.com/share/080f1d64e13c4acaa8025c1bcb992ce9

## Checklist:

### DIFF
--- a/app/helpers/report_helper.rb
+++ b/app/helpers/report_helper.rb
@@ -17,30 +17,30 @@ module ReportHelper
   end
 
   def conversations_count
-    (get_grouped_values scope.conversations).count
+    (get_grouped_values scope.conversations.where(account_id: account.id)).count
   end
 
   def incoming_messages_count
-    (get_grouped_values scope.messages.incoming.unscope(:order)).count
+    (get_grouped_values scope.messages.where(account_id: account.id).incoming.unscope(:order)).count
   end
 
   def outgoing_messages_count
-    (get_grouped_values scope.messages.outgoing.unscope(:order)).count
+    (get_grouped_values scope.messages.where(account_id: account.id).outgoing.unscope(:order)).count
   end
 
   def resolutions_count
-    (get_grouped_values scope.conversations.resolved).count
+    (get_grouped_values scope.conversations.where(account_id: account.id).resolved).count
   end
 
   def avg_first_response_time
-    grouped_reporting_events = (get_grouped_values scope.reporting_events.where(name: 'first_response'))
+    grouped_reporting_events = (get_grouped_values scope.reporting_events.where(name: 'first_response', account_id: account.id))
     return grouped_reporting_events.average(:value_in_business_hours) if params[:business_hours]
 
     grouped_reporting_events.average(:value)
   end
 
   def avg_resolution_time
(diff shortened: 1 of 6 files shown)


spec/controllers/api/v2/accounts/report_controller_spec.rb
```diff
@@ -223,6 +223,39 @@
...
+      let(:params) do
+        super().merge(
+          type: :agent,
+          since: 30.days.ago.to_i.to_s,
+          until: date_timestamp.to_s
+        )
+      end
+
+      it 'returns agent metrics from the current account' do
+        admin1 = create(:user, account: account1, role: :administrator)
```
sojan-official, line 240: we haven't added admin1 as a member of account 2 here I think. 
we have to add
```
create(:account_user, user: admin1, account: account2)
```

tejaswinichile (commented):
One small doubt, otherwise looks good to me.

spec/controllers/api/v2/accounts/report_controller_spec.rb
```diff
@@ -223,6 +223,39 @@
         expect(response).to have_http_status(:success)
       end
     end
+
+    context 'when an agent has access to multiple accounts' do
+      let(:account1) { create(:account) }
+      let(:account2) { create(:account) }
+
+      let(:params) do
+        super().merge(
```
pranavrajs, line 232: For my understanding, what does this do?

sojan-official: there is a params object defined in the parent's block.. super will access that and then the additional params will be merged into it

app/helpers/report_helper.rb
```diff
@@ -17,38 +17,38 @@ def scope
   end
 
   def conversations_count
-    (get_grouped_values scope.conversations).count
+    (get_grouped_values scope.conversations.where(account_id: account.id)).count
```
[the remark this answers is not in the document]
aswindevps: @sojan-official any thoughts here?

spec/controllers/api/v2/accounts/report_controller_spec.rb
aswindevps: Done

app/helpers/report_helper.rb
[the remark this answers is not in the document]
sojan-official: the link @tejaswinichile associated with the line isn't redirecting me to a specific line.  😅 .  So I am not sure. 
Would love to keep the account scoping in a unified place. But I don't mind taking it as refactoring for the future.

PR #4672, opened 2022-05-11
Feat: APIs to configure an auto assignment limit for inboxes

sojan-official:
This PR introduces the ability of an admin to configure a max assignment limit for auto assignments at an inbox level. 
Once it is configured, an agent wouldn't be auto-assigned further conversations from that inbox, once their assignment limit is reached.

https://user-images.githubusercontent.com/73185/167802526-10761ee5-4f69-4152-9556-5cf0fcb2f5eb.mp4

Fixes: https://github.com/chatwoot/product/issues/413

## QA
- [ ] Auto assignment works as usual 
- [ ] Configuring the limit works
- [ ] After the configuration of the limit, the limit is respected by auto-assignment.


app/helpers/api/v1/inboxes_helper.rb
```diff
@@ -74,4 +74,22 @@ def enable_openssl_mode(smtp_openssl_verify_mode)
     context.verify_mode = openssl_verify_mode
     context
   end
+
+  def account_channels_method
```
sojan-official, line 78: moved methods over her to satisfy rubocop

app/controllers/application_controller.rb
```diff
@@ -1,7 +1,7 @@
 class ApplicationController < ActionController::Base
   include DeviseTokenAuth::Concerns::SetUserByToken
   include RequestExceptionHandler
-  include Pundit
+  include Pundit::Authorization
```
sojan-official, line 4: fixing pundit deprecation warning

app/controllers/api/v1/accounts/inboxes_controller.rb
```diff
@@ -128,29 +132,13 @@ def channel_type_from_params
...
   end
-
-  def validate_limit
-    return unless Current.account.inboxes.count >= Current.account.usage_limits[:inboxes]
-
-    render_payment_required('Account limit exceeded. Upgrade to a higher plan')
-  end
 end
+
+Api::V1::Accounts::InboxesController.prepend_mod_with('Api::V1::Accounts::InboxesController')
```
sojan-official, line 144: will let us override methods in ee edition

app/models/account.rb
```diff
@@ -145,3 +144,5 @@ def remove_account_sequences
     ActiveRecord::Base.connection.exec_query("drop sequence IF EXISTS conv_dpid_seq_#{id}")
   end
 end
+
+Account.prepend_mod_with('Account')
```
sojan-official, line 148: following a unified style everywhere for prepend mod with

vishnu-narayanan:
@sojan-official If I understand correctly, `chatwootApp.enterprise?` is now relying on `DISABLE_ENTERPRISE` variable and/or the `enterprise` folder being present to enable/disable enterprise features. So at the moment, we will not be checking for a valid EE license to activate EE features?

sojan-official:
@vishnu-narayanan  That is correct. The license check will be implemented in the code and access to specific EE features will be limited by the features flags in the EE code.

enterprise/app/models/enterprise/inbox.rb
```diff
@@ -0,0 +1,16 @@
...
+    super - overloaded_agent_ids
+  end
+
+  private
+
+  def get_agent_ids_over_assignment_limit(limit)
+    # since we are checking for > than, lets reduce one from the number
+    # .abs will handle cases when the limit configured to be 0
+    validated_limit = (limit - 1).abs
+    conversations.open.select(:assignee_id).group(:assignee_id).having("count(*) > #{validated_limit}").filter_map(&:assignee_id)
```
pranavrajs, line 14: Can't we use `>=` instead of `>`?

sojan-official:
- added validations for assignment limit
- 
Screenshot 2022-06-07 at 4 34 05 PM

pranavrajs:
@sojan-official A few issues.

### Round robin assignment logic

**✅ Case 1: If all agents in the team are online**

After reaching the capacity of all agents, the system should move the new conversations to the unassigned queue.

**❌ Case 2: If only a couple of agents in the team are online**

Consider a team with 2 people. 1 of them is online and the other is offline. In this case, once the online agent's capacity is reached, the system should keep the new conversations unassigned. 

Right now, it is assigning the conversations to the offline agent.

**❌ Case 3: If all the agents in the team are offline**

Ideally, this should not assign to any agent in the team, and the system should keep the conversations in the unassigned queue. We have the current behavior in the system assigning the conversations to offline agents. This needs to be updated to the case where offline agents would not be considered in round-robin assignment. 


### Validation issue in the frontend
Screenshot 2022-06-08 at 5 48 44 PM

I can type any character in the input field and it doesn't show any error. The API calls don't pass this information to the backend. We need to add an additional validation to check whether the input is a number. Show error otherwise.

pranavrajs (changes requested):
I have identified some issues, please take a look at the comment above.

tejaswinichile (dismissed):
LGTM 💯

PR #4697, opened 2022-05-17
fix: Update report method to fix issues with special characters

pranavrajs:
- Updated the methods in frontend to fix the issue with special characters

### DIFF
--- a/app/javascript/dashboard/helper/specs/downloadHelper.spec.js
+++ b/app/javascript/dashboard/helper/specs/downloadHelper.spec.js
@@ -1,24 +1,4 @@
-import { downloadCsvFile, generateFileName } from '../downloadHelper';
-
-const fileName = 'test.csv';
-const fileData = `Agent name,Conversations count,Avg first response time (Minutes),Avg resolution time (Minutes)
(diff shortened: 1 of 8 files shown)


app/javascript/dashboard/helper/specs/downloadHelper.spec.js
```diff
@@ -1,24 +1,4 @@
...
-const fileName = 'test.csv';
-const fileData = `Agent name,Conversations count,Avg first response time (Minutes),Avg resolution time (Minutes)
-Pranav,36,114,28411`;
-
-describe('#downloadCsvFile', () => {
-  it('should download the csv file', () => {
-    const link = {
-      click: jest.fn(),
-    };
-    jest.spyOn(document, 'createElement').mockImplementation(() => link);
```
pranavrajs, line 12: Removed this test as `createObjectURL` is not testable with the version of jest we are using. Stubbing createObjectURL doesn't make sense as it is core of the method.

pranavrajs: https://stackoverflow.com/a/52969731

PR #4702, opened 2022-05-19
Destructuring response.data sometimes throws an exception

devinellis:
# Pull Request Template

## Description

I'm trying to get the SDK setUser working and I'm running into this problem:
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'data')
    at contacts.js:65:21
```

## Type of change

## How Has This Been Tested?

I'll admit, I didn't test this fix as I don't feel like getting a whole dev environment up and running.


pranavrajs:
@devinellis Sorry about the delay. I believe this fix would suppress the actual cause. The error might be propagated from some other place, that is why response.data is missing. Would you be able to share the error object you get in the catch block?

devinellis:
I fixed whatever was causing the underlying error, so this isn't a problem for me anymore. I'm curious though why you think it would suppress the error. My goal was to have the Error thrown reflect the issue, vs what I was getting which this a secondary error thrown within the catch. It seems like bad practice to assume keys like `data` are set within a catch block?

pranavrajs:
@devinellis 

I was under an assumption that `error.response` object is available for all Axios API calls, which is not the case, and that is a miss on my part. 

If you look at the `update` block, the only place where an error would be thrown is the API call. The rest of all of them shouldn't throw an error in our expectations. Hence my point was that if we do a safe operator here, the server/error logging mechanisms wouldn't be able to find out whether there are any other errors bubbling up from the dispatch calls although the users listening to `chatwoot:error` message would see a cryptic error.

I guess the change makes sense, as it is a better experience than a broken listener. So, merging it now. 

Would you be able to share what the issue was and how you fixed it?

pranavrajs (approved):

Issue #3428, opened 2021-11-20
feat: Update APIs for contact avatar


Stephen-Lobo:
**Describe the bug**

The thumbnail string for Contact is not getting updated. 

**To Reproduce**

1. Send PUT / PATCH request to update thumbnail of the contact through API
2. All other fields get updated, except THUMBNAIL

**Additional context**

Tried through POSTMAN and also through local PHP API

pranavrajs:
@Stephen-Lobo Can you please share the payload you are using?

Stephen-Lobo:
I am tried to use all PUT/PATCH/POST for {{host}}/{{api_version}}/accounts/{{account_id}}/contacts/some_contact_id through POSTMAN

with simple payload "
{
  "thumbnail": "url_of_some_picture.png"
}

This doesn't works. 

if I send similar request with complex payload like :

{
  "thumbnail": "url_of_some_picture.png",
   "phone_number": "1232323",
    "otherparam": "param_options"
}

All other param gets updated except thumbnail. 

Please try to test at your end.

sojan-official:
## Current behavior
- If an `avatar_url` parameter is passed in `setUser` method of widget SDK, chatwoot will update the contact Avatar. 
- if no `avater_url` is passed. But the contact has an email address, Chatwoot renders the gravatar for the contact if present ( chatwoot doesn't locally cache the gravatar. it generates the hashed URL for gravatar in backend instead ) 
ref: https://github.com/chatwoot/chatwoot/blob/9103fbbddeb24c2b5dba10f75087cb9b2ca6ef22/app/models/concerns/avatarable.rb#L17

## Suggested Changes 
- Contacts create and update endpoints in application APIs should be able to accept avatar attachment parameters [form data] 
- Contacts create and update endpoints should be able to accept the avatar_url parameter if users wish so. use  ContactAvatarJob
- update the API swagger docs 
- Instead of returning gravatar URLs on the fly, cache them as attachments on the contact. 
  -  this will stop the 404 errors while loading the dashboard 
  -  gives users the ability to remove avatars when deemed inappropriate. 


@Stephen-Lobo  I looked into the code and at the moment chatwoot application APIs don't support updating the contact avatar.  We will take this up for future releases.

akhilgkrishnan:
@sojan-official  I can pick this up

sojan-official:
@akhilgkrishnan  please go ahead

hasanbasri1993:
[...]

Issue #3853, opened 2022-01-26
feat: Option to disable gravatar usage


freddy36:
Due to EU GDPR regulations the current requests to gravatar are problematic (We would have to ask the chat visitor for consense first).

Due to this we would like to suggest an option to disable the usage of Gravatar completely.

iammattmartin:
Might be worth checking this as you'll only get the gravatar if they are already supplying their email, which you'd need their permission to do in the first place (i.e storage).

freddy36:
If an agent hasn't set an avatar, the client will do an gravatar request for the agent too.
The client request can be suppressed by setting some dummy `avatar_url` with `setUser()`.

jaime-rivas:
We are in a similar situation. It would be great to have the option to fully disable the usage of Gravatar.

Thanks for the great work!

ffroehlcke:
Same her! Please add an option to disable the usage of Gravatar.

sojan-official:
@jaime-rivas  @ffroehlcke  Could you share more details behind the requirement to disable gravatar? 
Is it aesthetics, or are there compliance issues that you are facing?

marcel-necker:
Here are a few things to consider:
- Compliance: As mentioned by @freddy36, this could be a compliance issue since a request is made to an external URL (gravatar.com) at repetitive intervals. This could potentially be used for tracking or analytics purposes (e.g. like Google Fonts does) that require an explicit user consent
- Aesthetics/Buggy: A request every minute which results in a 404 not found error in console doesn't seem ideal
- Usage: Gravatar is not that well known in some regions or customer audience. So they probably don't have an account there or use Gravatar at all

A way to optionally disable or enable (initially disabled) this feature would be nice!

jaime-rivas:
@sojan-official I concur with @marcel-necker

ffroehlcke:
@sojan-official Same here

KevinEduardo:
Same, gravatar also kinda sucks tbh

sojan-official:
Thanks a lot for the feedback. We will introduce an environment variable, which lets you disable gravatar once https://github.com/chatwoot/chatwoot/pull/4719/ is merged

PR #4719, opened 2022-05-23
feat: Update API for contact avatar

giquieu:
# Pull Request Template

## Description

Added the possibility to inform the contact's avatar via API and Dashboard.
Contacts create and update endpoints in app APIs must be able to accept avatar attachment parameters [form data].
Contacts that create and update endpoints must be able to accept the avatar_url parameter if users so choose.

Fixes # (issue)
#3428 

## Type of change

## How Has This Been Tested?
Insert avatar contact via Dashboard.
Updated the contact, changing the avatar.
Removed contact avatar from Dashboard.
Used API: /api/v1/accounts/{account}/contacts/{contact}
Via API it is possible to use avatar_url, informing the avatar's URL, Chatwoot will be in charge of downloading the image and attaching it to the contact.

## Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented on my code, particularly in hard-to-understand areas
- [X] I have added tests that prove my fix is effective or that my feature works


hasanbasri1993:
Nice feature, this is i wait for.

giquieu:
The POST return is the contact itself without the thumbnail. I came back like this to propagate change via action. However, it has the same effect with socket change notification.
Perhaps doing a DELETE would be more appropriate?

pranavrajs:
@sojan-official Can you please take a look at this?

sojan-official:
@giquieu apologies for the delay on this one. I will have it reviewed this week. Thanks a lot for working on this.

giquieu:
Np @sojan-official , Thanks for the review and I'm open to suggestions.

app/controllers/api/v1/accounts/contacts_controller.rb
```diff
@@ -131,19 +138,23 @@ def build_contact_inbox
     ContactInbox.create(contact: @contact, inbox: inbox, source_id: source_id)
   end
 
+  def permitted_params
+    params.permit(:name, :identifier, :email, :phone_number, :avatar, :avatar_url, additional_attributes: {}, custom_attributes: {})
+  end
+
   def contact_params
     params.require(:contact).permit(:name, :identifier, :email, :phone_number, additional_attributes: {}, custom_attributes: {})
   end
```
sojan-official, line 148:
```suggestion
```

Let's have it removed since we replaced it with `permitted_params`

app/models/concerns/avatarable.rb
```diff
@@ -12,10 +12,10 @@ module Avatarable
...
     return url_for(avatar.representation(resize: '250x250')) if avatar.attached? && avatar.representable?
 
-    if [SuperAdmin, User, Contact].include?(self.class) && email.present?
-      hash = Digest::MD5.hexdigest(email)
-      return "https://www.gravatar.com/avatar/#{hash}?d=404"
-    end
+    # if [SuperAdmin, User, Contact].include?(self.class) && email.present?
+    #   hash = Digest::MD5.hexdigest(email)
+    #   return "https://www.gravatar.com/avatar/#{hash}?d=404"
+    # end
```
sojan-official, line 18:
```suggestion
     if [SuperAdmin, User, Contact].include?(self.class) && email.present?
       hash = Digest::MD5.hexdigest(email)
       return "https://www.gravatar.com/avatar/#{hash}?d=404"
     end
```

let's keep this in since removing this will stop rendering gravatar for existing Contacts and Users. Let's take it up as a follow-up in https://github.com/chatwoot/chatwoot/issues/3853, with 
- ENV variable to control whether to use gravatar
- migration to populate the value for existing accounts 
- extend the contact avatar job to associate avatars to users on creation as well.

sojan-official (commented):
@giquieu  I have tested this PR, and it works well. Added a couple of comments. I will let someone from the FRONTEND team have another pass, and we will merge it.

sojan-official:
> The POST return from destroy_avatar is the contact itself without the thumbnail. Returns as the body of the response, so the EDIT_CONTACT mutation was used to propagate the change.
> 
> However, the expected result to notify listeners of the contact/avatar change is also done via socket, through CONTACT_CREATED and CONTACT_UPDATED.
> 
> Sending a request of type DELETE could be more appropriate?

For now, I think we can go ahead with the `CONTACT_CREATED\UPDATED` event since we don't have a specific event for `DELETE`. Let's revisit this in the future. 
and once could also think about it as making the contacts' avatar attribute as blank, which is an update :D

app/models/concerns/avatarable.rb
sojan-official: There have been some concerns about using external services. But we could look into providing it as configurable options.

sojan-official:
note: also switch the `post .../destroy_avatar` to `delete .../avatar` which stays consistent with other end points

giquieu:
@sojan-official 
It makes sense to keep the default for the other exclusions. Thanks for the improvements. :)

I sent an adjustment to the specification and a correction to the AvatarUploader's handleImageUpload, which when canceling the selection window returned an error.
createObjectURL file undefined

A short video of what happened:
https://www.loom.com/share/c7bc03804c5c4ba184a99f34e0585698

sojan-official:
@giquieu Thanks a lot for the changes. About the delete button, in #3853, we will remove the behavior of sending the gravatar URLs directly and switching entirely to the attached avatar. So, let's not do extra work on it now.

PR #4647, opened 2022-05-09
Feat: Bulk actions

fayazara:
# Pull Request Template

Demo
https://www.loom.com/share/6b55e165a17d498caae956fcd06bb617

## Description
This PR allows you to bulk update conversations.
1. Assigning conversations to agent
2. Resolving conversations

Fixes #3846 

## Type of change

Please delete options that are not relevant.
- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Test cases for Store, and QA.

## Checklist:
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #4722, opened 2022-05-23
chore: Add Assignable Agents API

pranavrajs:
With https://github.com/chatwoot/chatwoot/pull/4647, users can select conversations from multiple inboxes and assign them to someone. This change would make sure that the list of agents displayed has access to all selected inboxes. I also used a new path AssignableAgents instead of inbox namespace, as we might be adding permissions over the team as well. This would become a generic API to list the assignable agents.


sojan-official (approved):
LGTM

pranavrajs:
@sojan-official Merging this, I think @vishnu-narayanan's PR would fix the issue with caching.

PR #4764, opened 2022-05-30
fix: Disable marking IMAP connection as invalid for Standard Errors

pranavrajs:
With the current setup, the IMAP connection is marked as invalid for all types of errors. This is not the right way to handle the errors as an underlying data issue might also mark the IMAP connection as invalid.


tejaswinichile (commented):
LGTM 💯 

@pranavrajs Just one thing, this rails logged should go inside the error label or warn since we decided to move the logger info to `Rails.logger.warn`

tejaswinichile (dismissed):

aswindevps (approved):
LGTM

PR #4777, opened 2022-06-01
feat: Article search

### LINKED ISSUE #4802: Add a dashboard API to return all articles/edit/destroy in a category, all agents who have access to the portal can create/edit/destroy/view an article

tejaswinichile:
# Pull Request Template

## Description

Fixes #4802  

## Type of change

## How Has This Been Tested?

## Checklist:


app/models/article.rb
```diff
@@ -11,16 +11,26 @@
...
+#
+#  fk_rails_...  (author_id => users.id)
+#
 class Article < ApplicationRecord
+  include PgSearch::Model
+
   belongs_to :account
   belongs_to :category
   belongs_to :portal
-  belongs_to :folder
```
tejaswinichile, line 23: Just when we finalize the whole flow, I will remove the respective table for this.

app/views/api/v1/accounts/articles/_article.json.jbuilder
```diff
@@ -0,0 +1,12 @@
...
+json.category_id article.category_id
+json.title article.title
+json.content article.content
+json.description article.description
+json.status article.status
+json.account_id article.account_id
+json.portal_id article.portal_id
+json.views article.views
+json.author_id article.author_id
+json.category_id article.category_id
```
pranavrajs, line 11: duplicate category id.

app/views/api/v1/accounts/articles/_article.json.jbuilder
```diff
@@ -0,0 +1,12 @@
...
+json.folder_id article.folder_id
```
pranavrajs, line 12: Please remove folder_id. We will not be using it.

app/views/api/v1/accounts/articles/_article.json.jbuilder
```diff
@@ -0,0 +1,12 @@
+json.id article.id
+json.category_id article.category_id
+json.title article.title
+json.content article.content
+json.description article.description
+json.status article.status
+json.account_id article.account_id
+json.portal_id article.portal_id
+json.views article.views
+json.author_id article.author_id
```
pranavrajs, line 10: Should we include author instead of author_id. Similarly portal instead of portal_id?

spec/controllers/api/v1/accounts/articles_controller_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+  describe 'POST /api/v1/accounts/{account.id}/portals/{portal.slug}/articles' do
+    context 'when it is an unauthenticated user' do
+      it 'returns unauthorized' do
+        post "/api/v1/accounts/#{account.id}/portals/#{portal.slug}/articles", params: {}
+        expect(response).to have_http_status(:unauthorized)
+      end
+    end
+
+    context 'when it is an authenticated user' do
+      it 'creates article' do
```
pranavrajs, line 19: If the user is of agent role, then they should be a portal member to access it. 

If this case is not handled here, let us pick it as a separate ticket.

spec/controllers/api/v1/accounts/articles_controller_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+        article_params = {
+          article: {
+            category_id: category.id,
+            description: 'test description',
+            title: 'MyTitle',
+            content: 'This is my content.',
+            status: 1,
```
pranavrajs, line 26: Can we pass `published` string here instead of 1?

app/controllers/api/v1/accounts/articles_controller.rb
```diff
@@ -0,0 +1,38 @@
+class Api::V1::Accounts::ArticlesController < Api::V1::Accounts::BaseController
+  before_action :portal
+  before_action :fetch_article, except: [:index, :create]
+
+  def index
+    @articles = @portal.articles.where(status: :published)
+    @articles.search(params[:payload]) if params[:payload].present?
+  end
+
```
pranavrajs, line 9: Please add :show action as well.

tejaswinichile: I did it because we only did the show with edit action for automation pages.

app/models/article.rb
```diff
@@ -11,16 +11,25 @@
 #  created_at  :datetime         not null
 #  updated_at  :datetime         not null
 #  account_id  :integer          not null
-#  author_id   :integer
+#  author_id   :bigint
```
pranavrajs, line 14: Shouldn't we add locale to the article as well? If we need to search all the articles in the locale `en`, with this modeling we would have to find all the categories in the locale and find the articles in those categories.

This doesn't need to take from the user, it can be derived from the category during creation.

Thoughts?

db/migrate/20220527120826_add_index_on_category_slug_and_locale.rb
```diff
@@ -0,0 +1,6 @@
+class AddIndexOnCategorySlugAndLocale < ActiveRecord::Migration[6.1]
+  def change
+    add_column :categories, :slug, :string, null: false, default: 'common'
```
pranavrajs, line 3: What does the `default` mean here? This slug has to be unique for a (portal, locale), right?

db/migrate/20220527120826_add_index_on_category_slug_and_locale.rb
```diff
@@ -0,0 +1,6 @@
...
+    add_index :categories, [:slug, :locale], unique: true
```
pranavrajs, line 4: I think the uniqueness should be on [portal_id, slug, locale] not on [slug, locale]

app/controllers/api/v1/accounts/articles_controller.rb
```diff
@@ -0,0 +1,38 @@
+class Api::V1::Accounts::ArticlesController < Api::V1::Accounts::BaseController
+  before_action :portal
+  before_action :fetch_article, except: [:index, :create]
+
+  def index
+    @articles = @portal.articles.where(status: :published)
+    @articles.search(params[:payload]) if params[:payload].present?
```
pranavrajs, line 7: What is the payload here? Can we use permitted_params? 

Please add a spec for the same.

app/models/article.rb
tejaswinichile: @pranavrajs 
We will have an extra column to save the data copied from the category, but we need to join with the categories with articles on the listing page if the user wants to apply search on it.

I am not in favor of redundant information columns until and unless the join and search are very heavy to handle. We can take this if we face any such issues.

app/controllers/api/v1/accounts/articles_controller.rb
```diff
@@ -0,0 +1,38 @@
+class Api::V1::Accounts::ArticlesController < Api::V1::Accounts::BaseController
+  before_action :portal
+  before_action :fetch_article, except: [:index, :create]
+
+  def index
+    @articles = @portal.articles.where(status: :published)
```
pranavrajs, line 6: I don't think we should limit this to published articles only. This is the admin view, right? It list all the articles in the account.

db/migrate/20220527120826_add_index_on_category_slug_and_locale.rb
tejaswinichile: The slug values are not null, in that case, robocop gives an error that it should have some default values assigned to it, that's why added `common`, has no meaning as such.

pranavrajs (changes requested):
Overall nice work! I have some comments, please check the individual comments.

app/controllers/api/v1/accounts/articles_controller.rb
tejaswinichile: For index? We can use permitted params, but we do it by searching based on params name, so there won't be any injection while searching.

app/controllers/api/v1/accounts/articles_controller.rb
pranavrajs: We need an individual article route as well. We are not showing the information in a modal, instead, it is shown on a different page, fetching all the articles is not ideal in that case.

db/migrate/20220527120826_add_index_on_category_slug_and_locale.rb
pranavrajs: This means that if I added a category without a slug, it would update the slug as `common` right? This is not intended behavior. Can you please share the failing rubocop error?

app/controllers/api/v1/accounts/articles_controller.rb
tejaswinichile: Basically this would be the payload for search articles 

```
{
    "payload": { 
        "locale": "es",
        "category_slug": "agent-category-1"
    }
}
```

app/views/api/v1/accounts/articles/_article.json.jbuilder
```diff
@@ -0,0 +1,10 @@
+json.id article.id
+json.category_id article.category_id
+json.title article.title
+json.content article.content
+json.description article.description
+json.status article.status
+json.account_id article.account_id
+json.portal article.portal
+json.views article.views
+json.author article.author
```
pranavrajs, line 10: This would send all attributes of the user, right? I think we should use the agent.json.jbuilder view here.

app/views/api/v1/accounts/articles/_article.json.jbuilder
```diff
@@ -0,0 +1,10 @@
+json.id article.id
+json.category_id article.category_id
+json.title article.title
+json.content article.content
+json.description article.description
+json.status article.status
+json.account_id article.account_id
+json.portal article.portal
```
pranavrajs, line 8: Let us use the view made for portal just to make sure we are not sending unnecessary information.

app/views/api/v1/accounts/articles/_article.json.jbuilder
```diff
@@ -0,0 +1,14 @@
...
+json.title article.title
+json.content article.content
+json.description article.description
+json.status article.status
+json.account_id article.account_id
+json.portal do
+  json.partial! 'api/v1/accounts/portals/portal.json.jbuilder', portal: article.portal
+end
+json.views article.views
+json.author do
```
pranavrajs, line 12: If the author is nil, would it break the API?

pranavrajs: Also missed from the last review, if the author is deleted does it nullify the article author? What is the current behavior?

db/migrate/20220527120826_add_index_on_category_slug_and_locale.rb
tejaswinichile: I removed the `common` setting it to `category` now or we could set it to the `blank` value.
Sorry, not RoboCop Postgres gives an error if we don't send any value, the following is the error.

`PG::NotNullViolation: ERROR:  column "slug" of relation "categories" contains null values`

tejaswinichile: To resolve the rubocop error for the same https://github.com/rubocop/rubocop-rails/issues/237

sojan-official (commented):
LGTM mostly. small comments on the annotations.

sojan-official (approved):

PR #4790, opened 2022-06-03
Make "apk" commands consistent in Dockerfile

jordan-brough:
Just some minor tweaks that seemed like nice-to-haves while I was working w/ the Dockerfile.

1. Add "--no-cache" to some "apk" command that are missing it
2. ~Use convenience option "--update" consistently~
3. ~Use full option name "--update-cache" instead of "--update" (easier to
identify in "man" docs, etc)~
4. Remove `--update-cache` option because it's not needed with `--no-cache` ([see comment](https://github.com/chatwoot/chatwoot/pull/4790#issuecomment-1146682237))


paescuj:
I don't think `--update-cache` is necessery when using the `--no-cache` option. See https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache.

jordan-brough:
Nice! I've updated to remove `--update-cache`.

vishnu-narayanan (approved):
@jordan-brough Thank you for the PR.  😄 
@paescuj Thank you for your input. TIL.

PR #4795, opened 2022-06-06
chore: Remove flay, rspec from development/test group

pranavrajs:
`flay` and `rspec` are development dependencies of the `devise-secure_password` gem. It shouldn't generate an error message ideally if we have not installed it. So, I made a change [here](https://github.com/chatwoot/devise-secure_password/commit/d777b04f12652d576b1272b8f39857e3e0b3fc26) to avoid that. 

These 2 gems create a lot of warnings on the M1 mac. This PR removes these gems.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -156,11 +156,6 @@ group :test do
 end
 
 group :development, :test do
-  # TODO: is this needed ?
-  # errors thrown by devise password gem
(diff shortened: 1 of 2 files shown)


tejaswinichile (approved):
LGTM 💯

PR #4835, opened 2022-06-10
chore: Add support for message_templates in API inbox

pranavrajs:
- Save message templates in the additional attributes in the API inbox
- Render template selector if message templates are present


app/javascript/dashboard/routes/dashboard/settings/inbox/Settings.vue
```diff
@@ -562,7 +561,6 @@ export default {
   },
   validations: {
     webhookUrl: {
-      required,
```
pranavrajs, line 565: WebhookURL is not required in the API inbox, so removing this validation.

PR #4873, opened 2022-06-15
chore: Inbox management copy update

gk7700:
## Description

- Fixed the grammatical error.
- Modified the 2nd and 3rd paragraphs to make them look more user-friendly.


sojan-official:
@gk7700 Thanks for the contribution. Since it's a trivial copy change, let's skip updating this. Mostly because it will cause significant effort for the community translators as they will have to retranslate the string across all the languages.

ref: https://www.chatwoot.com/docs/handbook/engineering/pr-guidelines#avoid-cosmetic-changes

PR #4901, opened 2022-06-20
chore: Email improvements.

aswindevps:
## Description

- This change replace the timestamp logic with the message exists check in Chatwoot. 
- The email fetch count increased to 60.

Fixes https://github.com/chatwoot/product/issues/402
Fixes https://github.com/chatwoot/product/issues/371
Fixes https://github.com/chatwoot/chatwoot/issues/4932

## Type of change

## How Has This Been Tested?

- Verified email continuity with IMAP and SMTP.
- Emails are now available in Chatwoot after moving from spam to inbox.

## Checklist:


app/jobs/inboxes/fetch_imap_emails_job.rb
```diff
@@ -32,16 +32,11 @@ def fetch_mail_for_channel(channel)
                               enable_ssl: channel.imap_enable_ssl
     end
 
-    new_mails = false
-
-    Mail.find(what: :last, count: 10, order: :desc).each do |inbound_mail|
-      next unless inbound_mail.date.utc >= channel.imap_inbox_synced_at
+    Mail.find(what: :last, count: 60, order: :asc).each do |inbound_mail|
```
sojan-official, line 35: since we are switching to 1-minute intervals for mail fetch. Should we go for a less aggressive number for the count like the original `10` ?
if the jobs end up taking more time iterating over all the mails. it could clog up the workers?

sojan-official (commented):
The code looks good to me. minor comment on the count of emails that we are fetching.

app/jobs/inboxes/fetch_imap_emails_job.rb
aswindevps: @sojan-official what would be the possible number of emails per minute?

sojan-official: We can stick to 10 I guess. since we compensate with a quicker polling interval. lets get feedback and then iterate

sojan-official (commented):
LGTM lets test and get this merged.

aswindevps:
@sojan-official @pranavrajs Below scenarios are tested and verified.

- The user sends an email, the agent replies to the email, user replies to the email.
   **Expected**: The replies to the conversation should not create a separate conversation.

- User sends an email, the agent sends a reply, the agent sends a reply, user replies to the first mail.
   **Expected**: The replies to the conversation should not create a separate conversation.

- The user sends an email, the user sends a follow-up, the agent sends a reply, user replies to the mail.
   **Expected**: The replies to the conversation should not create a separate conversation.

- The user sends an email, and after that Chatwoot inbox is created. Send a follow-up to the first mail. Send another to the first mail.
   **Expected**: Even though the parent is not available in Chatwoot, it should show as the same thread.

- If I have more than one inbox connected in Chatwoot, the user sends the same message to both inboxes.
   **Expected**: It should create a conversation in both inboxes with the same message.

- User sends an email, but the email initially landed in spam. Then moved to the inbox and marked not spam.
   **Expected**: It should create a conversation after moving from spam to inbox.

PR #4906, opened 2022-06-21
chore: Append incoming SMS to last conversation

sojan-official:
>   the issue that has come up is the continuity of SMS conversations started from the contacts page. If a “New Message” is sent to contact from their contact page, instead of adding it to the current conversation with that contact, it creates a completely new conversation. However, when that contact then responds, it then adds it to the original conversation with that contact. We are ending up with straggler conversations that consist of only 1 outgoing message

Note: 

Look into retaining a single conversation for WhatsApp / SMS


sojan-official:
`Look into retaining a single conversation for WhatsApp / SMS` will be tracked over https://github.com/chatwoot/chatwoot/issues/4975

PR #4912, opened 2022-06-22
feat: CRUD operation for linked articles to current articles

tejaswinichile:
# Pull Request Template

## Description

Fixes # (issue)

## Type of change

## How Has This Been Tested?

## Checklist:


app/models/article.rb
```diff
@@ -2,31 +2,41 @@
...
+#  id                    :bigint           not null, primary key
+#  content               :text
+#  description           :text
+#  status                :integer
+#  title                 :string
+#  views                 :integer
+#  created_at            :datetime         not null
+#  updated_at            :datetime         not null
+#  account_id            :integer          not null
+#  associated_article_id :bigint
```
sojan-official, line 14: let's use the same naming conventions as we have used in categories. so `linked_article_id` ? 
or else the synonyms could start causing confusion

app/models/article.rb
```diff
@@ -2,31 +2,41 @@
...
 #
 # Foreign Keys
 #
+#  fk_rails_...  (associated_article_id => articles.id)
 #  fk_rails_...  (author_id => users.id)
 #
 class Article < ApplicationRecord
   include PgSearch::Model
 
+  has_many :associated_articles,
```
sojan-official, line 33: let's do linked articles

app/models/article.rb
```diff
@@ -2,31 +2,41 @@
...
+           class_name: :Article,
+           foreign_key: :associated_article_id,
+           dependent: :nullify,
+           inverse_of: 'associated_article'
+
+  belongs_to :associated_article, class_name: :Article, optional: true
```
sojan-official, line 39: little confused here since `linked_article` along with `linked_articles` could cause confusion. But let's sync up and ensure a consistent naming across categories and articles.

app/models/article.rb
```diff
@@ -71,6 +81,37 @@ def self.current_page(params)
     params[:page] || 1
   end
 
+  def associate_root_article(associated_article_id)
+    article = portal.articles.find(associated_article_id) if associated_article_id.present?
```
sojan-official, line 85: We might not need to do recursion with the query.

we could simplify the logic to 

- we fetch the associated article. 
- attach it to current articles if the associated article doesn't have another association ... 
- in the case of another association, we save that id instead.

this pattern would ensure that there won't be a deeper level association saved in the database.

sojan-official (commented):
looks good overall. added few comments

app/models/article.rb
tejaswinichile: @sojan-official Makes sense if we are already associated root article from the start to the current article.

app/models/article.rb
tejaswinichile: I couldn't think of a better inverse association for linked_articles.

app/models/article.rb
tejaswinichile: We have decided on associated_articles and will be raising a PR for category association changes.

sojan-official (approved):
LGTM

PR #4920, opened 2022-06-23
fix: Inbox name validation with regex format for special characters

tejaswinichile:


app/models/inbox.rb
```diff
@@ -33,7 +33,10 @@ class Inbox < ApplicationRecord
   include Avatarable
   include OutOfOffisable
 
-  validates :name, presence: true, format: { with: /^\b[\w\s]*\b$/, multiline: true }
+  # Not allowing characters:
+  validates :name, presence: true,
+                   format: { with: %r{^^\b[^/\\<>@]*\b$}, multiline: true,
+                             message: 'Should not start and end with symbols and should not have < > / \ @ symbols' }
```
pranavrajs, line 39:
```suggestion
                             message: 'The name should not start or end with symbols, and it should not have < > / \ @ characters.' }
```

tejaswinichile: @pranavrajs Name is auto included in the validation message, as it is the attribute name.

pranavrajs:
@tejaswinichile It looks like FB/Twitter allows some of these characters. We should scope the validation to email, website, and API channels only.

image

spec/models/inbox_spec.rb
```diff
@@ -164,13 +165,31 @@
...
 
         inbox.name = 'inbox_name.and_1'
         expect(inbox).to be_valid
       end
+
+      context 'when special characters allowed for some channel' do
+        let!(:tw_channel_val) { FactoryBot.create(:channel_twitter_profile) }
+        let(:inbox) { create(:inbox, channel: tw_channel_val) }
+
+        it 'does allow chacters like /\@<> for Facebook Channel' do
```
pranavrajs, line 188:
```suggestion
        it 'does allow special characters like /\@<> for Facebook Channel' do
```

pranavrajs (approved):

PR #4940, opened 2022-06-28
chore: Fix sentry issues

aswindevps:
Fixes https://github.com/chatwoot/chatwoot/issues/4810


app/jobs/inboxes/fetch_imap_emails_job.rb
```diff
@@ -9,8 +9,10 @@ def perform(channel)
     fetch_mail_for_channel(channel)
     # clearing old failures like timeouts since the mail is now successfully processed
     channel.reauthorized!
-  rescue Errno::ECONNREFUSED, Net::OpenTimeout, Net::IMAP::NoResponseError
+  rescue Errno::ECONNREFUSED, Net::OpenTimeout, Net::IMAP::NoResponseError, Errno::ECONNRESET, Errno::ENETUNREACH, Net::IMAP::ByeResponseError
     channel.authorization_error!
+  rescue EOFError => e
```
pranavrajs, line 14: What trigger EOFError? Do we need to notify the user?

app/mailboxes/mailbox_helper.rb
```diff
@@ -35,7 +35,7 @@ def add_attachments_to_message
 
   def create_contact
     @contact_inbox = ::ContactBuilder.new(
-      source_id: "email:#{processed_mail.message_id}",
+      source_id: processed_mail.original_sender,
```
pranavrajs, line 38: Existing conversations would break if we push this directly.

tejaswinichile: @tejaswinichile Can we change the `original_sender` method, how do we consider `message_id` further?

sojan-official: I am not sure this has an impact on existing conversations. In the case of email Inboxes, we are fetching contact inbox using a combination of contact & inbox, not source_id.  So this change should help us to make the behavior consistent: "source_id will store the identifier of the contact in the external system"

tejaswinichile: As per the discussion with @aswindevps email's `message_id`  can not be the source_id for the `contact_inbox` we are saving that as a message's source_id which would be the source of the truth for the message.

Updating the source_id for the contact_inbox to the original_sender makes sense here and as we dug through the code it won't have any further impact.

tejaswinichile (approved):
LGTM 💯

app/jobs/inboxes/fetch_imap_emails_job.rb
sojan-official: It seems random on the production amidst the account we observed. sometimes the server responds with EOFError, but on subsequent retries, it was working fine. 

ref: https://github.com/tpitale/mail_room/issues/15

PR #4946, opened 2022-06-29
feat: Portal and Category public APIs

### LINKED ISSUE #4600: Add a public API to return the details of a portal.

tejaswinichile:
# Pull Request Template

## Description

1. Return the portals available on the account
2. Return all the categories belonging to the portal with slug

Fixes #4600 

## Type of change

## How Has This Been Tested?

1. Written the test cases to get the information passed.
2. As it is a public API no check for authorization is involved.

## Checklist:

### DIFF
--- a/app/controllers/api/v1/accounts/portals_controller.rb
+++ b/app/controllers/api/v1/accounts/portals_controller.rb
@@ -26,6 +26,11 @@ class Api::V1::Accounts::PortalsController < Api::V1::Accounts::BaseController
     head :ok
   end
 
+  def archive
+    @portal.update(archive: true)
+    head :ok
+  end
+
   private
 
   def fetch_portal
--- a/app/controllers/public/api/v1/portals_controller.rb
+++ b/app/controllers/public/api/v1/portals_controller.rb
@@ -1,9 +1,5 @@
 class Public::Api::V1::PortalsController < PublicController
-  before_action :set_portal, except: [:index]
-
-  def index
-    @portals = Portal.all
-  end
+  before_action :set_portal
 
   def show; end
 
--- a/app/controllers/public/api/v1/portals_controller.rb
+++ b/app/controllers/public/api/v1/portals_controller.rb
@@ -1,0 +1,15 @@
+class Public::Api::V1::PortalsController < PublicController
+  before_action :set_portal, except: [:index]
+
+  def index
+    @portals = Portal.all
+  end
+
+  def show; end
+
+  private
+
+  def set_portal
+    @portal = ::Portal.find_by!(slug: params[:slug], archived: false)
+  end
+end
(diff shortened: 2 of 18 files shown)


app/controllers/api/v1/accounts/portals_controller.rb
```diff
@@ -2,7 +2,7 @@ class Api::V1::Accounts::PortalsController < Api::V1::Accounts::BaseController
   before_action :fetch_portal, except: [:index, :create]
 
   def index
-    @portals = Current.account.portals
+    @portals = Current.account.portals.active
```
pranavrajs, line 5: In the internal APIs, we should list all the portals, right?

app/controllers/public/api/v1/portals_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Public::Api::V1::PortalsController < PublicController
+  before_action :set_portal, except: [:index]
+
+  def index
+    @portals = Portal.all.active
+  end
```
pranavrajs, line 6: How does this work? If it lists all the portals available in the installation, please remove this end point.

We would need the show end point only.

app/controllers/api/v1/accounts/portals_controller.rb
tejaswinichile: Yes, I forgot to remove this after our discussion yesterday. 👍🏼

app/views/public/api/v1/models/_category.json.jbuilder
```diff
@@ -1,5 +1,25 @@
...
+
+json.related_categories do
+  if category.related_categories.any?
+    json.array! category.related_categories.each do |related_category|
+      json.partial! partial: 'associated_category', category: related_category
+    end
+  end
+end
+
+if category.parent_category.present?
```
sojan-official, line 15: root ?

app/views/public/api/v1/models/_category.json.jbuilder
```diff
@@ -1,5 +1,25 @@
...
+  json.parent_category do
+    json.partial! partial: 'associated_category', category: category.parent_category
+  end
+end
+
+if category.linked_category.present?
```
sojan-official, line 21: we will be taking the change for associated category naming as a separate PR?

sojan-official (commented):
Looks good to me. 
Maybe we can also take up the `root` and `associated` naming in this PR.

app/views/public/api/v1/models/_category.json.jbuilder
tejaswinichile: Yes. There is one more article public APIs PR is open, I am adding the changes in that.

app/views/public/api/v1/models/_category.json.jbuilder
tejaswinichile: No no, we have 3 different category associations
1. Parent - sub categories
2. linked_categories which we are going to change to associated categories.
3. related categories with many-to-many associations mostly work like tags.

tejaswinichile:
@sojan-official Can you have a look in this PR again?

PR #4966, opened 2022-07-02
Remove no-op condition in Api::V1::InboxesHelper

jordan-brough:
* `smtp` cannot be nil in this context
* `anything&.nil?` will never be truthy, even if `anything` is nil


sophiascto:
We're working on it

PR #4972, opened 2022-07-05
fix: last incoming message sort

tejaswinichile:
# Pull Request Template

## Description

I have faced the issue where I have no conversations with the last message as an incoming message, all were either activity messages or template messages, because of the previous change it sent an empty conversation list which is the wrong result. So to fix this

1.  I have removed the filer i.e `where clause` from the sort query.
2. Added sort based on message_type and message created_at on the conversation table. 

## Type of change


sojan-official:
Can we also a test case where " have no conversations with the last message as an incoming message, all were either activity messages or template messages "

tejaswinichile:
> Can we also a test case where " have no conversations with the last message as an incoming message, all were either activity messages or template messages "
 
Yes makes sense.

sophiascto (approved):

sojan-official (approved):

PR #4976, opened 2022-07-05
chore: Remove confirm password in the account signup step

pranavrajs:
To reduce the effort to signup, we would try to keep the fields to a minimum in the signup form. As a first step, I'm removing the confirm password option. It seems redundant. Even if the user forgets the password, they can reset it easily.

@iamsivin / @muhsin-k / @fayazara I did a quick test on the local machine, it seems to be working fine without issues.


iamsivin (approved):
Tested locally, LGTM 👍🏻

PR #4984, opened 2022-07-06
fix: Handling facebook client error for deleted story

tejaswinichile:
Fixing: https://sentry.io/share/issue/930fed19b99d4f45a95c13c1f2e60652/

This known exception that Facebook throws when we are fetching the deleted story data. So instead of logging this to sentry, we are logging it to rails logger as we can not handle or resolve this issue from our end.

```
type: OAuthException, code: 9000001, message: (#9000001) This Message has been deleted by the user or the business., x-fb-trace-id: AH8KdfvtSee [HTTP 400]
```

### DIFF
--- a/spec/builders/messages/instagram/message_builder_spec.rb
+++ b/spec/builders/messages/instagram/message_builder_spec.rb
@@ -5,12 +5,14 @@ describe  ::Messages::Instagram::MessageBuilder do
 
   before do
     stub_request(:post, /graph.facebook.com/)
+    stub_request(:get, 'https://imagekit.io/blog/content/images/2020/05/media_library.jpeg')
   end
 
   let!(:account) { create(:account) }
   let!(:instagram_channel) { create(:channel_instagram_fb_page, account: account, instagram_id: 'chatwoot-app-user-id-1') }
   let!(:instagram_inbox) { create(:inbox, channel: instagram_channel, account: account, greeting_enabled: false) }
   let!(:dm_params) { build(:instagram_message_create_event).with_indifferent_access }
+  let!(:story_mention_params) { build(:instagram_story_mention_event).with_indifferent_access }
   let(:fb_object) { double }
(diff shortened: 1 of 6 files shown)


spec/builders/messages/instagram/message_builder_spec.rb
```diff
@@ -5,12 +5,14 @@
 
   before do
     stub_request(:post, /graph.facebook.com/)
+    stub_request(:get, 'https://imagekit.io/blog/content/images/2020/05/media_library.jpeg')
```
sojan-official, line 8: can we replace this url with `https://example.com/.....` ?

sojan-official (approved):
Minor comment. otherwise LGTM

spec/builders/messages/instagram/message_builder_spec.rb
tejaswinichile: We have used it for story mention attachments to test actual attachments through story mention in [instagram_event_job_spec.rb](https://github.com/chatwoot/chatwoot/blob/4375a7646ee84a20cfd055e722b3aca68e5b34b0/spec/jobs/webhooks/instagram_events_job_spec.rb)

sojan-official: It would be better to leave out third-party references in the code which doesn't have an affiliation with us. 

let's replace the URLs used in 

https://github.com/chatwoot/chatwoot/blob/4375a7646ee84a20cfd055e722b3aca68e5b34b0/spec/factories/instagram/instagram_message_create_event.rb#L107

https://github.com/chatwoot/chatwoot/blob/4375a7646ee84a20cfd055e722b3aca68e5b34b0/spec/factories/instagram/instagram_message_create_event.rb#L141

since  webmock would prevent our specs from accessing outside network, we can stick  with something like `example.com` or `local.host` etc.

sojan-official (approved):

PR #4989, opened 2022-07-07
chore: Ability to Disable Sentry Transactions

sojan-official:
fixes: #4866


config/initializers/sentry.rb
```diff
@@ -5,7 +5,7 @@
 
     # To activate performance monitoring, set one of these options.
     # We recommend adjusting the value in production:
-    config.traces_sample_rate = 0.1
```
pranavrajs, line 8: Can we do `ENABLE_SENTRY_TRANSACTIONS`, by default turned off?

Issue #4775, opened 2022-06-01
Contact model silently discards invalid attributes as of v2.5.0


jordan-brough:
In #4519 the Contact model was updated to silently ignore invalid phone numbers and email addresses, in order to address #4233.

I think this is surprising behavior at the model level, and unusual for a Rails app.  If I call `contact.update!(phone: 'invalid')` I expect to get an error if my change is invalid, rather than having my command to update be silently ignored.  i.e. when I tell the app to update a value, I don't want to have to always check every attribute afterward to confirm that it was actually updated, and wonder why it was ignored.

Instead of the change in #4519, could we instead possibly update just the API (controller) layer somehow, so that just the widget api has this behavior? (And even in that case it seems like it could be nice to return some extra data to indicate that some attributes were ignored)

pranavrajs:
@jordan-brough The concern seems valid, @tejaswinichile / @sojan-official can you comment on this?

PR #4994, opened 2022-07-07
chore: Fix contact model silently discarding invalid attributes

sojan-official:
fixes: #4775


tejaswinichile (approved):
LGTM 💯

app/controllers/api/v1/widget/contacts_controller.rb
```diff
@@ -36,7 +36,8 @@ def destroy_custom_attributes
   def identify_contact(contact)
     contact_identify_action = ContactIdentifyAction.new(
       contact: contact,
-      params: permitted_params.to_h.deep_symbolize_keys
+      params: permitted_params.to_h.deep_symbolize_keys,
+      discard_invalid_attrs: true
```
tejaswinichile, line 40: Will it be always true? when are we setting it false in which case?

sojan-official: the default value is false in the ContactIdentify action. 
Only in the widget controllers we want this behaviour

Issue #4999, opened 2022-07-08
[APM and Tracing] Add support for Elastic APM


kylemclaren:
Add support for [Elastic APM](https://www.elastic.co/observability/application-performance-monitoring)

The Ruby gem is well-documented [here](https://www.elastic.co/guide/en/apm/agent/ruby/current/index.html).

PR #5004, opened 2022-07-08
chore: Support for Elastic APM

kylemclaren:
## Description

This PR adds support for Elastic APM for [APM and Tracing](https://www.chatwoot.com/docs/self-hosted/monitoring/apm-and-error-monitoring) in self-hosted installations.

Configuration via the `ELASTIC_APM_SERVER_URL` and `ELASTIC_APM_SECRET_TOKEN` env variables. Docs and further configuration [here](https://www.elastic.co/guide/en/apm/agent/ruby/current/getting-started-rails.html).

Fixes:
- https://github.com/chatwoot/chatwoot/issues/4999

## Type of change

## How Has This Been Tested?

Tested locally using docker-compose on the `development` branch. Remote APM installation running on Elastic Cloud. Confirmed that metrics and logs are being ingested into Elasticsearch and "Chatwoot" is added as a service in APM. 

Screenshots below:

Screenshot 2022-07-08 at 13 10 53
Screenshot 2022-07-08 at 13 11 09
Screenshot 2022-07-08 at 13 11 35
Screenshot 2022-07-08 at 13 12 03

## Checklist:


sojan-official:
@kylemclaren  seems like the gem tries to connect to the APM server even when the environment variable isn't configured. We need to see if we can prevent this behavior. The agent should start tracking only if the specific environment variables for elastic are present 

Screenshot 2022-07-12 at 11 55 47 AM

kylemclaren:
@sojan-official looks like we can set `ELASTIC_APM_ENABLED` to `false` in the env. It defaults to true which is why the agent is starting up and trying to send data to localhost

sojan-official:
@kylemclaren We need to look into a way to prevent that startup by default before we can merge this. A significant % of chatwoot installations will use other APMs or no APMs.

so the default configuration should be that the agent is not started unless the environment variable is present.

kylemclaren:
@sojan-official right, that makes sense.

According to the [docs](https://www.elastic.co/guide/en/apm/agent/ruby/current/configuration.html#configuration), using a `config/elastic_apm.yml` is preferable. Thereafter, environment variables take precedence.

So, we can set `enabled: false` in `elastic_apm.yml`. This will prevent the agent from starting by default. Then, this can be overridden by setting `ELASTIC_APM_ENABLED=true` in the env.

Would that satisfy the requirement? Testing locally this works.

sojan-official:
@kylemclaren sounds good. We could also look into whether the yml supports erb, so that we can do something similar to this
https://github.com/chatwoot/chatwoot/blob/d7be0ea08ea8149306f76253d403427483874ed8/config/newrelic.yml#L22

```
enabled: <%= ENV.fetch('ELASTIC_APM_SECRET_TOKEN',  false).present? %>
```

So they don't have to configure the extra environment variable

PR #5021, opened 2022-07-12
Fix: Add missing referrer-link key for i18n

fayazara:
## Description
Adds missing i18n key for referrer link attribute in filters.

## Type of change


iamsivin (approved):
LGTM

PR #5027, opened 2022-07-12
chore: Ability to Disable Gravatars

sojan-official:
fixes: #3853 

## Changelog 
- Introduced `DISABLE_GRAVATAR` Global Config, which will stop chatwoot from making API requests to gravatar
- Cleaned up avatar-related logic and centralized it into the avatarable concern
- Added specs for the missing cases 
- Added migration for existing installations to move the avatar to attachment, rather than making the API that results in 404.


tejaswinichile (dismissed):
LGTM 💯

tejaswinichile:
@sojan-official can you fix the conflicts.

PR #5047, opened 2022-07-14
feat: Macros CRUD api

### LINKED ISSUE #5012: Add an API to create macros

tejaswinichile:
# Pull Request Template

## Description

Macro CRUD operations.

Fixes #5012 

## Type of change

## How Has This Been Tested?

Local Postman and Test cases.

## Checklist:

### DIFF
--- a/app/controllers/api/v1/accounts/macros_controller.rb
+++ b/app/controllers/api/v1/accounts/macros_controller.rb
@@ -1,29 +1,29 @@
 class Api::V1::Accounts::MacrosController < Api::V1::Accounts::BaseController
-	before_action :check_authorization
-	before_action :fetch_macro, only: [:show, :update, :destroy]
+  before_action :check_authorization
+  before_action :fetch_macro, only: [:show, :update, :destroy]
 
-	def index
-		@macros = Macro.with_visibility(params)
-	end
+  def index
+    @macros = Macro.with_visibility(params)
+  end
 
-	def create
-		@macro = Current.account.macros.new(macros_permit)
-		@macro.set_visibility(current_user, macros_permit)
-		@macro.actions = params[:actions]
+  def create
+    @macro = Current.account.macros.new(macros_with_user)
+    @macro.set_visibility(current_user, macros_permit)
+    @macro.actions = params[:actions]
 
-		render json: { error: @macro.errors.messages }, status: :unprocessable_entity and return unless @macro.valid?
+    render json: { error: @macro.errors.messages }, status: :unprocessable_entity and return unless @macro.valid?
 
-		@macro.save!
-	end
+    @macro.save!
+  end
 
-	def show; end
+  def show; end
 
-	def destroy
-		@macro.destroy!
+  def destroy
+    @macro.destroy!
     head :ok
-	end
+  end
 
-	def update
+  def update
     ActiveRecord::Base.transaction do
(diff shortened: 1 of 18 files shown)


app/controllers/api/v1/accounts/macros_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+      @macro.update!(macros_permit)
+      @macro.set_visibility(current_user, macros_permit)
+      @macro.save!
+    rescue StandardError => e
+      Rails.logger.error e
+      render json: { error: @macro.errors.messages }.to_json, status: :unprocessable_entity
+    end
+  end
+
+  def macros_permit
```
sojan-official, line 37: lets do `permited_params` as the parameter name?
so that it stays consistent with all the newer controllers we have

app/models/custom_attribute_definition.rb
```diff
@@ -16,7 +16,7 @@
 #
 # Indexes
 #
-#  attribute_key_model_index                         (attribute_key,attribute_model,account_id) UNIQUE
```
sojan-official, line 19: Intended?

config/routes.rb
```diff
@@ -57,6 +57,7 @@
             post :clone
             post :attach_file, on: :collection
           end
+          resources :macros
```
sojan-official, line 60: let's also explicitly define the methods we use in controller using `only:` so that we don't accidentally leave any endpoints open

app/policies/macro_policy.rb
```diff
@@ -0,0 +1,21 @@
...
+    @account_user.administrator? || @account.users.include?(@user)
+  end
+
+  def update?
+    @account_user.administrator? || @account.users.include?(@user)
+  end
+
+  def destroy?
+    @account_user.administrator? || @account.users.include?(@user)
+  end
```
sojan-official, line 20: We can just do true here if all users can use macros. 
since we ensure that the user has access to the account at our auth layer. we don't need to make another db call.
ref: contact_policy

app/models/macro.rb
```diff
@@ -0,0 +1,49 @@
...
+             class_name: :User
+  enum visibility: { user: 0, account: 1 }
+
+  def set_visibility(user, params)
+    self.visibility = params[:visibility]
+    self.visibility = :user if user.agent?
+  end
+
+  def self.with_visibility(user, params)
+    records = user.administrator? ? Current.account.macros : Current.account.macros.account
```
sojan-official, line 40: let's rename the enum `visibility` options to something like `personal` and `global` . open to better names.. 
because `Current.account.macros.account` could be very confusing.

spec/controllers/api/v1/accounts/macros_controller_spec.rb
```diff
@@ -0,0 +1,133 @@
+require 'rails_helper'
+
+RSpec.describe 'Api::V1::Accounts::MacrosController', type: :request do
+  let(:account) { create(:account) }
```
sojan-official, line 4: lets add the spec for an agent here and make sure only the visible macros are returned.
or that the call is made to self.with_visibility in macros model

sojan-official (commented):
Looks good. added some minor comments

sojan-official (approved):

PR #5059, opened 2022-07-19
feat: Allow SaaS users to manage subscription within the dashboard

pranavrajs:
Currently, the cancel/upgrade is done manually. It creates a lot of support requests and drops. This PR would make it easy for users to cancel the subscription without reaching out to support and upgrade the subscription if needed.

- [ ] Add Enterprise API Client
- [ ] Move Stripe logic to Enterprise code
- [ ] Update the UI to have manage subscription which would redirect to customer portal


pranavrajs:
@sojan-official @tejaswinichile I need some help on the webhook specs. Otherwise, this PR is good to be reviewed. Also, let us sync up on the flow.

app/controllers/application_controller.rb
```diff
@@ -17,10 +17,6 @@ def set_current_user
     Current.user = @user
   end
 
-  def current_subscription
```
pranavrajs, line 20: Unused.

sojan-official (approved):

PR #5090, opened 2022-07-22
fix: Update browser_language logic to include all languages

pranavrajs:
- Add missing attribute in additional_attributes
- Display browser_language in the sidebar.
- Update the language list to use everything.

### DIFF
--- a/app/javascript/dashboard/components/widgets/conversation/advancedFilterItems/languages.js
+++ b/app/javascript/dashboard/components/widgets/conversation/advancedFilterItems/languages.js
@@ -1,131 +1,735 @@
 const languages = [
   {
-    id: 'eng',
-    name: 'English (en)',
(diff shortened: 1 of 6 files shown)


app/javascript/dashboard/components/widgets/conversation/advancedFilterItems/languages.js
```diff
@@ -1,131 +1,735 @@
...
+    name: 'Yoruba',
+    id: 'yo',
+  },
+  {
+    name: 'Zhuang, Chuang',
+    id: 'za',
+  },
+  {
+    name: 'Zulu',
+    id: 'zu',
```
pranavrajs, line 732: Display all the languages in ISO 639 list.

tejaswinichile (dismissed):
LGTM 💯

Issue #5107, opened 2022-07-26
Scroll resets to bottom when a new conversation is created


pranavrajs:
Scroll in the message view resets to the bottom when a new conversation is created even though there is no new message in the currently active conversation. This is annoying behavior if you get a lot of conversations in the application. The scroll would reset every 10-15s.

**Expected**: Scroll to the bottom would only occur if a new message is sent from the reply box or a new message is received from the customer.

**To reproduce:**

- Open a conversation with a lot of messages, scroll to the first message
- Create a new conversation
- Scroll would reset to the bottom

PR #5108, opened 2022-07-26
fix: Do not reset the scroll if there is no new message

pranavrajs:
Fixes https://github.com/chatwoot/chatwoot/issues/5107


iamsivin (approved):
Tested locally LGTM ⚡️

https://www.loom.com/share/db0d47fdd720461eb274d249571084cb

PR #5125, opened 2022-07-27
Fix: add articles metadata

tejaswinichile:
fixes: https://github.com/chatwoot/product/issues/497


app/views/api/v1/accounts/portals/_portal.json.jbuilder
```diff
@@ -17,3 +17,9 @@ json.portal_members do
     end
   end
 end
+
+json.meta do
+  json.articles_count portal.articles.published.size
```
muhsin-k, line 22: Shouldn't we return all the articles?

muhsin-k: Screenshot 2022-07-27 at 3 00 01 PM

tejaswinichile: It was mentioned in one of the tickets about the published articles count.

muhsin-k:
I have found 2 issues while testing the API. Please check this video

https://www.loom.com/share/030081b0e7f34dc08ee9c095140cc3a5

tejaswinichile:
@muhsin-k I have changed the filter params in the `payload` key, but removing that since we are not doing that in any other filters.

muhsin-k (approved):

Issue #4990, opened 2022-07-07
feat: Enable Capacity Config UI


sojan-official:
@pranavrajs  Accommodating the last two conditions with retaining support for existing behavior will require significant changes for the round-robin services. So let's merge this PR with the UI for capacity config disabled.  

Will open up a follow-up PR with the config to control assignment behavior for offline agents and let's introduce the capacity feature over that.

_Originally posted by @sojan-official in https://github.com/chatwoot/chatwoot/issues/4672#issuecomment-1152407225_

PR #5164, opened 2022-08-01
feat: Enable Capacity Config UI

sojan-official:
- Enables Capacity Config in UI
- Rewrite auto assignment Logic to consider only online agents

fixes: https://github.com/chatwoot/chatwoot/issues/4990


nithindavid (commented):
FE part looks good 🌝

PR #5179, opened 2022-08-02
chore: Add index on conversations.uuid

jordan-brough:
We search for conversations by this attribute (e.g. in ReplyMailbox and CsatSurveyController) so it seems like we should have an index on it.


app/models/conversation.rb
```diff
@@ -57,6 +58,7 @@ class Conversation < ApplicationRecord
   before_validation :validate_additional_attributes
   validates :additional_attributes, jsonb_attributes_length: true
   validates :custom_attributes, jsonb_attributes_length: true
+  validates :uuid, uniqueness: true
```
jordan-brough, line 61: Is this intended to be required to be unique?  I opened this PR in part to understand that.

jordan-brough: Or, should it be unique within the context of a single account?

sojan-official: @jordan-brough  this looks good to me. The idea is for the UUID to be unique across the installation since we do this look-up. 

https://github.com/chatwoot/chatwoot/blob/f7d4f39b5c8efb6e1f423b867c8c2845aabc73eb/app/mailboxes/reply_mailbox.rb#L47

sojan-official (dismissed):
Thanks for the changes :)

pranavrajs:
@sojan-official As discussed, let us run the migration separately and push this change.

sojan-official:
- performed this migration in chatwoot prod mannualy

PR #5180, opened 2022-08-02
Update .env.example

nikhilmuz:
Added default database name config in example env file

## Type of change

## How Has This Been Tested?

Fixed non-breaking documentation file so it won't affect the working of the codebase.

## Checklist:

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -38,7 +38,9 @@ REDIS_SENTINEL_MASTER_NAME=
 # REDIS_OPENSSL_VERIFY_MODE=none
 
 # Postgres Database config variables
-POSTGRES_DATABASE=chatwoot_production
+# You can leave POSTGRES_DATABASE blank. The default name of 
+# the database in the production environment is chatwoot_production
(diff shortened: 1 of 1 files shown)


pranavrajs:
Screenshot 2022-08-03 at 2 19 13 PM

@nikhilmuz Any context on why this was added would be helpful. There is already a default in the database.yml file.

nikhilmuz:
It would be helpful incase someone doesn’t walk through the codebase and
deploy multiple different instances of Chatwoot on a DB using production
docker image just having env file. One can just change sample env with
different DB name and extend production docker image

nikhilmuz:
Yes, the change works. I had a hard time changing the DB parameters for the name yesterday and had to dig deeper into ruby code to figure out the same so at least having params in sample .env makes sense. And for my case, the DB name which I got was as committed but now I realised that's different on different builds. Anyway, I have updated the same.

pranavrajs:
@nikhilmuz Sure, makes sense. Thanks for the contribution. ❤️

PR #5206, opened 2022-08-04
Fix: "presence" checks in Channel::TwilioSms

jordan-brough:
This reverts a change from https://github.com/chatwoot/chatwoot/pull/5055.

The other fixes in #5055 look good, but I think this part doesn't do what was intended.

`:phone_number.presence` is evaluated at class-definition time (not at validation time) and it evaluates to `:phone_number`, so:
```ruby
unless: :phone_number.presence
```
is the same as:
```ruby
unless: :phone_number
```

Which counts empty strings as truthy (which we don't want).

On the other hand, `:phone_number?` is a Rails-provided method that is the same as `phone_number.present?`.
So
```ruby
unless: :phone_number?
```
is the same as
```ruby
unless: -> { phone_number.present? }
```
which is what we want.


pranavrajs:
@tejaswinichile Can you please take a look at this change?

tejaswinichile:
The code changes pretty straightforward; I tested this locally with an empty `phone_number` and `messaging_service_id`.
I will merge this after the circle-ci is passed.

tejaswinichile (approved):
Works well 💯 

Note: The change is not necessary but good to have with this new standards.

PR #5211, opened 2022-08-05
Find by downcased email in SupportMailbox

jordan-brough:
## Description

I noticed duplicate `ContactInbox`s being created and I believe this may be the reason.

If this looks correct, I noticed some other areas that might need the same sort of change?

1. `ContactIdentifyAction`
https://github.com/chatwoot/chatwoot/blob/628f40c2abe97d31d9d70ffcb53ae183ce78f0b1/app/actions/contact_identify_action.rb#L62
2. `DataImportJob`
https://github.com/chatwoot/chatwoot/blob/628f40c2abe97d31d9d70ffcb53ae183ce78f0b1/app/jobs/data_import_job.rb#L29
3. `Imap::ImapMailbox`
https://github.com/chatwoot/chatwoot/blob/628f40c2abe97d31d9d70ffcb53ae183ce78f0b1/app/mailboxes/imap/imap_mailbox.rb#L68

## Type of change

## How Has This Been Tested?

In the updated specs


tejaswinichile:
@sojan-official This fix seems helpful.

But I haven't seen any sentry issues related to duplicate ContactInbox for Email inboxes.

tejaswinichile (approved):
LGTM 💯

pranavrajs:
@tejaswinichile Please make sure that we have updated all the touch points to use downcased instead of the actual email. If you could list those places where we need to update this, that would be helpful.

tejaswinichile:
> If you could list those places where we need to update this, that would be helpful.

@pranavrajs I have created one more issue, #5601, to track this to fix the case-sensitive matching for contact_inboxes. This PR ensures that the support mailbox is not creating duplicate contact_inbox records.

I was thinking of merging this PR before and then working on the created issue; let me know your thoughts.

PR #5220, opened 2022-08-08
fix: Update only the relevant agent only

pranavrajs:
Fixes https://github.com/chatwoot/satisfilabs/issues/10

The issue was that the `updatePresence` action updated all agents' availability. Only one agent information is passed down to the action from the `updateAvailability` method, so the rest of the agents were updated as offline.

- Added a new method updateSingleAgentPresence. 
- Removed duplicate updates in the live-agent reports. Instead, use the agent store itself.


app/javascript/shared/helpers/vuex/mutationHelpers.js
```diff
@@ -45,6 +45,16 @@ export const updatePresence = (state, data) => {
   });
 };
 
+export const updateSingleAgentPresence = (
```
nithindavid, line 48: Since this is resource specific we could move it helpers under agents module?

pranavrajs: @nithindavid `updatePresence` is already part of MutationHelper as it is used in multiple places. It makes sense to have it here rather than moving it to the agents module. A generic name, if needed, can be used.

pranavrajs: Renamed it to updateSingleRecordPresence

fayazara (approved):
Tested, working correctly.

PR #5222, opened 2022-08-08
chore: Lock closed threads automatically

sojan-official:
We often have cases where users would comment over stale closed Github Issues.
This creates unnecessary noise for the original reporter and makes it harder for triaging.
This action locks the closed threads once it is inactive for over a month.

fixes: https://github.com/chatwoot/product/issues/421

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -1,7 +1,7 @@
 # We often have the case where users would comment over stale closed Github Issues
-# This creates unecessary noise to the original reporter as well as makes it harder for triaging.
-# This actions locks the closed threds once it is inactive for over a month
-# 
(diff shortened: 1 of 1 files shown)


pranavrajs (approved):

nh2:
@sojan-official @pranavrajs I can only recommend to not do this -- communication blocker bots are a huge deterrant for open-source projects communities and often hurt productivity more than improve.

Often people comment useful things in closed pull requests, such as "This was later implemented in [other PR]". I am quite sure it is better to close the few problematic PRs manually where discussion goes overboard, rather than prevent useful post-close contributions in _all_ PRs.

Triage processes are also rarely helped when the system suggests/forces people to disperse the same topic across more and more separate issues.

PR #5240, opened 2022-08-10
Fix 'imap_enabled' and 'smtp_enabled' condition in inbox config page

0x7466:
## Description

Hi, I'm currently getting to know your awesome project. I have found a missing and a wrong condition in a settings page and I thought this could by my first contribution to this project.

Fixes # (issue)
* Missing condition for IMAP settings (`imap_enabled`).
* Wrong condition for SMTP settings component. `smtp_enabled` instead of `imap_enabled`.

## Type of change

## How Has This Been Tested?

Test suite is fine.

## Checklist:

### DIFF
--- a/app/javascript/dashboard/routes/dashboard/settings/inbox/settingsPage/ConfigurationPage.vue
+++ b/app/javascript/dashboard/routes/dashboard/settings/inbox/settingsPage/ConfigurationPage.vue
@@ -88,8 +88,8 @@
         <woot-code :script="inbox.forward_to_email" />
       </settings-section>
     </div>
-    <imap-settings :inbox="inbox" />
-    <smtp-settings v-if="inbox.imap_enabled" :inbox="inbox" />
+    <imap-settings v-if="inbox.imap_enabled" :inbox="inbox" />
+    <smtp-settings v-if="inbox.smtp_enabled" :inbox="inbox" />
   </div>
   <div v-else-if="isAWhatsappChannel && !isATwilioChannel">
     <div v-if="inbox.provider_config" class="settings--content">


pranavrajs:
@0x7466 Thanks for the contribution. I don't think this is by mistake. The SMTP is enabled only if there is a valid imap_setting. @aswindevps / @sojan-official can confirm.

0x7466:
Hi @pranavrajs, thanks for your comment. I think I don't get it. It doesn't make sense to me.

What if I only want to configure SMTP but not IMAP? Or the other way around? For example we use per account SMTP configuration, but forwarding in through ActionMailbox for incoming messages (instead of pulling messages from an email account with IMAP).

How would I be able to do that in the UI without setting it directly using the rails console?

I hope I don't make a fool of myself with that. :)

joejordanbrown:
@pranavrajs

We have the exact same issue, we use a custom Action Mailbox that we created for our incoming email relay. We do not want or need IMAP for the inboxes that use our relay.

We only want SMTP per inbox so we can send using the correct address.

This change should be accepted.

sojan-official:
@joejordanbrown @0x7466 The current product expectation in Chatwoot is that Email inboxes should be used in either of the following 2 ways

1) Email Forwarding  
 - incoming emails come through the forwarded to the email address 
 - outgoing emails are sent via the SMTP configuration of the installation

2) SMTP & IMAP 
- incoming emails through inbox's IMAP config
- outgoing emails through the inbox's SMTP config

So the code is currently written to satisfy this product requirement. The SMTP option is shown only when a valid IMAP configuration is established first. 
-----

Mixing cases 1 and 2 isn't intended in the initial product design.  So we will have to take it up as a research ticket and see the implications before enabling such behavior. We would also need to test this further as both the methods use different logic for appending the replies to a single thread.

1) Email Forwarding constructs independent email addresses for each thread ex: `{conversationid}@xyx.com`
2) SMTP & IMAP uses embed conversation-id in the email header

This would also require further documentation updates.  Hence we can take it up as a separate issue for further action items.
But the current logic is intended.

sojan-official:
Closing this PR since there aren't further actions to be done over this for now. Let's reopen this in future if required.

PR #5313, opened 2022-08-19
feat: Allow Dashboard App Frames to query the information 

pranavrajs:
Allow dashboard app frames to query information after the load event. Following [this tweet](https://twitter.com/ericmigi/status/1560321001405050880), the Tooljet team and I did a sync-up to figure out why the events are not working.

- Dashboard Apps expected the message event to be registered on the load event.
- The Tooljet Apps had delayed loading. Their main page would fire the load event even before the event listeners were registered.
- There needed a way to query the information after the page was loaded.

With this change, the child iframe pages can send an event `chatwoot-dashboard-app:fetch-info` using postMessage. Then Chatwoot would respond with the appContext payload.

### DIFF
--- a/app/javascript/dashboard/components/widgets/DashboardApp/Frame.vue
+++ b/app/javascript/dashboard/components/widgets/DashboardApp/Frame.vue
@@ -38,6 +38,18 @@ export default {
       return this.currentChat?.meta?.sender?.id;
     },
   },
+
+  mounted() {
+    window.onmessage = e => {
+      if (
+        typeof e.data !== 'string' ||
+        e.data === 'chatwoot-dashboard-app:fetch-info'
+      ) {
+        return;
+      }
+      this.onIframeLoad(0);
+    };
+  },
   methods: {
     onIframeLoad(index) {
       const frameElement = document.getElementById(
--- a/app/javascript/dashboard/components/widgets/DashboardApp/Frame.vue
+++ b/app/javascript/dashboard/components/widgets/DashboardApp/Frame.vue
@@ -43,7 +43,7 @@ export default {
     window.onmessage = e => {
       if (
         typeof e.data !== 'string' ||
-        e.data === 'chatwoot-dashboard-app:fetch-info'
+        e.data !== 'chatwoot-dashboard-app:fetch-info'
       ) {
         return;
       }


pranavrajs:
@fayazara Please test this out whenever you have time.

app/javascript/dashboard/components/widgets/DashboardApp/Frame.vue
```diff
@@ -38,6 +38,18 @@ export default {
       return this.currentChat?.meta?.sender?.id;
     },
   },
+
+  mounted() {
+    window.onmessage = e => {
+      if (
+        typeof e.data !== 'string' ||
+        e.data === 'chatwoot-dashboard-app:fetch-info'
```
fayazara, line 46: @pranavrajs I think this should be.
```suggestion
        e.data !== 'chatwoot-dashboard-app:fetch-info'
```

fayazara (changes requested):
Working perfectly, needs a small change on the message listener.

How it was tested
```html
<template>
  <div id="app" class="p-8 min-h-screen bg-slate-100">
    <div>Test App</div>
    <code>
      <pre class="bg-black text-white p-2">
        {{ eventData }}
      </pre>
    </code>
    <button
      @click="getData()"
      class="mt-8 bg-sky-500 text-white px-4 py-3 rounded-lg"
    >
      Get Data from Chatwoot
    </button>
  </div>
</template>

<script setup>
import { ref, onMounted } from "vue";

const eventData = ref({
  name: "Test",
});

function getData() {
  window.parent.postMessage("chatwoot-dashboard-app:fetch-info", "*");
}

onMounted(() => {
  window.addEventListener("message", function (event) {
    const chatwootEvent = event.data;
    if (JSON.parse(chatwootEvent)) {
      eventData.value = JSON.parse(chatwootEvent);
    }
  });
});
</script>
```

app/javascript/dashboard/components/widgets/DashboardApp/Frame.vue
pranavrajs: That makes sense. Thanks for testing this. While debugging with Tooljet, I used `includes`, later moved it to `===`.

PR #5317, opened 2022-08-22
Fix: Instagram webhook test event

tejaswinichile:
Instagram test webhook events have been updated, updating the webhook job accordingly.


app/services/instagram/message_text.rb
```diff
@@ -56,6 +57,10 @@ def contacts_first_message?(ig_scope_id)
     @inbox.contact_inboxes.where(source_id: ig_scope_id).empty? && @inbox.channel.instagram_id.present?
   end
 
+  def sent_via_test_webhook?
+    @messaging[:sender][:id] == '12334' && @messaging[:recipient][:id] == '23245'
```
sojan-official, line 61: @tejaswinichile, do we have documentation on Facebook about these test ids? They seem pretty random. What are the chances that it will be changed in future?

tejaswinichile: They have changed the whole test webhook JSON altogether on the recent release, that's the test webhook data wasn't getting populated. 
But for the current META version, it is constant.

Screenshot 2022-08-22 at 1 49 52 PM

sojan-official (approved):

PR #5357, opened 2022-08-30
chore: Default to rich content editor in website and email channel

pranavrajs:
As per the discussion https://github.com/chatwoot/chatwoot/discussions/5321, it doesn't look like we need an option to toggle. Everything that can be done in a simple editor could also be done in the rich content editor.


nithindavid (approved):

iamsivin (approved):
LGTM

PR #5358, opened 2022-08-31
Use "create!" and "save!" bang methods when not checking the result

jordan-brough:
I happened to notice this in one place and decided to grep the codebase for other instances.

Afaict, these are instances where we're not checking the result and we don't expect the create or save to fail, but we should raise an exception if it does fail.

I'm guessing they probably _don't_ ever actually fail in real usage, but best practice would be to either check the result or use the "bang" versions of the methods.


app/listeners/reporting_event_listener.rb
```diff
@@ -16,7 +16,7 @@ def conversation_resolved(event)
       event_start_time: conversation.created_at,
       event_end_time: conversation.updated_at
     )
-    reporting_event.save
+    reporting_event.save!
```
jordan-brough, line 19: On this one I wasn't sure whether the result of this method is checked somewhere.

sojan-official: This should be good.

PR #5359, opened 2022-08-31
chore: Moves portal slug as ID to query resources in vue store

nithindavid:
## Description
1. Our store was designed in a way that 'id' will be the key to query from store but our routes have only `portalSlug`. To solve this there was an additional code `selectedPortalId`. I have remove this and now this uses portal slug instead of id to query.
2. Cleans route files and renamed params `portal_slug`


muhsin-k:
@nithindavid Portal sidebar is not working 
Screenshot 2022-08-31 at 7 20 46 PM

muhsin-k (approved):

PR #5377, opened 2022-09-01
feat: Add the ability update/edit locale in portal 

muhsin-k:
Fixes https://github.com/chatwoot/product/issues/558 and https://github.com/chatwoot/product/issues/535

- Added the option set default locale in portal
- Added the option delete locale from the portal


nithindavid (approved):

iamsivin (approved):
LGTM

PR #5388, opened 2022-09-02
chore: Final touches for portals

nithindavid:
# Pull Request Template

## Description
image

Fixes # (issue)
1. Opening portal link
2. opening article link
3. selected locale style
4. button styles


app/javascript/dashboard/assets/scss/widgets/_buttons.scss
```diff
@@ -44,6 +44,52 @@ $default-button-height: 4.0rem;
...
+    }
+
+    &.alert {
+      border-color: var(--s-200);
+      color: var(--r-700)
+    }
+
+    &.warning {
+      border-color: var(--s-200);
+      color: var(--w-700)
```
iamsivin, line 68: @nithindavid Is this color right for warning?

nithindavid:
@iamsivin Fixed the comments!

muhsin-k (approved):

iamsivin:
@nithindavid When I hover the cancel button in popover looks like this, Is it okay?
Screenshot 2022-09-03 at 5 58 47 PM

app/javascript/dashboard/routes/dashboard/helpcenter/components/PortalSwitch.vue
```diff
@@ -111,12 +127,14 @@ export default {
       });
       this.$emit('open-portal-page');
     },
-    isLocaleActive(code) {
-      const isPortalActive = this.portalSlug === this.portal.slug;
+    isLocaleActive(code, slug) {
+      const isPortalActive = this.portal.slug === slug;
       const isLocaleActive = this.portal?.meta?.default_locale === code;
```
iamsivin, line 132:
```suggestion
      const isLocaleActive = this.$route.params.locale === code;
```

Instead of taking the current active locale from the store can we take it from route params? So I think the portal switch active state will work properly. Currently, there is some issue.

image

nithindavid: Yes, We will do this in a diffrerent PR. We need to do it like `selectedPortal` since there are URL's like `app.chatwoot.com/accounts/1/portals` where param is not available.

PR #5426, opened 2022-09-13
fix: Fixes bug with locale switching from popover

nithindavid:
## Description

The portal switch component was using the default locale, making the active locale to show a different locale.


iamsivin (dismissed):
LGTM

iamsivin (approved):

PR #5465, opened 2022-09-20
Bug: Adds support multiline text input on article edit page

nithindavid:
## Description
Adds multiline support for article title input


iamsivin (approved):
LGTM

PR #5500, opened 2022-09-26
fix: Add slug to articles

tejaswinichile:
Fixes #575

### DIFF
--- a/db/migrate/20220926164441_add_slug_to_article.rb
+++ b/db/migrate/20220926164441_add_slug_to_article.rb
@@ -1,0 +1,12 @@
+class AddSlugToArticle < ActiveRecord::Migration[6.1]
+  def change
+    add_column :articles, :slug, :string, index: true, default: ''
+
+    Article.all.each do |article|
+      slug = article.title.underscore.parameterize(separator: "_")
+      article.update!(slug: slug)
+    end
+
+    add_index :articles, :slug
+  end
+end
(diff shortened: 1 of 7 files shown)


db/migrate/20220926164441_add_slug_to_article.rb
```diff
@@ -0,0 +1,21 @@
...
+    change_column_null(:articles, :slug, false)
+  end
+
+  def down
+    remove_column(:articles, :slug)
+  end
+
+  def update_past_articles_with_slug
+    Article.all.each do |article|
+      slug = article.title.underscore.parameterize(separator: '_')
```
sojan-official, line 17: We might have to add a fallback condition here for the cases where someone created articles with the same title. 
We could catch the error and try updating the slug with numerical increments in the end.

db/migrate/20220926164441_add_slug_to_article.rb
```diff
@@ -0,0 +1,21 @@
+class AddSlugToArticle < ActiveRecord::Migration[6.1]
+  def up
+    add_column :articles, :slug, :string, default: ''
```
sojan-official, line 3: is the default: '' required ?

db/migrate/20220926164441_add_slug_to_article.rb
pranavrajs: The separator should be `-` not `_`

db/migrate/20220926164441_add_slug_to_article.rb
tejaswinichile: When null is `false` we need to have some default value. This usually helps with the previous records which are missed with the slug.

sojan-official (approved):

PR #5507, opened 2022-09-27
fix: Contact Inbox uniq index violation

tejaswinichile:
Fixes https://github.com/chatwoot/chatwoot/issues/4680


app/builders/messages/facebook/message_builder.rb
```diff
@@ -43,7 +43,8 @@ def build_contact
     return if contact.present?
 
     @contact = Contact.create!(contact_params.except(:remote_avatar_url))
-    @contact_inbox = ContactInbox.find_or_create_by!(contact: contact, inbox: @inbox, source_id: @sender_id)
+    @contact_inbox = ContactInbox.find_or_create_by!(inbox: @inbox, source_id: @sender_id)
```
sojan-official, line 46: @tejaswinichile, could you add more details on this change ? Do we know how we end up with contact inboxes with the same source_id but a different contact associated with it?

tejaswinichile: @sojan-official I am not so sure about this, because we enter this block when contact is not present.
But I was memorizing contact [here](https://github.com/chatwoot/chatwoot/pull/5507/files#diff-0d098b3b099261200b5b9c56cb08c4eaebfea4413d8200934c94cd286f7e4bf9R39) so I think that might be the case and also didn't add the attr_accessor for the same method.

I think some memory saved from the different contact might be affecting the same block. But with this fix, we are trying to avoid the situation.

tejaswinichile:
@sojan-official Can you help me to fix this, these test cases failing randomly.

app/builders/messages/facebook/message_builder.rb
sojan-official: @tejaswinichile let's try to clearly understand what's happening and how to replicate this case.

tejaswinichile: @sojan-official One of my guesses is what if an Instagram user or FB user changes their name later and then tries to send the message? I will test this out in our staging and add the investigation here.

tejaswinichile:
I tried to reproduce this issue but couldn't, and I was checking this https://sentry.io/share/issue/f87566ca7c594cc2a4475f2e5772c479/ issue on prod, and it's not raising the exception for the same record, but the messages shown in this issue is not available in the system, that means it failed at this exception.

One of the key observations is that it's happening for echo calls. I think it might be a memoising issue, so I am removing the extra changes I have added.

Screenshot 2022-10-06 at 5 47 00 PM
Screenshot 2022-10-06 at 5 47 23 PM

tejaswinichile:
ToDo: Deploy these changes in productions[without merging] and check sentry for a day.

cc @sojan-official

sojan-official:
@tejaswinichile  the issue seems to be this : https://github.com/rails/rails/issues/36027 
so rather than doing find_or_create by .. lets do find and create as separate steps

sojan-official:
closing in favor for https://github.com/chatwoot/chatwoot/pull/5617

PR #5509, opened 2022-09-27
fix: contact is not available for inaccessible sender

tejaswinichile:
Fixes: #5508 

We can not read contact information, because of this error, as the messages are echo messages when the sender sends the message to contacts and we don't have the consent of the user until and unless they send messages to us.

So after this result to get information about the contact is empty and we are trying to create a contact inbox for the same and the error appears.

type: OAuthException, code: 230, message: (#230) User consent is required to access user profile, x-fb-trace-id: AaitxF/whwY [HTTP 403] (Koala::Facebook::ClientError)

### DIFF
--- a/app/services/instagram/message_text.rb
+++ b/app/services/instagram/message_text.rb
@@ -42,7 +42,7 @@ class Instagram::MessageText < Instagram::WebhooksBaseService
       ChatwootExceptionTracker.new(e, account: @inbox.account).capture_exception
     end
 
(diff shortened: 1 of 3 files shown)


sojan-official (commented):
Can we also add a spec for the same?

tejaswinichile:
@sojan-official Added specs, that we are not creating contact or messages when FacebookClientError occurs.

sojan-official (approved):

PR #5520, opened 2022-09-28
fix: Updated IMAP errors add method

tejaswinichile:
fixes: #5519 

```
ArgumentError: wrong number of arguments (given 0, expected 1..2)
  app/models/message.rb:244:in `validate_attachments_limit'
    errors.add(attachments: 'exceeded maximum allowed') if attachments.size >= NUMBER_OF_PERMITTED_ATTACHMENTS
```


sojan-official:
@tejaswinichile  can we also add a spec that triggers this error condition to ensure no further breakages in the future?

sojan-official (approved):

PR #5524, opened 2022-09-29
fix: Article slug auto saves

tejaswinichile:
Article slug textbox is not available at frontend till then we are adding the slug from backend.

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -116,6 +116,6 @@ class Article < ApplicationRecord
   end
 
   def ensure_article_slug
-    self.slug = self.title.underscore.parameterize(separator: '-')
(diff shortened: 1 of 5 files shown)


app/models/article.rb
```diff
@@ -112,4 +114,8 @@ def draft!
   def ensure_account_id
     self.account_id = portal&.account_id
   end
+
+  def ensure_article_slug
+    self.slug = title.underscore.parameterize(separator: '-')
```
muhsin-k, line 119:
```suggestion
    self.slug = title.underscore.parameterize(separator: '-') if title.present?
```

If the title is empty, it will throw this error instead of the title validation message.

Screenshot 2022-09-29 at 1 45 26 PM

PR #5563, opened 2022-10-06
feat: Quickly create canned responses

pranavrajs:
Most of the time when we reply, the response could be saved as a canned response. Right now, we have to copy the content, go to settings, and add it manually. With this change, the agent can easily add a canned response by clicking on the message context menu.

Screenshot 2022-10-05 at 7 15 18 PM

Here is a quick video on how it works. https://www.loom.com/share/c2ea60637915495aac19ce3829ace05b


iamsivin (approved):
Tested locally LGTM

Issue #5561, opened 2022-10-05
feat: Ability to manage account features from API


voko3456:
Now we can manage the acount features via super admin ui it will be greate to manage it via api while creating the account

![Screenshot_20221005-175936_Chrome](https://user-images.githubusercontent.com/20530550/194093929-59fff017-7052-4428-8e3e-a215a37f2fda.jpg)

sojan-official:
duplicate https://github.com/chatwoot/chatwoot/issues/5529

PR #5589, opened 2022-10-11
chore: feature flags and agents limit parameters in account creation through platform API

smartdev58:
## Description

This pull request is about adding the ability to add feature flags and agents limit to Account creation platform API endpoint according to the issue #5561 and #5159. As I was working on API endpoints, I also needed to create the views for accounts API for the test spec and proper response data format.

Fixes # (5561, 5159)

## Type of change

## How Has This Been Tested?

https://www.loom.com/share/8c9cc3adfc224e57bed759df7076a5bf

I tested this updates via postman to my local Ubuntu self-hosted server.
I have cloned Chatwoot Platform API to my workspace in postman and added some minor changed and performed a test.
Here is the endpoint and post data:

localhost:3000/platform/v1/accounts

```json
{
    "name": "New Account",
    "enabled_features": [
        "feature_inbound_emails",
        "feature_channel_email",
        "feature_channel_facebook",
        "feature_channel_twitter",
        "feature_ip_lookup",
        "feature_help_center"
    ],
    "limits": {
        "agents": 5,
        "inboxes": 10
    }
}
```

The response data likes like the following:

```json
{
    "id": 35,
    "name": "New Account",
    "locale": "en",
    "domain": null,
    "support_email": "Chatwoot <accounts@chatwoot.com>",
    "enabled_features": {
        "inbound_emails": true,
        "channel_email": true,
        "channel_facebook": true,
        "channel_twitter": true,
        "ip_lookup": true,
        "help_center": true
    },
    "custom_attributes": {},
    "limits": {
        "agents": 5,
        "inboxes": 10
    },
    "status": "active"
}
```


app/controllers/platform/api/v1/accounts_controller.rb
```diff
@@ -3,16 +3,12 @@ def create
...
     @platform_app.platform_app_permissibles.find_or_create_by(permissible: @resource)
-    render json: @resource
   end
 
-  def show
-    render json: @resource
-  end
+  def show; end
 
   def update
```
sojan-official, line 10: @smartdev58  could you also add a spec for the update endpoint since this change can impact account update  API calls? We want to ensure that existing features aren't disabled if someone makes a patch request to the account endpoint without specifying the enabled feature flags.

sojan-official (commented):
Thanks for the changes. It looks good to me. we can add specs for update call and it should be good to go

app/controllers/platform/api/v1/accounts_controller.rb
smartdev58: @sojan-official Thank you for your review. I will add the spec for update endpoint.

smartdev58: @sojan-official I have added the required spec. Please review

PR #5647, opened 2022-10-17
Help center count fix

tejaswinichile:
- I have updated the help-center sidebar metadata with the local filter.
- Now we are getting metadata from the selected portals with local.
- I have added a new endpoint for the same, and we are fetching the respective data.

Issue: It's not getting updated with the portal/local change until I refresh the page.


tejaswinichile:
https://www.loom.com/share/4d10f9d1df914170ad0c49a281b02205

@iamsivin / @fayazara, Could you help me to point out where I should change the page article count mentioned in the video?

spec/models/article_spec.rb
```diff
@@ -138,7 +138,7 @@
         end
         params = { category_slug: 'category_2' }
         records = portal_2.articles.search(params)
-        expect(records.count).to eq(25)
+        expect(records.count).to eq(31)
```
sojan-official, line 141: Having a random number here would be pretty confusing and error prone in the specs. I.e. how the number `31` is derived or why did we change the number 25.    

Either let's create a new category within this specific case, build a list of articles associated with that category and ensure that the specific count is resolved within this block. or at least lets add comments on how the number 31 is derived

tejaswinichile: Let me add added articles_count here.

spec/models/article_spec.rb
```diff
@@ -138,7 +140,7 @@
         end
         params = { category_slug: 'category_2' }
         records = portal_2.articles.search(params)
-        expect(records.count).to eq(25)
+        expect(records.count).to eq(articles.size)
```
pranavrajs, line 143: @tejaswinichile This spec is useless, we have removed the pagination logic from the article model.

PR #5658, opened 2022-10-17
Fix: Fixes overflow issue for category name in article list

nithindavid:
**Why**
We recently made the category name in the article table clickable. This introduced an overflow issue for long category names.


app/javascript/dashboard/routes/dashboard/helpcenter/components/ArticleItem.vue
```diff
@@ -155,4 +160,9 @@ td {
     }
   }
 }
+
+.category-link-content {
+  max-width: 160px;
```
pranavrajs, line 165:
```suggestion
  max-width: 16rem;
```

PR #5669, opened 2022-10-18
Update

Justman10000:


.env.example
```diff
@@ -23,7 +23,7 @@ FORCE_SSL=false
 ENABLE_ACCOUNT_SIGNUP=false
 
 # Redis config
-REDIS_URL=redis://redis:6379
+REDIS_URL=redis://localhost:6379
```
sojan-official, line 26: We can't change this value because it is written to support the Chatwoot docker installations.

Justman10000: What should be different in the Docker image? Redis listing on localhost!

Issue #5717, opened 2022-10-23
platform API accounts creation endpoint not working with features false flags


smartdev58:
POST request for platform API accounts to create an account doesn't work.

POST and PATCH endpoints need to work for following data format.
```json
{
    "name": "New Account",
    "locale": "en",
    "features": {
        "inbound_emails": true,
        "channel_email": true,
        "channel_facebook": true,
        "channel_twitter": true,
        "ip_lookup": false,
        "macros": true,
        "agent_management": true,
        "team_management": true,
        "inbox_management": true,
        "labels": false,
        "custom_attributes": true,
        "automations": true,
        "canned_responses": true,
        "integrations": true,
        "voice_recorder": true
    },
    "limits": {
        "agents": 5,
        "inboxes": 10
    },
    "status": "active"
}

PR #5718, opened 2022-10-23
chore: Update data format for platform API feature management

smartdev58:
## Description

Updated json data format for Platform API for account creation update API endpoints features flags to accept false values on each feature. #5717 

Fixes # (issue)

## Type of change

## How Has This Been Tested?

https://www.loom.com/share/b0d9452d556f41738924239cba080e92

## Checklist:


spec/controllers/platform/api/v1/accounts_controller_spec.rb
```diff
@@ -42,29 +42,24 @@
         json_response = JSON.parse(response.body)
         expect(json_response['name']).to eq('Test Account')
         expect(json_response['locale']).to eq('es')
-        expect(json_response['enabled_features']).to eq(
```
sojan-official, line 45: I believe this line is intentional and shouldn't be removed. We want to make sure that the default features enabled by installation config should work when an account is created with out any specification through `features` params

spec/controllers/platform/api/v1/accounts_controller_spec.rb
```diff
@@ -144,14 +140,19 @@
 
         patch "/platform/api/v1/accounts/#{account.id}", params: {
           name: 'Test Account',
-          enabled_features: %w[feature_ip_lookup feature_help_center],
+          features: {
```
sojan-official, line 143: @smartdev58 when specifying  { `feature_name`:  false } , will it disable the feature if it was already enabled ? 
If so, can you update this test case to ensure that?

sojan-official (changes requested):
The new format looks good to me and seems more versatile. 
Added a comment on the patch request. lets process to merge once that is resolved.

spec/controllers/platform/api/v1/accounts_controller_spec.rb
```diff
@@ -42,29 +42,25 @@
...
-        expect(json_response['enabled_features']).to eq(
-          {
-            'agent_management' => true
-          }
-        )
+        expect(json_response['enabled_features']['agent_management']).to be(true)
       end
 
       it 'creates an account with feature flags' do
         InstallationConfig.where(name: 'ACCOUNT_LEVEL_FEATURE_DEFAULTS').first_or_create!(value: [{ 'name' => 'inbox_management',
```
sojan-official, line 49:
```suggestion
        InstallationConfig.where(name: 'ACCOUNT_LEVEL_FEATURE_DEFAULTS').first_or_create!(value: [{ 'name' => 'disable_branding',
```
let's have this added to ensure that `disable_branding`: false in the params is working properly.

spec/controllers/platform/api/v1/accounts_controller_spec.rb
smartdev58: > @smartdev58 when specifying { `feature_name`: false } , will it disable the feature if it was already enabled ? If so, can you update this test case to ensure that?

@sojan-official I have already done the update you mentioned for the patch. line: 146: `channel_facebook: false`

spec/controllers/platform/api/v1/accounts_controller_spec.rb
```diff
@@ -144,14 +140,19 @@
 
         patch "/platform/api/v1/accounts/#{account.id}", params: {
           name: 'Test Account',
-          enabled_features: %w[feature_ip_lookup feature_help_center],
+          features: {
+            ip_lookup: true,
+            help_center: true,
+            channel_facebook: false
```
smartdev58, line 146: @sojan-official Here is what I did.

spec/controllers/platform/api/v1/accounts_controller_spec.rb
smartdev58: I am done with this suggested changes.

smartdev58 (commented):
I am done with the requested changes.

smartdev58:
> The new format looks good to me and seems more versatile. Added a comment on the patch request. lets process to merge once that is resolved.

I am done with the changes in patch request test spec.

sojan-official:
@smartdev58, thanks a lot for the changes.  I have slightly refactored the logic so that the behaviour is consistent for creation and update. With this change, the currently enabled features in an account or features enabled through installation configuration are not affected unless the feature key is explicitly mentioned in features params. 

```
Example: 
Account 1 -> already has inbox_management / channel_facebook enabled
Patch /accounts/1  { features: {  inbox_management: false, channel_email: true } 

Based on the initial PR -> after the patch account will only have: channel_email to be true; the previously enabled channel_facebook is also used to get overwritten.

With the current changes: the account will have both channel_email and channel_facebook enabled after the patch.

```

PR #5726, opened 2022-10-24
fix: Add preference to choose browser lang for widget settings page

### LINKED ISSUE #5082: feat: Make Chatwoot truly multilingual based on browser language or other alternatives
Chatwoot already has a multilingual widget. This can be controlled via the "locale". But there are hard contents in the following which can not be translated.

Among them title & subtitle as well as the welcome message and the "Out of office hours" message. Even via the SDK, such messages are not adjustable.

If you now create a separate inbox for each language, which could well be an idea, you quickly run into the problem with changing guests. For example, a guest lands on an Italian-language website, starts the live chat and quickly realizes that his Italian is not as good as his English and then switches to the English website. This would load a new inbox and consequently start a new conversation. If the user is then still unidentified, the previous conversation is lost and cannot be merged either. A big chaos would be the consequence if you have a well frequented website.

If you use only one inbox for a multilingual website, there are further hurdles besides the non-translatable variables described above.

**Dialogflow**: Theoretically, Dialogflow is able to handle multiple languages. For this, however, Chatwoot would have to transmit the widget language. This does not happen, so that with multiple languages the English Dialogflow is always controlled.

**Automations**: Again, this cannot be delivered according to language. There is theoretically the option of the browser language, but in multiple debugging with different browsers this value is not transmitted. Also, it is not always identical to the widget language. Here there should also be a query / option of the widget language.

**Campaigns**: Here the campaign text can also be written only hard. There are no further options. If you have a multilingual inbox, you cannot address the customer in his native language. Either you write to all of them in English or you have to do without the campaign function.

Dialogflow: Transmitting the widget language
Campaigns: Widget language as condition
Automations: Widget language as condition
Translate widget variables: Either changeable via the SDK or the ability to create translations in the UI.

See initial text

nithindavid:
## Description
Fixes #5082

https://www.loom.com/share/97ff5ae9826345ed86d914be6e9cff9c

### DIFF
--- a/app/javascript/packs/sdk.js
+++ b/app/javascript/packs/sdk.js
@@ -18,6 +18,14 @@ const runSDK = ({ baseUrl, websiteToken }) => {
   }
 
   const chatwootSettings = window.chatwootSettings || {};
+  const localeWithVariation = window.navigator.language.replace('-', '_');
+  const defaultLocale = chatwootSettings.locale || 'en';
+
+  const locale =
+    localeWithVariation && chatwootSettings.localeFromBrowser
+      ? localeWithVariation
+      : defaultLocale;
+
   window.$chatwoot = {
     baseUrl,
     hasLoaded: false,
@@ -25,7 +33,8 @@ const runSDK = ({ baseUrl, websiteToken }) => {
     isOpen: false,
     position: chatwootSettings.position === 'left' ? 'left' : 'right',
     websiteToken,
-    locale: chatwootSettings.locale,
+    locale,
+    localeFromBrowser: chatwootSettings.localeFromBrowser || false,
     type: getBubbleView(chatwootSettings.type),
     launcherTitle: chatwootSettings.launcherTitle || '',
     showPopoutButton: chatwootSettings.showPopoutButton || false,
@@ -58,7 +67,7 @@ const runSDK = ({ baseUrl, websiteToken }) => {
       IFrameHelper.events.popoutChatWindow({
         baseUrl: window.$chatwoot.baseUrl,
         websiteToken: window.$chatwoot.websiteToken,
-        locale: window.$chatwoot.locale,
+        locale,
       });
     },
 
--- a/app/javascript/packs/sdk.js
+++ b/app/javascript/packs/sdk.js
@@ -18,13 +18,10 @@ const runSDK = ({ baseUrl, websiteToken }) => {
   }
 
   const chatwootSettings = window.chatwootSettings || {};
-  const localeWithVariation = window.navigator.language.replace('-', '_');
-  const defaultLocale = chatwootSettings.locale || 'en';
-
-  const locale =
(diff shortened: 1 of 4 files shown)


app/javascript/packs/sdk.js
```diff
@@ -18,14 +18,23 @@ const runSDK = ({ baseUrl, websiteToken }) => {
   }
 
   const chatwootSettings = window.chatwootSettings || {};
+  const localeWithVariation = window.navigator.language.replace('-', '_');
+  const defaultLocale = chatwootSettings.locale || 'en';
+
+  const locale =
+    localeWithVariation && chatwootSettings.useBrowserLanguage
+      ? localeWithVariation
+      : defaultLocale;
```
pranavrajs, line 27:
```suggestion
  let locale = chatwootSettings.locale || 'en';
  if (chatwootSettings.useBrowserLanguage) {
    locale = window.navigator.language.replace('-', '_');
  }
```

pranavrajs: @nithindavid Do we need to check whether `window.navigator.language.replace('-', '_')` exist or not? It is available in all browsers, right?

app/javascript/widget/App.vue
```diff
@@ -134,10 +134,20 @@ export default {
...
+      );
+      const hasLocaleWithVariation = enabledLanguages.some(
+        lang => lang.iso_639_1_code === localeWithVariation
+      );
+
+      if (hasLocaleWithVariation) {
+        this.$root.$i18n.locale = localeWithVariation;
+      } else if (hasLocaleWithoutVariation) {
+        this.$root.$i18n.locale = localeWithoutVariation;
       }
```
pranavrajs, line 151: I'm not sure if I understand this change. Why do we split the locale here? Also, I don't think it is ideal to use pt as a substitute for pt_BR or zh instead of zh_TW.

pranavrajs: @sojan-official Thoughts?

pranavrajs (changes requested):
Nice work, this has been one of the low-hanging fruits we have kept aside for a long time. I have some comments, please take a look.

app/javascript/packs/sdk.js
nithindavid: As seen here https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language

> Examples of valid language codes include "en", "en-US", "fr", "fr-FR", "es-ES", etc.

app/javascript/widget/App.vue
nithindavid: We have `pt` and `pt_BR`. With this change, if `pt_BR` is not available it'll show `pt`. If not correct, we could do a hard check for `pt_BR`.

app/javascript/packs/sdk.js
pranavrajs: `localeWithVariation &&` - This check is not required, right?

PR #5751, opened 2022-10-26
chore: send bot reply for resolved conversations when a bot is assigned

### LINKED ISSUE #5668: send conversations to agent bot if customer reopen conversation
The agent bot now is working fine when conversations created its go directly to pending status and agent bot will handle it

The problem is :
After I close the conversation then after some days customer send new message *re open* then the conversations it will not go to agent bot again..

Solution:
If the client reopen the conversation it should go always to agent bot if it's was closed

smartdev58:
## Description

Updated bot processor service to send reply from agent bot for the resolved conversations. It just turn the conversation as pending if there is active bot assigned to the inbox.
Fixes #5668

## Type of change

## How Has This Been Tested?
https://www.loom.com/share/a9184b8acd9844beb2ab14a824962752

## Checklist:


pranavrajs (changes requested):
@smartdev58 I think it would be better to put the conversation back to the pending if the agent bot is connected to an inbox rather than building logic around resolved status. 

You should be able to update the status here. When a new message is created in a conversation it would reopen the conversation. Instead of that, you can move it pending if the agent bot is connected. See the following code

https://github.com/chatwoot/chatwoot/blob/12cd15b6ad2783634308cdee2d03f70d07adbb3c/app/models/message.rb#L192-L197

smartdev58:
@pranavrajs That sounds great idea!. Actually I was trying to do it because it conflicted between my conditional to check status "resolved" and the message model callback "reopen_converstion" Thank you!

app/models/message.rb
```diff
@@ -192,6 +192,8 @@ def send_reply
   def reopen_conversation
     return if conversation.muted?
     return unless incoming?
+    # agent bot processor will open the conversation, stay resolved
+    return if conversation.inbox.agent_bot_inbox&.active?
```
pranavrajs, line 196: Shall we mark it as pending instead of returning from here?

smartdev58: > Shall we mark it as pending instead of returning from here?

yes doing it now

smartdev58:
@pranavrajs @voko3456 I have completed the above requested changes. Please review it.

app/models/message.rb
```diff
@@ -189,11 +189,24 @@ def send_reply
...
+    if conversation.resolved?
+      if conversation.inbox.active_bot?
+        conversation.pending!
+      else
+        conversation.open!
+      end
+    end
+  end
+
+  def mark_bot_conversation_pending
```
sojan-official, line 207: @smartdev58 are we using this method anywhere ?

app/models/message.rb
```diff
@@ -189,11 +189,24 @@ def send_reply
...
 
-  def reopen_conversation
+  def update_conversation_status
     return if conversation.muted?
     return unless incoming?
 
-    conversation.open! if conversation.resolved? || conversation.snoozed?
+    conversation.open! if conversation.snoozed?
+
+    if conversation.resolved?
```
sojan-official, line 198: lets move this logic to a separate method to avoid the nested if else

app/models/conversation.rb
```diff
@@ -210,7 +210,7 @@ def validate_additional_attributes
 
   def mark_conversation_pending_if_bot
     # TODO: make this an inbox config instead of assuming bot conversations should start as pending
-    self.status = :pending if inbox.agent_bot_inbox&.active? || inbox.hooks.pluck(:app_id).include?('dialogflow')
+    self.status = :pending if inbox.active_bot?
```
sojan-official, line 213: ❤️

sojan-official (changes requested):
Looks good overall. a couple of minor comments.

app/models/message.rb
smartdev58: @sojan-official  The method is being used here: https://github.com/chatwoot/chatwoot/blob/e1190fd9bf171a3e95215bacb9bbeb217d7b4a1a/app/models/conversation.rb#L92

pranavrajs: I think this must be a mistake while I merged the code. I have updated the method above. We could remove this.

pranavrajs: @smartdev58 I can update the method as per the suggestion and get it merged.

app/models/message.rb
smartdev58: @sojan-official This is fixed.

app/models/message.rb
smartdev58: @pranavrajs Thank you. I am done with other comments here.

Issue #5688, opened 2022-10-20
bug: Conversation read status is not working properly


muhsin-k:
https://www.loom.com/share/0ef52825032249549527327ab76fce5c

PR #5777, opened 2022-10-28
fix: Conversation read update

muhsin-k:
Fixes https://github.com/chatwoot/chatwoot/issues/5688

**How to test?**
1. Create two conversations by using a web widget.
2. Minimize these two conversations.
3. Send an outgoing message from these two conversations (Messages sent by the agent are outgoing).
4. Open one of the conversations from the widget side.
5. Check the read status of the last messages. It should update only for the messages which the user reads.


pranavrajs (approved):
https://www.loom.com/share/ed6951593da9414cb87f4164e4b23357

PR #5779, opened 2022-10-31
fix: Macros authorizations

### LINKED ISSUE #5730: Macros - Release v1
Macros CRUD operations and execution is completed already. During the test, we found a couple of improvements/bugs. The list is added below.

#### Permissions:
- [x] Agents can update/delete their own macros.
- [x] Any agent or admin can only delete the macros they created or orphan macros (any macros created by an agent who was previously a part of the team).
- [x] A public macro created by one agent should be available in read-only mode to other agents, it should not be editable. Admins can edit it.
- [x] Admin can change the visibility of all the macros.
- [x] Order macros by ID. (Let us move it to last used later on)

#### UI Update for the permissions

- [ ] If agents/admins are changing a public macro to private, alert them that “You are changing the visibility of the macro. This action would make it private and it would be only available to the agent who created it”. Once confirmed, update the macro.

#### Chores
- [x] Make sure that the macros created_by / edited_by become `null` if an agent is removed from the account.
- [ ] Rename the visibility in the code to public/private instead of global/personal in the code and external facing APIs.

#### New

- [x] #5733
- [x] #5735
- [x] #5736
- [x] #5734

### Bugs

- [x] #5731
- [x] #5732

tejaswinichile:
# Pull Request Template

## Description

Macros policy update.

Fixes #5730 

## Type of change

## How Has This Been Tested?

Tested on local

## Checklist:

### DIFF
--- a/app/policies/macro_policy.rb
+++ b/app/policies/macro_policy.rb
@@ -36,7 +36,7 @@ class MacroPolicy < ApplicationPolicy
   end
 
   def orphan_record?
-    return @record.created_by.nil? if @account_user.agent? || @account_user.administrator?
+    return @account_user.administrator? if @record.created_by.nil?
 
     false
   end
--- a/app/policies/macro_policy.rb
+++ b/app/policies/macro_policy.rb
@@ -8,15 +8,17 @@ class MacroPolicy < ApplicationPolicy
   end
 
   def show?
+    return @record.global? if !author? && @account_user.agent?
+
     true
   end
 
   def update?
-    true
+    author? || @account_user.administrator?
   end
 
   def destroy?
-    true
+    author? || orphan_record?
   end
 
   def execute?
@@ -25,5 +27,17 @@ class MacroPolicy < ApplicationPolicy
 
   def attach_file?
     true
+  end
+
+  private
+
+  def author?
+    @record.created_by == @account_user.user
+  end
+
+  def orphan_record?
+    return @record.created_by.nil? if @account_user.agent? || @account_user.administrator?
+
+    false
   end
 end
--- a/app/policies/macro_policy.rb
+++ b/app/policies/macro_policy.rb
@@ -8,13 +8,11 @@ class MacroPolicy < ApplicationPolicy
   end
 
   def show?
-    return @record.global? if !author? && @account_user.agent?
-
-    true
+    @record.global? || author?
   end
 
   def update?
-    author? || @account_user.administrator?
+    author? || (@account_user.administrator? && @record.global?)
   end
 
   def destroy?
@@ -22,7 +20,7 @@ class MacroPolicy < ApplicationPolicy
   end
 
   def execute?
-    true
+    @record.global? || author?
   end
 
   def attach_file?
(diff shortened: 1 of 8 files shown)


app/policies/macro_policy.rb
```diff
@@ -8,15 +8,17 @@ def create?
...
   end
 
   def update?
-    true
+    author? || @account_user.administrator?
   end
 
   def destroy?
-    true
+    author? || orphan_record?
```
sojan-official, line 21: @tejaswinichile  should we let agents delete orphan macros ? I think it makes more sense to give this permission only to admins.

sojan-official (changes requested):
Looks good mostly. Added a comment on destroying permissions. 
Also, can we have the specs for the same?  ( agents not able to delete )

app/policies/macro_policy.rb
tejaswinichile: Yes, that does make sense but if the admin is clogged up with the work then we should let the agent delete these records. May be in the future we can come up with the archived state for the orphan records and then let the agent archive it. @pranavrajs Can you confirm?

pranavrajs: Let us not keep the private macros for the admin to delete them. When an agent is deleted, all their private macros are deleted and public macros would get orphaned. Then, all the orphan records can be deleted by an admin. I hope this would solve the burden of an admin to delete everything. Public Macros are anyways helpful for most of the users, so it makes sense to keep them and giving control to an admin to update it.

tejaswinichile:
@sojan-official Please check, I have added the changes according to the discussed policy changes.

tejaswinichile:
Test cases: 

1. Admin can see and update any[public/private] macro. [Change the visibility and data]
2. Admin can delete the public orphan macro
3. Agent/Admin can delete their own macro.
4. Agent can update their own macro.
5. Agent can see all public macros, not other agents' macros.

sojan-official:
@tejaswinichile @pranavrajs  should the admin be able to delete private macros as well, since it's visible to them? 
In that case, we could simplify the destroy policy and don't have to add specific checks for orphan records 

```
def destroy?
    author? || @account_user.administrator?
  end
```

tejaswinichile:
@sojan-official No, only author can delete their macros.
Admin can delete once it is an orphan.

All personal macros are visible to the admin because they can update it further to change the visibility if they think the macro is helpful for other agents.

cc @pranavrajs

pranavrajs:
> should the admin be able to delete private macros as well

The admin can only see the macros that are either their private macros or public macros. For eg: if agent X created a private macro, that would not be visible to the admin.

db/schema.rb
```diff
@@ -577,14 +577,12 @@
...
     t.integer "visibility", default: 0
-    t.bigint "created_by_id", null: false
-    t.bigint "updated_by_id", null: false
+    t.bigint "created_by_id"
+    t.bigint "updated_by_id"
     t.jsonb "actions", default: {}, null: false
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.index ["account_id"], name: "index_macros_on_account_id"
-    t.index ["created_by_id"], name: "index_macros_on_created_by_id"
```
pranavrajs, line 586: Should we keep this?

tejaswinichile: Yes, we are removing the index as it can be a null value.

PR #5783, opened 2022-11-01
fix: Unsupported message type

tejaswinichile:


tejaswinichile:
@sojan-official I have added test cases as per our discussion.

sojan-official (approved):

PR #5797, opened 2022-11-04
chore: Add feature flag for mobile v2

muhsin-k:
- Added feature flag `mobile_v2` to roll out new UI changes in the mobile app.
- Added `features` in `sign_in` response payload.

### DIFF
--- a/app/views/api/v1/models/_user.json.jbuilder
+++ b/app/views/api/v1/models/_user.json.jbuilder
@@ -28,5 +28,6 @@ json.accounts do
     # availability derived from presence
     json.availability_status account_user.availability_status
     json.auto_offline account_user.auto_offline
+    json.features account_user.account.enabled_features
   end
 end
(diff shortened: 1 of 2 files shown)


app/views/api/v1/models/_user.json.jbuilder
```diff
@@ -28,5 +28,6 @@ json.accounts do
     # availability derived from presence
     json.availability_status account_user.availability_status
     json.auto_offline account_user.auto_offline
+    json.features account_user.account.enabled_features
```
pranavrajs, line 31: @muhsin-k I think this change is unnecessary. We use `/api/v1/accounts/{account_id}` to fetch more details about the account on the dashboard.

This is the flow we have 
`validate_token -> fetch_single_account_details -> render dashboard`

Adding more information on the user API call might not be a good option to have.

pranavrajs (changes requested):
Please see the individual comment.

pranavrajs (approved):

PR #5814, opened 2022-11-08
feat: Add a loader till dashboard app is loaded

fayazara:
## Description
Adds a loader for dashbaord apps, for better experience, as the current screen just shows a blank box.

Screenshot2022-11-08 at 10 53 43


nithindavid (approved):
🚀

iamsivin (approved):
@fayazara LGTM 🎉. Have a small doubt that the loader of the dashboard app is only showing for the first time after refreshing or the app load. Is it okay? 
https://www.loom.com/share/3c8a63fb09e84c06991fe26769e7b926

app/javascript/dashboard/components/widgets/DashboardApp/Frame.vue
```diff
@@ -16,7 +21,11 @@
 </template>
 
 <script>
+import LoadingState from 'dashboard/components/widgets/LoadingState.vue';
```
iamsivin, line 24:
```suggestion
import LoadingState from 'dashboard/components/widgets/LoadingState';
```

pranavrajs:
@iamsivin Nice catch! @fayazara It should reset the loading state is we are moving to another tab.

PR #5830, opened 2022-11-10
bug: Fixes broken dropdown issue

nithindavid:
## Description
- Adds missing method


app/javascript/shared/components/ui/MultiselectDropdownItems.vue
```diff
@@ -122,6 +122,10 @@ export default {
     focusInput() {
       this.$refs.searchbar.focus();
     },
+    isActive(option) {
+      const selected = this.selectedItem;
+      return selected ? option.id === selected.id : false;
+    },
```
pranavrajs, line 128:
```suggestion
    isActive(option) {
      return this.selectedItem ? option.id === this.selectedItem.id : false;
    },
```

Issue #5705, opened 2022-10-21
The full name is not updating when creating a conversation from the widget


Ghost108:
Self hosted via Docker

If you activate the pre chat option wit the fields fullname and email, it works fine:
<img width="971" alt="Bildschirmfoto 2022-10-21 um 15 22 22" src="https://user-images.githubusercontent.com/17011498/197206436-f2efca36-0f48-4deb-a583-31475ac2cc90.png">

<img width="333" alt="Bildschirmfoto 2022-10-21 um 15 21 25" src="https://user-images.githubusercontent.com/17011498/197206515-812c2860-cd54-468e-88bf-ce9d1a122a47.png">

As a agent, I get all values:
<img width="324" alt="Bildschirmfoto 2022-10-21 um 15 21 50" src="https://user-images.githubusercontent.com/17011498/197206620-d875f321-6cf6-40ae-ac07-2f00404c2a63.png">

But If I would only set fullname active in the pre chat:
<img width="955" alt="Bildschirmfoto 2022-10-21 um 15 22 32" src="https://user-images.githubusercontent.com/17011498/197206709-c790d544-4589-4592-a5fc-8deec8ef544e.png">

<img width="328" alt="Bildschirmfoto 2022-10-21 um 15 23 00" src="https://user-images.githubusercontent.com/17011498/197206739-2d75f9e8-1d40-40c9-8d10-5921563cdeda.png">

Do I not get the name transmitted
<img width="256" alt="Bildschirmfoto 2022-10-21 um 15 23 18" src="https://user-images.githubusercontent.com/17011498/197206917-99e0ab3b-36fd-4ff9-ae8c-a5b7e13821d4.png">

pranavrajs:
@Ghost108 Thanks for reporting the issue. @muhsin-k Can you please check if you can reproduce the behaviour?

muhsin-k:
The issue is reproducible.

muhsin-k:
Currently, the contact name will only add if an email ID exists.

Reference
https://github.com/chatwoot/chatwoot/blob/develop/app/controllers/api/v1/widget/base_controller.rb#L53 

We could do something like this,
```
def contact_name
   permitted_params.dig(:contact, :name) || (contact_email.split('@')[0] if contact_email.present?)
end
```

 The issue with this approach is that the contact name will be rewritten if the contact already exists.

@pranavrajs @sojan-official

muhsin-k:
[...]

PR #5832, opened 2022-11-10
fix: Full name update when creating a conversation without an email id

muhsin-k:
Fixes https://github.com/chatwoot/chatwoot/issues/5705

**How to test the PR?**

- Enable the `fullName` field and marked as mandatory in the pre-chat form.
- Create a new conversation from the web widget.
- Check name is updating in the dashboard or not.


iamsivin (commented):
I tested in staging and LGTM 🎉
https://www.loom.com/share/96ad9086a5d6450bbabcc802560bce0b

app/controllers/api/v1/widget/base_controller.rb
```diff
@@ -50,7 +50,9 @@ def contact_email
   end
 
   def contact_name
-    params[:contact][:name] || contact_email.split('@')[0] if contact_email.present?
+    return if @contact.email.present? || @contact.phone_number.present?
```
pranavrajs, line 53:
```suggestion
    return if @contact.email.present? || @contact.phone_number.present? || @contact.identifier.present?
```

PR #5847, opened 2022-11-13
chore: Better scrollbars

fayazara:
## Description

Adds better looking and standard scrollbars. 

Used a custom styled scrollbar using
```css
-webkit-overflow-scrolling: touch;
-ms-overflow-style: -ms-autohiding-scrollbar;
 ```
 And the ` overflow: overlay;` property with a fallback of `overflow: auto`

Firefox does not support this, have added a fallback for this in `app/javascript/dashboard/assets/scss/_mixins.scss`

| Before                                                                                                           | After                                                                                                           |
| ---------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------- |
| before | after |


pranavrajs:
@nithindavid Can you please take the next steps on this?

scmmishra:
@fayazara @pranavrajs @nithindavid this seems like a nice change to have, are we planning to roll it out anytime soon?

fayazara:
> @fayazara @pranavrajs @nithindavid this seems like a nice change to have, are we planning to roll it out anytime soon? At the moment, there is a lot of layout shift during interactions, this can help solve that

@nithindavid we can merge this. Will do another round of test to make sure it works as expected.

scmmishra:
@fayazara seems like the styles have drifted a bit since the PR was open. We would need to take a look at the changes again. Also, seems the like the property `overlay` is not standard, and is considered deprecated according to MDN. Not sure if we should depend on it, need your take on this @fayazara, @nithindavid

scmmishra:
@sojan-official I think for now we can close this PR. It's a good to have feature we can work on after we complete v3 tasks.

@nithindavid WDYT?

sojan-official:
lets reopen if required

PR #5849, opened 2022-11-14
Instagram mocked webhook

tejaswinichile:


tejaswinichile:
We are not doing this mocked service thing anymore. It just needs the app in the production status and then makes the same app in the development state.

PR #5853, opened 2022-11-15
fix: send label list not object in event data presenter

tejaswinichile:
Fixes https://github.com/chatwoot/product/issues/659

## Type of change

It was added for dispatching an event when the label is changing through macros or API.

- Test macros add_label for the impact.

### DIFF
--- a/spec/listeners/action_cable_listener_spec.rb
+++ b/spec/listeners/action_cable_listener_spec.rb
@@ -149,7 +149,7 @@ describe ActionCableListener do
     end
 
     it 'broadcast event with label data' do
-      expect(conversation.reload.push_event_data[:labels]).to eq(conversation.labels.pluck(:id, :name))
+      expect(conversation.reload.push_event_data[:labels]).to eq(conversation.labels.pluck(:name))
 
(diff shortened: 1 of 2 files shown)


spec/listeners/action_cable_listener_spec.rb
```diff
@@ -149,7 +149,7 @@
     end
 
     it 'broadcast event with label data' do
-      expect(conversation.reload.push_event_data[:labels]).to eq(conversation.labels.pluck(:id, :name))
+      expect(conversation.reload.push_event_data[:labels]).to eq(conversation.labels.pluck(:name))
```
pranavrajs, line 152: Shouldn't this be conversation.label_list?

fayazara (approved):
https://user-images.githubusercontent.com/15716057/201832352-816db04c-d9c5-44ad-b7c3-907530635240.mp4

Working as expected.

spec/listeners/action_cable_listener_spec.rb
tejaswinichile: @pranavrajs `label_list` is what we are fetching in the `push_event_data['labels']` and this gives more idea of what it contains.

Issue #5501, opened 2022-09-27
Editing canned response breaks formatting in reply box


nithindavid:
We have line breaks in the Canned Response message, but once you use the snippet and edit it, all line breaks and formatting breaks.

https://user-images.githubusercontent.com/1277421/192401857-4ddfd810-7dad-4270-beb2-cb19c7f20d53.mov

nithindavid:
Closing in favour of #5533

PR #5880, opened 2022-11-16
fix: Updates logic to insert canned response into editor

nithindavid:
- Refactoring the Hacky fix for #5501 to remove unnecessary code
- Renders incoming canned text as markdown

**What it does**
- Creates a text node with the canned response 

**Loom video**

https://www.loom.com/share/aaee97d2b83c47128fd1f6ff14b320de


iamsivin (approved):
Tested LGTM. 
https://www.loom.com/share/f8338b1ef16c40df9c612bb93acb4614

@nithindavid Do we support saving the canned response as rich text?

nithindavid:
@iamsivin We can now use raw markdown in canned responses. Once this is merged we could add the support for rich text

Issue #5078, opened 2022-07-21
bug: Sending Audio messages not working in Whatsapp Cloud


sojan-official:

giquieu:
@sojan-official 
Anyone working on this? I can investigate if needed.

sojan-official:
@giquieu  not at the moment. That would be super helpful. ❤️

drdaga:
> @sojan-official Anyone working on this? I can investigate if needed.

i checked actually audio file type should not pass caption parameter

i written simple if else 

`def send_attachment_message(phone_number, message)

    attachment = message.attachments.first
    type = %w[image audio video].include?(attachment.file_type) ? attachment.file_type : 'document'
    attachment_url = attachment.download_url
    if type == "audio"
      response = HTTParty.post(
      "#{phone_id_path}/messages",
      headers: api_headers,
      body: {
        messaging_product: 'whatsapp',
        'to' => phone_number,
        'type' => message.attachment_type,
        message.attachment_type.to_s => {
          'link': attachment_url,
        }
      }.to_json
    )
    else
      response = HTTParty.post(
        "#{phone_id_path}/messages",
        headers: api_headers,
        body: {
          messaging_product: 'whatsapp',
          'to' => phone_number,
          'type' => message.attachment_type,
          message.attachment_type.to_s => {
            'link': attachment_url,
            'caption': message.content
          }
        }.to_json
      )
      end
    process_response(response)
  end`

sojan-official:
@drdaga Thanks for the findings. Would you be interested in raising a PR? or should we take it up?

giquieu:
@drdaga @sojan-official 
If you don't mind, I sent the PR.
As we send the media type as audio, Whatsapp renders it as a single audio file. To render as a recording, we would need to change the media type to voice to send as a voice recording.

drdaga:
> I have consulted the documentation, and you are right not to send the caption to Cloud API. For Twilio it is not necessary. For D360 within the documentation it says nothing about captioning.
> 
> About the .amr serialization, did you get to validate or implement something?
> 
> Twilio

abt Twilio & D360 they are hosted whatsapp api so even if we send caption they will ignore it
and cloud whatsapp api it its send directly to meta server so it will have an error
@giquieu u can send the pr no problem as i dont knw much abt coding its jst as hobby by profession i am doctor implemented chatwoot for our hospital and our consulting doctors so using chatwoot

drdaga:
[...]

Issue #4481, opened 2022-04-15
bug: Files in Whatsapp arrives with a different name


mattiamarras:
Hi, if I send a file via WhatsApp to the end customer it arrives with a different name (null.pdf or Untitled).

sojan-official:
@mattiamarras  could you add more details on which provider you are using for whatsapp ?

mattiamarras:
Ciao Sojan, i am using https://www.360dialog.com/, using the native integration of Chatwoot.

sojan-official:
### Notes 

ref: https://github.com/chatwoot/chatwoot/blob/a788e7ac7ab4b8bc358981124694676b413849fd/app/services/whatsapp/providers/whatsapp_360_dialog_service.rb#L81

- [ ] Replace caption with filename when attachment type is document
- [ ] Send message content as a separate message before the attachment ( since whatsapp doesn't support message and attachments attached together )
- [ ] Implement the same for whatsapp cloud
- [ ] Verify behaviour of audio messages for 360dialog when caption is present: ref: https://github.com/chatwoot/chatwoot/issues/5078

vagner-dias:
@sojan-official - FYI - I am using Whasapp Cloud API and I am having the same synptom.

If I send this on chatwoot:
![image](https://user-images.githubusercontent.com/86367839/201949151-8eb3c485-9011-4875-a12e-abd617f784ac.png)

the user gets this:
![image](https://user-images.githubusercontent.com/86367839/201949355-792b954f-98e0-4fee-9cb4-7ef5325abb59.png)

giquieu:
Hello @sojan-official 
I sent a PR about this Issue: https://github.com/chatwoot/chatwoot/pull/5907
Could you please review the PR? Also a test with 360Dialog.
I tested it only on WhatsApp Cloud and it's working.

sojan-official:
@giquieu  Thanks a lot for working on this. We will have it tested and merged.

PR #5907, opened 2022-11-21
Fix: Files in Whatsapp arrives with a different Name

giquieu:
# Pull Request Template

## Description

Adjusted the way the Cloud service and Dialog360 send the attachment name.
According to documentation:
image

- [X]  Replace caption with filename when attachment type is document
- [X] Implement the same for whatsapp cloud
- [X] Verify behaviour of audio messages for 360dialog when caption is present: ref: (https://github.com/chatwoot/chatwoot/issues/5078)

About this:
- Send message content as a separate message before the attachment ( since whatsapp doesn't support message and attachments attached together )

Only audio/sticker type attachments do not allow text.

The other models allow text along with the attachment (caption), but when there is content (text) and it is not audio/sticker, the text will be sent as a caption.

Fixes # (issue)
#4481 

## Type of change

## How Has This Been Tested?

Attached image with and without content in message.
Attached PDF with and without content in the message.
Recorded Audio (Recorded audio does not allow message text).

In both cases the messages were received on WhatsApp.

## Checklist:

### DIFF
--- a/app/services/whatsapp/providers/whatsapp_360_dialog_service.rb
+++ b/app/services/whatsapp/providers/whatsapp_360_dialog_service.rb
@@ -69,17 +69,18 @@ class Whatsapp::Providers::Whatsapp360DialogService < Whatsapp::Providers::BaseS
   def send_attachment_message(phone_number, message)
     attachment = message.attachments.first
     type = %w[image audio video].include?(attachment.file_type) ? attachment.file_type : 'document'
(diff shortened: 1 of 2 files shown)


sojan-official:
QA

https://user-images.githubusercontent.com/73185/204526653-d3727cff-f908-4d9e-9d8e-de1229b98365.mov

Issue #5552, opened 2022-10-05
feat: Ability to mark conversations as unread


sojan-official:
As a customer service agent, I might often open a conversation and leave it for later to respond. In this case, it is helpful if we can mark the conversation as unread so that other members of the team can pick it up. 

> strongly want to have the ability to “unread” messages in the inbox and make them bold again. Is this something that could be added using the multiselect and/or context menu?

sojan-official:
- context menu will have mark unread option
 - it will hit conversation/id/unread endpoint 
 - system will change conversation agent last seen timestamp to last message timestamp - 1

PR #5924, opened 2022-11-23
feat: Allow users to mark a conversation as unread

sojan-official:
fixes: #5552 

Loom video: https://www.loom.com/share/ab70552d3c9c48b685da7dfa64be8bb3


pranavrajs:
@muhsin-k @fayazara @iamsivin Please review the PR and test out the functionality. 

@tejaswinichile Please take a look at the backend changes.

app/javascript/dashboard/store/modules/conversations/actions/messageReadActions.js
```diff
@@ -0,0 +1,35 @@
+import { throwErrorMessage } from 'dashboard/store/utils/api';
```
pranavrajs, line 1: Created a new file as the action.js has passed 300 lines. Modularizing parts of the code would be required.

fayazara (approved):
FE code looks.

Tested locally - working as expected.

https://www.loom.com/share/d4aa13d885de462883ed89cf7788749a

muhsin-k (approved):

nithindavid:
Should we keep a button next to the resolve button(adjacent to mute) to toggle the same and with a tooltip?

tejaswinichile:
The backend looks good, why did you add skipModelValidation in robocop?

sojan-official:
@tejaswinichile we are doing an update column here, which rubocop doesn't like. but it is intended in this case, as we don't want the callbacks.

sojan-official:
> Should we keep a button next to the resolve button(adjacent to mute) to toggle the same and with a tooltip?

We could take it as a subsequent task if there is enough usage for this functionality. The current customer requirement only wanted it in the context menu.  will convert it to a follow up issue.

pranavrajs:
@nithindavid This could be a good enhancement. Let us rework on the top panel and place important items there.

PR #5976, opened 2022-11-29
feat: add lockable functionality for user login attempts

tejaswinichile:
Security issues: If the user applies brute force for login with multiple passwords.

We should limit the login attempts for the user, starting with user/agents not for super admin yet.


tejaswinichile:
I am closing this PR, may be we will revisit this later. 

cc @sojan-official @pranavrajs

PR #5978, opened 2022-11-29
fix: Custom attr filter on conversations

tejaswinichile:
# Pull Request Template

## Description

Custom attribute filter on conversations.

Fixes https://github.com/chatwoot/product/issues/680

## Type of change

## How Has This Been Tested?

- Equal to filter works fine
- Not equal to filter works only if custom attributes are present.

## Checklist:


tejaswinichile:
https://www.loom.com/share/b900a9b3ec0d48c8a409fd441370020d

sojan-official (approved):

pranavrajs (changes requested):
The problem still exists. It returns only the conversations with the attribute present. 

To reproduce:

Add a new attribute in custom attributes and apply the Not Equal To filter. The expectation is that it should return all the conversations because none of the conversations have this attribute added to them. The actual result is that it returns empty results.

pranavrajs:
@tejaswinichile Looked at the loom video. I don't get the logic. Whether the custom attributes key is present or not, it should return all the objects which doesn't match the filter.

Here `Not Equal` implies `Not Equal or not present.` If the key is not present, it should match.

tejaswinichile:
@pranavrajs Please check the changes with the updated query

https://www.loom.com/share/17d2672a5c0c491b99c41352b6e43416

pranavrajs:
@tejaswinichile Tested this and it works locally now. Nice work. @sojan-official Can you please check the query builder changes and merge this?

PR #5984, opened 2022-11-29
fix: Save hostname for the custom domain in the portal

### LINKED ISSUE #5983: Save the portal custom_domain hostname in the backend
When we are redirecting the portal's data to custom_domain, the Portal checks custom_domain against the hostname and we are saving the whole URL.

```
domain = request.host
@portal = Portal.find_by(custom_domain: domain)
```

This validation over the domain URL we added after the functionality is deployed on the front end.
`return this.$t('HELP_CENTER.PORTAL.ADD.DOMAIN.ERROR');`

To fix this: Whenever we are saving the custom-domain URL, we need to save the hostname only.

tejaswinichile:
Saving the custom-domain with validation of URL, impacting the backend functionality to check the custom-domain against portal records.

So while saving the records we are only saving the hostname not the whole URL in the backend.

Fixes #5983 

### DIFF
--- a/app/controllers/api/v1/accounts/portals_controller.rb
+++ b/app/controllers/api/v1/accounts/portals_controller.rb
@@ -21,6 +21,7 @@ class Api::V1::Accounts::PortalsController < Api::V1::Accounts::BaseController
 
   def create
     @portal = Current.account.portals.build(portal_params)
+    @portal.custom_domain = parsed_custom_domain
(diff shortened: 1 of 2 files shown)


vishnu-narayanan (approved):
Tested and verified on https://pfx.chatwoot.dev

portal --> `https://support.chatwoot.dev`

PR #6001, opened 2022-12-01
fix: Avoid conversationId getting undefined in unattended view

### LINKED ISSUE #6000: Draft replies not working with unattended / mentions view
Draft reply only works if the user is all conversations, inbox views, or label views. Suppose you try to write anything while in the unattended view. It doesn't save it.

pranavrajs:
Fixes #6000 

https://www.loom.com/share/34935da28b854fb1a7f17156cb5a9db3


fayazara (approved):
Looks good. Tested locally

iamsivin (approved):
LGTM

PR #6005, opened 2022-12-01
chore: Fix connection pool

sojan-official:
We want to allocate as much database connection to the pool to match the sidekiq concurrency configuration. 
ref: https://maxencemalbois.medium.com/the-ruby-on-rails-database-connections-pool-4ce1099a9e9f 

fixes: #6004


sojan-official:
@vishnu-narayanan  could you do a quick QA for this? let's try it out on staging and maybe on a self-hosted instance as well and merge this.

vishnu-narayanan:
@sojan-official Not able to verify this. The Sidekiq concurrency variable is not taking effect.

```
3.0.4 :005 > ENV.fetch('SIDEKIQ_CONCURRENCY')
 => "12"
3.0.4 :006 > Sidekiq.options[:concurrency]
 => 10
```

image

vishnu-narayanan (commented):
Can we also have this variable on the `.env.example` file with a comment?

tejaswinichile (approved):

sojan-official:
Screenshot 2023-01-13 at 4 29 22 PM

Verified on QA instance that the configuration is now working

deployment/chatwoot-worker.1.service
```diff
@@ -7,7 +7,7 @@ Type=simple
 User=chatwoot
 WorkingDirectory=/home/chatwoot/chatwoot
 
-ExecStart=/bin/bash -lc 'bundle exec sidekiq -C config/sidekiq.yml'
+ExecStart=/bin/bash -lc 'dotenv bundle exec sidekiq -C config/sidekiq.yml'
```
sojan-official, line 10: @vishnu-narayanan, since we load environment variables in Linux installations using dotenv. we will need to have this. Could you access the impact? also, see if we need to update any documentation.

vishnu-narayanan: @sojan-official LGTM

vishnu-narayanan (approved):
@sojan-official Should the change also be made on `Procfile`? https://github.com/chatwoot/chatwoot/blob/develop/Procfile

vishnu-narayanan:
Tested and verified on a linux installation
```
SIDEKIQ_CONCURRENCY=20
```

image

sojan-official:
> @sojan-official Should the change also be made on `Procfile`? https://github.com/chatwoot/chatwoot/blob/develop/Procfile

In other environments we be fetching this values directly from environment and not using dotenv. So it should be good.

PR #6021, opened 2022-12-05
feat: Search improvements

### LINKED ISSUE #5913: API Updates to support conversation search

tejaswinichile:
Fixes #5913 

Deployed on QA instance: https://www.loom.com/share/5a8f77a8fe49402f8f42e49c320f7288

https://www.loom.com/share/23a32c889cb940229aeb55ccf9cd3ab6


app/views/api/v1/models/_conversation.json.jbuilder
```diff
@@ -10,7 +10,7 @@ json.inbox do
   json.channel_type conversation.inbox.channel_type
 end
 json.messages do
-  json.array! conversation.messages do |message|
+  json.array! conversation.messages.last(1) do |message|
```
tejaswinichile, line 13: This is a temporary thing that I have added to make the UI show some records.

tejaswinichile:
https://www.loom.com/share/037a7deab5af4fb6aeb6816635f05cf1

fayazara:
@tejaswinichile How do I query conversation content?

I tried `/api/v1/accounts/3/conversations/text_search?q=Thank%20you` - it returns me an internal error

tejaswinichile:
@fayazara This has been fixed. Can you please check?

Screenshot 2022-12-13 at 9 26 26 AM

tejaswinichile:
@fayazara Please check this response with the latest changes

Conversation with the contact email filter is being return and agent details along with conversation if it is present.

Screenshot 2022-12-16 at 6 52 45 PM

app/finders/text_search.rb
```diff
@@ -0,0 +1,36 @@
...
+      contacts: filter_contacts
+    }
+  end
+
+  private
+
+  def filter_conversations
+    conversation_ids = PgSearch.multisearch((@params[:q]).to_s).where(account_id: @current_account,
+                                                                      searchable_type: 'Conversation').pluck(:searchable_id)
+    @conversations = Conversation.where(id: conversation_ids)
```
sojan-official, line 24: We should do pagination here. this could be the culprit behind the slow search for shorter terms. 
We can also look into doing a join query so that the pagination works effectively.  

The same goes for messages and contacts as well.

db/migrate/20221212061802_enable_multi_searchable.rb
```diff
@@ -0,0 +1,12 @@
+class EnableMultiSearchable < ActiveRecord::Migration[6.1]
+  def up
+    Contact.rebuild_pg_search_documents
+    PgSearch::Multisearch.rebuild(Conversation)
+    PgSearch::Multisearch.rebuild(Message)
+    execute 'CREATE EXTENSION IF NOT EXISTS pg_trgm;'
+  end
```
sojan-official, line 7: - We need to estimate the time of this migration while executing on prod. Since the message table could be quite big, I am concerned about the time this migration may take, so we need to test this thoroughly. 

- I think it's better to do it as a sidekiq job than directly inline. 

- we need to ensure the pg_trgm extension is available and working on the cloud platforms which we support like ubuntu, AWS, GCP etc.

spec/finders/text_search_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        expect(result[:contacts].length).to be 2
+      end
+
+      it 'filter conversations by contact details' do
+        params = { q: 'pot' }
+        result = described_class.new(user_1, params).perform
+        expect(result[:conversations].length).to be 1
+      end
+
+      it 'filter conversations by contact email' do
```
sojan-official, line 56: Can we also have a spec where deleting objects removes the data from the index?

app/models/contact.rb
```diff
@@ -140,6 +147,24 @@ def discard_invalid_attrs
...
+                CONCAT_WS(' ', contacts.id, contacts.email, contacts.name, contacts.phone_number, contacts.account_id) AS content,
+                contacts.account_id::int AS account_id,
+                conversations.id AS conversation_id,
+                now() AS created_at,
+                now() AS updated_at
+        FROM contacts
+        LEFT OUTER JOIN conversations
+          ON conversations.contact_id = contacts.id
+    SQL
+  end
```
sojan-official, line 166: Could you add more context to the SQL statement?

sojan-official (commented):
A couple of comments after the quick pass. I will test this out again later based on the bandwidth.

db/migrate/20221212061802_enable_multi_searchable.rb
tejaswinichile: @sojan-official This Makes sense; let me move it to sidekiq job.

tejaswinichile: pg_trgm available:

GCP extension https://cloud.google.com/alloydb/docs/reference/extensions
AWS extension https://docs.aws.amazon.com/AmazonRDS/latest/PostgreSQLReleaseNotes/postgresql-extensions.html
Ubuntu apt-get installs are available with the pg_trgm extension.

app/finders/text_search.rb
tejaswinichile: `PgSearch.multisearch` joins with conversations? 

@sojan-official I have discussed this with @fayazara  pagination wouldn't work for a multipage search he said, instead, we are limiting the result to the top 20.

tejaswinichile: @sojan-official MultiSearch generates query like this

```PgSearch.multisearch('test').to_sql
=> "SELECT \"pg_search_documents\".* FROM \"pg_search_documents\" INNER JOIN (SELECT \"pg_search_documents\".\"id\" AS pg_search_id, (ts_rank((to_tsvector('simple', coalesce(\"pg_search_documents\".\"content\"::text, ''))), (to_tsquery('simple', ''' ' || 'test' || ' ''')), 0)) AS rank FROM \"pg_search_documents\" WHERE ((to_tsvector('simple', coalesce(\"pg_search_documents\".\"content\"::text, ''))) @@ (to_tsquery('simple', ''' ' || 'test' || ' ''')))) AS pg_search_ce9b9dd18c5c0023f2116f ON \"pg_search_documents\".\"id\" = pg_search_ce9b9dd18c5c0023f2116f.pg_search_id ORDER BY pg_search_ce9b9dd18c5c0023f2116f.rank DESC, \"pg_search_documents\".\"id\" ASC"```

I will check if joins are possible.

**Comment by sojan-official** [comment]
Note: check whether messages and conversations are only fetched from inboxes to which the agent has access.

**Comment by sojan-official** [comment]
## Next Steps

- [ ] Add a filter for allowed inboxes for users
- [ ] Split the indexing job into sub-jobs based on accounts
- [ ] Change the priority for the migration jobs, add a new queue async database migration job

-----

- [ ] Setup an instance with a snapshot of the staging
- [ ] Test the availability of pg extension in self-hosted environments ( Cloud SQL, Ubuntu installation )
- [ ] Load test the search endpoint with locust

**Comment by tejaswinichile** [comment]
This is after fixing the conversation link in searched messages payload: https://www.loom.com/share/de1a33c361554369a86b5eb6a542a69d

File: db/schema.rb [hunk 6]
```diff
@@ -838,6 +853,9 @@
     t.jsonb "custom_attributes", default: {}
     t.string "type"
     t.text "message_signature"
+    t.datetime "locked_at"
+    t.integer "failed_attempts"
+    t.string "unlock_token"
```
sojan-official, line 858: not a part of this PR i believe

app/models/concerns/multi_searchable_helpers.rb
```diff
@@ -0,0 +1,71 @@
...
+                  conversations.inbox_id::int AS inbox_id,
+                  now() AS created_at,
+                  now() AS updated_at
+          FROM conversations
+          INNER JOIN contacts
+            ON conversations.contact_id = contacts.id
+          WHERE conversations.account_id = #{account_id}
+      SQL
+    end
+  end
```
sojan-official, line 46: look and see if we can get rid of the SQL statements and use `record.update_pg_search_document ` instead. 
otherwise move the logic to migrations and make sure we escape values before running the sql.

app/models/contact.rb
```diff
@@ -26,6 +26,13 @@ class Contact < ApplicationRecord
   include Avatarable
   include AvailabilityStatusable
   include Labelable
+  include PgSearch::Model
+  include MultiSearchableHelpers
```
sojan-official, line 30: ensure to add an after update method which will reindex the conversation records if contact email phone number etc gets updated

app/models/contact.rb
```diff
@@ -140,6 +147,28 @@ def discard_invalid_attrs
...
+                now() AS created_at,
+                now() AS updated_at
+        FROM contacts
+        INNER JOIN conversations
+          ON conversations.contact_id = contacts.id
+        WHERE contacts.account_id = #{account_id}
+    SQL
+  end
+
   private
```
sojan-official, line 172: same comment as the helper

app/models/message.rb
```diff
@@ -166,6 +174,25 @@ def email_notifiable_message?
...
+                messages.account_id::int AS account_id,
+                messages.conversation_id::int AS conversation_id,
+                messages.inbox_id::int AS inbox_id,
+                now() AS created_at,
+                now() AS updated_at
+        FROM messages
+        WHERE messages.account_id = #{account_id}
+    SQL
+  end
+
```
sojan-official, line 195: same comments as in helpers

config/routes.rb
```diff
@@ -72,6 +72,7 @@
               get :meta
               get :search
               post :filter
+              get :text_search
```
sojan-official, line 75: lets have a dedicated search controller instead

db/migrate/20221205081737_create_pg_search_documents.rb
```diff
@@ -0,0 +1,22 @@
+class CreatePgSearchDocuments < ActiveRecord::Migration[6.1]
+  def up
+    say_with_time('Creating table for pg_search multisearch') do
+      create_table :pg_search_documents do |t|
+        t.text :content
+        t.bigint 'conversation_id'
+        t.bigint 'account_id'
+        t.bigint 'inbox_id'
```
sojan-official, line 8: conversation id and inbox id might not be required

db/schema.rb
```diff
@@ -401,7 +402,7 @@
     t.datetime "agent_last_seen_at"
     t.jsonb "additional_attributes", default: {}
     t.bigint "contact_inbox_id"
-    t.uuid "uuid", default: -> { "gen_random_uuid()" }, null: false
+    t.uuid "uuid", default: -> { "public.gen_random_uuid()" }, null: false
```
sojan-official, line 405: required ?

sojan-official (commented):
## Summary of the review discussion 

- [x] Move the search to a dedicated controller / route [ account/id/search] 
- [ ] Revisit the helpers and see if we can remove the raw SQL statements if not escape the values
- [ ] trim the methods in search helpers if deemed unnecessary
- [ ] ensure to have conversation records are updated if the contact gets updated 
- [ ] remove the inbox id and conversation id from search index.
- [ ] Ensure to test on AWS, and GCP before merging and see if need any additional steps and have it documented before merging this

tejaswinichile:
@sojan-official Could you please re-review this?

app/views/api/v1/models/_message.json.jbuilder
```diff
@@ -2,7 +2,8 @@ json.id message.id
 json.content message.content
 json.inbox_id message.inbox_id
 json.echo_id message.echo_id if message.echo_id
-json.conversation_id message.conversation.display_id
+# For deleted conversation, messages are not yet deleted [because of destroy_async] for this we added try block
```
sojan-official, line 5: 👍

app/models/contact.rb
```diff
@@ -180,10 +193,15 @@ def dispatch_create_event
...
   def dispatch_update_event
+    update_conversation_pgsearch_document
     Rails.configuration.dispatcher.dispatch(CONTACT_UPDATED, Time.zone.now, contact: self)
   end
 
   def dispatch_destroy_event
     Rails.configuration.dispatcher.dispatch(CONTACT_DELETED, Time.zone.now, contact: self)
   end
+
+  def update_conversation_pgsearch_document
```
sojan-official, line 204: let's move this to a background job.

app/helpers/api/v1/accounts/search_helper.rb
```diff
@@ -0,0 +1,2 @@
+module Api::V1::Accounts::SearchHelper
```
sojan-official, line 1: required?

sojan-official:
##  next steps 
- [ ] estimate the time for migration on QA instance
- [ ] Merge to develop on Friday and deploy to prod over the weekend

pranavrajs:
@tejaswinichile Sounds good. Let us ship the API first and get the migrations up to date, and once it is ready, let us swap the UI.

tejaswinichile:
We couldn't deploy on Friday as On Thursday, we found an issue related to performance on search query with 4Cr+ records.

So I tried to add an index with different migration which failed, with the server connection error. I tried to roll back the migration, which also gave the same issue on the QA instance.
  
So this rollover to the next weekend.

cc @pranavrajs @sojan-official 

The new index on content[PgSearch::Document] will help to improve performance on the search, 

```
Sort  (cost=33.18..33.19 rows=3 width=108)
   Sort Key: (ts_rank(to_tsvector('simple'::regconfig, COALESCE(pg_search_documents_1.content, ''::text)), '''hi'''::tsquery, 0)) DESC, pg_search_documents.id
   ->  Nested Loop  (cost=8.17..33.16 rows=3 width=108)
         ->  Bitmap Heap Scan on pg_search_documents pg_search_documents_1  (cost=8.02..15.88 rows=3 width=40)
               Recheck Cond: (to_tsvector('simple'::regconfig, COALESCE(content, ''::text)) @@ '''hi'''::tsquery)
               ->  Bitmap Index Scan on pg_multisearch_index  (cost=0.00..8.02 rows=3 width=0)
                     Index Cond: (to_tsvector('simple'::regconfig, COALESCE(content, ''::text)) @@ '''hi'''::tsquery)
         ->  Index Scan using pg_search_documents_pkey on pg_search_documents  (cost=0.15..5.50 rows=1 width=104)
               Index Cond: (id = pg_search_documents_1.id)
```
We just need to check out once again on QA intance.

sojan-official:
Closing in favour of https://github.com/chatwoot/chatwoot/pull/6564 
Let's reopen later if required for  pgsearch

glauberssoutodossantos:
Hi @sojan-official and @tejaswinichile, how are you?

I'll leave my 2 cents for this topic.

I think the search bar is very little visible in chatwoot. With a simple application of gray in the background it could be much better.

See the example below.

Chatwoot:

image


WhatsApp:

image


Chatwoot with gray background

image

I know it's not the focus of this task, but since it's moving towards V3, I could already highlight this at task.

PR #6024, opened 2022-12-06
Fix performance of report builder spec

matuszewskijan:
## Description

Hey, I was writing an article for my blog about optimizing test suite speed and have been looking at open-source projects. I've found out that one of the tests: https://github.com/chatwoot/chatwoot/blob/develop/spec/builders/v2/report_builder_spec.rb is extremely inefficient as it's recreating the test data for every example while we could reuse the data for all examples.

I am not sure if you like the introduction of `test-prof` gem, we could also use `before(:all)` and `after(:all)` to achieve similar results.

## Type of change

## How Has This Been Tested?

Running `bundle exec rspec spec/builders/v2/report_builder_spec.rb` now takes ~30 seconds instead of 3-4 minutes. The CI seems to execute 6-8 minutes faster.

## Checklist:


tejaswinichile (approved):
LGTM 👍🏼 

And there is a significant improvement in the test of the report.

One thought: Should we add a gem for one spec?

Issue #6092, opened 2022-12-18
Slack Integration is not working if there is a thumbnail for the contact or the user


pranavrajs:
Slack messages are not created when there is a thumbnail present for the contact or the user. 

The core reason for this is that the Rails Attachment URLs are very long, which Slack doesn't accept. It returns a 500 Internal server error if a URL of the following length is shared.

```
<URL>/rails/active_storage/disk/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaDdDVG9JYTJWNVNTSnJkbUZ5YVdGdWRITXZlVzkxYVdwamJIWnRhSHB1Y25oNmFuVnlhemt5YUcxMGRtTTFaaTgzWVdVMk1EWTBZMlZtT0RWaE1EazBOalU1Tmpaa01XRTVZVFJpTWprMk1EYzBPVEJsWmpsalpHWmtPV05qT1RZek1HUXdaRGt3Tm1WbVptWmxaR1ZpQmpvR1JWUTZFR1JwYzNCdmMybDBhVzl1U1NJL2FXNXNhVzVsT3lCbWFXeGxibUZ0WlQwaWNISmhibUYyTG1wd1p5STdJR1pwYkdWdVlXMWxLajFWVkVZdE9DY25jSEpoYm1GMkxtcHdad1k3QmxRNkVXTnZiblJsYm5SZmRIbHdaVWtpRDJsdFlXZGxMMnB3WldjR093WlVPaEZ6WlhKMmFXTmxYMjVoYldVNkNteHZZMkZzIiwiZXhwIjoiMjAyMi0xMi0xN1QyMzowNDo0MC4zNjNaIiwicHVyIjoiYmxvYl9rZXkifX0=--bc19c0544c9f95b9adf881086ca0f60c6e9d0437/<file-name>.jpg
```

**Todo:**

- [ ] Reduce the length of the URL.

sojan-official:
https://github.com/chatwoot/chatwoot/pull/6093

PR #6093, opened 2022-12-18
fix: Update Slack integration to fix message delivery issues

pranavrajs:
Fixes https://github.com/chatwoot/chatwoot/issues/6090

- Fixes message delivery issue with Slack.
- Added a temporary fix for https://github.com/chatwoot/chatwoot/issues/6092


lib/integrations/slack/incoming_message_builder.rb
```diff
@@ -22,13 +22,23 @@ def perform
...
-    supported_event_type? && supported_event?
+    supported_event_type? && supported_event? && should_process_event?
   end
 
   def supported_event_type?
     SUPPORTED_EVENT_TYPES.include?(params[:type])
   end
 
+  # Discard all the subtype of a message event
+  # We are only considering the acutal message sent by a Slack user
```
pranavrajs, line 33:
```suggestion
  # We are only considering the actual message sent by a Slack user
```

PR #5778, opened 2022-10-29
chore: feature flags for campaigns and website channel

smartdev58:
## Description

Added two more feature flags to disable/enable website inbox channel and campaign. Also updated platform accounts test specs to test the new flags params

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

https://www.loom.com/share/c27d0a0a05154c09b98f50e6f95c3db0

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #6096, opened 2022-12-19
chore: Cleanup feature flags

pranavrajs:
I made a mistake while merging this https://github.com/chatwoot/chatwoot/pull/5778 PR. I didn't realize that the order of the flags got changed. I made a fix on Cloud as soon as I found out by running an update script in the console. This PR would fix the feature flag updates in the self-hosted installations as well.

**Changes:** 
- Migration to enable all the features that were supposed to be enabled in an account in a self-hosted installation
- Add more feature flags for CRM, auto_resolution, and reports
- Add a SuperAdmin link in the sidebar if the user is a super-admin
- SuperAdmin would be able to view all the features on an account irrespective of whether the feature is enabled on the account.

Screenshot 2022-12-18 at 9 46 56 PM


db/migrate/20221219051529_fix_feature_flags.rb
```diff
@@ -0,0 +1,46 @@
...
+    Account.find_in_batches do |account_batch|
+      account_batch.each do |account|
+        if installation_config == 'cloud'
+          account.enable_features(
+            'reports',
+            'crm',
+            'auto_resolve_conversations'
+          )
+        else
+          account.enable_features(
```
sojan-official, line 15: Wouldn't this rewrite the configuration in self-hosted accounts where a user might have already saved some configurations to existing accounts?

PR #5157, opened 2022-08-01
feat: Read/Delivery status for Whatsapp Cloud API

clairton:
ref: #1021

# Read/Delivery whatsapp cloud api status

## Description

Process field statuses receives in webhook whatsapp cloud api

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Creating a spec

PR #6107, opened 2022-12-19
fix: Update the logic to show read status for web

pranavrajs:
With https://github.com/chatwoot/chatwoot/pull/5157, the read status in the website channel went missing. I have refactored the code to ensure the status works on both the website and WhatsApp channels.

### DIFF
--- a/app/javascript/dashboard/components/widgets/conversation/bubble/Actions.vue
+++ b/app/javascript/dashboard/components/widgets/conversation/bubble/Actions.vue
@@ -1,8 +1,14 @@
 <template>
   <div class="message-text--metadata">
-    <span class="time" :class="{ delivered: messageRead }">{{
-      readableTime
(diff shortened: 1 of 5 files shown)


app/javascript/dashboard/components/widgets/conversation/bubble/Actions.vue
```diff
@@ -174,28 +185,52 @@ export default {
...
-      );
+      if (!this.showStatusIndicators) {
+        return false;
+      }
+
+      if (this.isAnEmailChannel) {
+        return !!this.sourceId;
+      }
+
+      if (this.isAWhatsAppChannel) {
```
nithindavid, line 203: Shouldn't we add a case web widget to show sent markers?

nithindavid (approved):

PR #6125, opened 2022-12-22
fix: whatsapp cloud to consider multiple attachments

### LINKED ISSUE #6120: Failure to deliver multiple images per message (WhatsApp Cloud API)
**Describe the bug**

Chatwoot fails to deliver multiple images in one messages on Whatsapp (via WhatsApp Cloud API). Instead, only one of the images will be sent. This is particularly bad, because there is no indication of this happening on the sender side.

On the sender side:
image

Receiver:
image

**To Reproduce**

1. Attach multiple images to one messages
2. Send it through the Web Interface
3. Check on the receiving WA account

**Environment**

chatwoot 2.12 (was also present in 2.11)

- [x] Self-hosted
- [x] Linux VM

tejaswinichile:
Fixes #6120


app/services/whatsapp/providers/whatsapp_cloud_service.rb
```diff
@@ -66,14 +69,20 @@ def send_text_message(phone_number, message)
...
-    attachment = message.attachments.first
-    type = %w[image audio video].include?(attachment.file_type) ? attachment.file_type : 'document'
-    type_content = {
-      'link': attachment.download_url
-    }
-    type_content['caption'] = message.content unless %w[audio sticker].include?(type)
+  def process_attachments(phone_number, message)
+    responses = []
+    message.attachments.each do |attachment|
+      responses << send_attachment_message(phone_number, message, attachment)
```
scmmishra, line 75: Should we wrap this in a `begin`-`rescue` block? Or is it handled somewhere in the above levels?

Also, since we're not using the `responses` in the `send_message` method, does it make sense to return it in the first place?

tejaswinichile: We can rescue the individual attachment-sending functionality. Not doing it here, just to tracking the standard exceptions.

tejaswinichile:
@sojan-official 

> Whatsapp API only supports sending one attachment at a time. So let's limit the ability to create multiple attachments on a single message for WhatsApp Channel APIs.

Does it mean showing some sort of error message?

nithindavid:
Closing in favour of https://github.com/chatwoot/chatwoot/pull/6428

Issue #5332, opened 2022-08-24
Use expanded layout as the default layout in the mobile view


voko3456:
hi after i test switch layout on mobile browser the view become more responsive can you make the switch layout as default layout on mobile browser

also there's small bug : when switch layout I'm not bale to scroll down  

![image](https://user-images.githubusercontent.com/20530550/186400348-c47f58db-6af6-415a-9cad-a5cbf36a8d0c.png)

BhavyaT-135:
Hi @pranavrajs, can I take up this issue if no one else is working on it? 😊

pranavrajs:
@BhavyaT-135 Sure, please go ahead.

BhavyaT-135:
Hi @pranavrajs, would changing the default value of `isOnExpandedLayout` props, in the `SwitchLayout` file, to `true` be a solution to this issue? Sorry for the late response, had been occupied with college 😅

pranavrajs:
@BhavyaT-135 It requires a little more work. Assigning this to @iamsivin for now, as it needs a fix soon. Let me know if you are interested, we can find another issue for you.

BhavyaT-135:
Hey @pranavrajs, thanks for asking. I'll go through the documentation and look for an issue that I could contribute too. 😃

PR #6134, opened 2022-12-26
feat: Uses expanded layout as default if the breakpoint is less than or equal to 992px

iamsivin:
# Pull Request Template

## Description

1. If the display breakpoint is less than or equal to 992px, then the conversation screen default layout will be expanded.
2. if the display breakpoint is 640px, then the contact sidebar will take full width instead of showing it on the right side.
3. if the display breakpoint is 640px, the chat header will change to column view and align it to the center.

Fixes https://github.com/chatwoot/chatwoot/issues/5332

## Type of change

## How Has This Been Tested?
https://www.loom.com/share/432d73c7c68345d1a90b347c6213f61c

## Checklist:


app/javascript/dashboard/components/widgets/conversation/ConversationBox.vue
```diff
@@ -166,11 +166,15 @@ export default {
 .conversation-sidebar-wrap {
   height: auto;
   flex: 0 0;
-  overflow: hidden;
+  z-index: var(--z-index-low);
   overflow: auto;
   background: white;
   flex-basis: 28rem;
 
+  @include breakpoint(768px down) {
```
nithindavid, line 174: Please use existing breakpoints mentioned foundation-settings.scss

nithindavid: `@include breakpoint(large up) {`

app/javascript/dashboard/components/widgets/conversation/ConversationHeader.vue
```diff
@@ -165,14 +167,26 @@ export default {
...
+}
+
+.conversation-header--details {
+  display: flex;
+  justify-content: center;
+  flex-direction: row;
+  align-items: center;
+  width: 100%;
+
+  @include breakpoint(640px down) {
```
nithindavid, line 187: Use pre defined breakpoints or create new ones in foundation settings `@include breakpoint(large up) {`

app/javascript/dashboard/routes/dashboard/Dashboard.vue
```diff
@@ -77,14 +79,40 @@ export default {
...
+    this.handleResize();
     bus.$on(BUS_EVENTS.TOGGLE_SIDEMENU, this.toggleSidebar);
   },
   beforeDestroy() {
     bus.$off(BUS_EVENTS.TOGGLE_SIDEMENU, this.toggleSidebar);
+    window.removeEventListener('resize', this.handleResize);
   },
   methods: {
+    handleResize() {
+      if (window.innerWidth <= 992) {
```
nithindavid, line 110: Create a constant files for breakpoints.

nithindavid (changes requested):

iamsivin:
I tested these changes on mobile.

https://user-images.githubusercontent.com/64252451/210327486-e9d33eba-f7a7-4ab8-b5db-bf1aefa08697.MP4

app/javascript/dashboard/components/ChatList.vue
```diff
@@ -794,16 +794,16 @@ export default {
...
   }
   @include breakpoint(xlarge up) {
-    width: 35rem;
+    flex-basis: 38rem;
   }
   @include breakpoint(xxlarge up) {
-    width: 38rem;
+    flex-basis: 42rem;
   }
   @include breakpoint(xxxlarge up) {
```
scmmishra, line 808: Curious question? Are these values standardized or documented somewhere, if not we can use a formula-based width for this using a combination of calc and clamp.

```css
.conversations-list-wrap {
    flex-shrink: 0;
    /* Where 1vw is 1% of the viewport */
    flex-basis: clamp(34rem, 1vw + 34rem, 42rem);
}
``` 

Take this suggestion with a pinch of salt, **this would make the width of the list fluid**, and it can break in other places where we depend on the predictable and fixed widths of the list. The main benefit this adds is reducing the code footprint of the CSS we write

P.S. CSS `clamp` is widely supported, [ref](https://caniuse.com/css-math-functions)

app/javascript/dashboard/components/widgets/conversation/ConversationHeader.vue
```diff
@@ -1,53 +1,55 @@
...
-            @click="$emit('contact-panel-toggle')"
-          >
-            {{ contactPanelToggleText }}
-          </woot-button>
+    <div class="conversation-header--details">
+      <div class="user">
+        <back-button v-if="showBackButton" :back-url="backButtonUrl" />
+        <Thumbnail
+          :src="currentContact.thumbnail"
+          size="40px"
```
scmmishra, line 8:
```suggestion
```

`Thumbnail` already has a default size of `40px`. 

On another note, not actionable for this PR. 
Passing value for CSS styles as a prop is generally considered an anti-pattern, given one cannot reliably verify the value, WDYT?

CC: @fayazara

app/javascript/dashboard/routes/dashboard/Dashboard.vue
```diff
@@ -77,14 +79,41 @@ export default {
...
+        conversation_display_type: !this.isDesktopView
+          ? LAYOUT_TYPES.EXPANDED
+          : this.previouslyUsedDisplayType,
+      });
+    },
   },
   mounted() {
+    this.handleResize();
+    window.addEventListener('resize', this.handleResize);
     bus.$on(BUS_EVENTS.TOGGLE_SIDEMENU, this.toggleSidebar);
```
scmmishra, line 102: There should be a CSS-only way of doing this, I see we're hiding some elements, we can use `min-inline-size` to trigger these I guess (need to verify). Do we have transitions in the UI? Having an active listener can cause some performance issues for lower-end laptops and tablets. 

If a CSS-only way is not possible, we can perhaps throttle the event and see, it can potentially make the UI a bit janky though. If it does we can let the current changes stay as is, else we can consider throttling. 

Here's a possible way to throttle it

```js
let throttled = false,
const delay = 150 // 150ms
window.addEventListener('resize', function() {
  if (!throttled) {
    this.toggleSidebar();
    throttled = true;
    setTimeout(function() {
      throttled = false;
    }, delay);
  }
});
```

iamsivin: @scmmishra I think we couldn't do it with CSS (correct me if I was wrong) because when the display breakpoint <= 992px, it will update the conversation display layout as 'expanded' to UISettings. We don't have transitions in UI. I will try it the other way that you mentioned. 
CC: @nithindavid

iamsivin: Done https://github.com/chatwoot/chatwoot/pull/6134/commits/f84b05b16cdc66f1241555c71f00c4f555e9301e

app/javascript/dashboard/components/ChatList.vue
iamsivin: Done https://github.com/chatwoot/chatwoot/pull/6134/commits/f84b05b16cdc66f1241555c71f00c4f555e9301e

nithindavid (approved):

app/javascript/dashboard/routes/dashboard/Dashboard.vue
scmmishra: This won't remove the listener, since the function signature won't match with the one given originally, we can add the throttle in the handle resize itself

Another option is to write a throttle handler, it will return a throttled function

```js
function throttle(fn, delay) {
  let throttled = false;
  return function (...args) {
    if (throttled) {
      return;
    }
    throttled = true

    setTimeout(() => {
      throttled = false;
      fn(...args);
    }, delay);
  };
}
```

But for now, maybe we can just add the throttle to the `handleResize` itself. Upto you, on what seems better

app/javascript/dashboard/routes/dashboard/Dashboard.vue
```diff
@@ -77,14 +79,51 @@ export default {
...
+        throttled = true;
+        setTimeout(() => {
+          throttled = false;
+        }, delay);
+      }
+    });
     bus.$on(BUS_EVENTS.TOGGLE_SIDEMENU, this.toggleSidebar);
   },
   beforeDestroy() {
+    window.removeEventListener('resize', this.handleResize);
```
scmmishra, line 115: This won't remove the listener, since the function signature won't match with the one given originally, we can add the throttle in the handle resize itself

Another option is to write a throttle handler, it will return a throttled function

```js
function throttle(fn, delay) {
  let throttled = false;
  return function (...args) {
    if (throttled) {
      return;
    }
    throttled = true

    setTimeout(() => {
      throttled = false;
      fn(...args);
    }, delay);
  };
}
```

But for now, maybe we can just add the throttle to the `handleResize` itself. Upto you, on what seems better

scmmishra (changes requested):

app/javascript/dashboard/routes/dashboard/Dashboard.vue
iamsivin: I pushed the changes by this commit https://github.com/chatwoot/chatwoot/pull/6134/commits/a7664eaa0959e89673f92f67b7930c8bac334920 Can you check it?

nithindavid:
Tested locally and found fix the issues mentioned in the PR.

The scrolling issue mentioned in the Original issue is fixed too (or not reproducible )
Found a minor issue - The sidebar expanded state is not shared across other parts of app like portals. This might be fixed by https://github.com/chatwoot/chatwoot/pull/6191

image


We could hide the view toggle switch for tablets/small devices as it wouldn't make sense to use the other view. We could do this as a separate PR

image

nithindavid (approved):

scmmishra (approved):

Issue #5938, opened 2022-11-24
Production Deployment Error


pranavrajs:
Due to one of these 4 commits the production deployment on Heroku is broken 

86ca7f4a8d5e2b7d45033e85d5e100870ed0bfa0, 8b659de73d3813489489b249a9b3f8e95dbd2999, f2753df8df424b5b9137cad0da2e3f6da7ad2dda, c121b44df419696cd44601468e36064c680356d9

`REDIS_OPENSSL_VERIFY_MODE` is not working with these changes.

### Error message
SSL_connect returned=1 errno=0 state=error: certificate verify failed (self-signed certificate in certificate chain)

```
/app/vendor/bundle/ruby/3.0.0/gems/redis-4.7.1/lib/redis/connection/ruby.rb:264:in `connect_nonblock'
/app/vendor/bundle/ruby/3.0.0/gems/redis-4.7.1/lib/redis/connection/ruby.rb:264:in `connect'
/app/vendor/bundle/ruby/3.0.0/gems/redis-4.7.1/lib/redis/connection/ruby.rb:306:in `connect'
/app/vendor/bundle/ruby/3.0.0/gems/redis-4.7.1/lib/redis/client.rb:385:in `establish_connection'
/app/vendor/bundle/ruby/3.0.0/gems/redis-4.7.1/lib/redis/client.rb:115:in `block in connect'
/app/vendor/bundle/ruby/3.0.0/gems/redis-4.7.1/lib/redis/client.rb:344:in `with_reconnect'
/app/vendor/bundle/ruby/3.0.0/gems/redis-4.7.1/lib/redis/client.rb:114:in `connect'
/app/vendor/bundle/ruby/3.0.0/gems/newrelic_rpm-8.9.0/lib/new_relic/agent/instrumentation/redis/prepend.rb:20:in `block in connect'
/app/vendor/bundle/ruby/3.0.0/gems/newrelic_rpm-8.9.0/lib/new_relic/agent/instrumentation/redis/instrumentation.rb:30:in `block in connect_with_tracing'
/app/vendor/bundle/ruby/3.0.0/gems/newrelic_rpm-8.9.0/lib/new_relic/agent/instrumentation/redis/instrumentation.rb:45:in `block in with_tracing'
/app/vendor/bundle/ruby/3.0.0/gems/newrelic_rpm-8.9.0/lib/new_relic/agent/tracer.rb:351:in `capture_segment_error'
[...]

PR #6182, opened 2023-01-05
Fix: Errors in Heroku deployments

sojan-official:
Due to https://github.com/chatwoot/chatwoot/commit/86ca7f4a8d5e2b7d45033e85d5e100870ed0bfa0 the Redis configuration in some Heroku deployments was breaking, temporarily reverting the part until we identify a fix.

fixes: #5938

### DIFF
--- a/config/initializers/01_redis.rb
+++ b/config/initializers/01_redis.rb
@@ -1,16 +1,14 @@
-Rails.application.reloader.to_prepare do
-  # Alfred
-  # Add here as you use it for more features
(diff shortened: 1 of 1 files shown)


sojan-official:
# Testing conditions 

- Enable Heroku Redis premium plans 
- Configure environment variable # REDIS_OPENSSL_VERIFY_MODE=none 

The current changes fix the error. Before this commit, deployment in Heroku will fail when Heroku Redis is upgraded to premium.

PR #6191, opened 2023-01-06
fix: default article sidebar state

scmmishra:
# Pull Request Template

## Description

The collapse sidebar inside the articles page editor was not corresponding to the actual state on load. The PR fixes it

https://user-images.githubusercontent.com/18097732/211002350-776f060c-fa2f-40e3-9943-43fb9b224b5c.mov

#### After fix

https://user-images.githubusercontent.com/18097732/211002086-4bef65b0-55da-469b-87e3-492b2104324e.mov

## Type of change

## How Has This Been Tested?

This has been tested locally, doesn't seem to require staging testing

## Checklist:


scmmishra:
This is not the best solution for this, here the state is being "declared" in two separate components. Ideally, the context of the state should be passed from the parent instead of the component assuming it. @pranavrajs @fayazara WDYT? I can make those changes in this PR

pranavrajs:
@scmmishra Sure, this makes sense. I think this should be a prop. Let us make that change.

scmmishra:
> @scmmishra Sure, this makes sense. I think this should be a prop. Let us make that change.

@pranavrajs made the suggested changes, @nithindavid PTAL

pranavrajs (approved):

PR #6207, opened 2023-01-09
feat: Add video call option in the dashboard

pranavrajs:
- Added a video call option in the reply box 
- Added an integration-type message


app/javascript/dashboard/components/widgets/WootWriter/ReplyBottomPanel.vue
```diff
@@ -269,7 +277,7 @@ export default {
       }
     },
     showMessageSignatureButton() {
-      return !this.isPrivate && this.isAnEmailChannel;
+      return !this.isOnPrivateNote && this.isAnEmailChannel;
```
pranavrajs, line 280: There is no variable `isPrivate` defined in this component.

app/javascript/dashboard/components/widgets/WootWriter/ReplyBottomPanel.vue
```diff
@@ -87,6 +87,10 @@
         :title="'Whatsapp Templates'"
         @click="$emit('selectWhatsappTemplate')"
       />
+      <video-call-button
```
pranavrajs, line 90: Load only for website inbox and API inbox. We need to get clarity on whether we can send a link to other channels. Until then it is disabled.

app/javascript/dashboard/components/widgets/conversation/Message.vue
```diff
@@ -274,6 +285,9 @@ export default {
     isTemplate() {
       return this.data.message_type === MESSAGE_TYPE.TEMPLATE;
     },
+    isAnIntegrationMessage() {
+      return this.contentType === 'integrations';
```
pranavrajs, line 289: contentType constants are not defined in the frontend code. We can pick it up as a chore.

app/javascript/dashboard/components/widgets/conversation/bubble/integrations/Dyte.vue
```diff
@@ -0,0 +1,102 @@
...
+      :is-loading="isLoading"
+      @click="joinTheCall"
+    >
+      {{ $t('INTEGRATION_SETTINGS.DYTE.CLICK_HERE_TO_JOIN') }}
+    </woot-button>
+    <div v-if="dyteAuthToken" class="video-call--container">
+      <iframe
+        :src="meetingLink"
+        allow="camera;microphone;fullscreen;display-capture;picture-in-picture;clipboard-write;"
+      />
```
pranavrajs, line 18: The integration is loaded via the iFrame option they provide so that we are not embedding the SDKs unnecessarily. This option needs to be revisited.

app/javascript/dashboard/components/widgets/conversation/bubble/integrations/Dyte.vue
```diff
@@ -0,0 +1,102 @@
...
+      return `${DYTE_MEETING_LINK}${this.meetingData.room_name}?authToken=${this.dyteAuthToken}&showSetupScreen=true&disableVideoBackground=true`;
+    },
+  },
+  methods: {
+    async joinTheCall() {
+      this.isLoading = true;
+      try {
+        const {
+          data: { authResponse: { authToken } = {} } = {},
+        } = await DyteAPI.addParticipantToMeeting(this.messageId);
```
pranavrajs, line 61: I didn't add a key in the store since the lifecycle of the token or the meeting id below is only till the component is destroyed.

app/javascript/dashboard/components/widgets/conversation/bubble/integrations/Dyte.vue
```diff
@@ -0,0 +1,102 @@
...
+  iframe {
+    width: 100%;
+    height: 100%;
+    border: 0;
+  }
+
+  button {
+    position: absolute;
+    top: var(--space-smaller);
+    right: 16rem;
```
pranavrajs, line 99: We don't have a very large value in `spacing.scss`, hence used a custom value here.

app/javascript/dashboard/api/integrations/dyte.js
```diff
@@ -0,0 +1,23 @@
...
+
+import ApiClient from '../ApiClient';
+
+class DyteAPI extends ApiClient {
+  constructor() {
+    super('integrations/dyte', { accountScoped: true });
+  }
+
+  createAMeeting(conversationId) {
+    return axios.post(`${this.url}/create_a_meeting`, {
```
scmmishra, line 11: Just curious, how is `axios` globally available? Is it a rails thing? Or we just attach axios to the window?

fayazara: I think it comes from this comment - `/* global axios */`

pranavrajs: Attached to the window object.

nithindavid: Yes, it's a global window variable. Is there a better way to do this?
image

app/javascript/dashboard/components/widgets/conversation/bubble/integrations/Dyte.vue
```diff
@@ -0,0 +1,102 @@
...
+  margin: var(--space-small) 0;
+}
+
+.video-call--container {
+  position: fixed;
+  bottom: 0;
+  right: 0;
+  width: 100%;
+  height: 100%;
+  z-index: 10000;
```
nithindavid, line 86:
```suggestion
  z-index: var(  --z-index-high);
```

app/javascript/dashboard/api/integrations/dyte.js
pranavrajs: @nithindavid The API client should have taken the Axios object within it instead of making it a global variable. Essentially we are using axios in a similar way fetch works. 

This was on my list for refactoring, I couldn't get enough time to redo all the APIClients.

app/javascript/dashboard/components/widgets/conversation/bubble/integrations/Dyte.vue
```diff
@@ -0,0 +1,102 @@
...
+      >
+        {{ $t('INTEGRATION_SETTINGS.DYTE.LEAVE_THE_ROOM') }}
+      </woot-button>
+    </div>
+  </div>
+</template>
+<script>
+import DyteAPI from 'dashboard/api/integrations/dyte';
+
+const DYTE_MEETING_LINK = 'https://app.dyte.in/meeting/stage/';
```
scmmishra, line 34: Should we have this in an `env`, with a fallback to the `app.dyte.in` domain? 
It makes sense to do this only if it is a possibility that dyte can be hosted on a separate domain? Also given their website is `dyte.io` and API is `dyte.in` there is a possibility the domains may change in the future

app/javascript/dashboard/components/widgets/conversation/bubble/integrations/Dyte.vue
```diff
@@ -0,0 +1,102 @@
...
+      type: Object,
+      default: () => ({}),
+    },
+  },
+  data() {
+    return { isLoading: false, dyteAuthToken: '', isSDKMounted: false };
+  },
+  computed: {
+    meetingLink() {
+      return `${DYTE_MEETING_LINK}${this.meetingData.room_name}?authToken=${this.dyteAuthToken}&showSetupScreen=true&disableVideoBackground=true`;
```
scmmishra, line 52:
```suggestion
      return this.dyteAuthToken ? `${DYTE_MEETING_LINK}${this.meetingData.room_name}?authToken=${this.dyteAuthToken}&showSetupScreen=true&disableVideoBackground=true` : '';
```

This is more defensive, and ensures that the `meetingLink` is a proper formed URL, it's a minor nit-pick, feel free to ignore it

app/javascript/dashboard/components/widgets/conversation/bubble/integrations/Dyte.vue
```diff
@@ -0,0 +1,102 @@
...
+  methods: {
+    async joinTheCall() {
+      this.isLoading = true;
+      try {
+        const {
+          data: { authResponse: { authToken } = {} } = {},
+        } = await DyteAPI.addParticipantToMeeting(this.messageId);
+        this.dyteAuthToken = authToken;
+      } catch (err) {
+        // Ignore the error
```
scmmishra, line 64: Should we add a toast? Or perhaps log this to Sentry?

app/javascript/dashboard/components/widgets/WootWriter/ReplyBottomPanel.vue
```diff
@@ -124,13 +128,13 @@ import {
   ALLOWED_FILE_TYPES,
   ALLOWED_FILE_TYPES_FOR_TWILIO_WHATSAPP,
 } from 'shared/constants/messages';
-
+import VideoCallButton from '../VideoCallButton.vue';
```
iamsivin, line 131:
```suggestion
import VideoCallButton from '../VideoCallButton';
```

scmmishra (commented):
A few minor suggestions and comments, the rest looks good to me!

app/javascript/dashboard/components/widgets/conversation/bubble/integrations/Dyte.vue
pranavrajs: This is changed in the subsequent PR. Also, the link is used only if there is an authToken present. 

I understand that this would be computed, but that shouldn’t be an issue.

app/javascript/dashboard/api/integrations/dyte.js
scmmishra: > The API client should have taken the Axios object within it instead of making it a global variable.

This is ideal

> I couldn't get enough time to redo all the APIClients.

We can start by having axios proxied within the client and then slowly refactoring others, I can pick this up if you think it's worth doing

app/javascript/dashboard/components/widgets/conversation/bubble/integrations/Dyte.vue
pranavrajs: I don’t think so.

pranavrajs: I have not found the self-hosted option in their docs. If it is there, I can pick it up as an enhancement.

app/javascript/dashboard/components/widgets/VideoCallButton.vue
```diff
@@ -0,0 +1,55 @@
...
+      this.$store.dispatch('integrations/get');
+    }
+  },
+  methods: {
+    async onClick() {
+      this.isLoading = true;
+      try {
+        await DyteAPI.createAMeeting(this.conversationId);
+      } catch (error) {
+        // Ignore Error
```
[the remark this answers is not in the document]
pranavrajs: Toast would be a nice idea, however I’m yet to find the possible cases where the API call would break. I can add the toasts as an improvement.

nithindavid:
Tested without live chat code pulled in. https://www.loom.com/share/32403808639243c190cee196582212c7

**Minor Suggestion** It Would be nice to show a call summary as private note/message/activity once call is ended. We could do as an improvement.

app/javascript/dashboard/components/widgets/WootWriter/ReplyBottomPanel.vue
```diff
@@ -87,6 +87,10 @@
         :title="'Whatsapp Templates'"
         @click="$emit('selectWhatsappTemplate')"
       />
+      <video-call-button
+        v-if="(isAWebWidgetInbox || isAPIInbox) && !isOnPrivateNote"
```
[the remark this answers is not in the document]
pranavrajs: This was done to separate the logic. The component is only rendered if the inbox is website or APi inbox and the editor is not on private note.

The video integration requires a check on the connected integration. Adding these logic to the parent would make it bloated. 

If it is required I can move these checks to child component.

pranavrajs:
> Minor Suggestion It Would be nice to show a call summary as private note/message/activity once call is ended. We could do as an improvement.

I wanted to do this, but there is no option in the iFrame SDK to do this.

nithindavid (approved):
Please add the toast suggested by Shivam if possible.

iamsivin (approved):
🔥

app/javascript/dashboard/components/widgets/WootWriter/ReplyBottomPanel.vue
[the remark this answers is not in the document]
scmmishra: > The video integration requires a check on the connected integration. Adding these logic to the parent would make it bloated.

This makes sense

> If it is required I can move these checks to child component.

Nopes, I wouldn't say it's critical as of now. I'll resolve the conversation

app/javascript/dashboard/components/widgets/VideoCallButton.vue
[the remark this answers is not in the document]
pranavrajs: Added the toast.

app/javascript/dashboard/components/widgets/conversation/bubble/integrations/Dyte.vue
pranavrajs: Added the toast.

PR #6208, opened 2023-01-09
feat: Add video call support in live-chat widget

pranavrajs:
- Added a controller for adding the participant to the meeting
- Added the components for rendering the video call on the widget

**How to test:** https://www.loom.com/share/e0ce77953bb24ef0a97e8d0dd76f8785


app/javascript/widget/components/template/IntegrationCard.vue
```diff
@@ -0,0 +1,118 @@
+<template>
```
pranavrajs, line 1: This component is repeated and is similar to the dashboard component, however, the issues with the styles and missing shared components make it hard to build a common component.

app/javascript/widget/components/template/IntegrationCard.vue
```diff
@@ -0,0 +1,118 @@
...
+    },
+  },
+};
+</script>
+<style lang="scss" scoped>
+@import '~widget/assets/scss/variables.scss';
+
+.video-call--container {
+  position: fixed;
+  top: 72px;
```
pranavrajs, line 91: Defined `Space variables` in the widget doesn't match the header size.

app/javascript/sdk/IFrameHelper.js
```diff
@@ -56,7 +56,8 @@ export const IFrameHelper = {
       widgetUrl = `${widgetUrl}&cw_conversation=${cwCookie}`;
     }
     iframe.src = widgetUrl;
-
+    iframe.allow =
```
pranavrajs, line 59: Added this to make sure that the website LiveChat widget asks for permissions when the SDK is loaded.

app/javascript/widget/components/template/IntegrationCard.vue
pranavrajs: This piece suggests that there is an absolute need for standardization between the dashboard and widget components.

app/javascript/widget/components/template/IntegrationCard.vue
```diff
@@ -0,0 +1,118 @@
...
+  <div>
+    <button
+      class="button join-call-button"
+      color-scheme="secondary"
+      :is-loading="isLoading"
+      :style="{
+        background: widgetColor,
+        borderColor: widgetColor,
+        color: textColor,
+      }"
```
pranavrajs, line 11: This is how it is done in most places where we had to customize the colors in the widget based on the user-selected widget brand color. With the latest tailwind, I believe we could do `text-${textColor}` and `bg-${widgetColor}`. However, I don't think this works with the version we are using.

fayazara (approved):
Tested locally, works as expected. 

Tested on Chrome, FF and Safari - no issues.
Tested on ios safari, and chrome android - no issues.

Few questions.
1. There are two leave call options currently, Dyte's and Chatwoots, maybe we can listen to Dyte's call end and listen then close the Dyte component.
2. Can we show an activity message "The call was ended by xyz" from either parties

The rest looks good 🔥

pranavrajs:
> There are two leave call options currently, Dyte's and Chatwoots, maybe we can listen to Dyte's call end and listen then close the Dyte component. Can we show an activity message "The call was ended by xyz" from either parties? 

This is an enhancement I wanted to add. The current iFrame implementation and postMessages seems to be broken from Dyte's end. I have been in conversation with them. If they fix it, we could add an event and an activity message. 

From what I understand, there doesn't seem to be a concept of "ending the meeting". When the user joins the meeting again after everyone leaves, it will start the meeting timer again. It falls on our end to define how to end the meeting. We need some trigger for the agent to cancel the meeting after it is completed. We can pick up this as an enhancement ticket once the v1 is merged.

app/javascript/widget/components/template/IntegrationCard.vue
scmmishra: > I don't think this works with the version we are using.

Yeah it does not, the processor is static, the hacky way to do this is having all possible combinations of this in a comment some where

> With the latest tailwind, I believe we could do text-${textColor} and bg-${widgetColor}

Custom properties are allowed like "bg-[#ff22ee]" but the processing is still static in nature, so having JS templates won't work

scmmishra: In this case, the computed methods can simply generate the CSS class names, which can then be attached to the `class` property, instead of being attached to `style`

In case we need to use something like `text-${textColor}` we can add it to the `allowList` / `whiteList` in the purgeCSS options in the tailwind config

app/javascript/widget/components/template/IntegrationCard.vue
```diff
@@ -0,0 +1,118 @@
...
+    width: 100%;
+    height: calc(100% - 72px);
+
+    border: 0;
+  }
+}
+
+.join-call-button {
+  margin: $space-small 0;
+  border-radius: 4px;
```
scmmishra, line 108: Do we have a variable for this?

scmmishra (approved):
LGTM! 🚀

app/javascript/widget/components/template/IntegrationCard.vue
iamsivin: @scmmishra I think we don't have a variable for `border-radius` on the widget. We have it on the dashboard side.

sojan-official (approved):

nithindavid (approved):

iamsivin (approved):
💯

app/javascript/widget/components/template/IntegrationCard.vue
pranavrajs: I'm not sure if I understand. We cannot add an allow list of the classes as the color is selected by the user, we don't have control over what color they would choose.

PR #6216, opened 2023-01-10
feat: Use rich text editor for automation and macros message action

fayazara:
Uses rich message editor instead of native textarea for message actions in automations and macros editors
Screenshot2023-01-10 at 14 14 41@2x
Screenshot2023-01-10 at 14 14 19@2x


nithindavid (approved):

iamsivin (approved):
LGTM 💯
https://www.loom.com/share/5b622e4876ce45308837dca33c37eaa5

PR #6219, opened 2023-01-10
fix: Fixes inconsistent spacing in message bubble

nithindavid:
## Description
This PR will fix the space issue with the timestamp in every bubble. It's showing an extra margin, which looks inconsistent.

**Before**
image

**After**
image


iamsivin (approved):
LGTM

PR #6222, opened 2023-01-10
fix: Duplicate conversations and contacts WA and Brazil numbers

### LINKED ISSUE #5840: Duplicate conversations and contacts (WA and Brazil numbers)
**Describe the bug**

Hello. Here in Brazil, we had a change on mobile numbers a few years ago, they've added an extra leading digit to account the total numbers growth over the last years. The general change was to add a leading 9 over all current numbers. Most WA users have their accounts registered to the old number, without the leading digit. When we start a conversation with Chatwoot, we add the correct number, with the leading digit but when the user responds, Chatwoot creates both a new contact and a new conversation ticket with the registered, old number, without the leading number. This is a major problem, because it breaks the ticket flow and duplicates contacts.

**To Reproduce**

Create a contact with the "new", correct number, then start a conversation, wait for them to respond, check conversations and contacts.

**Expected behavior**

Contacts and conversations should not duplicate.

**Environment**

I'm currently using Chatwoot cloud.

jacsonsantospht:
## Description

Resolves issue when receiving message from Brazil number.
Fixes #5840 

## Type of change

## How Has This Been Tested?

1 - Create inbox type WhatsApp
2 - Open Postman to test
3 - Send a message to WhatsApp inbox with number using 9 or without 9

Test without number 9 
Captura de tela de 2023-01-10 09-55-43

Test with number 9
Captura de tela de 2023-01-10 09-55-16

## Checklist:


fgrep:
Will this commit force a number 9 to all contacts?

If so, it will break whatsapp for landline and toll free numbers.

jacsonsantospht:
> 
@fgrep 
This commit does not force a 9 for all Brazilian numbers. Just check if there is already a contact with 9 or without 9 previously registered. Avoiding duplicity.

Example:
- This number already exists in contacts (note that it does not have the 9 at the beginning): +55 41 8799-XXXX.
- When you receive a message from this number (note that there is a 9) +55 41 9 8799-XXXX

Chatwoot will not register a new contact, because there is already a contact without the number 9 in contacts.

sojan-official:
@jacsonsantospht  I have made some changes to this PR. Could you ensure that the assumptions I have made over here are correct and it covers the basic cases? 

> Note: When looking up contacts connected to incoming messages from external channels, We rely on the contact inboxes table and the source Id instead of doing a direct look-up into the Contacts table
> ref: https://github.com/chatwoot/chatwoot/wiki/Building-on-Top-of-Chatwoot:-Importing-Existing-Contacts-and-Creating-Conversations

## Case: Incoming WhatsApp message  with a `wa_id` in a new format 
>  (13 digits `country code-DDD-9-number`)
-  No changes in behaviour

## Case: Incoming WhatsApp message with a `wa_id` in the old format 
> ( 12 digits `country code-DDD-number`)
- We check whether a contact inbox exists with the new format; if it does, we use the new format as the `wa_id` and while processing as the `source_id`
- If the contact inbox doesn't exist. We use the original 12-digit wa_id for the wa_id as well as the source id

--------------------

This covers the case described in #5840 and, in other cases, falls back to the default behaviour. 

- Messages from new contacts with the `old phone number` format will create contact with their phone number in `old format` in Chatwoot. 
- The agent can update the phone number to `new format` from the UI. even in this case, the subsequent messages will correctly get appended to the same contact since the underlying `source_id` of the contact inbox wouldn't change.

sojan-official:
@ASchneiderBR  @glauberssoutodossantos  Would love to hear your thoughts as well.

PR #6227, opened 2023-01-10
Feat: Support for Microsoft Oauth in Email Channel

tejaswinichile:
https://www.loom.com/share/e930184057d549da813f97bc66fff390


app/controllers/microsoft/callbacks_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+      {
+        site: 'https://login.microsoftonline.com',
+        authorize_url: 'https://login.microsoftonline.com/common/oauth2/v2.0/authorize',
+        token_url: 'https://login.microsoftonline.com/common/oauth2/v2.0/token'
+      }
+    )
+
+    @response = client.auth_code.get_token(
+      oauth_code,
+      redirect_uri: "#{base_utl}/microsoft/callback"
```
pranavrajs, line 17:
```suggestion
      redirect_uri: "#{base_url}/microsoft/callback"
```

app/controllers/microsoft/callbacks_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+    )
+
+    ActiveRecord::Base.transaction do
+      inbox = create_inbox
+      ::Redis::Alfred.delete(permitted_params['current_account_id'])
+      # ::microsoft::WebhookSubscribeService.new(inbox_id: inbox.id).perform
+      redirect_to app_twitter_inbox_agents_url(account_id: account.id, inbox_id: inbox.id)
```
pranavrajs, line 24: Let us rename this path.

Gemfile.lock
```diff
@@ -249,6 +249,8 @@ GEM
     gli (2.21.0)
     globalid (1.0.0)
       activesupport (>= 5.0)
+    gmail_xoauth (0.4.2)
```
pranavrajs, line 252: @tejaswinichile net::imap latest versions support xoauth automatically. We can upgrade the package. No need to do it in this PR. We can take it  as a separate task.

pranavrajs (approved):

pranavrajs (changes requested):
Moving to "request changes" status so that we don't accidently merge this PR.

.env.example
```diff
@@ -213,3 +213,8 @@ STRIPE_WEBHOOK_SECRET=
 # Set to true if you want to upload files to cloud storage using the signed url
 # Make sure to follow https://edgeguides.rubyonrails.org/active_storage_overview.html#cross-origin-resource-sharing-cors-configuration on the cloud storage after setting this to true.
 DIRECT_UPLOADS_ENABLED=
+
+#MS OAUTH creds
+AZURE_APP_ID=
+AZURE_APP_SECRET=
+AZURE_SCOPES = 'Mail.Send offline_access https://graph.microsoft.com/IMAP.AccessAsUser.All https://graph.microsoft.com/SMTP.Send'
```
pranavrajs, line 220: @tejaswinichile Why do we need to take this as an env variable?

app/controllers/api/v1/accounts/microsoft/authorizations_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+                                                            https://outlook.office.com/SMTP.Send openid',
+                                                              prompt: 'consent'
+                                                            })
+    if redirect_url
+      ::Redis::Alfred.setex('current_account_id', Current.account.id)
+      render json: { success: true, url: redirect_url }
+    else
+      render json: { success: false }, status: :unprocessable_entity
+    end
+  end
```
sojan-official, line 18: - would be better to redirect them microsoft/callback/uniquecode  and the `::Redis::Alfred.setex('current_account_id', Current.account.id)`  to use that unique code as the key. Otherwise multiple users trying to connect inbox at the same time would create problems.

app/controllers/microsoft/callbacks_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def show
+    @response = microsoft_client.auth_code.get_token(
+      oauth_code,
+      redirect_uri: "#{base_url}/microsoft/callback"
+    )
+
+    ActiveRecord::Base.transaction do
+      inbox = find_or_create_inbox
+      ::Redis::Alfred.delete('current_account_id')
```
sojan-official, line 12: let's not use `current_account_id` as the Redis key here.

app/controllers/api/v1/accounts/microsoft/authorizations_controller.rb
tejaswinichile: I tried to do that, even more, I was trying to redirect it to the `{account.id}/microsoft/callback`, but we don't get to set any code in the redirect_uri else it gives the redirect_url mismatch error, we get the `auth_code` as params after the authorization. 

I get your point, but I tried to set the `wildcard` in the URL and couldn't.

Whatever redirect_url we pass should match the redirect_uri we set in the `azure-portal` app.

tejaswinichile: @sojan-official I was thinking of having one more step before the authorization step to add emails[as a uniqcode] and then after the authorization callback we get an 'authorized email' from users_data. We can get the account based on that users_data['email']

app/controllers/api/v1/accounts/microsoft/authorizations_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+  def create
+    email = params[:authorization][:email]
+    redirect_url = microsoft_client.auth_code.authorize_url({
+                                                              redirect_uri: "#{base_url}/microsoft/callback",
+                                                              scope: 'offline_access https://outlook.office.com/IMAP.AccessAsUser.All
+                                                            https://outlook.office.com/SMTP.Send openid',
+                                                              prompt: 'consent'
+                                                            })
+    if redirect_url
+      ::Redis::Alfred.setex(email, Current.account.id)
```
sojan-official, line 14: let's also pass the expiry as 5 minutes here or shorter. No need to persist the value for 1 day

app/jobs/inboxes/fetch_imap_emails_job.rb
```diff
@@ -41,9 +46,51 @@ def fetch_mail_for_channel(channel)
...
+  end
+
   def process_mail(inbound_mail, channel)
     Imap::ImapMailbox.new.process(inbound_mail, channel)
   rescue StandardError => e
     ChatwootExceptionTracker.new(e, account: channel.account).capture_exception
   end
+
+  # Making sure the access token is valid for microsoft provider
+  def valid_access_token(channel)
```
sojan-official, line 93: We will be doing this step every time when we fetch the mail right ? ( in 1 minute ). In case we won't be needing to run the job separately in a schedule ?

config/schedule.yml
```diff
@@ -19,3 +19,9 @@ trigger_imap_email_inboxes_job:
   cron: '*/1 * * * *'
   class: 'Inboxes::FetchImapEmailInboxesJob'
   queue: scheduled_jobs
+
+# refresh ms oauth token for the channel
+trigger_ms_oauth_token_refresh:
```
sojan-official, line 24: We will be doing this step every time when we fetch the mail right ? ( in 1 minute ). In case we won't be needing to run the job separately in a schedule ?

spec/controllers/api/v1/accounts/microsoft/authorization_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+             as: :json
+
+        expect(response).to have_http_status(:unauthorized)
+      end
+
+      it 'creates a new authorization and returns the redirect url' do
+        allow(::OAuth2::Client).to receive(:new).and_return(microsoft_client)
+        allow(microsoft_client).to receive(:auth_code).and_return(auth_code)
+        allow(auth_code).to receive(:authorize_url).and_return(
+          microsoft_client.authorize_url({ redirect_uri: 'http:0.0.0.0:3000/microsoft/callback' })
```
sojan-official, line 39:
```suggestion
          microsoft_client.authorize_url({ redirect_uri: 'http//:0.0.0.0:3000/microsoft/callback' })
```

app/jobs/inboxes/fetch_imap_emails_job.rb
tejaswinichile: Oh yes! you are right :D.. This is happening every 5 minute to check if the token is still valid and if not it will request for new token.

Removing the refresh token job.

app/jobs/channels/refresh_ms_oauth_token_job.rb
```diff
@@ -0,0 +1,62 @@
...
+  queue_as :low
+
+  def perform
+    Channel::Email.where(provider: 'microsoft').each do |channel|
+      # refresh the token here, with microsoft offline_access scope
+      provider_config = channel.provider_config || {}
+
+      refresh_tokens(channel, provider_config.with_indifferent_access) if provider_config[:refresh_token].present?
+    end
+  end
```
sojan-official, line 17: let's move this to `services` instead. since we won't be using the perform block nor need to inherit the application job. We only need the access_token  method, right?

tejaswinichile: Right! makes sense 👍🏼

sojan-official (commented):
- refactored the refresh token service a bit and added some addition specs. 
- noted that specs for callbacks controller is missing

app/javascript/dashboard/i18n/locale/en/inboxMgmt.json
```diff
@@ -53,6 +53,10 @@
           "ENABLE": "Create conversations from mentioned Tweets"
         }
       },
+      "MICROSOFT": {
```
sojan-official, line 56: should this be a part of the frontend PR ?

PR #6233, opened 2023-01-11
chore: Adds a settings button to the notification settings from the notification popup

iamsivin:
# Pull Request Template

## Description

This PR allows you to go to the notification setting from the unread conversation popup.

## Type of change

## How Has This Been Tested?
https://www.loom.com/share/f2e093b089e4485ea770502d428f8f9b

## Checklist:

### DIFF
--- a/app/javascript/dashboard/routes/dashboard/notifications/components/NotificationPanel.vue
+++ b/app/javascript/dashboard/routes/dashboard/notifications/components/NotificationPanel.vue
@@ -27,6 +27,14 @@
           </woot-button>
           <woot-button
             color-scheme="secondary"
+            variant="smooth"
+            size="tiny"
+            class-names="action-button"
+            icon="settings"
+            @click="openAudioNotificationSettings"
+          />
+          <woot-button
+            color-scheme="secondary"
             variant="link"
             size="tiny"
             icon="dismiss"
@@ -199,6 +207,21 @@ export default {
     onMarkAllDoneClick() {
       this.$store.dispatch('notifications/readAll');
     },
+    openAudioNotificationSettings() {
+      this.$router.push({ name: 'profile_settings_index' });
+      this.closeNotificationPanel();
+      this.$nextTick(() => {
+        const audioSettings = document.getElementsByClassName(
+          'audio-settings--view'
+        );
+        if (audioSettings) {
+          audioSettings[0].scrollIntoView(
+            { behavior: 'smooth', block: 'start' },
+            150
+          );
+        }
+      });
+    },
     closeNotificationPanel() {
       this.$emit('close');
     },
@@ -244,13 +267,13 @@ export default {
   .total-count {
     padding: var(--space-smaller) var(--space-small);
     background: var(--b-50);
-    border-radius: var(--border-radius-rounded);
+    border-radius: var(--border-radius-normal);
     font-size: var(--font-size-micro);
     font-weight: var(--font-weight-bold);
+    margin-left: var(--space-smaller);
   }
 
   .action-button {
-    padding: var(--space-micro) var(--space-small);
     margin-right: var(--space-small);
   }
 }
--- a/app/javascript/dashboard/routes/dashboard/settings/profile/NotificationSettings.vue
+++ b/app/javascript/dashboard/routes/dashboard/settings/profile/NotificationSettings.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <div class="profile--settings--row row">
+    <div class="profile--settings--row audio-settings--view row">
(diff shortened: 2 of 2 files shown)


fayazara (approved):
LGTM

app/javascript/dashboard/routes/dashboard/notifications/components/NotificationPanel.vue
```diff
@@ -199,6 +207,21 @@ export default {
...
+          'audio-settings--view'
+        );
+        if (audioSettings) {
+          audioSettings[0].scrollIntoView(
+            { behavior: 'smooth', block: 'start' },
+            150
+          );
+        }
+      });
+    },
```
scmmishra, line 224: Instead of this can we just add `id` to all the blocks in the profile settings? That way a shareable link can be created for any section of the settings page. We would avoid this logic too with that. 

P.S. We will have to check how will the `$router` respond to this, we might have to make some changes based on this [documentation](https://router.vuejs.org/guide/advanced/scroll-behavior.html)

iamsivin: @scmmishra By default, the Vue router (v.3) uses hash mode, but we are using history mode in our configuration. Do I want to change it? Thoughts? 
CC: @pranavrajs @nithindavid

scmmishra: We won't have to change the router mode, just a combination of history mode and using a `hash` to route to a part of the page. So the router declaration will be something like this

```js
export const router = new VueRouter({
  mode: "history",
  scrollBehavior(to, from, savedPosition) {
    if (to.hash) {
      return {
        selector: to.hash,
        // , offset: { x: 0, y: 10 }
      };
    }
  },
  routes,
});

```

iamsivin: @scmmishra I tried adding it, and it's not working. You can see it in the loom video.
https://www.loom.com/share/e403351c920442148f9f67240ebe8e12

Correct me if anything is wrong.

scmmishra: @iamsivin I tried to do this at well, seems like it's not working because the scroll is not on the body element, it's an overflowing inner element instead. Perhaps this can be fixed (Separately of course)

For now the current implementation should be fine, but maybe we can add a TODO there referencing this conversation

app/javascript/dashboard/routes/dashboard/settings/profile/NotificationSettings.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <div class="profile--settings--row row">
+    <div class="profile--settings--row audio-settings--view row">
```
nithindavid, line 3:
```suggestion
    <div id="profile-settings-notifications" class="profile--settings--row row">
```

app/javascript/dashboard/routes/dashboard/notifications/components/NotificationPanel.vue
```diff
@@ -199,6 +207,22 @@ export default {
     onMarkAllDoneClick() {
       this.$store.dispatch('notifications/readAll');
     },
+    openAudioNotificationSettings() {
+      this.$router.push({ name: 'profile_settings_index' });
+      this.closeNotificationPanel();
+      this.$nextTick(() => {
+        const audioSettings = document.getElementsByClassName(
+          'audio-settings--view'
```
nithindavid, line 215:
```suggestion
        const audioSettings = document.getElementById(
          'profile-settings-notifications'
```

app/javascript/dashboard/routes/dashboard/notifications/components/NotificationPanel.vue
```diff
@@ -199,6 +207,22 @@ export default {
...
+        );
+        if (audioSettings) {
+          // TODO [ref](https://github.com/chatwoot/chatwoot/pull/6233#discussion_r1069636890)
+          audioSettings[0].scrollIntoView(
```
nithindavid, line 219:
```suggestion
          audioSettings.scrollIntoView(
```

nithindavid (changes requested):
Can we use Id instead of class as suggested by Shivam?

scmmishra (approved):

scmmishra:
@iamsivin is this good to merge?

iamsivin:
> @iamsivin is this good to merge?

Yes, for me, the merge is blocked. If it works for you, can you?

nithindavid (approved):

PR #6236, opened 2023-01-11
fix: Stop processing broken liquid tags

sojan-official:
It will not throw an error if messages have invalid liquid syntax.


muhsin-k (approved):
Tested locally and working as expected.

PR #6241, opened 2023-01-12
chore: Create a resuable component for selecting a channel

pranavrajs:
To add Microsoft and other email providers as an option for #6227, I have created a reusable selector component. There is visual change. The component channel-selector can be used anywhere right now.

- Move the images to the public folder so that it is not bundled with the dashboard pack and are fetched on demand.
- Removed the SCSS file and used component scoped style.
- Reduce the lines of code.

Screenshot 2023-01-11 at 7 25 28 PM


app/javascript/dashboard/components/ChannelSelector.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <div class="small-6 medium-4 large-3 channel" @click="$emit('click')">
+    <img :src="src" />
+    <h3 class="channel__title">
+      {{ title }}
+    </h3>
+  </div>
```
scmmishra, line 7:
```suggestion
  <button class="small-6 medium-4 large-3 channel" @click="$emit('click')">
    <img :src="src" />
    <h3 class="channel__title">
      {{ title }}
    </h3>
  </button>
```

Should be a button if it's clickable

app/javascript/dashboard/components/ChannelSelector.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <div class="small-6 medium-4 large-3 channel" @click="$emit('click')">
+    <img :src="src" />
```
scmmishra, line 3:
```suggestion
    <img :src="src" :alt="title" />
```

scmmishra (changes requested):
Just a few suggestions, LGTM after that!

app/javascript/dashboard/components/ChannelSelector.vue
pranavrajs: Done.

app/javascript/dashboard/components/ChannelSelector.vue
pranavrajs: Done

PR #6248, opened 2023-01-12
fix: Use account locale as the default locale on widget

nithindavid:
Fixes https://github.com/chatwoot/product/issues/714

## How Has This Been Tested?

https://www.loom.com/share/3e241f230ed04747b743d1de8a9a6ebd

### DIFF
--- a/app/javascript/packs/sdk.js
+++ b/app/javascript/packs/sdk.js
@@ -18,7 +18,8 @@ const runSDK = ({ baseUrl, websiteToken }) => {
   }
 
   const chatwootSettings = window.chatwootSettings || {};
-  let locale = chatwootSettings.locale || 'en';
+  let locale = chatwootSettings.locale;
+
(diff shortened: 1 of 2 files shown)


app/javascript/packs/sdk.js
```diff
@@ -18,7 +18,8 @@ const runSDK = ({ baseUrl, websiteToken }) => {
   }
 
   const chatwootSettings = window.chatwootSettings || {};
-  let locale = chatwootSettings.locale || 'en';
+  let locale = chatwootSettings.locale;
```
nithindavid, line 21: The bug was happening as `App.vue` will set the locale from the chatwoot account correctly. But the widget sets 'en' as the default language and then updates locale on `App.vue` through postMessage. Which should not happen if no language is set by user in `window.chatwootSettings`

muhsin-k (approved):
LGTM

PR #6249, opened 2023-01-12
fix: add permissible when creating user through agents' controller

victor-torres:
# Add permissible when creating user through Agents' controller

## Description

When you create a user and assign it to an account with the agent role, the system automatically creates a permissible giving access to the newly created user to the platform account.

Turns out there's another way to create an user/agent, and it's through the Agents' controller, but when you follow this path, no permissible is created, and the platform account cannot interact with that resource properly.

## Type of change

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/app/controllers/api/v1/accounts/agents_controller.rb
+++ b/app/controllers/api/v1/accounts/agents_controller.rb
@@ -3,6 +3,7 @@ class Api::V1::Accounts::AgentsController < Api::V1::Accounts::BaseController
   before_action :check_authorization
   before_action :find_user, only: [:create]
   before_action :validate_limit, only: [:create]
+  before_action :set_platform_app, only: [:create]
   before_action :create_user, only: [:create]
   before_action :save_account_user, only: [:create]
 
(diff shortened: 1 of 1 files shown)


sojan-official:
@victor-torres, thanks for the changes. But the platform app tokens are only intended to be used with the platform APIs. We want to avoid polluting the application APIs by mixing both. 

The application endpoints should only be accessed via user/bot tokens. 

Also, in future, we will be deprecating the permissible check for platform apps and giving them global privilege in an installation.

victor-torres:
No problem, @sojan-official. Thanks for the quick answer. 

> Also, in future, we will be deprecating the permissible check for platform apps and giving them global privilege in an installation.

That's awesome! While it doesn't happen, do you suggest another way to create the permissible when using the Agents' Application API?

Creating a new User and assigning it the agent role requires a Platform API token with higher permissions, and we don't really wanna give these credentials to the systems that are intended to execute this operation. 

Also, it's more simple and concise to create agents using the Application API than the Platform API.

> But the platform app tokens are only intended to be used with the platform APIs. We want to avoid polluting the application APIs by mixing both.
>
> The application endpoints should only be accessed via user/bot tokens.

Sorry if I'm mistaken here, I'm quite new to the project 😅 but I can't see how we're using a Platform token here. 

The token used is still an Application token, I just need to identify the app / platform account id in order to create a database record that's going to give permissions to the platform account itself to manipulate a newly created resource under its account.

Otherwise, the theoretical "owner" of the resource won't be able to manipulate it. Makes sense? 

---

Anyways, if the removal of the permissibles is going to be worked soon, I can just close this pull request and commit the changes locally to address this use case while this major refactor is executed here. 

Thanks once again!

sojan-official:
> Sorry if I'm mistaken here, I'm quite new to the project 😅 but I can't see how we're using a Platform token here.

> The token used is still an Application token, I just need to identify the app/platform account id in order to create a database record that's going to give permissions to the platform account itself to manipulate a newly created resource under its account.


@ victor-torres My bad, I overlooked the PR. I can see that you are using application access tokens.  But again we would be deprecating the Platform app permissions of resources in future as there has been limited usage over that.  The permissions were built originally as a POC for us to add an integration to a SaaS management platform on Chatwoot Cloud.  But the project got scraped. If your use case is super dependent on platform tokens as the way forward, it's better to retain it over a fork.

sojan-official:
Closing the pull request as stated in the above comment.

PR #6252, opened 2023-01-12
fix: Adds animation to live chat trigger button

nithindavid:
**Fixes** https://github.com/chatwoot/product/issues/741

https://user-images.githubusercontent.com/1277421/212156931-22af9455-6939-468f-9b62-304335453cb2.mov

https://user-images.githubusercontent.com/1277421/212157096-b72716a7-3a91-4fc2-845f-1d7aebfd1bd3.mov

https://user-images.githubusercontent.com/1277421/212157120-219fa71f-8605-42ee-a5e0-e5f361142038.mov

### DIFF
--- a/app/javascript/sdk/sdk.js
+++ b/app/javascript/sdk/sdk.js
@@ -203,6 +203,7 @@ export const SDK_CSS = `
 
 .woot-widget-bubble {
   transform-origin: center;
+  will-change: transform, opacity;
   transform: translateX(0) scale(1) rotate(0deg);
   transition: transform 300ms ease, opacity 100ms ease, visibility 0ms linear 0ms, bottom 0ms linear 0ms;
 }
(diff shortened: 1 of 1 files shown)


nithindavid:
We could add https://developer.mozilla.org/en-US/docs/Web/CSS/will-change but since the SDK needs to cover different browsers of varying performance, I don't know if this would get into trouble. 
cc: @pranavrajs @scmmishra

scmmishra:
>  I don't know if this would get into trouble.

The property won't do any harm. Just test it on Safari I guess and ensure we respect [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) preference from the users. 

If performance is a major concern we can ensure that the click action that expands the chatbox is wrapped in a [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)

CC: @nithindavid

nithindavid:
@scmmishra I have added `will-change` to code.

scmmishra (approved):
LGTM!

iamsivin (approved):
LGTM

PR #6267, opened 2023-01-17
feat: Shows the last activity, created at timestamp in the same row

iamsivin:
# Pull Request Template

## Description
This PR includes - 
1. Shows the last activity, created at the timestamp in the same row in 24h - 3m format.
2. Adds a tooltip on hover (delay:1500ms), which includes the last activity, created at the time in the ago format.
3. Added some transparency for the tooltip background color

Fixes https://github.com/chatwoot/product/issues/756

## Type of change

## How Has This Been Tested?

**Screencast**

https://user-images.githubusercontent.com/64252451/212892674-5353a0b9-8f73-4723-9c4b-0640b37d7494.mov

## Checklist:


app/javascript/dashboard/components/ui/TimeAgo.vue
```diff
@@ -19,20 +28,40 @@ export default {
...
   },
+  computed: {
+    lastActivityTime() {
+      return this.shortTimestamp(this.lastActivityAtTimeAgo);
+    },
+    createdAtTime() {
+      return this.shortTimestamp(this.createdAtTimeAgo);
+    },
+    tooltipText() {
+      return `Created at: ${this.createdAtTimeAgo}
```
scmmishra, line 55: If we are formatting this as "X days ago" we should not use "Created at:"

The UI can be more conversational like

Created 5 days ago, last activity 3 days ago

And we can put a limit of 3 months, if the date is older than 3 months, we show the exact date on hover, something like

Created on 5th August 2022, last activity today

Because after a point it's difficult to mentally infer dates and months

CC: @nithindavid what do you think?

scmmishra (changes requested):

app/javascript/dashboard/assets/scss/_foundation-custom.scss
```diff
@@ -17,9 +17,10 @@
 }
 
 .tooltip {
+  background-color: #000000d5;
```
scmmishra, line 20: Can we use a variable here?

app/javascript/dashboard/assets/scss/_foundation-custom.scss
```diff
@@ -17,9 +17,10 @@
...
   border-radius: $space-smaller;
   font-size: $font-size-mini;
-  max-width: 15rem;
+  max-width: 20rem;
```
scmmishra, line 23: Are there token variables that can be used instead?

app/javascript/dashboard/mixins/time.js
```diff
@@ -12,5 +12,29 @@ export default {
       const unixTime = fromUnixTime(time);
       return formatDistanceToNow(unixTime, { addSuffix: true });
     },
+    shortTimestamp(time) {
```
scmmishra, line 15: Can we write tests for this? Correct me if I am wrong, I am assuming the input to this function comes from date-fns distanceInWords method, so if we change the underlying library this could break

app/javascript/shared/components/FluentIcon/dashboard-icons.json
```diff
@@ -51,6 +51,7 @@
   "chevron-left-outline": "M15.53 4.22a.75.75 0 0 1 0 1.06L8.81 12l6.72 6.72a.75.75 0 1 1-1.06 1.06l-7.25-7.25a.75.75 0 0 1 0-1.06l7.25-7.25a.75.75 0 0 1 1.06 0Z",
   "chevron-right-outline": "M8.47 4.22a.75.75 0 0 0 0 1.06L15.19 12l-6.72 6.72a.75.75 0 1 0 1.06 1.06l7.25-7.25a.75.75 0 0 0 0-1.06L9.53 4.22a.75.75 0 0 0-1.06 0Z",
   "chevron-up-outline": "M4.22 15.53a.75.75 0 0 0 1.06 0L12 8.81l6.72 6.72a.75.75 0 1 0 1.06-1.06l-7.25-7.25a.75.75 0 0 0-1.06 0l-7.25 7.25a.75.75 0 0 0 0 1.06Z",
+  "circle-outline": "M2 12C2 6.477 6.477 2 12 2s10 4.477 10 10s-4.477 10-10 10S2 17.523 2 12Z",
```
scmmishra, line 54: In case we have to include a new icon, it's probably better to use the bullet ascii character instead, since this will be rendered in a list

app/javascript/dashboard/mixins/time.js
```diff
@@ -12,5 +12,29 @@ export default {
...
+      const timestamp = time;
```
scmmishra, line 16: Any reason to proxy this variable? The replace method returns a new string anyway

app/javascript/dashboard/mixins/time.js
```diff
@@ -12,5 +12,29 @@ export default {
...
+      const convertToShortTime = timestamp
+        .replace('about', '')
```
scmmishra, line 18: These can be collapsed in lesser replaces with regex

replace(/about|another|word|here/g, 'world');

scmmishra (changes requested):

pranavrajs:
@scmmishra @iamsivin The tooltips are too distracting. Shall we use the `title` attribute instead `v-tooltip`?

app/javascript/dashboard/mixins/time.js
pranavrajs: Shall we use this helper and take the first letter after the space? https://date-fns.org/v2.29.3/docs/formatDistanceToNowStrict

app/javascript/dashboard/components/ui/TimeAgo.vue
nithindavid: Totally makes sense. Can we fall back to normal (MM/DD/YY)  if it's created before a month?  I find the `2 months ago` not usefull after a point as, 1 month 28 days would also be shown as more than a month ago; which doesn't contribute to the visual sorting logic + slows the urgency one might need to do.

nithindavid:
> @scmmishra @iamsivin The tooltips are too distracting. Shall we use the `title` attribute instead of `v-tooltip`?

Since title takes 2-3 seconds delay before showing up, wouldn't it be frustrating for the user as it's something people check often?

iamsivin:
> @scmmishra @iamsivin The tooltips are too distracting. Shall we use the `title` attribute instead of `v-tooltip`?

@pranavrajs We can add a delay for the tooltip. Now it is 300ms. Do I want to increase it?

app/javascript/dashboard/assets/scss/_foundation-custom.scss
iamsivin: No, we don't have it

app/javascript/dashboard/components/ui/TimeAgo.vue
iamsivin: @nithindavid @scmmishra Thoughts on this? If the conversation is created before a month and the user activity is not there for more than one month, it will show in `MM/DD/YYYY.` Else, it will show in the ago format.


https://user-images.githubusercontent.com/64252451/213091983-44a28242-91b9-4bef-9e24-c4ff782b5bf4.mov

app/javascript/dashboard/mixins/time.js
iamsivin: @pranavrajs Here, we want to replace `" year ago"` with `"y"`, is it possible with [formatDistanceToNowStrict](https://date-fns.org/v2.29.3/docs/formatDistanceToNowStrict)? Please let me know if I'm wrong.

app/javascript/dashboard/components/ui/TimeAgo.vue
iamsivin: @scmmishra @nithindavid Updated the PR with the last commented changes.

pranavrajs:
> @iamsivin @pranavrajs We can add a delay for the tooltip. Now it is 300ms. Do I want to increase it?

Please increase the delay. Can we make it less prominent as well?

app/javascript/dashboard/components/ui/TimeAgo.vue
scmmishra: Looks great!

app/javascript/dashboard/mixins/time.js
scmmishra: > Shall we use this helper and take the first letter after the space? https://date-fns.org/v2.29.3/docs/formatDistanceToNowStrict

@pranavrajs This will not work for `month` and `minute`

nithindavid (approved):

nithindavid:
@vishnu-narayanan The circle tests are failing, is this the same upgrade issue?

vishnu-narayanan:
@nithindavid No; this one was different. This was a bundle audit failure. Fixed by @tejaswinichile via  https://github.com/chatwoot/chatwoot/pull/6313

PR #6293, opened 2023-01-19
fix: Whatsapp template picker bug

fayazara:
## Description

Enforce lowercasing the template status value before checking the value


sojan-official (approved):
tested and working

PR #6326, opened 2023-01-23
fix: Identity JSON response header

tejaswinichile:
Screenshot 2023-01-23 at 12 47 37 PM


app/views/microsoft/identity_association.json.erb
```diff
@@ -1,7 +1 @@
-{
```
sojan-official, line 1: Can we use a JBuilder template instead?
To stay consistent with what we have in the codebase. 

lets also update the specs to verify the response headers.

spec/controllers/microsoft_controller_spec.rb
```diff
@@ -6,7 +6,7 @@
       with_modified_env AZURE_APP_ID: 'azure-application-client-id' do
         get '/.well-known/microsoft-identity-association.json'
         expect(response).to have_http_status(:success)
-        expect(response.body).to include '"applicationId":  "azure-application-client-id"'
+        expect(response.body).to include '"applicationId":"azure-application-client-id"'
```
sojan-official, line 9: @tejaswinichile  if the header is a mandatory case, can we add a spec for the header as well .  other wise looks good to me.

sojan-official (commented):
Minor comment. looks good otherwise.

vishnu-narayanan (approved):

PR #6340, opened 2023-01-24
fix: validate instagram story only while saving the message

tejaswinichile:
https://www.loom.com/share/3b6e461c92cc4f539b2061a9c9d58354


app/javascript/dashboard/components/widgets/conversation/bubble/Image.vue
```diff
@@ -1,6 +1,11 @@
 <template>
   <div class="image message-text__wrap">
-    <img :src="url" @click="onClick" @error="onImgError()" />
+    <img
+      :src="url"
+      alt="Resource is not available"
```
nithindavid, line 5: Please use i18n here

fayazara (approved):
LGTM.

Tested locally with TJ

Please also have a look at Nithins comment

PR #4576, opened 2022-04-28
chore: Add first_reply_created event in conversation

aswindevps:
Fixes #4645

PR #6409, opened 2023-02-07
fix: bots included in time to response metrics

scmmishra:
## Description

Time to first response included the metrics from bots as well, this PR fixes that. 

The changes in this PR

1. ~https://github.com/chatwoot/chatwoot/pull/6409/commits/bb9171945d5e3b2f6015f4f96dd1b76b3efb6987 & https://github.com/chatwoot/chatwoot/pull/6409/commits/de1977c219a2e7a9180dd02272244fe3b3f7ce89 ignore bot values in conversations, **this will affect the current values the users see on their reports**, we can probably remove this, need thoughts?~ Reverted these commits
1. https://github.com/chatwoot/chatwoot/pull/6409/commits/4a236930f8f9cc036d9e6e9f2a956b3740455193 & https://github.com/chatwoot/chatwoot/pull/6409/commits/9f4cfea4d787b62f889ea03325cf85866170e9ec logs the bot handoff event explicitly in the `response_events` table
1. https://github.com/chatwoot/chatwoot/pull/6409/commits/33f434ac72288b1a68a73bdd71a188d3caa1ec93 ensure that `FIRST_REPLY_CREATED` is dispatched only if the response is made by a human and not `AgentBot`, **this makes the first change redundant over time as new data accumulates**
1. https://github.com/chatwoot/chatwoot/pull/6409/commits/606d0b0f469611f7d03d6004766800bd4816273a this ensures that the bot_handoff time is taken into account when calculating the first_response. 

Issue for reference: https://github.com/chatwoot/product/issues/790

## Type of change

## How Has This Been Tested?

- Setup CSML locally, and add a basic bot.
- Try starting a conversation from the widget using [the test page](http://localhost:3000/widget_tests)
- Observe the time to response time in the conversation report

Apply this patch and try it again, it should disregard the bot conversations.

## Checklist:


scmmishra:
I have merely ignored the conversations with no `user_id` in the report builder. I figured that the `user_id` for bot conversations is `NULL`.

image

If this approach is fine, I can write tests for this.

---

**Another way to do this would be not capture the event if the `user_id` is not present in the first place**. The benefit is that it will avoid a lot of redundant rows in the table, but (assuming it's not used elsewhere) the downside is that this data won't be available to use elsewhere if needed in the future

@pranavrajs @sojan-official @tejaswinichile LMK what you folks think

tejaswinichile:
If `first_response_metrics` should depend on the human response, then this PR makes sense. You can go ahead and add the test cases for the same.

pranavrajs:
@scmmishra 

IMO this wouldn't solve the issue. We have only one first_response_time for a conversation at the moment. If there is a bot response, then it wouldn't record human response time. 

There are caveats to the following design. However, let us start with the basics. 

Let us consider the first Response Time metrics to be calculated based only on the human. The current design of the bot is that the response would be immediate, so saving those metrics is not really helpful. However, the agent wouldn't see the message until it came in the open status.

The time that needs to be computed here is the time between the first time appearance in open status and the first human response time. 

If the conversation is created in open status directly, then it becomes the time difference between created_time and first human response time.

If the agent replies to the conversation before moving it to open status, then we can set the response time as 0.

To summarize:

Now we have the following events for every conversations:

- first_response
- conversation_resolved

After the change, this would become.
- first_appreance_in_open_status: This would also tell us how much time was spent with the bot.
- first_response: First human response after they could see the conversation
- conversation_resolved: Total time taken to resolve = Time taken by bot + Time taken by the agent

Let me know if you have any questions.

scmmishra:
@pranavrajs this makes sense. Perhaps instead of `first_appreance_in_open_status` we call it `bot_handoff`, I figured the the conversation stays in `pending` state until the bot hands it off. 

**Is it safe to assume that if the status of a conversation changes from `pending` to `open` it was a bot handoff and nothing else?** 

Assuming this is not the case, here's the solution I propose which is much more implicity

1. Add a `bot_handoff_at` field in the `conversation` table
1. We create a `handoff` method in the conversation model. This will be used in [`bot_processor_service.rb`](https://github.com/chatwoot/chatwoot/blob/develop/lib/integrations/bot_processor_service.rb#L58) directly
1. The handoff method will set the `conversation.status` to `open` and set the `conversation.bot_handoff_at` to current timestamp
1. We can then alter the [`first_reply_created`](https://github.com/chatwoot/chatwoot/blob/8a6f64702779cc252713716b5554ae112f97303b/app/listeners/reporting_event_listener.rb#L22-L44) method with two changes
    1. Do not record `first_response` event if the `conversation` is in pending
    1. Change the first_response calculation to use the `bot_handoff_at` instead of `created_at`
    
So the updated method will look like this

```ruby
def first_reply_created(event)
  message = extract_message_and_account(event)[0]
  conversation = message.conversation
  # ignore pending conversations
  if conversation.status == 'pending'
    return
  end

  # updated calculation
  last_human_activity = conversation.bot_handoff_at || conversation.created_at
  first_response_time = message.created_at.to_i - last_human_activity.to_i

  reporting_event = ReportingEvent.new(
    name: 'first_response',
    value: first_response_time,
    value_in_business_hours: business_hours(conversation.inbox, conversation.created_at,
                                            message.created_at),
    user_id: conversation.assignee_id,
    conversation_id: conversation.id,
    event_start_time: conversation.created_at,
    event_end_time: message.created_at
  )

  conversation.update(first_reply_created_at: message.created_at)

  reporting_event.save!
end
```

pranavrajs:
@scmmishra As discussed, we will have the `bot_handoff` event in the reporting table. The first response will be computed as the time difference from the `bot_handoff` event if the event exists for the conversation. If not, we will calculate the time difference from the created time.

scmmishra:
@pranavrajs I've updated the logic based on our discussion, lmk when it's good to go

scmmishra:
@sojan-official I added a method to `Conversation` model, but adding that breaks code climate with the following error, `Class Conversation has 33 methods (exceeds 32 allowed). Consider refactoring.` 

We will probably have to abstract some methods out as a concern, but most methods implement very specific business logic for conversations and are not re-usable. What would you suggest? 

An option is to move all the status related methods to a `conversation_status_handler` concern and include it here, but again this won't be re-usable

app/helpers/report_helper.rb
```diff
@@ -33,7 +33,8 @@ def resolutions_count
   end
 
   def avg_first_response_time
-    grouped_reporting_events = (get_grouped_values scope.reporting_events.where(name: 'first_response', account_id: account.id))
+    grouped_reporting_events = (get_grouped_values scope.reporting_events.where(name: 'first_response',
```
pranavrajs, line 36: Do we need this change since we would not record the bot responses in the first_response metric anymore? If this is done to support the previous data. Let's update the previous data using a migration.

app/helpers/report_helper.rb
```diff
@@ -59,6 +60,7 @@ def avg_resolution_time_summary
   def avg_first_response_time_summary
     reporting_events = scope.reporting_events
                             .where(name: 'first_response', account_id: account.id, created_at: range)
+                            .where.not(user_id: nil) # exclude bot responses
```
pranavrajs, line 63: Same as previous

app/helpers/report_helper.rb
scmmishra: Yeah, this change becomes redundant over time.

> Let's update the previous data using a migration.
Should I drop the old rows?

scmmishra: @pranavrajs we can add a migration to drop the old rows with incorrect data like this

```ruby
class DropReportingEventsWithIncorrectFirstResponses < ActiveRecord::Migration[6.1]
  def change
    Reporting::Event.where(name: 'first_response').where(user_id: nil).delete_all
  end
end
```

Does this work?

sojan-official:
> @sojan-official I added a method to `Conversation` model, but adding that breaks code climate with the following error, `Class Conversation has 33 methods (exceeds 32 allowed). Consider refactoring.`
> 
> We will probably have to abstract some methods out as a concern, but most methods implement very specific business logic for conversations and are not re-usable. What would you suggest?
> 
> An option is to move all the status related methods to a `conversation_status_handler` concern and include it here, but again this won't be re-usable

We could just have a helper module which groups a specific set of methods into a seperate file so that total line length is less and each of this files are more readable.  Doesn't have to focus too much on whther those methods are re-usable. you can approach it purely from an aesthetics stand point.

scmmishra:
@pranavrajs @sojan-official I've added a migration, PTAL. @sojan-official the build is failing because there are pending migrations; how do I fix the build, [here](https://app.circleci.com/pipelines/github/chatwoot/chatwoot/55781/workflows/b55626b8-e668-4dc4-976d-da388b39c8cd/jobs/57711/parallel-runs/0/steps/0-120) is the broken step for the [build](https://app.circleci.com/pipelines/github/chatwoot/chatwoot/55781/workflows/b55626b8-e668-4dc4-976d-da388b39c8cd/jobs/57711)

scmmishra:
@pranavrajs @sojan-official I've made the changes and tested the migration locally. There's one flaw, however, with what the migration is doing.

The migration assumes that the bot handoff happens at the `last_bot_reply` created time

```
 conversation       bot                         conversation
 start              handoff                     resolved
 |                  |                           |
 |____|___|_________|____|_______|_____|________|
      bot reply     ^    ^  human reply
                    |    |
                    |    |
       last_bot_reply    first_human_reply

```

However, this is not always the case. We can have a CSML bot that hands-off after user input, see the following screenshot.

<details>
<summary>
Here, the handoff happens after the user has sent their email address.
</summary>
image
</details>

## Solution explorations

We could look for the immediate message after the last bot-message
- if it is made by an Agent, it is safe to assume that the bot message was the moment of hand-off
- if the message was made by a contact, we could assume that it was the moment of hand-off and recompute accordingly. However, this is not a very safe assumption. Given it will take out outliers where this is not the case. Say the bot-handoff was done, and no agent replied for an hour, and the contact sent a follow-up message. 

My suggestion is that we let the migration stay as is, this will show some skewed data for the past, but it will be normalized over the period of time.

pranavrajs:
> I would suggest we let the migration stay as is, this will show some skewed data for the past, but it will be normalized over the period of time.

This sounds good. Let us go ahead as it is. If there is any concern raised after the migration, let us rebuild and rerun this migration again.

scmmishra:
> This sounds good. Let us go ahead as it is. If there is any concern raised after the migration, let us rebuild and rerun this migration again.

@pranavrajs, noted, just need your review now and we're good to go

pranavrajs:
@sojan-official @scmmishra Please get the review completed and get it out. The logic explained looks good enough to me.

db/schema.rb
```diff
@@ -10,908 +10,908 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2023_02_09_033203) do
```
sojan-official, line 13: Lets not alter the db:schema file generated by rails.  I think only the version number change should happen with the migration that is added. so lets revert the other changes.

db/migrate/20230214025901_update_reporting_events_with_incorrect_first_responses.rb
```diff
@@ -0,0 +1,45 @@
...
+  include ReportingEventHelper
+
+  # rubocop:disable Metrics/AbcSize
+  def change
+    ReportingEvent.where(name: 'first_response', user_id: nil).find_in_batches do |bot_responses|
+      bot_responses.each do |event|
+        conversation = event.conversation
+        next if conversation.nil?
+
+        last_bot_reply = conversation.messages.where(sender_type: 'AgentBot').order(created_at: :asc).last
```
sojan-official, line 11: We have over a million records in our cloud. Running this migration sequentially wouldn't be effective. Instead, lets move the logic to data migration sidekiq jobs.  lets go with this strategy 

- iterate over accounts and enqueue a recompute job for each account
- the logic in this migration happens in the recompute job for scoped for that account.  

ref: https://github.com/chatwoot/chatwoot/pull/4576/files

sojan-official (commented):
It seems we are not considering the case where the agent bot is not a native csml or dialogflow bot for recording the `bot_handoff` and first_reply_created.  Is it a conscious decision ? @pranavrajs  

Also, if an agent switches a conversation back to pending state and the conversation triggers handoff again. This will record additonal `conversation_bot_handoff` for the conversation, where the time values won't be accurate. Should we record these events in subsequent handoff cases ?

scmmishra:
> This will record additonal conversation_bot_handoff for the conversation

Right now the `CONVERSATION_BOT_HANDOFF` event is triggered in one place only https://github.com/chatwoot/chatwoot/blob/35aadb77614a237a6adfb116c3458ed28b906252/lib/integrations/bot_processor_service.rb#L55-L62

> if an agent switches a conversation back to pending state and the conversation triggers handoff again

In this case, does the `bot_processor_service` come into action? If it does not, then there should not be another handoff. Right now the handoff is contingent on `bot_processor_service` triggering it.

> it seems we are not considering the case where the agent bot is not a native csml or dialogflow 

Is there a separate handler for Dialogflow other than `bot_processor_service`? I saw that the DialogFlow handler extends `Integrations::BotProcessorService` I am assuming the `handoff` action is triggered with DialogFlow as well

@sojan-official LMK what should I do next

db/schema.rb
scmmishra: Reverted this. PTAL

db/migrate/20230214025901_update_reporting_events_with_incorrect_first_responses.rb
scmmishra: This commit resolves it https://github.com/chatwoot/chatwoot/pull/6409/commits/1992fc2c187bf3e8ef2024bb69ae33b5eb6b0253 PTAL

sojan-official:
> Is there a separate handler for Dialogflow ? 

 we also have the case of AgentBots 
ref: https://www.chatwoot.com/docs/product/others/agent-bots

the bot_processor_service only takes care of the native bot integrations. but in case of AgentBots, the handoff is triggered by the customer by calling our conversation API to toggle status from `pending` to `open`

----


> if an agent switches a conversation back to pending state and the conversation triggers handoff again

In case of the processor service it comes back to the handoff block. In case of agent bots its just API toggling conversations back to pending  from open. and vice versa.

in either case, 
- should we  update reports with another handoff event ?
- if yes, i believe the current time calculations will be off.

pranavrajs:
> It seems we are not considering the case where the agent bot is not a native csml or dialogflow bot for recording the bot_handoff and first_reply_created. Is it a conscious decision ? @pranavrajs

Yes, We are solving this only for the native bots. Fixing the other integration requires more discussions. 

> Also, if an agent switches a conversation back to pending state and the conversation triggers handoff again. This will record additonal conversation_bot_handoff for the conversation, where the time values won't be accurate. Should we record these events in subsequent handoff cases ?

At the moment, we only need to consider the first handoff and the first human reply time.

app/jobs/migration/update_first_response_time_in_reporting_events_job.rb
```diff
@@ -0,0 +1,44 @@
...
+  def perform(account)
+    account.reporting_events.where(name: 'first_response', user_id: nil).each do |event|
+      conversation = event.conversation
+      next if conversation.nil?
+
+      last_bot_reply = conversation.messages.where(sender_type: 'AgentBot').order(created_at: :asc).last
+      first_human_reply = conversation.messages.where(sender_type: 'User').order(created_at: :asc).first
+
+      # accomodate for campaign if required
+      # new_value = difference between the first_human_reply and the first_bot_reply if it exists or first_human_reply and created at
```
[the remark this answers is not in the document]
scmmishra: Fixed here https://github.com/chatwoot/chatwoot/pull/6409/commits/01e02e9e8d6ad9ef1358244e551309c45d460f06, lmk if this is good?

app/listeners/reporting_event_listener.rb
```diff
@@ -22,23 +23,41 @@ def conversation_resolved(event)
...
+      value: time_to_handoff,
+      value_in_business_hours: business_hours(conversation.inbox, conversation.created_at, conversation.updated_at),
+      account_id: conversation.account_id,
+      inbox_id: conversation.inbox_id,
+      user_id: conversation.assignee_id,
+      conversation_id: conversation.id,
+      event_start_time: conversation.created_at,
+      event_end_time: conversation.updated_at
+    )
+    reporting_event.save!
```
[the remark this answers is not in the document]
scmmishra: Fixed here: https://github.com/chatwoot/chatwoot/pull/6409/commits/7def2da1ae723667b3a10b2d3e0100dcc2425a29

scmmishra:
@sojan-official lmk if this is good to go?

spec/listeners/reporting_event_listener_spec.rb
```diff
@@ -60,5 +60,51 @@
...
+      end
+    end
+  end
+
+  describe '#conversation_bot_handoff' do
+    it 'creates conversation_bot_handoff event' do
+      expect(account.reporting_events.where(name: 'conversation_bot_handoff').count).to be 0
+      event = Events::Base.new('conversation.bot_handoff', Time.zone.now, conversation: conversation)
+      listener.conversation_bot_handoff(event)
+      expect(account.reporting_events.where(name: 'conversation_bot_handoff').count).to be 1
```
sojan-official, line 92: - lets also add a case for duplicate handoff events not getting created

sojan-official (commented):
LGTM over all. 
- a minor comment on an additional test case to be added. 

Lets deploy this over staging once to test before merging it to develop.

sojan-official (approved):

scmmishra:
@sojan-official added some minor changes, and tested the migration locally. Let me know if you have some time to pair up and test the migration on staging

pranavrajs (approved):

PR #6433, opened 2023-02-10
chore: Limit conversation resolution Job

sojan-official:
We will be adding a limit to the resolution job so that it is performed at a spaced interval. This will ensure there won't be a sudden spike in resource usage

fixes: https://github.com/chatwoot/product/issues/707

ref customer report 

> An account had -20k unresolved conversations. Those conversations got resolved at once due to inactivity. This event triggered around 400k jobs in Sidekiq, eventually leading to increased memory and CPU usage. The number of jobs is common (not unusual), considering each resolution event can lead to 8-20 smaller jobs performing several activities like webhook notifications, email alerts, or realtime messages to agents.

### DIFF
--- a/app/jobs/conversations/resolution_job.rb
+++ b/app/jobs/conversations/resolution_job.rb
@@ -1,8 +1,9 @@
 class Conversations::ResolutionJob < ApplicationJob
-  queue_as :medium
+  queue_as :low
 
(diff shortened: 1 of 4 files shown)


tejaswinichile (approved):
LGTM 💯

Issue #5592, opened 2022-10-11
[CW-42] CSAT Not Showing After Conversation Closed When Connected to External Bot


punyaikhwan:
**Describe the bug**
I connect the Chatwoot to external bot (RASA).
After the conversation closed, CSAT not showing even in the settings CSAT is enabled.

**To Reproduce**

1. Create Inbox.
2. Connect that inbox to external bot like RASA, and activate.
3. Enable CSAT in that inbox
4. Make a conversation and make sure the external bot is active.
5. Close the conversation, CSAT is not showing.

If I deactivate the external bot, CSAT appears.

**Expected behavior**
1. If CSAT is enabled in setting, show it even the inbox is connected to external bot.
2. Or, make API to trigger CSAT, so we can call it when conversation is closed.

**Screenshots**
![image](https://user-images.githubusercontent.com/14848437/194993593-7606e600-83de-4444-9eee-3988521a0207.png)
Image 1: bot is not activated, CSAT appears

![image](https://user-images.githubusercontent.com/14848437/194993502-8d63d25f-7c74-4c48-9275-26401843bee7.png)
Image 2: bot is activated, CSAT not showing

**Environment**
- Self hosted Chatwoot
- Ubuntu 22.04 64-bit,  using Docker 20.10.18

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser: Chrome
 - Version : 106.0.5249.91 (Official Build) (64-bit)

<sub>[CW-42](https://linear.app/chatwoot/issue/CW-42/csat-not-showing-after-conversation-closed-when-connected-to-external)</sub>

eremeye:
CSAT is disabled if agent bot connected to inbox.

take a look at app/services/message_templates/hook_execution_service.rb

```
def perform
    **return if inbox.agent_bot_inbox&.active?**
    return if conversation.campaign.present?

    trigger_templates
end
```

sojan-official:
ref: https://github.com/chatwoot/chatwoot/issues/5514

sojan-official:
This has been resolved in #5592

PR #6455, opened 2023-02-15
chore: Ensure Template messages work even when bot is connected

sojan-official:
ref: #5592


scmmishra (approved):
Approving this, should we add tests for this case?

sojan-official:
@scmmishra 😅 ideally, we should have had tests for the this condition which would be removed along with this change. but we missed it back then. 

Bot connection doesn't have specific impact, otherwise for the condition which we are removing now. So the existing specs should be suffice. 

waiting to see if @pranavrajs  have any other thoughts on this behaviour.

PR #6477, opened 2023-02-17
chore: Remove `mobile-v2` from feature flags

muhsin-k:
This PR removes the feature `mobile-v2` since the updated mobile app is already released to all the users.

### DIFF
--- a/config/features.yml
+++ b/config/features.yml
@@ -37,8 +37,6 @@
   enabled: true
 - name: voice_recorder
   enabled: true
-- name: mobile_v2
-  enabled: false
 - name: channel_website
   enabled: true
 - name: campaigns


config/features.yml
```diff
@@ -37,8 +37,6 @@
   enabled: true
 - name: voice_recorder
   enabled: true
-- name: mobile_v2
```
sojan-official, line 40: @muhsin-k  feature flag, once added, can't be removed from the yml file. We can update the flag for customers by running a migration instead.  if there is no checks in code based on this flag. then lets just write a comment over this key that its depricated.

muhsin-k: Makes sense.

PR #6482, opened 2023-02-17
chore: Support plus forwarding in email channel

tejaswinichile:
- One of our community customers said they could not get the IMAP emails in our dashboard if the email has been sent to `+` extension email address.
- I think we could provide this support, I fixed it while investigating it


pranavrajs:
@sojan-official This is the expected IMAP email behavior. I don't think there is a need to restrict that.

tejaswinichile:
@sojan-official Could you review this? Mostly regex changes to support `+` email inbox finder.

app/finders/email_channel_finder.rb
```diff
@@ -7,7 +7,11 @@ def perform
     channel = nil
     recipient_mails = @email_object.to.to_a + @email_object.cc.to_a
     recipient_mails.each do |email|
-      channel = Channel::Email.find_by('lower(email) = ? OR lower(forward_to_email) = ?', email.downcase, email.downcase)
+      mail_id, domain = email.split('@')
+      original_mail_address = mail_id.split('+')[0]
+      channel = Channel::Email.where('lower(email) ~* ? OR lower(forward_to_email) ~* ?',
```
sojan-official, line 12: Missed the case where we do .downcase on email. Pushing it alongside some refactoring.

PR #6498, opened 2023-02-20
fix: fetch only unread emails

sajangupta74:
# Pull Request Template

## Description

Fixes # (issue)

## Type of change

## How Has This Been Tested?

## Checklist:


tejaswinichile:
@sajangupta74 Hey I will have it tested today and get back to you with the next steps.

tejaswinichile:
This works fine, but one copy should be in the chatwoot dashboard.

As inbox is shared between agents, and they are handling customer support queries from the dashboard, then anyone who doesn't have access to emails but access to inbox should be able to respond to the conversation.

cc @pranavrajs @sojan-official what do you guys think?

sajangupta74:
> This works fine, but I think one copy should be in the chatwoot dashboard even if it's read email.
> 
> As inbox is shared between agents, and they are handling customer support queries from the dashboard, then anyone who doesn't have access to emails but access to inbox should be able to respond to the conversation.
> 
> cc @pranavrajs @sojan-official what do you guys think?

@tejaswinichile anyone having access to inbox can reply to the email. As this change only reads unread emails, read emails will be in the inbox so the user having access can reply from inbox as well

tejaswinichile:
@sajangupta74 

I tested this scenario:

1. User sent an email to inbox-email
2. I(suppose super-admin) opened inbox email from the email dashboard, not the chatwoot dashboard
3. So that particular email doesn't reach to chatwoot inbox because it gets filtered out with the NOT SEEN key

We don't want this particular behavior.

tejaswinichile:
https://www.loom.com/share/40781dfcd8d4423c9216f64f2bcba9d1

This is the behavior that is wrong and not expected from the customers as well as the agents.

sajangupta74:
yep that's right, consider following scenario:
we had an issue in our environment where in tickets were not getting processed into chatwoot dashboard, if this issue continue for significant amount of time then what happens is there will be 1000s of unproccessed emails. After fixing the issue what happened is latest 10 emails was processed but remaining 1000s - 10 were not getting processed as the worker fetches latest 10 emails only. 

So its a tradeoff where in we supposed no one would open email from inbox(that's more likely) and whenever similar prod issue happens then all unprocessed emails will be processed automatically(no intervention of developer required).

tejaswinichile:
@sajangupta74 This is a very specific issue with your specific instance; in this case, you can raise the count to 100 or 1000 to fetch the last 1000 emails may be in batches if you have that large amount of emails. The contribution PR should keep up with the existing functionality.

If you want our repo should have this change may add some installation config which would ask client if they are ready for this tradeoff, if the installation config is true then only add this key filter.

Otherwise, As of now, we can't merge this PR.

pranavrajs:
>  we had an issue in our environment where in tickets were not getting processed into chatwoot dashboard, if this issue continue for significant amount of time then what happens is there will be 1000s of unproccessed emails. After fixing the issue what happened is latest 10 emails was processed but remaining 1000s - 10 were not getting processed as the worker fetches latest 10 emails only.

@tejaswinichile @sajangupta74 I believe this is a valid case and should have a fix. We can discuss this separately. The idea of the email channel with Chatwoot is that should work similarly to the regular email clients. So, the point @tejaswinichile made still exists. If someone accidentally opens the email, it will not be available in Chatwoot, which leads to an inconsistent state which is hard to debug.

tejaswinichile:
@sajangupta74 Can you add this as a GitHub issue?

tejaswinichile:
@sajangupta74 Please mention this PR there so we can come up with a solution by considering the comments on this PR and please close this PR.

PR #6508, opened 2023-02-22
feat: Conversation heatmap [DRAFT]

scmmishra:
## Description

This PR adds a conversation heatmap to reports. 

## Type of change

## How Has This Been Tested?

Tested locally, needs unit tests

## Checklist:


scmmishra:
@pranavrajs @nithindavid a few questions about the traffic heatmap

**1. How to apply filters**

Assuming it is in the conversations page. The problem here is that the original chart in the page has a time filter. Those filters are not always applicable to the heatmap. Right now we have a weekly heatmap only. There's two options. 

1. We create heatmaps for monthly, yearly as well. But then if someone selects a custom time-range, we may not be able to handle that elegantly
2. The other option is to give it a separate filter where the user can see data from the past week with a filter
    
    image

**2. Where should we place it?**

Right now I've added it to the conversation page

<details>
<summary>
Screenshot
</summary>

image

</details>

Assuming we go with approach two from the previous point. We will have to isolate the heatmap from the page level filter visually. **We can either do it on the same page or move this graph to the overview page**

pranavrajs:
@scmmishra Putting it on the overview page is an easy fix. However, I think we might need to consider having multiple charts which might not conform to the filter. 

I like how Stripe does this. They have a page-level filter which would change the metrics in some of the charts. Some other doesn't,  but there is a clear differentiation between the one which does and the one which doesn't. The sub-title makes it clear.

<details>
<summary>
An example from Stripe.
</summary>
Screenshot 2023-02-23 at 7 05 21 PM

</details>

nithindavid:
@scmmishra Moving this to the overview page sounds good

app/javascript/dashboard/routes/dashboard/settings/reports/components/Heatmap.vue
```diff
@@ -0,0 +1,235 @@
...
+    <div class="heatmap-container">
+      <div
+        v-for="dateKey in processedData.keys()"
+        :key="dateKey"
+        class="heatmap-container-row"
+      >
+        <div
+          v-for="data in processedData.get(dateKey)"
+          :key="data.timestamp"
+          v-tooltip.top="`${data.value} conversations`"
```
scmmishra, line 22: @pranavrajs @muhsin-k @nithindavid how do I translate text like this?

app/builders/v2/report_builder.rb
```diff
@@ -74,7 +74,7 @@ def get_grouped_values(object_scope)
       :created_at,
       default_value: 0,
       range: range,
-      permit: %w[day week month year],
+      permit: %w[day week month year hour],
```
scmmishra, line 77: @sojan-official I am permitting `hour` here, I need it for the heatmap. Is it okay to do so?

app/javascript/dashboard/i18n/locale/en/report.json
```diff
@@ -411,5 +414,14 @@
...
     }
+  },
+  "DAYS_OF_WEEK": {
+    "SUNDAY": "Sunday",
+    "MONDAY": "Monday",
+    "TUESDAY": "Tuesday",
+    "WEDNESDAY": "Wednesday",
+    "THURSDAY": "Thursday",
+    "FRIDAY": "Friday",
+    "SATURDAY": "Saturday"
```
scmmishra, line 425: @pranavrajs, there's [another place](https://github.com/chatwoot/chatwoot/blob/c998c84bc090b75f282a7b0a95b11c158e65a614/app/javascript/dashboard/routes/dashboard/settings/inbox/components/WeeklyAvailability.vue#L97-L105) where we have days of the week. But it's not translated there. What's the best place to add these?

app/javascript/dashboard/routes/dashboard/settings/reports/LiveReports.vue
```diff
@@ -108,6 +125,41 @@ export default {
...
+          metric: 'conversations_count',
+          from: getUnixTime(from),
+          to: getUnixTime(to),
+          groupBy: 'hour',
+          businessHours: false,
+        },
+        _options: {
+          reconcile,
+        },
+      });
```
scmmishra, line 162: Is this over-engineering 👀 
CC: @pranavrajs @nithindavid

app/javascript/dashboard/routes/dashboard/settings/reports/components/Heatmap.vue
nithindavid: @scmmishra i18n has support for variables. See `app/javascript/dashboard/i18n/locale/en/conversation.json:86` and usage will be like 
```
this.$t(
              'CONVERSATION.CARD_CONTEXT_MENU.API.AGENT_ASSIGNMENT.SUCCESFUL',
              {
                agentName: agent.name,
                conversationId,
              }
            )
          );
```

app/javascript/dashboard/i18n/locale/en/report.json
pranavrajs: Can we have a common place in i18n for these kinds of items? I'm not sure about the best practices. Can you please do some research and see how we can forward?

app/javascript/dashboard/routes/dashboard/settings/reports/LiveReports.vue
pranavrajs: @scmmishra I'm not sure if I understand this. Can you please explain a bit more on what you are trying to do?

nithindavid:
@scmmishra I'll test and review this tomorrow.

app/javascript/dashboard/routes/dashboard/settings/reports/LiveReports.vue
scmmishra: Sure, in the first case, if there's no data, the entire weeks data is fetched from the API. Since this method is called in the callback of an event, this can result in multiple API calls for this. This can be heavy for the server, so for subsequent API calls only fetch data for today. This is a significantly lighter query, the data is then reconciled with the existing data

scmmishra:
@nithindavid let me know if you had the time to test this out

app/javascript/dashboard/i18n/locale/en/report.json
scmmishra: At the moment, since all the JSON files are imported as is and the global translation directory is flattened. It's okay to store it in the `reports.json` itself. Since all the translations are loaded in their entirety, it's fine.

An alternative is to add a `global.json` file and add it there.

nithindavid:
@scmmishra During the first load before plotting the tiles, can we add loading text `Loading...`?

image

nithindavid:
@scmmishra This PR works well 🔥 Can you fix the minor I have added?

scmmishra:
@nithindavid added a custom loading state, PTAL

https://user-images.githubusercontent.com/18097732/223064815-e2b3e5ed-4c5b-4930-ae49-b2ba2586d363.mov

iamsivin (commented):
Here is the loom video. Looks good 🔥
https://www.loom.com/share/d1b8b5cbcc104d9eb12dab2dcef377fd

scmmishra:
@iamsivin resolved all the css issues. Also tweaked the color assignment a bit.

scmmishra:
> @scmmishra When will the heatmap level color change? I created five conversations, and still, the color is in level-6. image

It will change as the distribution gets wider. Since out of the 168 data points, 160 of them have zero value. Any number greater than zero will potentially be level-6

app/javascript/dashboard/routes/dashboard/settings/reports/components/Heatmap.vue
```diff
@@ -0,0 +1,307 @@
...
+  display: grid;
+  grid-template-rows: 1fr;
+  gap: $tile-gap;
+  flex-shrink: 0;
+
+  .heatmap-axis-label {
+    height: $tile-height;
+    min-width: 70px;
+    font-size: var(--font-size-micro);
+    font-weight: 600;
```
iamsivin, line 221:
```suggestion
    font-weight: var(--font-weight-bold);
```

PR #6509, opened 2023-02-22
Customizable limit (perPage) for /messages

dortonway:
## Description

Adds the ability to change the messages amount returned by the endpoint. It's required when it's necessary to unload messages via the API.


sojan-official:
@dortonway would you be able to add some specs for these changes ?

pranavrajs:
@dortonway @sojan-official I'd advise setting the max limit to 100 so that if you pass more than 100, it returns only 100 records. This change can introduce a potential vulnerability because it would take the entire system down if the messages count in a conversation is huge and the rack-timeout is disabled.

pranavrajs:
Closing this as the PR has been stale for a while. We can reopen if there is interest.

PR #6528, opened 2023-02-23
Add sender_name in the SMTP reply mails

tejaswinichile:
Fixes: https://github.com/chatwoot/product/issues/804

Screenshot 2023-02-23 at 11 34 41 PM

### DIFF
--- a/app/mailers/conversation_reply_mailer.rb
+++ b/app/mailers/conversation_reply_mailer.rb
@@ -35,7 +35,6 @@ class ConversationReplyMailer < ApplicationMailer
     init_conversation_attributes(message.conversation)
     @message = message
     reply_mail_object = prepare_mail(true)
-
     message.update(source_id: reply_mail_object.message_id)
   end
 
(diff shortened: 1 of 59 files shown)


app/mailers/conversation_reply_mailer.rb
```diff
@@ -115,6 +115,14 @@ def from_email_with_name
     end
   end
 
+  def channel_email_with_name
+    if @conversation.assignee.present?
+      I18n.t('conversations.reply.channel_email.header.reply_with_name', assignee_name: assignee_name, from_email: @channel.email)
```
pranavrajs, line 120: Let us use the same format "Pranav from Chatwoot" instead of "Pranav via hello@chatwoot.com".

app/mailers/conversation_reply_mailer.rb
```diff
@@ -115,6 +115,14 @@ def from_email_with_name
...
+    else
+      @channel.email
```
pranavrajs, line 122: I would suggest "Inbox Name <@channel.email>"

app/mailers/conversation_reply_mailer.rb
tejaswinichile: Okay 👍🏼

tejaswinichile:
Screenshot 2023-02-24 at 11 40 26 AM

vishnu-narayanan (approved):

Issue #812, opened 2020-05-03
Feature: Integrate typing indicators in Facebook channel


pranavrajs:
Integrate Typing indicators with Facebook

pranavrajs:
Closing this as we will not be picking this up anytime soon. We will reopen if we reprioritize the feature.

PR #6533, opened 2023-02-24
fix: Email subject automation issue

tejaswinichile:
fix: [#812](https://github.com/chatwoot/product/issues/812)

https://www.loom.com/share/ff9d30e0ce8841f09ebb6f6176069595

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -5,7 +5,7 @@
 npx --no-install lint-staged
 
 # lint only staged ruby files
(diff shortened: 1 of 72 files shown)


sojan-official (commented):
@tejaswinichile can we add some specs for the same ?

pranavrajs:
@tejaswinichile I'm not sure if I understand the change. Was that the bug happening because we missed adding the filter in the file?

tejaswinichile:
> I'm not sure if I understand the change. Was the bug happening because we missed adding the filter in the file?

@pranavrajs The attribute type was necessary else. It was matching with everything.

sojan-official (approved):

PR #6535, opened 2023-02-24
chore: add `--force-exclusion` option to rubocop pre-commit hook

scmmishra:
## Description

The pre-commit hook would format event those files excluded in the config. This is because the files were passed as args instead of the linter running on the entire project. When this happens, the exclude is not respect. 

The `--force-exclusion` flag instructs our machine overlords to force exlucde files specified in the configuration `Exclude` even if they are explicitly passed as arguments.

## Type of change

## Tested 

image

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -5,7 +5,7 @@
 npx --no-install lint-staged
 
 # lint only staged ruby files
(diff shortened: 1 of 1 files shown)


sojan-official (approved):
LGTM

Issue #5482, opened 2022-09-22
Bug: Meta tags input in help center clears the input value after clicking outside.


fayazara:
**Describe the bug**
Happens in help center article editor page.  When we add some meta tags and click outside the input, the tags just get cleared on its own.

https://user-images.githubusercontent.com/15716057/191674697-d946530f-2998-4f81-b206-3940bbe9b47b.mov

**To Reproduce**

1. Go to Help center
2. Click on "New Article"
3. Add some meta tags and click outside
4. You can see that the meta tags get cleared.

BikashSah999:
Hello @fayazara  @pranavrajs, from my spec I found that we need to support auto adding new tags when user press comma, and also add the tags in the input when user blurs the input. WDYT, would this be an acceptable resolution strategy?

iamsivin:
@BikashSah999 Thanks for the findings. I think we can go with the `preserveSearch` prop in multi-select. 
CC: @pranavrajs @nithindavid 

@BikashSah999 Are you working on this issue? If you are, I can assign this issue to you, and you can raise a PR.

PR #6552, opened 2023-02-27
fix : Meta tags input in help center clears the input value after clicking outside

gitstart:
Fixes https://github.com/chatwoot/chatwoot/issues/5482

## Description

When we enter something in the meta tags input of the article and then clicked outside, the entered value of meta tags input gets cleared. This PR implements the preserving the search value in meta tags input.

## Type of change

## How Has This Been Tested?

- Go to an article
- Enter something in meta tags input and click outside
- When we again go to meta tags input, the previously entered input is preserved

## Checklist:

## Demo video
https://www.loom.com/share/09080e29cea1438baf5ed8a17363cb61


iamsivin:
Issue https://github.com/chatwoot/chatwoot/issues/5482

app/javascript/dashboard/routes/dashboard/helpcenter/pages/articles/ArticleSettings.vue
```diff
@@ -87,6 +87,7 @@
             track-by="name"
             :multiple="true"
             :taggable="true"
+            :preserve-search="true"
```
scmmishra, line 90: What does the `preserve-search` do here? Doesn't seem like a valid prop for the element? As far as I know, there is no standard HTML attribute or property named `preserve-search` either

scmmishra (commented):
About the behaviour of this, given there is no explicit "save" button. On blur, we should save what the user has entered as tags. 

WDYT @pranavrajs @nithindavid?

pranavrajs:
> About the behaviour of this, given there is no explicit "save" button. On blur, we should save what the user has entered as tags.

Yes, we should save the tags on blur.

app/javascript/dashboard/routes/dashboard/helpcenter/pages/articles/ArticleSettings.vue
scmmishra: Noted

app/javascript/dashboard/routes/dashboard/helpcenter/pages/articles/ArticleSettings.vue
```diff
@@ -184,13 +186,18 @@ export default {
...
-      this.metaTags.push(tag);
-      this.$refs.tagInput.$el.focus();
+      const tags = tagValue.split(',').map(value => ({
+        name: value.trim(),
+      }));
+      this.metaTags.push(...tags);
       this.saveArticle();
     },
+    onBlur() {
+      const currentSearch = this.$refs.tagInput.$refs.search.value;
```
scmmishra, line 196: This looks a bit hacky, but the `vue-multiselect` is not the ideal candidate for this kind of component with arbitrary tags and no list to select form. 

An alternative approach is that we could listen to the `@search-change` event. And store the result in a variable, onBlur we just assign the value to `currentSearch` and then reset the variable.

This is cleaner and does not leak the implementation details of the underlying component

scmmishra (changes requested):

app/javascript/dashboard/routes/dashboard/helpcenter/pages/articles/ArticleSettings.vue
gitstart: @scmmishra Thanks for the suggestion, it works fine.

scmmishra (approved):

PR #6559, opened 2023-02-28
fix: Fix i18n issues with help center

pranavrajs:
Fixes the following

- Fix search in the public portal with multiple locales.
- Fix the i18n selection in the portal.
- Fix the issues with locale variants used in the portal.
- Fix the language, and plurals.


iamsivin:
Issue https://github.com/chatwoot/product/issues/805

Issue #6570, opened 2023-03-01
[CW-15] Dyte Video Call Button Only Shown if User Role is Administrator


punyaikhwan:
**Describe the bug**
If an agent is set as Agent (not administrator), the Dyte video call button is not shown.

**To Reproduce**
1. As administrator, setup Dyte integration.
2. Go to an Inbox, select a conversation, then see that video call button is shown next to the "record audio" button.
![image](https://user-images.githubusercontent.com/14848437/222059119-2430fb89-3baf-410a-9a08-aeecaa26f2bc.png)
3. Login as another agent who has Agent role, not administrator.
4. Do step number 2. The video call button is not shown.
![image](https://user-images.githubusercontent.com/14848437/222059295-2556bb8b-9485-4778-b45a-c8a96e5b45a5.png)

**Expected behavior**
The video call button should be shown in all agent role type.

**Environment**
Self hosted in Linux. Chatwoot v2.14.0.

**Desktop (please complete the following information)** (If applicable)
 - OS: Windows 11
 - Browser: Edge, Chrome

<sub>[CW-15](https://linear.app/chatwoot/issue/CW-15/dyte-video-call-button-only-shown-if-user-role-is-administrator)</sub>

PR #6587, opened 2023-03-02
fix: Allow integration apps to be listed by an agent

pranavrajs:
Fixes https://github.com/chatwoot/chatwoot/issues/6570

- Allow listing the connected apps without exposing credentials


scmmishra (approved):
LGTM

PR #6406, opened 2023-02-07
fix: Setting a fallback name from the email address for MS email channel

moxvallix:
## Description
Fixes #6353, by using the email address (part before the "@") as a fallback name.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?
I made this change, then managed to have the inbox successfully create in my own deploy.
As it is such a small change, no formal tests have been written.

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #6591, opened 2023-03-02
Fallback name branch

tejaswinichile:
Added specs for https://github.com/chatwoot/chatwoot/pull/6406

Fixes: https://github.com/chatwoot/chatwoot/pull/6406


app/controllers/microsoft/callbacks_controller.rb
```diff
@@ -45,13 +45,20 @@ def find_or_create_inbox
     channel_email.inbox
   end
 
+  # Fallback name, for when name field is missing from users_data
+  def fallback_name
+    return 'New Inbox' if users_data['email'].blank?
+
```
sojan-official, line 51: do we need this condition. users_data['email'] can't be blank in any case right. we will need it for creating the inbox.

tejaswinichile: I think this is just extra check that they have added, but this will not and should not be the case.

Let me update it.

sojan-official (approved):

PR #6615, opened 2023-03-06
feat: add trigram index to tags

scmmishra:
## Description

Fixes: https://github.com/chatwoot/product/issues/796

This migration adds a trigram index to the tags table on the name column. This is used to speed up the search for tags by name, this is a heavy sub-query when generating reports where labels are involved.

Trigram indexes are used to speed up LIKE queries, they are not used for equality queries. This is because the index is not a btree index, it is a GIN index. This means that the index is not ordered, and so cannot be used for equality queries.

Read more: https://www.postgresql.org/docs/current/pgtrgm.html

The index is created concurrently, [read more here](https://thoughtbot.com/blog/how-to-create-postgres-indexes-concurrently-in)

## Type of change

#### Before Index

```sql
EXPLAIN ANALYZE SELECT tags.id FROM tags WHERE LOWER(tags.name) ILIKE 'fake-enim-155' ESCAPE '!';
```

```sql
"QUERY PLAN"
"Seq Scan on public.tags  (cost=0.00..136.31 rows=1 width=4) (actual time=12.572..12.586 rows=1 loops=1)"
"  Output: id"
"  Filter: (lower((tags.name)::text) ~~* 'fake-enim-155'::text)"
"  Rows Removed by Filter: 6220"
"  Buffers: shared hit=43"
"Planning:"
"  Buffers: shared hit=13 dirtied=1"
"Planning Time: 0.426 ms"
"Execution Time: 12.603 ms"
```

### After Index 

```sql
CREATE INDEX tags_name_trgm_idx ON tags USING gin (LOWER(name) gin_trgm_ops);
EXPLAIN ANALYZE SELECT tags.id FROM tags WHERE LOWER(tags.name) ILIKE 'fake-enim-155' ESCAPE '!';
```

```sql
"QUERY PLAN"
"Bitmap Heap Scan on public.tags  (cost=116.00..120.02 rows=1 width=4) (actual time=0.622..0.623 rows=1 loops=1)"
"  Output: id"
"  Recheck Cond: (lower((tags.name)::text) ~~* 'fake-enim-155'::text)"
"  Heap Blocks: exact=1"
"  Buffers: shared hit=35"
"  ->  Bitmap Index Scan on tags_name_trgm_idx  (cost=0.00..116.00 rows=1 width=0) (actual time=0.574..0.574 rows=1 loops=1)"
"        Index Cond: (lower((tags.name)::text) ~~* 'fake-enim-155'::text)"
"        Buffers: shared hit=34"
"Planning:"
"  Buffers: shared hit=1"
"Planning Time: 0.313 ms"
"Execution Time: 0.664 ms"
```

We can see a drastic reduction in the execution time of the query. With the new index, instead of sequential scan. A Bitmap index scan is used, which is much more effecient 

## Checklist:


db/migrate/20230306100327_add_trgm_index_to_tag_name.rb
```diff
@@ -0,0 +1,24 @@
...
+#
+# Read more: https://www.postgresql.org/docs/current/pgtrgm.html
+
+class AddTrgmIndexToTagName < ActiveRecord::Migration[6.1]
+  # disabling ddl transaction is required for concurrent index creation
+  # https://thoughtbot.com/blog/how-to-create-postgres-indexes-concurrently-in
+  disable_ddl_transaction!
+
+  def change
+    enable_extension 'pg_trgm'
```
sojan-official, line 18: This should already be enabled from the previous migration. we can skip this line. otherwise looks good to me.

PR #6641, opened 2023-03-09
chore: Set statement timeout for Postgres

sojan-official:
By default, Rails does not set a timeout on database statements. For example, this will run for a full day, even if your ruby process goes away. But it's configurable in the `database.yml` with the `statement_timeout` variable.

Hence we are enforcing a `14s` timeout by default. Migration commands inside chatwoot will run with a `10 minutes` timeout. For specific cases like migrations, we can override this timeout using the environment variable `POSTGRES_STATEMENT_TIMEOUT` while starting a new rails console. 

Test the timeouts from rails console using 
```
ActiveRecord::Base.connection.execute("SELECT pg_sleep(15);")
```

ref: https://github.com/ankane/the-ultimate-guide-to-ruby-timeouts#postgresql
ref: https://til.hashrocket.com/posts/b44baf657d-railspg-statement-timeout-

### DIFF
--- a/app/controllers/api/v1/accounts/articles_controller.rb
+++ b/app/controllers/api/v1/accounts/articles_controller.rb
@@ -55,9 +55,9 @@ class Api::V1::Accounts::ArticlesController < Api::V1::Accounts::BaseController
 
   def article_params
     params.require(:article).permit(
-      :title, :slug, :content, :description, :position, :category_id, :author_id, :associated_article_id, :status, meta: [:title,
-                                                                                                                          :description,
-                                                                                                                          { tags: [] }]
+      :title, :slug, :position, :content, :description, :position, :category_id, :author_id, :associated_article_id, :status, meta: [:title,
+                                                                                                                                     :description,
+                                                                                                                                     { tags: [] }]
     )
   end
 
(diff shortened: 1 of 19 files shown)


config/database.yml
```diff
@@ -5,7 +5,9 @@ default: &default
   port: <%= ENV.fetch('POSTGRES_PORT', '5432') %>
    # ref: https://github.com/mperham/sidekiq/issues/2985#issuecomment-531097962
   pool: <%= Sidekiq.server? ? Sidekiq.options[:concurrency] : ENV.fetch('RAILS_MAX_THREADS', 5) %>
-
+  variables:
+    statement_timeout: <%= ENV["STATEMENT_TIMEOUT"] || "5s" %>
```
tejaswinichile, line 9: You need to update the `env.example` file for this it says`POSTGRES_STATEMENT_TIMEOUT`

sojan-official: Nice catch. we intend to use `POSTGRES_STATEMENT_TIMEOUT` .will update it here.

.env.example
```diff
@@ -52,6 +52,8 @@ POSTGRES_HOST=postgres
 POSTGRES_USERNAME=postgres
 POSTGRES_PASSWORD=
 RAILS_ENV=development
+# Customise this only when required
```
vishnu-narayanan, line 55:
```suggestion
# Changes the Postgres query timeout limit. The default is 5 seconds. Modify only when required.
```

pranavrajs:
@sojan-official Can we do a sweep of the slowest queries on our prod just to make sure that none of it is crossing the 5s time limit?

sojan-official:
@pranavrajs  There is no specific query that requires 5 seconds in all cases. But I have seen instances in APM where some queries overrun this limit in specific cases.  Hence, switching the timeout to 14s to be around the same as the rack timeout limit and then working backwards, looking into logs for cases with statement timeout errors, fixing on case to case basis.

PR #6646, opened 2023-03-09
fix: hook exception for empty hook

tejaswinichile:
Sentry Issue: [EXTERNAL-SOCIALBOT-5R](https://chatwoot-p3.sentry.io/issues/3990919574/?referrer=github_integration)

### DIFF
--- a/lib/integrations/bot_processor_service.rb
+++ b/lib/integrations/bot_processor_service.rb
@@ -1,4 +1,5 @@
 class Integrations::BotProcessorService
+  # TODO: In CSML processor service, the argument is agent bot, update initializers accordingly.
   pattr_initialize [:event_name!, :hook!, :event_data!]
 
   def perform
@@ -7,7 +8,7 @@ class Integrations::BotProcessorService
 
     process_content(message)
   rescue StandardError => e
-    ChatwootExceptionTracker.new(e, account: hook.account).capture_exception
+    ChatwootExceptionTracker.new(e, account: (hook&.account || agent_bot&.account)).capture_exception
   end
 
   private


lib/integrations/bot_processor_service.rb
```diff
@@ -7,7 +7,7 @@ def perform
```
tejaswinichile, line 7: @sojan-official `:hook!` attributes shouldn't it match `:agent_bot!` in the child class in the initializer, changing the sequence here. Because seems hook is always nil in this case.

lib/integrations/bot_processor_service.rb
```diff
@@ -7,7 +7,7 @@ def perform
 
     process_content(message)
   rescue StandardError => e
-    ChatwootExceptionTracker.new(e, account: hook.account).capture_exception
+    ChatwootExceptionTracker.new(e, account: hook&.account).capture_exception
```
tejaswinichile, line 10: This will fix the sentry issue, but it will not answer why the hook is nil.

lib/integrations/bot_processor_service.rb
sojan-official: @tejaswinichile  it wouldn't match `hook` . the gem defines a new setter and getter methods named `agent_bot` .
https://github.com/barsoom/attr_extras/blob/4913a40d2cea9cd90fbdc444441e35268cf3c6c3/lib/attr_extras/attr_initialize.rb#L23

So in case of CMSL processor service hook will be nill.  ( we might have to look into the naming ) . but for now we could do  `(hook&.account || agent_bot&.account) `

also add a comment mentioning that in CSML processor service, the argument is agent bot. 

or we could just rename it to `bot_hook` or `bot_object` something and make this consistent across the base classes and inherited classes.

tejaswinichile: > Doing this for quick fix (hook&.account || agent_bot&.account) 

> or we could just rename it to bot_hook or bot_object something and make this consistent across the base classes and inherited classes.

<< I will take this one as a later ticket.

vishnu-narayanan (approved):

PR #6658, opened 2023-03-14
feat(poc): Disable widget based on country

pranavrajs:
- Disable widget based on country code from IP address

https://www.loom.com/share/f2363462485b4aa8a6826f77a8964bf7

### Changes:

The database download has been moved to a rake task and will be downloaded during boot up. Dynamically fetching (async fetches via Sidekiq) the database does not work well with synchronous calls since each server requires a copy of the file.

I updated the Procfile to download the DB initially and then run the service afterwards. This change may increase the bootup time by 3-4 seconds, depending on network latency.

I removed the IP_LOOKUP_SERVICE variable because none of the other services will work without the lookup parameter. Therefore, managing this variable is pointless.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
 release: POSTGRES_STATEMENT_TIMEOUT=600s bundle exec rails db:chatwoot_prepare
-web: bin/rails server -p $PORT -e $RAILS_ENV
-worker: bundle exec sidekiq -C config/sidekiq.yml
+web: bundle exec rails ip_lookup:setup && bin/rails server -p $PORT -e $RAILS_ENV
+worker: bundle exec rails ip_lookup:setup && bundle exec sidekiq -C config/sidekiq.yml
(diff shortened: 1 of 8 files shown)


tejaswinichile (approved):
LGTM 👍🏼

Procfile
```diff
@@ -1,3 +1,3 @@
 release: POSTGRES_STATEMENT_TIMEOUT=600s bundle exec rails db:chatwoot_prepare
-web: bin/rails server -p $PORT -e $RAILS_ENV
-worker: bundle exec sidekiq -C config/sidekiq.yml
+web: bundle exec rails ip_lookup:setup && bin/rails server -p $PORT -e $RAILS_ENV
+worker: bundle exec rails ip_lookup:setup && bundle exec sidekiq -C config/sidekiq.yml
```
sojan-official, line 3: We could trigger this from the initializer. It would ensure that it stays consistent across all installation environments with out updating the documentation.

sojan-official (approved):
LGTM

Procfile
pranavrajs: @sojan-official Can you please make the change if it is a small one?

PR #6660, opened 2023-03-14
Feat: detect language of the message content

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-46/reduce-spam

https://user-images.githubusercontent.com/10827585/224950011-488eeb82-c907-4d95-963a-31d9f6076c78.mov


sojan-official (commented):
- added some comments

app/jobs/conversations/detect_language_job.rb
```diff
@@ -0,0 +1,37 @@
+class Conversations::DetectLanguageJob < ApplicationJob
+  queue_as :default
+
+  def perform(message_id)
+    return unless GlobalConfigService.load('ENABLE_LANGUAGE_DETECTION', '')
+
+    @message = Message.find(message_id)
+
+    return unless @message.content.size < 1500
```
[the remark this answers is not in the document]
tejaswinichile: @sojan-official When we checked the spam messages, it was more like less content messages, that's why added this.

app/jobs/conversations/detect_language_job.rb
```diff
@@ -0,0 +1,37 @@
+class Conversations::DetectLanguageJob < ApplicationJob
+  queue_as :default
+
+  def perform(message_id)
+    return unless GlobalConfigService.load('ENABLE_LANGUAGE_DETECTION', '')
+
```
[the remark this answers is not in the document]
tejaswinichile: We have a hook present for translation purposes in our cloud account, but we want this check to only take place when  ENABLE_LANGUAGE_DETECTION is on our account.

app/jobs/conversations/detect_language_job.rb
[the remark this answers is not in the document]
tejaswinichile: I am hoping the long message in another language might be the real concerned message.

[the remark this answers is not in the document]
sojan-official: makes sense

app/jobs/conversations/detect_language_job.rb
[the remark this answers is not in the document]
sojan-official: `ENABLE_LANGUAGE_DETECTION` is a global config, ie works across the installation. But not all the accounts in the installation would have a translation integration configured.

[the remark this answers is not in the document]
tejaswinichile: @sojan-official This makes sense. I will have a hook base check; in addition to that, I am thinking of adding account-level settings with `settings_flags[ACCOUNT_SETTINGS_FLAGS]` . Even if the translation is turned on, only some accounts should flag/report spam if the query is from a different language.

tejaswinichile:
@sojan-official We can tie this over automation rule and report spam or resolve it. I am removing the settings; in this PR we can just have detectLanguageJob.

pranavrajs:
@tejaswini I don't think this is developed as intended. 

1/ The language can be an attribute of the conversation.
2/ The logic to resolve the conversation can be a new automation.

@sojan-official I'm not sure if I understand the following.

> Make message language detection a part of the google translate integration

Isn't it already taking the values from Google Translate Integration?

tejaswinichile:
> Isn't it already taking the values from Google Translate Integration? 

Yes, it is.

> 2/ The logic to resolve the conversation can be a new automation.

This has been handled in the latest commit.

> 1/ The language can be an attribute of the conversation.

Do you mean for language detection or save it as an additional attribute in the conversation instead of resolving it?

pranavrajs:
@tejaswinichile 
```
conversation.update_columns(status: :resolved) unless current_language == 'en'
```

Here the status and the language is hard coded. Instead this could have been an automation where condition is "conversation_language = en", then actions: "resolve / snooze" or what ever is required.

sojan-official:
> @tejaswini I don't think this is developed as intended.
> 
> 1/ The language can be an attribute of the conversation. 2/ The logic to resolve the conversation can be a new automation.
> 
> @sojan-official I'm not sure if I understand the following.
> 
> > Make message language detection a part of the google translate integration
> 
> Isn't it already taking the values from Google Translate Integration?

The PR diverged from the time this comment was made.

tejaswinichile:
Event Type: Message.Created
Condition: Based on anything
Action: Detect Language will have params array of valid language codes.

1. Detect language, will be visible only for the Mesage.Created event.
2. If detect_language doesn't have any action param then default we will consider `en`

tejaswinichile:
@pranavrajs @sojan-official 

I have handled this in the automation rule. I still need clarification about ensuring the admin/agent knows he is selecting a valid language.

I have renamed the action from report_spam to validate_language, but it still needs clarification.
Can we have a page for action information, or should I add it to the documentation?

https://user-images.githubusercontent.com/10827585/225859454-a5e07a91-589d-499e-8ecf-cb9c0280b4c2.mov

lib/automation_rules/conditions.json
```diff
@@ -63,6 +63,13 @@
       "filter_operators": [ "equal_to", "not_equal_to", "contains", "does_not_contain" ],
       "attribute_type": "additional_attributes"
     },
+    "current_language": {
+      "attribute_name": "Conversation Languages",
```
sojan-official, line 67: `Conversation Language` ?

app/javascript/dashboard/routes/dashboard/settings/automation/constants.js
```diff
@@ -132,6 +132,13 @@ export const AUTOMATIONS = {
         inputType: 'multi_select',
         filterOperators: OPERATOR_TYPES_1,
       },
+      {
```
sojan-official, line 135: let's call the attribute `conversation_language` itself for consistency. `current_language` might be confusing to the end users

app/javascript/dashboard/routes/dashboard/settings/automation/constants.js
```diff
@@ -249,6 +256,13 @@ export const AUTOMATIONS = {
...
+        key: 'current_language',
```
sojan-official, line 260: `conversation_language`

lib/filters/filter_keys.json
```diff
@@ -63,6 +63,13 @@
       "filter_operators": [ "equal_to", "not_equal_to", "contains", "does_not_contain" ],
       "attribute_type": "additional_attributes"
     },
+    "current_language": {
+      "attribute_name": "Conversation Language",
+      "input_type": "textbox",
+      "data_type": "text",
+      "filter_operators": [ "equal_to", "contains" ],
```
sojan-official, line 70: "filter_operators": [ "equal_to", "not_equal_to" ] ?

lib/filters/filter_keys.json
```diff
@@ -142,13 +149,6 @@
       "filter_operators": [ "equal_to", "not_equal_to", "contains", "does_not_contain" ],
       "attribute_type": "additional_attributes"
     },
-    "browser_language": {
-      "attribute_name": "Browser Language",
-      "input_type": "textbox",
-      "data_type": "text",
-      "filter_operators": [ "equal_to", "not_equal_to", "contains", "does_not_contain" ],
-      "attribute_type": "additional_attributes"
-    },
```
sojan-official, line 151: any specific reason why browser language is removed ?

spec/listeners/automation_rule_listener_spec.rb
```diff
@@ -584,6 +588,41 @@
...
+        conditions: [{ 'values': ['en'], 'attribute_key': 'current_language', 'query_operator': nil, 'filter_operator': 'not_equal_to' }],
+        actions: [{ 'action_name' => 'resolve_conversation' }.with_indifferent_access]
+      )
+    end
+
+    let(:client) { double }
+    let(:message) { create(:message, account: account, conversation: conversation, message_type: 'incoming', content: 'muchas muchas gracias') }
+
+    context 'when rule matches' do
+      it 'triggers automation rule to report spam and resolves the conversation' do
```
sojan-official, line 606:
```suggestion
      it 'triggers the automation rule to resolves the conversation' do
```

spec/listeners/automation_rule_listener_spec.rb
```diff
@@ -584,6 +588,41 @@
     end
   end
 
+  describe '#conversation_updated to to detect current language' do
```
sojan-official, line 591:
```suggestion
  describe '#conversation_updated with conversation_language' do
```

spec/listeners/automation_rule_listener_spec.rb
```diff
@@ -584,6 +588,41 @@
...
+    context 'when rule matches' do
+      it 'triggers automation rule to report spam and resolves the conversation' do
+        perform_enqueued_jobs do
+          create(:integrations_hook, :google_translate, account_id: account.id)
+          response = Google::Cloud::Translate::V3::DetectLanguageResponse.new({ languages: [{ language_code: 'es', confidence: 0.71875 }] })
+
+          allow(Google::Cloud::Translate).to receive(:translation_service).and_return(client)
+          allow(client).to receive(:detect_language).and_return(response)
+          message
+        end
```
sojan-official, line 614: We don't need to process the `translate` related jobs here? We can test them separately in a test for `detect_language_job` .  In this case lets create the conversation object with `conversation_language` populated.

spec/listeners/automation_rule_listener_spec.rb
```diff
@@ -584,6 +588,41 @@
...
+
+        event = Events::Base.new('conversation_updated', Time.zone.now, { conversation: conversation, message: message })
+
+        listener.conversation_updated(event)
+
+        expect(message.conversation.reload.status).to eq('resolved')
+        expect(message.conversation.additional_attributes['current_language']).to eq('es')
```
sojan-official, line 621: we don't need to test the current language of conversation since that is not the responsibility of automation listener.

sojan-official (commented):
I added some comments. Lets do the following


- [x]  - Rename the attribute to `conversation_language`
- [x]   - Move the logic to hook listener / hook job since the logic is super tied to an integration hook.
- [x]   - Could reduce the API calls after initial language detection
- [x]   - Specs for Detect Language job
- [x]   - Lets also test the functionality end to end by configuring and automation with this rule before merging

app/jobs/conversations/detect_language_job.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  private
+
+  def update_conversation(response)
+    conversation = @message.conversation
+    current_language = response.languages.first.language_code
+    additional_attributes = conversation.additional_attributes.merge({ current_language: current_language })
+    # rubocop:disable Rails/SkipsModelValidations
+    conversation.update_columns(additional_attributes: additional_attributes)
```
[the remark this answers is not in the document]
tejaswinichile: > The automation rule listener is tied to conversation_updated?

No. It is available for all the events

> if we use update_column, how will the subsequent automation rule be triggered?

We want to avoid triggering automation while updating the language. As this is happening as a background job to update additional_attribute for conversation_language when we create the message.

spec/listeners/automation_rule_listener_spec.rb
```diff
@@ -1,5 +1,9 @@
 require 'rails_helper'
+require 'google/cloud/translate/v3'
```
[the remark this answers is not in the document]
tejaswinichile: Ye, to use the ` Google::Cloud::Translate::V3`

spec/listeners/automation_rule_listener_spec.rb
tejaswinichile: Oh with `conversation_language` attribute then we don't need `require 'google/cloud/translate/v3'` here.

app/jobs/conversations/detect_language_job.rb
```diff
@@ -0,0 +1,38 @@
...
+    @message = Message.find(message_id)
+
+    return if hook.blank?
+
+    text = @message.content[0...1500]
+
+    response = client.detect_language(
+      content: text,
+      parent: "projects/#{hook.settings['project_id']}"
+    )
```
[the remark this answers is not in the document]
tejaswinichile: We are calling it on the first incoming message.

app/jobs/conversations/detect_language_job.rb
```diff
@@ -0,0 +1,42 @@
...
+  def update_conversation(response)
+    conversation = @message.conversation
+    conversation_language = response.languages.first.language_code
+    additional_attributes = conversation.additional_attributes.merge({ conversation_language: conversation_language })
+
+    conversation.update!(additional_attributes: additional_attributes)
+
+    event = Events::Base.new('conversation_updated', Time.zone.now, { conversation: conversation })
+
+    AutomationRuleListener.instance.conversation_updated(event)
```
tejaswinichile, line 30: @sojan-official I Had to add this manually as we don't support the 'conversation_updated' event in additional_attributes.

```
def notify_conversation_updation
    return unless previous_changes.keys.present? && (previous_changes.keys & %w[team_id assignee_id status snoozed_until custom_attributes label_list first_reply_created_at]).present?

    dispatcher_dispatch(CONVERSATION_UPDATED, previous_changes)
end
  ```

I didn't want to call it for all other listeners and grab the impact on this change.

Without this, it will work for other attributes on the conversation when you update something manually on the dashboard, not when you detect and update the language and then want to run the automation based on that. So for this case, I added the line.

Let me know wdyt?

sojan-official:
Note to self:

ref: https://github.com/chatwoot/chatwoot/pull/6660#discussion_r1149106533 
Pending: We need to address this case. 

It feels wrong to trigger the update event directly from the integration processor.  Lets do that from the conversation model itself by expanding update event to `additional_attribute` ->  `conversation_language` case alone, with relevant specs


https://github.com/chatwoot/chatwoot/blob/21da03fe5bbcda0e886f0f2ec787ce09917191b6/app/models/conversation.rb#L214


https://github.com/chatwoot/chatwoot/blob/21da03fe5bbcda0e886f0f2ec787ce09917191b6/spec/models/conversation_spec.rb#L157

scmmishra (approved):

tejaswinichile:
Screenshot 2023-04-04 at 12 28 26 PM
Screenshot 2023-04-04 at 12 28 14 PM

PR #6661, opened 2023-03-14
fix: assignment is not added as participant

scmmishra:
Fixes: https://github.com/chatwoot/product/issues/800


sojan-official:
- closing since we already merged the fix for this

PR #6667, opened 2023-03-14
fix: save twitter profile for inbox

### LINKED ISSUE #737: Chore: Twitter Profile images are not stored to the inbox
Chatwoot usually stores an avatar for each of the inboxes connect, this is missing for the twitter inboxes. 

to get the user avatar we might need to make a separate API call to https://developer.twitter.com/en/docs/accounts-and-users/follow-search-get-users/api-reference/get-users-show 

hence the method will have to be added to https://github.com/chatwoot/twitty

tejaswinichile:
Fixes #737 

Can be merged only when [Twitty Changes for User Show](https://github.com/chatwoot/twitty/compare/master...develop) is merged.

This has been merged to develop: https://github.com/chatwoot/twitty/pull/26


app/controllers/twitter/callbacks_controller.rb
```diff
@@ -49,10 +49,18 @@ def create_inbox
...
+    inbox = account.inboxes.create!(
       name: parsed_body['screen_name'],
       channel: twitter_profile
     )
+    save_profile_image(inbox)
+    inbox
+  end
+
+  def save_profile_image(inbox)
+    response = twitter_client.user_show(screen_name: inbox.name)
```
pranavrajs, line 61: Should we rescue from failures here?

tejaswinichile: This is already being handled in the transaction. The main function has exception handling.

sojan-official: @tejaswinichile  lets add a rescue for the function or add a condition. We don't need to prevent the inbox creation in case some error occurs while fetching the picture

```suggestion
    response = twitter_client.user_show(screen_name: inbox.name)
    if response. success?
       parsed_user_profile = JSON.parse(response.read_body)
       ::Avatar::AvatarFromUrlJob.perform_later(inbox, parsed_user_profile['profile_image_url_https'])
    end
```

spec/controllers/twitter/callbacks_controller_spec.rb
```diff
@@ -32,6 +38,7 @@
       allow(webhook_service).to receive(:perform).and_raise StandardError
       get twitter_callback_url
       account.reload
+      # expect(Avatar::AvatarFromUrlJob).not_to receive(:perform_later).with(account.inboxes.last, '')
```
sojan-official, line 41: This shouldn't be commented out ? 
We should also add  avatar job expectation for the inbox creation condition

sojan-official (commented):
Can we also have a loom video?

sojan-official (commented):
Let's have the tests ensuring that `::Avatar::AvatarFromUrlJob` is enqueued, which is a core change in this PR

tejaswinichile:
> Let's have the tests ensuring that `::Avatar::AvatarFromUrlJob` is enqueued, which is a core change in this PR

@sojan-official This has been addressed

sojan-official (approved):
LGTM

PR #6683, opened 2023-03-16
[CW-53] feat: allow downloading heatmap report 

scmmishra:
## Description

This PR allows users to download the data from the heatmap.

## Type of change

## Changes

https://user-images.githubusercontent.com/18097732/225861738-d6555ebb-8127-49d4-a313-8e9d520f81f8.mov

## Checklist:


scmmishra:
@nithindavid @pranavrajs I've tried something different for the UI for the button, the opacity of it is set to `0.2` and it goes to `1` when the card is hovered. Need your thoughts on it.

The logic behind this was eventually for the reports, when we have multiple cards, each report can have their individual actions in the cards, say "download" or adding specific filters, in such a case having multiple cards with buttons can look distracting. 

Since now there's just one of those, I can remove this effect.

app/helpers/api/v2/accounts/reports_helper.rb
```diff
@@ -27,6 +27,32 @@ def generate_labels_report
...
+    #
+    # since the data from the report builder is already timezone aware,
+    # the conversion to date and hour is done without timezone
+    # rubocop:disable Rails/TimeZone
+    data.map do |d|
+      date = Time.at(d[:timestamp]).to_date
+      hour = Time.at(d[:timestamp]).hour
+      [date, hour, d[:value]]
+    end
+    # rubocop:enable Rails/TimeZone
```
scmmishra, line 53: @sojan-official need your thoughts on disabling this rule, I've added the reasons in the comments, is there a better way to do this without `Time.at`

sojan-official: @scmmishra  The comment makes sense to me. But if we can get through with using `Time.zone.at` or `Time.at.in_time_zone` ref: https://www.rubydoc.info/gems/rubocop/0.57.2/RuboCop/Cop/Rails/TimeZone 

It would be better if we could get rid of the comment as well as disabling the rule.

scmmishra: I tried using `Time.zone.at` but it would apply the timezone calculation to already existing timezone adjusted value, so the 12:00 AM records show up in 5:30 AM records

pranavrajs:
> @nithindavid @pranavrajs I've tried something different for the UI for the button, the opacity of it is set to 0.2 and it goes to 1 when the card is hovered. Need your thoughts on it.

Looks good to me.

nithindavid:
LGTM

nithindavid:
Tested locally and the reports are downloading correct data.

nithindavid (approved):
FE part looks good

sojan-official (approved):

PR #6685, opened 2023-03-16
chore: Upgrade prosemirror package to fix link rendering issue

nithindavid:
Fixes: https://github.com/chatwoot/product/issues/781


pranavrajs:
Tested, it seems to be working properly.

PR #6686, opened 2023-03-17
Added PATCH support on  for conversations

nikhilmuz:
# Pull Request Template

## Description

This allows updating a particular attribute in conversation without sending the entire custom attribute payload as POST request. This comes in handy when editing different custom attributes too frequently and it avoids race conditions between two API hits where we first get the custom attribute and mutate the JSON payload and then push it.

## Type of change

## How Has This Been Tested?

I have checked the API on Postman and it seems to be working with Status 204

## Checklist:


app/controllers/api/v1/accounts/conversations_controller.rb
```diff
@@ -89,6 +89,12 @@ def custom_attributes
     @conversation.save!
   end
 
+  def patch_custom_attributes
+    custom_attributes = params.permit(custom_attributes: {})[:custom_attributes]
+    sanitized_json = custom_attributes.to_json.gsub("'", "''")
+    Current.account.conversations.where(id: @conversation.id).update_all("custom_attributes = custom_attributes || '#{sanitized_json}'::jsonb")
```
sojan-official, line 95: @nikhilmuz We already have `@conversation` available.  so the `update_all` doesn't look the right way to go about this. 
lets do something like 

```
@conversation.custom_attributes = (@conversation.custom_attributes || {}).deep_merge(params[:custom_attributes].stringify_keys)
@conversation.save!
```

ref: https://github.com/chatwoot/chatwoot/blob/71c5a1e1d4d46170abf4d19c708cc3cba745a949/app/actions/contact_identify_action.rb#L129

sojan-official (commented):
Could you also add the relevant specs?

app/controllers/api/v1/accounts/conversations_controller.rb
nikhilmuz: The idea was to avoid race condition so doing it on the ruby layer doesn't seems right to me. Basically, it fetches conversation somewhere in the controller and we do a deep merge with a new object and reinsert it overwriting all the existing data on parallel API calls. This way we are ensuring lock on the Postgres layer.

nikhilmuz: @sojan-official

sojan-official: @nikhilmuz We want the callbacks to happen on custom attribute updates for automation/webhooks to be triggered.

sojan-official:
Closing this since there hasn't been any activity on this for a while. Let's reopen if required.

Issue #4196, opened 2022-03-16
Create InstallationConfig migration breaks


matteomelotti:
**Describe the bug**

Migration '20200510112339_create_installation_config' crashed when launch rails migrations.
Bug introduced with commit 18d3c40fb3a6a2ebec62c4e8fa44f3eb50f6c982.
1. The before_validation: set_lock method inside InstallationConfig model. breaks the migration.
2. If fixed the point before, there are this error can't dump `serialized_value`: was supposed to be a ActiveSupport::HashWithIndifferentAccess, but was a String. -- "{}" for the `serialized_value` field. This is why at this moment of the migration the field has a default with '{}' and the fixing migration is after.

**To Reproduce**

1. Start fresh new project from Develop branch
3. Setup .env postgres credentials
4. Create Database
5. Launch migration

**Expected behavior**

All migrations runs without errors

**Screenshots**

![Schermata 2022-03-16 alle 17 38 50](https://user-images.githubusercontent.com/30950613/158641529-c4df48e5-f13e-452a-91ba-6484651b7c51.png)

![Schermata 2022-03-16 alle 17 59 18](https://user-images.githubusercontent.com/30950613/158645849-0b9f9ff0-199c-4e64-be6c-e3b1cb43280a.png)

**Environment**
Ruby: 3.0.2
Rails: rails 6.1.4.7
Postgres: 9.6 and 12.4

**Possible solutions**
1. add the respond_to? before the field assignment
2. create the value with HashWithIndifferentAccess before assignment

If you agree, I would like to work on this as first issue. Thanks

sojan-official:
@matteomelotti this is a very old migration so not much value add in fixing that( as the underlying code has changed ). We don't recommend newer installations to run all the migrations from scratch, instead, use ` rake db:chatwoot_prepare`. which is equivalent to the following

```
rake db:schema:load
rake db:seed
```

--- 

we could also consider squashing all the migrations before the last 6 months using gem https://github.com/jalkoby/squasher

ref: ref: https://github.com/chatwoot/chatwoot/issues/3638

matteomelotti:
@sojan-official thank you! Nice gem!
I tried locally before 2021-09 but it gaves me the same two errors from ConfigLoader and InstallationConfig

sojan-official:
@matteomelotti  you might have to comment out some of the data migrations to run squasher.  We will take care of this issue, since it has to be worked by someone who would have context about all the migrations.

[...]

PR #6716, opened 2023-03-21
fixed ConfigLoader process

tianlu1677:
# Fixed   ConfigLoader.new.process

## Description

When rails db migrate, current scheme version will raise error. Because `ConfigLoader.new.process` use the latest code.

Fixes # (issue)

## Type of change

## How Has This Been Tested?

```ruby
rails db:drop
rails db:create
rails db:migrate

# will raise error
```

### DIFF
--- a/db/migrate/20200510112339_create_installation_config.rb
+++ b/db/migrate/20200510112339_create_installation_config.rb
@@ -2,7 +2,8 @@ class CreateInstallationConfig < ActiveRecord::Migration[6.0]
   def change
     create_table :installation_configs do |t|
       t.string :name, null: false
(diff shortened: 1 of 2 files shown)


sojan-official:
@tianlu1677 this is a very old migration, so not much value add to fixing that( as the underlying code has changed ). We don't recommend newer installations to run all the migrations from scratch; instead, use ` rake db:chatwoot_prepare`. Which is equivalent to the following

```
rake db:schema:load
rake db:seed
```

--- 

ref: https://github.com/chatwoot/chatwoot/issues/4196#issuecomment-1069396658

Issue #6621, opened 2023-03-07
Conversation custom attributes should be shown again when the prechat screen is shown for the new conversation


blazejstarosta:
**Is your enhancement request related to a problem? Please describe.**
I'm using the pre-chat form to allow users to pick the topic of the conversation. Based on the topic, I have automation rules to assign the conversation to specific team responsible for that topic. 

It works fine when the user initiates the first conversation, but when the conversation is marked as resolved (either by user or by the agent) and then the user clicks the button to start a new conversation ("Allow messages after conversation resolved" is set to disabled), they cannot pick a different topic anymore, because the custom attributes are not showing in the pre-chat form anymore. They can only enter a new message and start a new conversation with previous settings.  That's because all the custom attributes are stored permanently in the cookie, so the user is only allowed to choose a different topic or change their data after they manually remove cookie.

**Describe the solution you'd like**
Very simple - when the user starts the new conversation, the full pre-chat form with all custom attributes should be shown. Or there could be a setting in the administration panel to let administrators choose how the pre-chat form should behave, so everyone is happy (either allow the custom attributes to be remembered or have them reset when the new conversation is initiated).

**Describe alternatives you've considered**
Alternatively, add a setting to allow specific custom attributes to be remembered in the cookie, that way individual attributes could be remembered while the others would be reset. So when the user initiates the conversation again, the pre-chat form would contain only the fields that are not set to be permanently stored in the cookie.

scmmishra:
@muhsin the [linked issue](https://github.com/chatwoot/chatwoot/issues/3789) is from a user who is facing a problem related to that

PR #6735, opened 2023-03-23
feat: Show custom attributes in pre-chat form on new conversation

muhsin-k:
Fixes https://github.com/chatwoot/chatwoot/issues/6621

https://www.loom.com/share/bce7aae2c8334aaf99a47785b8182302

**How to test this PR?**
1. Add one conversation custom attribute.
2. Enable pre-chat form.
3. Enable the custom attribute created in step 1.
5. Enable "Allow messages after conversation resolved" from the inbox settings.
6. Create a conversation from the widget.
7. Resolve the conversation from the dashboard.
8. Click on the new conversation from the widget and check that the custom attribute is visible, not in the pre-chat form.


app/javascript/widget/components/PreChat/Form.vue
```diff
@@ -114,7 +114,7 @@ export default {
       return this.options.preChatMessage;
     },
     preChatFields() {
-      return this.disableContactFields ? [] : this.options.preChatFields;
+      return this.options?.preChatFields || [];
```
scmmishra, line 117: In this case, I see that the option `disableContactFields` comes from the parent component "PreChatForm", which infers the data from `$route.params` 

Is there an case where the value is used except when the user is returning? Just want to ensure there are no unwanted side effects of ignoring that. 

If there aren't we could remove that check anyway, since it's not used

scmmishra (approved):

PR #4923, opened 2022-06-24
feat: Conversation context menu

fayazara:
## Description
Demo
https://www.loom.com/share/b18f0792fb044807a3fb01fa03491094

A right click menu on conversation to quickly resolve, reopen, mark as pending, snooze convresations and also assign agents and labels.

Fixes #5074

## Type of change

Please delete options that are not relevant.
- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #6750, opened 2023-03-24
feat: Setup context menu for message

pranavrajs:
- Use context menu for message options
- Add a confirmation modal for deletion
- Fix button spacing with the first button in modal footer.
- Hide context menu on scroll
- Move translate modal to context menu
- Allow right click on message bubble
Screenshot 2023-03-24 at 2 32 44 PM


app/javascript/dashboard/assets/scss/widgets/_modal.scss
```diff
@@ -86,6 +86,12 @@
       margin-left: var(--space-small);
     }
 
+    &:first-child {
```
pranavrajs, line 89: Screenshot 2023-03-24 at 3 49 53 PM

To fix this.

app/javascript/dashboard/components/ui/ContextMenu.vue
```diff
@@ -39,7 +37,6 @@ export default {
   },
   mounted() {
     this.$nextTick(() => this.$el.focus());
-    this.show = true;
```
pranavrajs, line 42: Feels like a left over code from [here](https://github.com/chatwoot/chatwoot/pull/4923/commits/4bff6002e2dc2984221af9b56000a28b37d1be6d).

app/javascript/dashboard/components/widgets/conversation/MessagesView.vue
```diff
@@ -355,6 +355,7 @@ export default {
     },
 
     handleScroll(e) {
+      bus.$emit(BUS_EVENTS.ON_MESSAGE_LIST_SCROLL);
```
pranavrajs, line 358: OnScroll emit an event, so that we can hide the context menu.

PR #6752, opened 2023-03-24
fix: Move subscription calls to after initialize

pranavrajs:
The notification subscription is called after every presence update due to it being implemented in the currentUser watcher. I move it current account watcher.

### DIFF
--- a/app/javascript/dashboard/App.vue
+++ b/app/javascript/dashboard/App.vue
@@ -72,13 +72,6 @@ export default {
       if (!this.hasAccounts) {
         this.showAddAccountModal = true;
       }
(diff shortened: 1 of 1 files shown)


scmmishra (approved):
@pranavrajs approving this PR, can you write a one-liner description on why the change is made, in case we need to refer this in the future

pranavrajs:
@scmmishra The notification subscription was called when the currentUser object was changed. The availability status would get changed every 20-30s as the status would get updated whenever there is a presence update. This resulted in unnecessary subscription calls. Right now, I moved it to the current Account update which happens during the initial load only.

PR #6757, opened 2023-03-27
feat: add index to conversation `id` and `account_id`

scmmishra:
## Description

This PR adds index to conversations `id` and `account_id`. This improves the performance for some reports query

## How Has This Been Tested?

Before adding index: https://explain.dalibo.com/plan/696hf6a313g927aa#plan
After adding index: https://explain.dalibo.com/plan/5334757gf3d62721

The cost of the query reduces by half, this will help across the app. The ultimate bottleneck is the sequential scans on the tags table. ~Gotta figure that out, will do that in a separate PR~

The `ILIKE` query is the holy grail of this, if that can be fixed, the query becomes extremely light.

Heres the query in question

```sql
EXPLAIN (ANALYZE, COSTS, VERBOSE, BUFFERS, FORMAT JSON) SELECT 
  "conversations"."id" 
FROM 
  "conversations" 
  INNER JOIN "taggings" "conversation_taggings_acd14c7" ON "conversation_taggings_acd14c7"."taggable_id" = "conversations"."id" 
  AND "conversation_taggings_acd14c7"."taggable_type" = 'Conversation' 
  AND "conversation_taggings_acd14c7"."tag_id" IN (
    SELECT 
      "tags"."id" 
    FROM 
      "tags" 
    WHERE 
      LOWER("tags"."name") ILIKE 'bugs' ESCAPE '!'
  ) 
WHERE 
  "conversations"."account_id" = 1;
```


Gemfile
```diff
@@ -165,6 +165,10 @@ group :development do
 
   # When we want to squash migrations
   gem 'squasher'
+
+  # profiling
+  gem 'rack-mini-profiler', require: false
+  gem 'stackprof'
```
scmmishra, line 171: @sojan-official I add this gem, gives neat profiles like this 
image

Is it worth keeping it? I can remove it, it's very simple to add whenever I need to test

sojan-official: should be good to go for development group

sojan-official (approved):

PR #6762, opened 2023-03-27
fix: mentions are not rendered properly in slack

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-1352/mentions-are-not-rendered-properly-in-slack

### DIFF
--- a/lib/integrations/slack/send_on_slack_service.rb
+++ b/lib/integrations/slack/send_on_slack_service.rb
@@ -34,12 +34,15 @@ class Integrations::Slack::SendOnSlackService < Base::SendOnChannelService
 
   def message_content
     private_indicator = message.private? ? 'private: ' : ''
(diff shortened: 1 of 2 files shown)


lib/integrations/slack/send_on_slack_service.rb
```diff
@@ -33,10 +34,11 @@ def perform_reply
 
   def message_content
     private_indicator = message.private? ? 'private: ' : ''
+    message_text = message.content.gsub(MENTION_REGEX, '\1')
```
pranavrajs, line 37: Message content can be null for file only messages. Can you please check whether this code works if I send image without text?

tejaswinichile: Good catch! It works for empty string, not for nil. Handling the case.

tejaswinichile:
Screenshot 2023-03-28 at 12 13 08 PM

sojan-official (approved):

Issue #5855, opened 2022-11-15
[CW-2348] Ability to send the user identification on whatsapp messages


vagner-dias:
Is your enhancement request related to a problem? Please describe.
When a user has a signature defined, this signature (or the user display name) could be appended to the whatsapp text sent to the customer

Describe the solution you'd like
The "signature" button could be enabled on the text field to append the user signature when writing whatsapp messages

![](https://uploads.linear.app/2847f7b9-2caa-474c-b261-1c9bdc2a5da6/3741e8ff-7e7f-4821-9fe6-5cb045c5eaff/bbbdc909-ff46-45f3-adae-0f5f7636b3cd?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzI4NDdmN2I5LTJjYWEtNDc0Yy1iMjYxLTFjOWJkYzJhNWRhNi8zNzQxZThmZi03ZTdmLTQ4MjEtOWZlNi01Y2IwNDVjNWVhZmYvYmJiZGM5MDktZmY0Ni00NWYzLWFkYWUtMGY1Zjc2MzZiM2NkIiwiaWF0IjoxNjkxMTY1NTg3LCJleHAiOjE2OTEyNTE5ODd9.aeUqNOTGfWZt92OHQMRvFvAqMoX4oRkNHcw8PhHvou4)

Example:
When the signature is set, like this example

![](https://uploads.linear.app/2847f7b9-2caa-474c-b261-1c9bdc2a5da6/46a65fcf-6f14-4682-9cb0-e4f83e782e20/b9973a9c-daa4-4d08-b30e-4385df103c04?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzI4NDdmN2I5LTJjYWEtNDc0Yy1iMjYxLTFjOWJkYzJhNWRhNi80NmE2NWZjZi02ZjE0LTQ2ODItOWNiMC1lNGY4M2U3ODJlMjAvYjk5NzNhOWMtZGFhNC00ZDA4LWIzMGUtNDM4NWRmMTAzYzA0IiwiaWF0IjoxNjkxMTY1NTg3LCJleHAiOjE2OTEyNTE5ODd9.CWyIia5PHNyCwMylD56cr_dklJRUnD6ZixkAWTVNGio)

The "signature" button can be enabled when the user writes the message and the signature appended to the message:

[...]

PR #6765, opened 2023-03-27
feat: message signature on whatsapp

clairton:
# Pull Request Template

## Description

fixes: https://github.com/chatwoot/chatwoot/issues/5855

Prefix agent on whatsapp message

## Type of change

## How Has This Been Tested?

Send message with whatsapp cloud api

Captura de Tela 2023-03-27 às 17 05 13
Captura de Tela 2023-03-27 às 17 05 00

### DIFF
--- a/app/controllers/api/v1/accounts/inboxes_controller.rb
+++ b/app/controllers/api/v1/accounts/inboxes_controller.rb
@@ -114,7 +114,7 @@ class Api::V1::Accounts::InboxesController < Api::V1::Accounts::BaseController
   def inbox_attributes
     [:name, :avatar, :greeting_enabled, :greeting_message, :enable_email_collect, :csat_survey_enabled,
      :enable_auto_assignment, :working_hours_enabled, :out_of_office_message, :timezone, :allow_messages_after_resolved,
-     :lock_to_single_conversation]
+     :lock_to_single_conversation, :send_message_signature]
   end
 
   def permitted_params(channel_attributes = [])
--- a/app/javascript/dashboard/i18n/locale/en/inboxMgmt.json
+++ b/app/javascript/dashboard/i18n/locale/en/inboxMgmt.json
@@ -65,6 +65,12 @@
           "PLACEHOLDER": "Enter your Webhook URL",
           "ERROR": "Please enter a valid URL"
         },
+        "SEND_SIGNATURE": {
+          "LABEL": "Send message signature",
+          "HELP_TEXT": "Send a signature configured in profile after each message.",
+          "ENABLED": "Enabled",
+          "DISABLED": "Disabled"
+        },
         "CHANNEL_DOMAIN": {
           "LABEL": "Website Domain",
           "PLACEHOLDER": "Enter your website domain (eg: acme.com)"
--- a/app/javascript/dashboard/routes/dashboard/settings/inbox/Settings.vue
+++ b/app/javascript/dashboard/routes/dashboard/settings/inbox/Settings.vue
@@ -281,6 +281,21 @@
           </p>
         </label>
 
+        <label v-if="isAWhatsAppChannel" class="medium-9 columns settings-item">
+          {{ $t('INBOX_MGMT.ADD.WEBSITE_CHANNEL.SEND_SIGNATURE.LABEL') }}
+          <select v-model="sendMessageSignature">
+            <option :value="true">
+              {{ $t('INBOX_MGMT.ADD.WEBSITE_CHANNEL.SEND_SIGNATURE.ENABLED') }}
+            </option>
+            <option :value="false">
+              {{ $t('INBOX_MGMT.ADD.WEBSITE_CHANNEL.SEND_SIGNATURE.DISABLED') }}
+            </option>
+          </select>
+          <p class="help-text">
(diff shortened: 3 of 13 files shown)


scmmishra:
@clairton while I understand the use case, I don't think this feature is helpful for the broad user base. The change is too opinionated to be the default (and the only) behaviour of WhatsApp messaging; it could be considered if this was available as a setting that users could enable, but even with that, it's hard for me to understand the value of this change

sojan-official:
ref: https://github.com/chatwoot/chatwoot/issues/5855

scmmishra:
@clairton, seeing this thread, It seems like there is enough conversation in the community for this; I would suggest changing the approach. Let this be an inbox configuration for WhatsApp, "Send message signature", and if enabled, we can format the message to include the message signature or the agent's name, whichever is available at the end of the message.

This will make this an entirely opt-in feature. @sojan-official @pranavrajs need your thoughts on this

CC: @tejaswinichile @nithindavid

pranavrajs:
Folks, I suggest extending the personal signature option rather than adding a new field in the inbox or anything. Right now, it is enabled only the email channels. 

Screenshot 2023-03-29 at 12 55 07 PM

If we enable it on WhatsApp, that will make the feature consistent across the channels. As a further step, we can try enabling it on Facebook, Twitter and Instagram.

clairton:
I think create a toggle in whatsapp inbox “identify agent in message” disabled by default, when it enable prefix message with agent name like so “agent name: the message content”. E-mail signature use more information witch WhatsApp message. But if your prefer use the same email signature ok, I do with signature email 😀

@sojan-official @pranavrajs @scmmishra

vagner-dias:
up

pranavrajs:
@clairton @glauberssoutodossantos I understand the struggle, but having more options to do the same thing is not ideal. Let's stick with the signature for now and see if there are any issues caused by it. If so, we can take it up in the following PRs. 

@clairton If you can update the PR with the suggested change soon, we can get that in Monday's release.

clairton:
> @clairton @glauberssoutodossantos I understand the struggle, but having more options to do the same thing is not ideal. Let's stick with the signature for now and see if there are any issues caused by it. If so, we can take it up in the following PRs.
> 
> @clairton If you can update the PR with the suggested change soon, we can get that in Monday's release.

@pranavrajs done! =)

clairton:
Settings:
Captura de Tela 2023-04-14 às 14 10 51


Signature:
Captura de Tela 2023-04-14 às 14 11 23

Message With Signature:
Captura de Tela 2023-04-14 às 14 11 41

vagner-dias:
@clairton, Question: If we configure the signature to be in bold, will the code set it in bold too?

clairton:
> @clairton, Question: If we configure the signature to be in bold, will the code set it in bold too?
🙄, it put *

Settings:
Captura de Tela 2023-04-14 às 16 22 50

In whatsapp message:
Captura de Tela 2023-04-14 às 16 23 08

clairton:
> @clairton, Question: If we configure the signature to be in bold, will the code set it in bold too?

Would be transform html to markdown...

clairton:
It works like this, I don't know how it could improve.

Settings:
Captura de Tela 2023-04-14 às 16 22 50

Postgres:
Captura de Tela 2023-04-14 às 17 29 56

Whatsapp

Captura de Tela 2023-04-14 às 16 23 08

vagner-dias:
IMHO the double asterisk ended up having a nice effect.

See: so far we do not have this feature and this is already a important improvement. It can be also be improved overtime.

vagner-dias:
> @clairton @glauberssoutodossantos I understand the struggle, but having more options to do the same thing is not ideal. Let's stick with the signature for now and see if there are any issues caused by it. If so, we can take it up in the following PRs.
> 
> @clairton If you can update the PR with the suggested change soon, we can get that in Monday's release.

Hello @pranavrajs , is the PR good to include on today's release?

Thank you!

pranavrajs:
@vagner-dias There are a couple of outstanding questions on this one.

@clairton I didn't get the changes completely. Sorry if it is a repeated question. I am trying to understand why we need the `Send message signature` flag. 

If you see the editor in the email channel (shown below), the user can disable the signature. If the user doesn't need to add the signature, they can avoid it. 

Screenshot 2023-04-19 at 6 53 51 PM

An inbox-level setting is not really required here.

pranavrajs:
@clairton I would suggest turning on the message signature option for the WhatsApp channel first and seeing if it solves the problem.

https://github.com/chatwoot/chatwoot/blob/527042afd1ca917f37e4cdcdbeef4cf7d6e5055c/app/javascript/dashboard/components/widgets/WootWriter/ReplyBottomPanel.vue#L279-L281

vagner-dias:
HI @clairton,

Maybe this is "simpler" to implement than the initial approach? I mean, less code, etc?

WallacyBR:
This feature was released?

vagner-dias:
Hello @sojan-official, @clairton @pranavrajs 

Do you guys plan to release this feature or even something similar?

Is there anything I can help with?

Regards.

clairton:
Hello, email subscription is an option, but my company's whatsapp subscription is required. That's why I ended up focusing on other subjects and putting this in a fork 😅

WallacyBR:
Someone can review this?

sojan-official:
Closing this since we have messages signature enabled for WhatsApp accounts in our latest version.

PR #6770, opened 2023-03-28
chore: Re-add "public" prefix to "public.gen_random_uuid()" in schema.rb

jordan-brough:
## Description

When I run `rails db:schema:dump` this is what is generated for me. It was previously in the chatwoot codebase _with_ the prefix, but was removed **[in this PR](https://github.com/chatwoot/chatwoot/pull/5338#discussion_r957645071)** without explanation as to why.

Should `public.` be in schema.rb or not? What was the reason for its removal? Do I have something unique about my setup that's causing it to be added?


sojan-official:
@jordan-brough Think the change crept in unintentionally from the previous PR. Both versions should work.

PR #6772, opened 2023-03-28
fix memory leek

s00d:
## Description

Hello, your CustomAttribute component is never cleared, which leads to memory leaks of several GB over time. All events must be turned off when the component destroyed.


pranavrajs:
@s00d This is a good find. Can you please share how you tested this? I mean, how did you find that there is a leak?

Merging this as the change is straightforward.

s00d:
I debugged manually using Chrome devtools, specifically the Memory tab. I noticed that after some time, there were a lot of FOCUS_CUSTOM_ATTRIBUTE objects in memory, and then it became clear what the cause was.

PR #6775, opened 2023-03-29
chore: Update the design of the help center portal

pranavrajs:
// TODO

### DIFF
--- a/app/views/public/api/v1/portals/_hero.html.erb
+++ b/app/views/public/api/v1/portals/_hero.html.erb
@@ -1,10 +1,10 @@
 
-<section class="bg-gradient-to-b from-white to-slate-75 pt-8 pb-16 md:py-16 flex flex-col items-center justify-center">
-  <div class="mx-auto w-full max-w-2xl flex flex-col items-center">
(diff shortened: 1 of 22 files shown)


app/views/public/api/v1/portals/_hero.html.erb
```diff
@@ -1,10 +1,10 @@
 
-<section class="bg-gradient-to-b from-white to-slate-75 pt-8 pb-16 md:py-16 flex flex-col items-center justify-center">
-  <div class="mx-auto w-full max-w-2xl flex flex-col items-center">
-    <h1 class="text-2xl px-5 md:text-4xl text-slate-900 font-semibold subpixel-antialiased leading-normal text-center">
+<section class="pt-8 pb-16 md:py-16 flex flex-col w-full bg-woot-50" style="background-color: <%= @portal.config['light_color'] %>">
```
scmmishra, line 2: We could use css variables globally instead of this, have a class called `woot-user-<property>` 

Have a global CSS

```css
:root {
    --user-light-bg: @portal.config['light_color']
}

.woot-user-bg {
    background: var(--user-light-bg)
}
```
---

Most portal-like sites allow users to pass custom CSS, we can expose a set of CSS variables the user can modify, that would greatly help in customizing the help center according to the brand

On second thoughts, we can let the current changes stay. And pick this up as a enhancement later

PR #6788, opened 2023-03-29
fix: Handle spaces in CC/BCC email lists

jordan-brough:
## Description

When the CC field is generated in the UI, the email values are joined together with ", ":
https://github.com/chatwoot/chatwoot/blob/d8604107aa7aa9f3188eac81b84e81932062b523/app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue#L1057-L1058

but when they are parsed we currently split by just ",":
https://github.com/chatwoot/chatwoot/blob/d8604107aa7aa9f3188eac81b84e81932062b523/app/builders/messages/message_builder.rb#L52-L55
https://github.com/chatwoot/chatwoot/blob/d8604107aa7aa9f3188eac81b84e81932062b523/app/javascript/dashboard/components/widgets/conversation/helpers/emailHeadHelper.js#L5

This causes an error on the backend and on the frontend.

I think we should update the code to allow whitespace in the input and to split by `\s*,\s` and also to trim leading and trailing whitespace when we parse.

If this looks OK, I'll look into adding tests.

Frontend:
image

Backend error going to frontend:
image

Backend reply:
image

## Type of change


app/builders/messages/message_builder.rb
```diff
@@ -49,10 +49,10 @@ def process_emails
     return unless @conversation.inbox&.inbox_type == 'Email'
 
     cc_emails = []
-    cc_emails = @params[:cc_emails].split(',') if @params[:cc_emails].present?
+    cc_emails = @params[:cc_emails].strip.split(/\s*,\s*/) if @params[:cc_emails].present?
```
sojan-official, line 52:
```suggestion
    cc_emails = @params[:cc_emails].gsub(/\s+/, '').split(',') if @params[:cc_emails].present?
```

app/builders/messages/message_builder.rb
```diff
@@ -49,10 +49,10 @@ def process_emails
...
 
     bcc_emails = []
-    bcc_emails = @params[:bcc_emails].split(',') if @params[:bcc_emails].present?
+    bcc_emails = @params[:bcc_emails].strip.split(/\s*,\s*/) if @params[:bcc_emails].present?
```
sojan-official, line 55:
```suggestion
    bcc_emails = @params[:bcc_emails].gsub(/\s+/, '').split(',') if @params[:bcc_emails].present?
```

sojan-official (commented):
@jordan-brough  Thanks for the changes. suggesting a minor tweak for readability. We can go ahead with the specs and get this merged.

app/javascript/dashboard/components/widgets/conversation/helpers/emailHeadHelper.js
```diff
@@ -2,6 +2,6 @@ import emailValidator from 'vuelidate/lib/validators/email';
 
 export const validEmailsByComma = value => {
   if (!value.length) return true;
-  const emails = value.split(',');
+  const emails = value.trim().split(/\s*,\s*/);
```
sojan-official, line 5:
```suggestion
  const emails = value.replace(/\s+/g, '').split(',');
```

app/builders/messages/message_builder.rb
jordan-brough: Just checking: Is it safe to eliminate all spaces here? Are these values ever in a format like `"John Doe" <john@doe.com>`?  Also I feel like I've heard that something like `"john doe"@doe.com` is a valid email address?

I wonder if these values should be sent in JSON format or something instead of being `join`ed with whitespace?

sojan-official: @jordan-brough  You are right about the spaces being allowed in email  ref: https://stackoverflow.com/questions/2049502/what-characters-are-allowed-in-an-email-address

Generally, the approach we have been following in Chatwoot for email validation is based on the addresses supported by Gmail. ie (letters numbers and + ). We could take it as a separate effort to see whether we want to support all cases per RFC.

For this particular case

- We currently expect an array of coma-separated emails which has  alphanumeric characters 
- We don't support the format with the name in the email ex:  "John Doe<john@doe.com>"

---

I wonder if these values should be sent serialized in JSON format or something instead of being joined with commas? 

We will have to do some work for backward compatibility for the change. There will be older messages in the database with cc emails and bcc emails in the current format. We can go ahead with the current fix and take it up separately.

sojan-official:
- Added the specs 👍

jordan-brough:
Thank you! I forgot to get back to this.

PR #6790, opened 2023-03-29
feat: Refetch the active conversation messages on action cable reconnect

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-1513/dashboard-on-cable-reconnect-fetch-messages


app/javascript/shared/helpers/BaseActionCableConnector.js
```diff
@@ -28,17 +32,46 @@ class BaseActionCableConnector {
...
+        'PRESENCE_INTERVAL-->',
+        isConnectionActive,
+        navigator.onLine,
+        new Date()
+      );
+      // this.subscription.consumer.connection.subscriptions.consumer.connection;
+      // .disconnected
+      const isAllGood =
+        BaseActionCableConnector.isDisconnected &&
+        navigator.onLine &&
```
pranavrajs, line 46: Do we need this check?

app/javascript/shared/helpers/BaseActionCableConnector.js
```diff
@@ -28,17 +32,46 @@ class BaseActionCableConnector {
...
+      const isConnectionActive = this.consumer.connection.isOpen();
+      console.log(
+        'PRESENCE_INTERVAL-->',
+        isConnectionActive,
+        navigator.onLine,
+        new Date()
+      );
+      // this.subscription.consumer.connection.subscriptions.consumer.connection;
+      // .disconnected
+      const isAllGood =
```
pranavrajs, line 44:
```suggestion
      const shouldRefetch =
```

app/javascript/shared/helpers/BaseActionCableConnector.js
muhsin-k: It is not needed.

app/javascript/shared/helpers/BaseActionCableConnector.js
```diff
@@ -1,11 +1,15 @@
 import { createConsumer } from '@rails/actioncable';
 import { BUS_EVENTS } from 'shared/constants/busEvents';
 
-const PRESENCE_INTERVAL = 20000;
+// TODO: Change this value to 20 seconds after the QA
+const PRESENCE_INTERVAL = 10000;
```
muhsin-k, line 5: I will change this value to 20 seconds after completing the QA.

scmmishra:
I gave some suggestions over-call to simplify the websocket part. @muhsin-k I'll review the store part and let you know

app/javascript/shared/helpers/BaseActionCableConnector.js
```diff
@@ -28,17 +32,35 @@ class BaseActionCableConnector {
 
     setInterval(() => {
```
scmmishra, line 33: Another unrelated comment, can we see if we can replace this with `setTimeout`?  Generally, `setInterval` is not considered safe unless we know when to trigger a `clearInterval`. For `setInterval`, the callback function stays in memory until `clearInterval` is called. This can cause potential memory issues. 

We could make a function `triggerPresenceInterval.`

```js
triggerPresenceInterval() {
    setTimeout(() => {
      this.subscription.updatePresence();
      const isConnectionActive = this.consumer.connection.isOpen();
      const shouldRefetch =
        BaseActionCableConnector.isDisconnected && isConnectionActive;
      if (shouldRefetch) {
        this.refreshActiveConversationMessages();
        BaseActionCableConnector.isDisconnected = false;
      }
      this.triggerPresenceInterval();
    }, PRESENCE_INTERVAL);
  }
```
and then call `this.triggerPresenceInterval()` in the constructor

app/javascript/dashboard/store/modules/conversations/actions.js
```diff
@@ -86,6 +86,62 @@ const actions = {
...
+        after: lastMessageId,
+      });
+      commit(`conversationMetadata/${types.SET_CONVERSATION_METADATA}`, {
+        id: conversationId,
+        data: meta,
+      });
+      const missingMessages = payload.filter(
+        message => !messages.find(item => item.id === message.id)
+      );
+      commit(types.SET_MISSING_MESSAGES, {
```
scmmishra, line 114: Do we need a separate mutation here? Since we have already filtered the messages previously, is it possible to re-use the mutation that was used before?

app/javascript/dashboard/store/modules/conversations/actions.js
```diff
@@ -86,6 +86,62 @@ const actions = {
...
+        id: conversationId,
+        data: missingMessages,
+      });
+      commit(types.CLEAR_LAST_MESSAGE_ID_FROM_SYNC_CONVERSATION, {
+        conversationId,
+      });
```
scmmishra, line 120: We can re-use the same mutation

```suggestion
      commit(types.SET_LAST_MESSAGE_ID_IN_SYNC_CONVERSATION, {
        conversationId,
        messageId: null
      });
```

app/javascript/dashboard/store/modules/conversations/index.js
```diff
@@ -202,6 +214,19 @@ export const mutations = {
...
+    { conversationId, messageId }
+  ) {
+    _state.syncConversationsMessages[conversationId] = messageId;
+  },
+  [types.CLEAR_LAST_MESSAGE_ID_FROM_SYNC_CONVERSATION](
+    _state,
+    { conversationId }
+  ) {
+    _state.syncConversationsMessages[conversationId] = null;
+  },
```
scmmishra, line 229: Do you know if we need to maintain this separately? Can we not just infer it from the `allConversations` state?

scmmishra (changes requested):
@muhsin-k added some comments on the store, lmk if you have any questions :)

app/javascript/dashboard/store/modules/conversations/actions.js
muhsin-k: The current mutation is used for setting all the messages. It can't put messages in a particular order (Missing messages can come at any position.)

app/javascript/dashboard/store/modules/conversations/actions.js
muhsin-k: Sounds good.

app/javascript/dashboard/store/modules/conversations/index.js
muhsin-k: Yes. We need to maintain this separation. The reconnect logic is not reliable at the moment. It would be better to store the last message id at the disconnect time so that it can fetch all the messages after the last id when reconnect happens.

app/javascript/dashboard/store/modules/conversations/actions.js
muhsin-k: https://github.com/chatwoot/chatwoot/blob/develop/app/javascript/dashboard/store/modules/conversations/index.js#L54

muhsin-k:
@scmmishra All the review comments are addressed. Please take a look when you are free.

app/javascript/shared/helpers/BaseActionCableConnector.js
```diff
@@ -25,19 +29,39 @@ class BaseActionCableConnector {
...
+      this.onReconnect();
+      BaseActionCableConnector.isDisconnected = false;
+    }
   }
 
-  // eslint-disable-next-line class-methods-use-this
-  onDisconnected() {
+  onReconnect = () => {
+    this.events.reconnect();
+  };
```
scmmishra, line 54:
```suggestion
  onReconnect = () => {};
```

scmmishra (approved):

PR #6793, opened 2023-03-30
slack-bot-message

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-1371/bot-messages-are-not-delivered-to-slack


app/listeners/hook_listener.rb
```diff
@@ -1,14 +1,14 @@
...
-    return unless message.webhook_sendable?
+    return unless message.slack_hook_sendable?
 
     execute_hooks(event, message)
   end
 
   def message_updated(event)
     message = extract_message_and_account(event)[0]
-    return unless message.webhook_sendable?
+    return unless message.slack_hook_sendable?
```
sojan-official, line 11: There might be a better place to add this. This the underlying hooks vary.  There might be an integration which would want to listen to all message created/updated events. 

Let's remove this condition check from the listener and have the checks in individual integration processor services instead.

app/jobs/hook_job.rb
```diff
@@ -24,6 +24,10 @@ def process_slack_integration(hook, event_name, event_data)
   def process_dialogflow_integration(hook, event_name, event_data)
     return unless ['message.created', 'message.updated'].include?(event_name)
 
+    message = event_data[:message]
+
+    return unless message.webhook_sendable?
+
```
sojan-official, line 30: Let's have this check inside the processor service so that all logic related to Dialogflow will remain encapsulated in a single place.

sojan-official (commented):
- added comments. We will have to tweak the specs accordingly ( ensure there are appropriate tests in the processor services )

app/listeners/hook_listener.rb
tejaswinichile: Cool!

app/jobs/hook_job.rb
tejaswinichile: We do have this check here https://github.com/chatwoot/chatwoot/blob/develop/lib/integrations/bot_processor_service.rb#L46

PR #6796, opened 2023-03-30
chore: Refactor Automation Specs [CW-1447]

sojan-official:
The current way of writing specs for automation in one single file automation_listener isn't effective. Hence we are breaking down the specs for each class into separate spec files.

fixes: CW-1447


tejaswinichile (approved):
LGTM 👍🏼

PR #6817, opened 2023-04-03
fix: inconsistency in report and summary for metric counts

scmmishra:
## Description

The reports would generate inconsistent values, when the timezone of the request was farther away from UTC. The cause of this was a hidden "feature/bug" with the [groupdate gem](https://github.com/ankane/groupdate). 

The gem handled the timerange specifically if the `exclude_end` was included. If included an extra day was added, see [groupdate/series_builder.rb](https://github.com/ankane/groupdate/blob/57150dd7a91e4aa3e3f9c9f4acf5dfb701cc4085/lib/groupdate/series_builder.rb#L117-L124)

This cause the SQL query to yield data for an extra day. The frontend clamped the data between timestamps, however the summary which was calculated on the server side didn't have any check like this. The solution proposed is to exclude the end of the range. 

Have got to test this with more conditions, but prima facie, the solution seems to work.

**Note:** Made another critical change, making the summary timezone sensitive. I guess it makes sense that way. 

**This PR needs needs a thorough manual testing**

## Type of change

### Before fix

Screenshot 2023-04-03 at 4 03 02 PM

### After fix

Screenshot 2023-04-03 at 4 01 35 PM

## Checklist:

### DIFF
--- a/app/controllers/api/v2/accounts/reports_controller.rb
+++ b/app/controllers/api/v2/accounts/reports_controller.rb
@@ -67,14 +67,16 @@ class Api::V2::Accounts::ReportsController < Api::V1::Accounts::BaseController
   def current_summary_params
     common_params.merge({
                           since: range[:current][:since],
-                          until: range[:current][:until]
+                          until: range[:current][:until],
+                          timezone_offset: params[:timezone_offset]
                         })
   end
 
   def previous_summary_params
     common_params.merge({
                           since: range[:previous][:since],
-                          until: range[:previous][:until]
+                          until: range[:previous][:until],
+                          timezone_offset: params[:timezone_offset]
                         })
   end
 
--- a/app/helpers/date_range_helper.rb
+++ b/app/helpers/date_range_helper.rb
@@ -7,7 +7,7 @@ module DateRangeHelper
   def range
     return if params[:since].blank? || params[:until].blank?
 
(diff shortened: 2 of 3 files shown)


spec/controllers/api/v2/accounts/report_controller_spec.rb
```diff
@@ -31,8 +32,8 @@
         super().merge(
           metric: 'conversations_count',
           type: :account,
-          since: date_timestamp.to_s,
-          until: date_timestamp.to_s
+          since: start_of_today.to_s,
+          until: end_of_today.to_s
```
scmmishra, line 36: @sojan-official I had to make this change because of the range of changes we made. Earlier, when including the ends of the range, an extra day was added, which made this test work fine.

Right now, excluding the ends, make he test break, but I believe the current behaviour is correct. LMK whenever you have time to discuss this

CC: @pranavrajs @tejaswinichile

tejaswinichile (approved):
LGTM 💯

scmmishra:
Still needs discussion on how we manage date limits.

CC: @sojan-official @pranavrajs lmk whenever you folks have the time

scmmishra:
@sojan-official @tejaswinichile I've updated this PR with tests, can you folks take a look whenever you have the time?

sojan-official (approved):
The code changes look good to me. @tejaswinichile  could you work with Shivam and get this merged?

PR #6820, opened 2023-04-04
chore: Use feature_flags attribute instead of settings_flags

pranavrajs:
The custom domain feature uses the settings_flag attribute (an old implementation), which was not used anywhere else. I have migrated the flag to the feature_flags attribute so that it can be updated via the UI.

This PR:
- Migrates custom_reply_domain and custom_reply_email to feature_flags.
- Remove the settings_flags attribute from the account model.


tejaswinichile (approved):
LGTM 💯 

I had one doubt for account which has settings_flags already set.
But then I checked the production data.

```
 Account.where('settings_flags > 0')
=> []```

There are no such accounts.

pranavrajs:
> I had one doubt for account which has settings_flags already set.

No, none. This feature is enabled only in a couple of accounts. We did that via console.

PR #6833, opened 2023-04-05
feat: allow sorting of articles

scmmishra:
## Reordering from the UI 

https://www.loom.com/share/cea10e3e08f84d00b6a1e9d3af0bcc05

## Creating a new article and/or moving articles between categories

https://www.loom.com/share/e327203a064f4f32a0ae2e3bc533ecbc


app/javascript/dashboard/routes/dashboard/helpcenter/components/ArticleTable.vue
```diff
@@ -89,4 +148,9 @@ export default {
     border: 0;
   }
 }
+
+.article-ghost-class {
+  opacity: 0.5;
+  background-color: var(--s-50);
```
nithindavid, line 154:
```suggestion
  background-color: var(--s-50);
  border-bottom: 1px solid var(--w-300);
```

Can you please see if the border color gives more  visual cue about the position it's getting inserted ?

nithindavid (approved):
Tested on local FE Part works as expected

nithindavid:
@scmmishra Any cue to let the user know we can re-order articles? 
We could add `    cursor: grab;` in `app/javascript/dashboard/routes/dashboard/helpcenter/components/ArticleItem.vue:141`

scmmishra:
@nithindavid yeah, I've been thinking of the UX part too. 

> We could add `cursor: grab;`

this looks like a good idea. 


I also though adding a drag-handle at the start of the list item, wdyt of that?

app/javascript/dashboard/routes/dashboard/helpcenter/components/ArticleTable.vue
scmmishra: Sure, we could do more improvements to this. But the table format is limiting. For instance, the `<tr>` cannot have left and right padding. If we could make that a grid/flex, we can make the styling much better suited for a dragging case

app/models/article.rb
```diff
@@ -115,6 +120,38 @@ def increment_view_count
...
   private
 
+  def category_id_changed_action
+    # We need to update the position of the article in the new category
+    return unless persisted?
+
+    # this means the article is just created
+    # and the category_id is newly set
+    # and the position is already present
+    return if created_at_before_last_save.nil? && position.present? && category_id_before_last_save.nil?
```
scmmishra, line 130: @tejaswinichile @sojan-official need your thoughts on this logic.. This is to ensure the category ID is not overridden on create, if the position is supplied when creating the article

nithindavid:
@muhsin-k
The icons can be improved with fluent suggested ones

1. Sort 
image

2. Filter
image
cc: @scmmishra

app/controllers/api/v1/accounts/articles_controller.rb
```diff
@@ -68,4 +78,12 @@ def list_params
   def set_current_page
     @current_page = params[:page] || 1
   end
+
+  def update_positions(positions_hash)
```
tejaswinichile, line 82: Can you move this method to `Article.rb`?

tejaswinichile (changes requested):

app/controllers/api/v1/accounts/articles_controller.rb
scmmishra: @tejaswinichile done, can you check

db/migrate/20230405104405_update_position_for_articles.rb
```diff
@@ -0,0 +1,30 @@
+class UpdatePositionForArticles < ActiveRecord::Migration[6.1]
```
tejaswinichile, line 1: 💯

app/models/article.rb
tejaswinichile: This seems okay.

tejaswinichile (approved):

PR #6835, opened 2023-04-05
Feat: Fetch whatsapp templates till next cursor

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-1464/message-templates-pagination-is-missing

### DIFF
--- a/app/services/whatsapp/providers/whatsapp_cloud_service.rb
+++ b/app/services/whatsapp/providers/whatsapp_cloud_service.rb
@@ -24,7 +24,11 @@ class Whatsapp::Providers::WhatsappCloudService < Whatsapp::Providers::BaseServi
 
   def sync_templates
     response = HTTParty.get("#{business_account_path}/message_templates?access_token=#{whatsapp_channel.provider_config['api_key']}")
-    whatsapp_channel.update(message_templates: response['data'], message_templates_last_updated: Time.now.utc) if response.success?
+
+    return unless response.success?
+
+    whatsapp_channel.update(message_templates: response['data'], message_templates_last_updated: Time.now.utc)
+    fetch_next_templates(response)
   end
 
   def validate_provider_config?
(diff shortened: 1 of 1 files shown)


app/services/whatsapp/providers/whatsapp_cloud_service.rb
```diff
@@ -40,8 +44,21 @@ def media_url(media_id)
...
+
+      response = HTTParty.get(response['paging']['next'])
+      save_templates(response)
+    end
+  end
+
   private
 
+  def save_templates(response)
+    whatsapp_channel.update(message_templates: response['data'], message_templates_last_updated: Time.now.utc) if response.success?
```
pranavrajs, line 59: Wouldn't this overwrite the previous template? I believe we should do a concat of all the responses here.

tejaswinichile: Yeah figured that updating the PR. 👍🏼

app/services/whatsapp/providers/whatsapp_cloud_service.rb
```diff
@@ -40,8 +45,25 @@ def media_url(media_id)
     "#{api_base_path}/v13.0/#{media_id}"
   end
 
+  def fetch_next_templates(response)
+    loop do
+      break unless response['paging'] && response['paging']['next']
+
+      response = HTTParty.get(response['paging']['next'])
+      push_templates(response)
+    end
```
scmmishra, line 54:
```suggestion
    while response['paging'] && response['paging']['next']
      response = HTTParty.get(response['paging']['next'])
      push_templates(response)
    end
```

Just one minor suggestion, everything else looks great

scmmishra (approved):

app/services/whatsapp/providers/whatsapp_cloud_service.rb
tejaswinichile: I added `while` before, but Rubocop gives an issue.

PR #6860, opened 2023-04-07
fix: Invalidate cache if any of the related objects change

pranavrajs:
The changes in channel object (for eg: whatsapp templates or config in the website live chat) doesn't trigger an cache invalidation right now. The user has to logout and log back in to get the changes. 

To avoid this, we have to invalidate the inbox cache if there is any change in the associated channel object.

To reproduce the issue. 
Open rails console, get the first web widget and update any attribute in the channel object. Check if the inbox cache invalidation event is triggered.


pranavrajs:
@scmmishra Please take a look at this when you have time.

scmmishra (approved):

PR #6728, opened 2023-03-22
Azure single-tenant application support, using the Graph API

moxvallix:
## Description
This pull request adds support to Chatwoot for using single tenant Azure applications.

When the AZURE_TENANT_ID environment variable is set, Chatwoot will now use it within it's calls to Microsoft's API, rather than `common`, which only works for multi tenant applications.

As well, this pull request also adds support for sending and retrieving emails through the Microsoft Graph API, rather than using IMAP / SMTP. This was because I was unable to get SMTP working with a single tenant application. As well, using this API will likely be needed in the future, as Microsoft are deprecating SMTP Auth.

https://learn.microsoft.com/en-us/exchange/clients-and-mobile-in-exchange-online/deprecation-of-basic-authentication-exchange-online#pop-imap-and-smtp-auth

I would appreciate some help with getting this pull request up to standards. As it is currently not, I will leave it as a draft for now.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## How Has This Been Tested?
I have been hand testing this all myself in my local environment, but some help/guidance with writing proper tests would be appreciated.

## Checklist:

- [ ] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #6878, opened 2023-04-11
feat: Support Azure single-tenant application using the Graph API (#6728)

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-1400/getting-microsoft-email-working-with-chatwoot-in-2023


app/jobs/inboxes/fetch_imap_email_inboxes_job.rb
```diff
@@ -2,8 +2,19 @@ class Inboxes::FetchImapEmailInboxesJob < ApplicationJob
...
-      ::Inboxes::FetchImapEmailsJob.perform_later(inbox.channel) if inbox.channel.imap_enabled
+      next unless inbox.channel.imap_enabled?
+
+      fetch_emails(inbox)
+    end
+  end
+
+  def fetch_emails(inbox)
+    if inbox.channel.microsoft? && ENV.fetch('AZURE_TENANT_ID', false)
+      ::Inboxes::FetchMsGraphEmailsJob.perform_later(inbox.channel)
```
sojan-official, line 15: `fetchMsGraphEmailForTenantJob` ?  in the name to depicts this works on for the case of single tenant ?

app/jobs/inboxes/fetch_ms_graph_emails_job.rb
```diff
@@ -0,0 +1,101 @@
+require 'net/http'
+
+class Inboxes::FetchMsGraphEmailsJob < ApplicationJob
```
sojan-official, line 3: prev comment. name that denotes single clients

app/jobs/inboxes/fetch_ms_graph_emails_job.rb
```diff
@@ -0,0 +1,101 @@
...
+  queue_as :low
```
sojan-official, line 4: lets queue this on scheduled_jobs queue

sojan-official (commented):
Minor comments. also, lets add some specs for 
- authorization controller 
- reply mailer
- lib/microsoft_graph_api  ( get from API and post to API methods )

also there seems to be some build issues

spec/jobs/inboxes/fetch_imap_email_inboxes_job_spec.rb
```diff
@@ -20,4 +26,24 @@
       described_class.perform_now
     end
   end
+
+  context 'when microsoft inbox' do
+    it 'calls ms graph channel for single tenant app' do
```
sojan-official, line 31:
```suggestion
    it 'calls fetch ms graph email job for single tenant app' do
```

sojan-official:
note : @sojan-official to take a look into the issue during staging deploy

sojan-official:
note: the issue was with autoloading.  verified that the server stars in RAILS_ENV=production

tejaswinichile:
Update documentation with this link: 
https://learn.microsoft.com/en-us/graph/api/user-list-messages?view=graph-rest-1.0&tabs=http#code-try-2
And all new screenshots for single tenant app.

tejaswinichile:
@sojan-official I have found one issue in single tenant changes, that's if the same user have one multi-tenant app and one single tenant app, the microsoft-grap call is not getting executed. I don't know how relevant is this could be for the real time user. 
We can merge this by keeping ENV['AZURE_TENANT_ID'] check to false, so single tenant changes will never get called till we fix the issue.
Else we can icebox this one, because I spent 2 hours for this to find the real cause and I did not find any.

sojan-official:
We need to make sure
1. if an account has an approved app  and an unapproved app. this wouldn't break for the approved app. ( works without tenant id )
2. we need to make sure for an account with just an unapproved app (it works with the tenant id)

-------
We won't be looking into the case where the client has both approved and unapproved, and then they try to configure with the unapproved app for now.

sojan-official (approved):

PR #6881, opened 2023-04-11
fix: validate `template_params` for WhatsApp

scmmishra:

1. Add `JsonSchemaValidator`, that takes a declarative schema and validates it for a given property
2. Add specs for `JsonSchemaValidator`
3. Enable the validator for `template_params`

### Needs more testing


app/models/concerns/json_schema_validator.rb
scmmishra, line 1: @sojan-official made a generic validator, we can use it to validate for other channels also. Also maybe replace `ContentAttributeValidator` with a schema. It takes a declarative schema and validates against it.

app/models/message.rb
```diff
@@ -37,12 +37,30 @@ class Message < ApplicationRecord
...
+      properties: {
+        'name' => { required: true, type: 'string' },
+        'category' => { required: true, type: 'string' },
+        'language' => { required: true, type: 'string' },
+        'namespace' => { required: true, type: 'string' },
+        'processed_params' => { required: true, type: 'hash' }
+      }
+    }
+  }.freeze
+
```
scmmishra, line 53: @sojan-official can you verify the schema?

spec/models/concerns/json_schema_validator_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    # rubocop:disable Lint/ConstantDefinitionInBlock
+    # rubocop:disable RSpec/LeakyConstantDeclaration
+    TestModelForJSONValidation = Struct.new(:additional_attributes) do
+      include ActiveModel::Validations
+
+      validates_with JsonSchemaValidator, schema: schema
+    end
+    # rubocop:enable Lint/ConstantDefinitionInBlock
+    # rubocop:enable RSpec/LeakyConstantDeclaration
+  end
```
scmmishra, line 29: Couldn't solve these Rubocop warnings, would need help if we have to solve this 🥲

app/models/concerns/json_schema_validator.rb
```diff
@@ -0,0 +1,97 @@
...
+      value = json_data[key] # get the value for the current key
+
+      # check if the value is required and not present
+      if rule[:required] && value.nil?
+        record.errors.add(path, "#{key} is required")
+      elsif !value.nil? # check if the value is present
+        validate_type(value, rule[:type], "#{path}.#{key}", record) # validate the type of the value
+
+        # validate the nested schema if the value is a Hash
+        validate_schema(value, rule[:properties], "#{path}.#{key}", record) if rule[:properties] && value.is_a?(Hash)
```
sojan-official, line 56: @scmmishra seems like we already have the `JSONSchemer` gem included in the code base to validate with JSONSchema, so some of the logic here would be redundant. 

Example: https://github.com/chatwoot/chatwoot/blob/5cdc7d654a461d9b20d583e19ed01e33a31f202d/app/models/integrations/hook.rb#L57

sojan-official (commented):
realised we already have a gem added in the codebase to validate json Schema. We should be able to trim down the logic around that by using the gem.

app/models/message.rb
```diff
@@ -37,12 +37,33 @@ class Message < ApplicationRecord
...
+          'name': { 'type': 'string' },
+          'category': { 'type': 'string' },
+          'language': { 'type': 'string' },
+          'namespace': { 'type': 'string' },
+          'processed_params': { 'type': 'object' }
+        },
+        'required': %w[name category language namespace processed_params]
+      }
+    }
+  }.to_json.freeze
```
sojan-official, line 55: the schema looks good for Whatsapp template_params.  just making sure that this validation won't fail for the additional keys which we have like `campaign_id`  which can also be a part of the additional attributes

app/models/concerns/json_schema_validator.rb
sojan-official: sounds good to me

app/services/whatsapp/send_on_whatsapp_service.rb
```diff
@@ -28,13 +28,14 @@ def send_template_message
     message.update!(source_id: message_id) if message_id.present?
   end
 
+  # rubocop:disable Metrics/CyclomaticComplexity
```
sojan-official, line 31: Is there cyclomatic complexity here? 😳

sojan-official (commented):
Mostly LGTM. just a comment of the additional keys that could appear in additional attributes

app/models/message.rb
scmmishra: Cool, will test this out

app/services/whatsapp/send_on_whatsapp_service.rb
scmmishra: Yes 🥲

```
ℹ No staged files match any configured task.
npm notice
npm notice New patch version of npm available! 9.6.2 -> 9.6.5
npm notice Changelog: https://github.com/npm/cli/releases/tag/v9.6.5
npm notice Run npm install -g npm@9.6.5 to update!
npm notice
Inspecting 1 file
C

Offenses:

app/services/whatsapp/send_on_whatsapp_service.rb:31:3: C: Metrics/CyclomaticComplexity: Cyclomatic complexity for processable_channel_message_template is too high. [8/7]
  def processable_channel_message_template ...
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

1 file inspected, 1 offense detected
husky - pre-commit hook exited with code 1 (error)
```

app/models/message.rb
scmmishra: @sojan-official tested this out here

https://www.loom.com/share/5e7984e43e154016a4ae4615f4b72274

spec/models/concerns/json_schema_validator_spec.rb
sojan-official: let's go ahead with this for now. I will be doing an audit for all the disabled cops during the rails upgrade. will have a look at this then. 
ref: https://linear.app/chatwoot/issue/CW-1574/renable-the-disabled-rubocop-rules

sojan-official (approved):

app/models/message.rb
pranavrajs: @sojan-official @scmmishra This change broke in production. The namespace key is not a mandatory one.

PR #6899, opened 2023-04-12
fix: Pagination bug in chat list

iamsivin:
# Pull Request Template

## Description

Fixes https://linear.app/chatwoot/issue/CW-1424/pagination-not-working-properly

## Type of change

## How Has This Been Tested?
https://www.loom.com/share/5c9d4f1583f045b3965ef5d809bc423f

## Checklist:


app/javascript/dashboard/components/ChatList.vue
```diff
@@ -449,6 +470,9 @@ export default {
         this.resetAndFetchData();
       }
     },
+    chatLists() {
+      this.chatList = this.conversationList;
```
nithindavid, line 474: @iamsivin, any reason we could avoid this? Can we use `conversationList` instead of `chatList`

iamsivin: @nithindavid If we use the conversationList watcher faced a warning "You may have an infinite update loop in watcher with the expression conversationList", that's the reason used chatLists getter as the watch method.

app/javascript/dashboard/components/ChatList.vue
```diff
@@ -347,18 +348,39 @@ export default {
...
+        page: this.conversationListPagination,
         labels: this.label ? [this.label] : undefined,
         teamId: this.teamId || undefined,
         conversationType: this.conversationType || undefined,
         folders: this.hasActiveFolders ? this.savedFoldersValue : undefined,
       };
     },
+    conversationListPagination() {
+      const conversationsPerPage = 25;
+      if (!this.hasAppliedFiltersOrActiveFolders && this.chatList !== []) {
```
nithindavid, line 372: We can move the conditions inside as variables. So we could combine the two if's into a single if condition as in both cases we are returning `return this.currentPage + 1;`.

app/javascript/dashboard/components/ChatList.vue
nithindavid: No for line 372, can we use `conversationList`

app/javascript/dashboard/components/ChatList.vue
iamsivin: Done https://github.com/chatwoot/chatwoot/pull/6899/commits/f0be840e4db6b3b149d874105749ca53925e0aca

nithindavid (approved):
Tested and working as expected.

PR #6904, opened 2023-04-14
fix: Remove fallback phone_number search in WhatsApp event processing

pranavrajs:
The WhatsApp webhook URL is of the format `<installation-url>/webhooks/whatsapp/<phone-number>`

Previously, jobs were processed based on the phone number in the payload (`entry[0].changes[0].value.metadata.phone_number`), with the phone_number from the URL used as a fallback to find the channel.

However, due to a bug on Meta's end (it's unclear if this was intentional or not 🧛), the same webhook URL would receive events for all other phone numbers connected to the business manager.

This means that if there are multiple phone numbers connected, and the user has only connected one phone number in Chatwoot, all conversations related to the other phone numbers would also get added to the inbox in the URL.

This PR removes the fallback and relies only on the payload. Even if the webhook event is received, it would not be processed as the phone number in the payload will not be present in Chatwoot.

Fixes https://linear.app/chatwoot/issue/CW-1510/conversations-from-other-phone-numbers-are-visible-in-the-same-inbox


tejaswinichile (approved):
LGTM 👍🏼

scmmishra (approved):

sojan-official:
Note: This will break for 360dialog. We will have to handle the case seperately when it's a payload from 360 dialogue.  
ref: https://github.com/chatwoot/chatwoot/pull/3112/files#diff-38c6c16543f7d38564b9fdbc0ec8cf455cf6208da5185eabd31eee3a3e2ac631

The `params[:phone_number]` is from the previous PR which we wrote for 360dialog.   360dialog payloads are formatted differently without the `entry` nesting

sojan-official (commented):
might break for 360dialog

PR #6921, opened 2023-04-18
feat: add priority field to conversation

scmmishra:
Seems like this can be done with minimal impact, tried it on staging instance with 2M rows, happened within 200ms

## References

- [StackOverflow](https://stackoverflow.com/a/19527223 )
- [PostgreSQL Docs](https://www.postgresql.org/docs/current/sql-altertable.html#:~:text=When%20a%20column,the%20table%20required.)

> When a column is added with ADD COLUMN and a non-volatile DEFAULT is specified, the default is evaluated at the time of the statement and the result stored in the table's metadata. That value will be used for the column for all existing rows. If no DEFAULT is specified, NULL is used. In neither case is a rewrite of the table required.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -458,7 +458,9 @@ ActiveRecord::Schema.define(version: 2023_04_18_100944) do
     t.index ["first_reply_created_at"], name: "index_conversations_on_first_reply_created_at"
     t.index ["inbox_id"], name: "index_conversations_on_inbox_id"
     t.index ["last_activity_at"], name: "index_conversations_on_last_activity_at"
+    t.index ["priority"], name: "index_conversations_on_priority"
     t.index ["status", "account_id"], name: "index_conversations_on_status_and_account_id"
+    t.index ["status", "priority"], name: "index_conversations_on_status_and_priority"
     t.index ["team_id"], name: "index_conversations_on_team_id"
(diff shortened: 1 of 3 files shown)


scmmishra:
@sojan-official raised this PR against a separate branch

sojan-official (approved):

db/schema.rb
```diff
@@ -437,7 +437,7 @@
     t.datetime "agent_last_seen_at"
     t.jsonb "additional_attributes", default: {}
     t.bigint "contact_inbox_id"
-    t.uuid "uuid", default: -> { "public.gen_random_uuid()" }, null: false
+    t.uuid "uuid", default: -> { "gen_random_uuid()" }, null: false
```
pranavrajs, line 440: Is this intentional?

pranavrajs:
@scmmishra Any reason why this is merged on a branch?

scmmishra:
> @scmmishra Any reason why this is merged on a branch?

Wanted to merge the API side also before merging it to develop

pranavrajs:
@scmmishra Let us only do feature branches if it is absolutely required. In this case, having the change in develop doesn't hurt.

scmmishra:
@pranavrajs noted, makes sense. CC: @muhsin-k

PR #6924, opened 2023-04-18
feat: Support multiple emails in email transcript automation

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-1432/allow-variables-in-automation

https://www.loom.com/share/e41a20766ab74e88bc4938688f540be7

### DIFF
--- a/app/services/action_service.rb
+++ b/app/services/action_service.rb
@@ -52,6 +52,8 @@ class ActionService
   end
 
   def send_email_transcript(emails)
+    emails = emails[0].split(',').compact
+
     emails.each do |email|
       ConversationReplyMailer.with(account: @conversation.account).conversation_transcript(@conversation, email)&.deliver_later
     end
--- a/app/services/action_service.rb
+++ b/app/services/action_service.rb
@@ -52,7 +52,7 @@ class ActionService
   end
 
   def send_email_transcript(emails)
-    emails = emails[0].split(',').compact
+    emails = emails[0].gsub(/\s+/, '').split(',')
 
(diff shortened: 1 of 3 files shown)


app/services/action_service.rb
```diff
@@ -52,6 +52,8 @@ def remove_assigned_team(_params)
   end
 
   def send_email_transcript(emails)
+    emails = emails[0].split(',').compact
```
sojan-official, line 55: Since are allowing emails as comma separated values. there could case that extra spaces could creep in. 
example : "test1@test.com.  ,        test2@test.com" . lets make sure to handle that as well. 

ref: https://github.com/chatwoot/chatwoot/commit/026e03c3073284be4e5d02b71d7027be95f9f8b2 

We could add a common filter function that we use to conversat comma separated emails string to array.

spec/services/automation_rules/action_service_spec.rb
```diff
@@ -87,5 +87,20 @@
         described_class.new(rule, account, conversation).perform
       end
     end
+
+    describe '#perform with send_email_transcript action' do
+      before do
+        rule.actions << { action_name: 'send_email_transcript', action_params: ['contact@example.com,agent@example.com'] }
```
sojan-official, line 93: let's add some spaces to this email list, so that this case is also covered.

sojan-official (commented):
Looks good to me.. a minor suggestion on handling an edge case.

sojan-official (approved):

PR #6925, opened 2023-04-18
feat: Add an action on automation to change the priority

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-1529/add-an-action-on-automation-to-change-the-priority

**Preview**

https://www.loom.com/share/0febd4e92b49499891c0c7e5f900de7a

### DIFF
--- a/app/javascript/dashboard/helper/automationHelper.js
+++ b/app/javascript/dashboard/helper/automationHelper.js
@@ -16,21 +16,25 @@ const MESSAGE_CONDITION_VALUES = [
   },
 ];
 
-const PRIORITY_CONDITION_VALUES = [
+export const PRIORITY_CONDITION_VALUES = [
   {
-    id: '0',
+    id: 'nil',
+    name: 'None',
+  },
+  {
+    id: 'low',
     name: 'Low',
   },
   {
-    id: '1',
+    id: 'medium',
     name: 'Medium',
   },
   {
-    id: '2',
+    id: 'high',
     name: 'High',
   },
   {
-    id: '3',
+    id: 'urgent',
     name: 'Urgent',
   },
 ];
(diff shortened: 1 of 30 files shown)


app/javascript/dashboard/helper/automationHelper.js
```diff
@@ -16,6 +16,25 @@ const MESSAGE_CONDITION_VALUES = [
...
+    name: 'Medium',
+  },
+  {
+    id: '2',
+    name: 'High',
+  },
+  {
+    id: '3',
+    name: 'Urgent',
+  },
```
pranavrajs, line 35: Please add none as well.

app/services/action_service.rb
```diff
@@ -19,6 +19,10 @@ def change_status(status)
     @conversation.update!(status: status[0])
   end
 
+  def change_priority(priority)
+    @conversation.update!(priority: (priority[0] == 'nil' ? nil : priority[0]))
```
muhsin-k, line 23: The `priority[0]` values are "low", "medium", "high", "urgent", or "nil". If it is "nil", we must set it to `nil`, Otherwise, we need to set the priority.

tejaswinichile: Send `null` instead of `nil` from the frontend. It will convert it to `nil` automatically.

app/javascript/dashboard/helper/automationHelper.js
```diff
@@ -16,6 +16,29 @@ const MESSAGE_CONDITION_VALUES = [
   },
 ];
 
+export const PRIORITY_CONDITION_VALUES = [
+  {
+    id: 'nil',
```
tejaswinichile, line 21: Set this to `null`.

app/services/action_service.rb
muhsin-k: We could do that. But we need to tweak some changes in the frontend for automation payload, which may also require changes in other automation rules.

**Reference**
https://github.com/chatwoot/chatwoot/blob/develop/app/javascript/dashboard/helper/actionQueryGenerator.js#L21
Here we always expect some values in id; if we set it to null, the automation will save it in a different format.

tejaswinichile (approved):

PR #6976, opened 2023-04-24
chore: Reorganize Sidekiq Queues

sojan-official:
- Rearrange and reprioritize current sidekiq queues 
- Trim the unnecessary queues

ref: https://linear.app/chatwoot/issue/CW-1480/chore-run-all-sidekiq-jobs-async


app/jobs/action_cable_broadcast_job.rb
```diff
@@ -1,5 +1,5 @@
 class ActionCableBroadcastJob < ApplicationJob
-  queue_as :default
+  queue_as :critical
```
vishnu-narayanan, line 2: @sojan-official This job has the highest throughput. Should this be `critical`?

sojan-official: @vishnu-narayanan, the time to execute the job is minimal, and that's the one job we want to get processed near real-time.  we can switch this around if we observe any issues.

PR #6996, opened 2023-04-26
feat: Refetch the latest messages on action cable reconnect in widget

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-1513/dashboard-on-cable-reconnect-fetch-messages

### DIFF
--- a/app/controllers/api/v1/widget/messages_controller.rb
+++ b/app/controllers/api/v1/widget/messages_controller.rb
@@ -48,7 +48,8 @@ class Api::V1::Widget::MessagesController < Api::V1::Widget::BaseController
   def message_finder_params
     {
       filter_internal_messages: true,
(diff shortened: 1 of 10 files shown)


app/controllers/api/v1/widget/messages_controller.rb
```diff
@@ -48,7 +48,8 @@ def set_conversation
   def message_finder_params
     {
       filter_internal_messages: true,
-      before: permitted_params[:before]
+      before: permitted_params[:before],
+      after: permitted_params[:after]
```
muhsin-k, line 52: Note: Specs for param `after` is already added https://github.com/chatwoot/chatwoot/blob/develop/spec/finders/message_finder_spec.rb#L51

scmmishra (approved):

PR #7010, opened 2023-04-28
fix: CC conditions when the last email is from someone else

scmmishra:
Fix a case when the last sent email is from a different sender, and not from the conversation contact. In that case, we reply to the author of the last email and put the conversation contact in CC.

Note in the following screenshots, the sender of the original mail is `someone-else@chatwoot.com` which is different from the contact email. When this happens, the reply box auto-populates the correct values and sends it across to the backend. Where the email is sent to the person tagged in `to_emails`

Here's the conversation

CleanShot 2023-05-22 at 19 21 14@2x

Here's the mail

CleanShot 2023-05-22 at 19 19 54@2x

### Pending Tasks

- [ ] Unit Tests for Backend
- [ ] Unit Tests for Frontend
- [ ] To Mail UI in the chatbox

Fixes https://github.com/chatwoot/chatwoot/pull/7010


nithindavid:
image
image

@scmmishra The first pic shows the mail sent from nithin@chatwoot.com but on chatwoot dashboard the cc: field has the original sender address. Is this expected?

PR #7028, opened 2023-05-04
feat(perf): improve performance of conversation filter API

scmmishra:
## Description

This PR fixes some N+1 queries for the conversation filter service. This caches the labels and precomputes the total message count. 

## Type of change

#### Before

CleanShot 2023-05-04 at 13 42 20@2x

#### After

CleanShot 2023-05-04 at 13 41 09@2x


app/views/api/v1/conversations/partials/_conversation.json.jbuilder
```diff
@@ -17,7 +17,7 @@ json.meta do
 end
 
 json.id conversation.display_id
-if conversation.messages.count.zero?
+if conversation.message_count.zero?
   json.messages []
 elsif conversation.unread_incoming_messages.count.zero?
```
pranavrajs, line 22: @scmmishra Can we preload this as well?

scmmishra: I managed to do it, however in `models/messages.rb`, it is calculated again in `push_event_data` method. So the query is inevitable. Couldn't find a way to cache that

scmmishra:
@sojan-official this is ready to go

app/views/api/v1/conversations/partials/_conversation.json.jbuilder
sojan-official: We can avoid doing count here but rather go for conversation.messages.any? Instead. 
The count will trigger an additional count query, and tbh we don't need the specific value of count here anyways. 

https://twitter.com/nateberkopec/status/1654199212479860768?s=20 
https://www.speedshop.co/2019/01/10/three-activerecord-mistakes.html

app/services/conversations/filter_service.rb
```diff
@@ -60,6 +60,12 @@ def conversations
     @conversations = @conversations.includes(
```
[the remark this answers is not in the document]
scmmishra: Not sure. I can try itout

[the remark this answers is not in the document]
sojan-official: we can pair on this.

scmmishra:
Closing this PR, this history has too many unrelated changes. Raised a new PR here: https://github.com/chatwoot/chatwoot/pull/7120

PR #7035, opened 2023-05-06
feat: add feature flag for Audit Log UI

scmmishra:
This PR adds a feature flag for Audit logs

## Type of change


vishnu-narayanan (approved):
LGTM

enterprise/app/controllers/api/v1/accounts/audit_logs_controller.rb
```diff
@@ -21,6 +21,14 @@ def show
   private
 
   def fetch_audit
-    @audit_logs = Current.account.associated_audits.order(created_at: :desc)
+    @audit_logs = if audit_logs_enabled?
+                    Current.account.associated_audits.order(created_at: :desc)
+                  else
+                    Current.account.associated_audits.none
```
pranavrajs, line 27: @scmmishra You cannot return an empty array as there would be a chained request after this method. Use `Model.none` to return chainable relation with zero records

enterprise/app/controllers/api/v1/accounts/audit_logs_controller.rb
```diff
@@ -21,6 +21,14 @@ def show
...
+                  end
+  end
+
+  def audit_logs_enabled?
```
pranavrajs, line 31: If the method is returning a boolean, the method name with `?` is preferred.

PR #7052, opened 2023-05-10
chore: Add migration to set the default empty string value for contact name

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-1650/issue-with-the-contact-name


pranavrajs:
@sojan-official @muhsin-k This wouldn't change the existing records, right? The existing records would still have the null value.

sojan-official:
@pranavrajs  That is correct. Altering existing values would require scanning over the whole table and performing the update.

PR #7053, opened 2023-05-11
feat: Order conversation by priority

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-1714/ui-sort-conversation-by-priority

https://github.com/chatwoot/chatwoot/assets/12408980/584e24e1-0038-4ca2-bdd1-eab3096ef716

### DIFF
--- a/app/javascript/dashboard/store/modules/conversations/getters.js
+++ b/app/javascript/dashboard/store/modules/conversations/getters.js
@@ -1,4 +1,7 @@
-import { MESSAGE_TYPE } from 'shared/constants/messages';
+import {
+  MESSAGE_TYPE,
+  CONVERSATION_PRIORITY_ORDER,
+} from 'shared/constants/messages';
 import { applyPageFilters } from './helpers';
 
 export const getSelectedChatConversation = ({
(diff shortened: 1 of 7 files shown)


app/javascript/dashboard/store/modules/conversations/getters.js
```diff
@@ -13,6 +16,12 @@ const getters = {
     const comparator = {
       latest: (a, b) => b.last_activity_at - a.last_activity_at,
       sort_on_created_at: (a, b) => a.created_at - b.created_at,
+      sort_on_priority: (a, b) => {
+        return (
+          CONVERSATION_PRIORITY_ORDER[a.priority] -
+          CONVERSATION_PRIORITY_ORDER[b.priority]
+        );
+      },
```
scmmishra, line 24: Why the need to add this here? Do we not get it from the server?

muhsin-k: @scmmishra If we are adding any new conversation to the state via action cable events, we should also include the necessary sorting in the getter.

scmmishra: Ahh, makes sense!

scmmishra (approved):
LGTM! just take a look at the failing build

Issue #7000, opened 2023-04-26
[CW-1728] Wrong cache information for GCS storage


hugodeco:
### Describe the bug

Attachments in conversations wont be available after a few minutes;

![image](https://user-images.githubusercontent.com/2934665/234691570-17c997b2-70be-4dc4-aa4f-90751f89b704.png)
![image](https://user-images.githubusercontent.com/2934665/234691708-22905b8b-b754-49ad-ba8b-b511d141a62b.png)
![image](https://user-images.githubusercontent.com/2934665/234691819-408603c4-363d-4913-bfbb-4ecfb2ab844d.png)

### To Reproduce

1. Using the GCS storage
2. Receive an Audio Attachment OGG
3. Wait a few minutes
4. It wont be available anymore

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

Browser

<sub>[CW-1728](https://linear.app/chatwoot/issue/CW-1728/wrong-cache-information-for-gcs-storage)</sub>

artbras:
Os áudios não estão sendo reproduzidos + de uma 1x. O atendente não consegue ouvir os áudios das mensagens pela segunda vez. Vocês poderiam verificar isso por favor?

pranavrajs:
I can reproduce this. There is a report of the same from a paid customer as well.

PR #7062, opened 2023-05-11
fix: Avoid audio URLs getting cached at the frontend

pranavrajs:
Fixes https://linear.app/chatwoot/issue/CW-1728/wrong-cache-information-for-gcs-storage
Fixes https://github.com/chatwoot/chatwoot/issues/7000

- Force reload audio on each re-render.

@scmmishra Do you have any other ideas here? The issue here is that the redirect gets cached on the frontend, which eventually expires and the audio becomes in-accessible

### DIFF
--- a/app/javascript/dashboard/components/widgets/conversation/Message.vue
+++ b/app/javascript/dashboard/components/widgets/conversation/Message.vue
@@ -59,7 +59,7 @@
               controls
               class="skip-context-menu"
             >
-              <source :src="attachment.data_url" />
+              <source :src="`${attachment.data_url}?t=${Date.now()}`" />
             </audio>
             <bubble-video
               v-else-if="attachment.file_type === 'video'"


pranavrajs:
Merging this for the time being, we can revert or come up with a better solution later.

civilcoder55:
@pranavrajs 
I faced this issue and it happens only on Chrome browsers, active storage redirect URLs have Cache-Control with max-age
and browsers don't respect it and never invalidate so I fixed it by removing the cache by adding Cache-Control with no-store
what do you think?

pranavrajs:
> I faced this issue and it happens only on Chrome browsers, active storage redirect URLs have Cache-Control with max-age
and browsers don't respect it and never invalidate so I fixed it by removing the cache by adding Cache-Control with no-store
what do you think?

@civilcoder55 This makes sense. I forced the refresh by adding a timestamp param. No-store seems like a better solution. Would you be able to raise a PR? I'm happy to test it out.

PR #7063, opened 2023-05-11
fix: Avoid blob URLs getting cached

civilcoder55:
Fixes https://github.com/chatwoot/chatwoot/issues/7000

related: https://github.com/chatwoot/chatwoot/pull/7062#issuecomment-1544601839

I fixed the issue by overriding the active storage blob redirect controller and removing the cache

Original rails 7 controller https://github.com/rails/rails/blob/7-0-stable/activestorage/app/controllers/active_storage/blobs/redirect_controller.rb

@pranavrajs

### DIFF
--- a/app/controllers/active_storage/blobs/redirect_controller.rb
+++ b/app/controllers/active_storage/blobs/redirect_controller.rb
@@ -1,0 +1,8 @@
+class ActiveStorage::Blobs::RedirectController < ActiveStorage::BaseController
+  include ActiveStorage::SetBlob
+
+  def show
+    response.set_header('Cache-Control', 'no-store')
+    redirect_to @blob.url(disposition: params[:disposition]), allow_other_host: true
+  end
+end


pranavrajs:
@civilcoder55 Thanks for the PR. We would have to check with the non-audio/video uploads to make sure that this doesn't break it. We will take it forward after the investigation.

sojan-official:
Holding this for now since we already  merged https://github.com/chatwoot/chatwoot/pull/7062 .
Need to look further into the implications of overriding the default controller just for this specific case.

civilcoder55:
@sojan-official 
I agree with you, but I guess we need to do the same for video as the cache affects audio and videos only

pranavrajs:
@civilcoder55 Closing this as it has been slate for a while. We will re-open this if there is a need for this. Sorry about the wasted effort.

PR #7101, opened 2023-05-17
Customer filter per account per user

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-1609/limit-the-number-of-custom-views-to-50-per-person-per-account

### DIFF
--- a/app/models/custom_filter.rb
+++ b/app/models/custom_filter.rb
@@ -21,4 +21,11 @@ class CustomFilter < ApplicationRecord
   belongs_to :account
 
   enum filter_type: { conversation: 0, contact: 1, report: 2 }
+  validate :validate_number_of_filters
+
+  def validate_number_of_filters
+    return true if self.class.where(account_id: account_id, user_id: user_id).size <= 50
+
+    errors.add :account_id, 'We do not allow creating >50 custom views per account per user'
+  end
 end
--- a/spec/controllers/api/v1/accounts/custom_filters_controller_spec.rb
+++ b/spec/controllers/api/v1/accounts/custom_filters_controller_spec.rb
@@ -66,6 +66,8 @@ RSpec.describe 'Custom Filters API', type: :request do
 
     context 'when it is an authenticated user' do
       let(:user) { create(:user, account: account) }
+      let(:new_account) { create(:account) }
+      let(:new_user) { create(:user, account: new_account) }
 
       it 'creates the filter' do
         expect do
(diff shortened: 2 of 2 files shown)


app/models/custom_filter.rb
```diff
@@ -21,4 +21,11 @@ class CustomFilter < ApplicationRecord
   belongs_to :account
 
   enum filter_type: { conversation: 0, contact: 1, report: 2 }
+  validate :validate_number_of_filters
+
+  def validate_number_of_filters
+    return true if self.class.where(account_id: account_id, user_id: user_id).size <= 50
+
+    errors.add :account_id, 'We do not allow creating >50 custom views per account per user'
```
pranavrajs, line 29: i18n.

pranavrajs, line 29:
```suggestion
    errors.add :account_id, 'The maximum number of allowed custom filters per account is 50.'
```

app/models/custom_filter.rb
```diff
@@ -21,4 +21,11 @@ class CustomFilter < ApplicationRecord
   belongs_to :account
 
   enum filter_type: { conversation: 0, contact: 1, report: 2 }
+  validate :validate_number_of_filters
+
+  def validate_number_of_filters
+    return true if self.class.where(account_id: account_id, user_id: user_id).size <= 50
```
sojan-official, line 27: account.custom_filters.where(user_id: user_id).size ?

sojan-official: also this could be a constant `MAX_FILTER_PER_USER` . so that you can mock it in the spec to a lower value  like 2 or 3. will reduce the complexity of creating 51 objects for the spec

spec/controllers/api/v1/accounts/custom_filters_controller_spec.rb
```diff
@@ -77,6 +79,20 @@
         json_response = JSON.parse(response.body)
         expect(json_response['name']).to eq 'vip-customers'
       end
+
+      it 'gives the error for 51st record' do
+        51.times do
```
sojan-official, line 84: above comment in class about making this a constant

app/models/custom_filter.rb
```diff
@@ -17,8 +17,16 @@
 #  index_custom_filters_on_user_id     (user_id)
 #
 class CustomFilter < ApplicationRecord
+  MAX_FILTER_PER_USER = 50
```
scmmishra, line 20: Should we make this a global config?

tejaswinichile: @scmmishra We can, but if the admin changes it to >50, it will be hard to handle the filter considering the performance.

Let's revisit this when the filter is ready and change it according to the performance.
Right now, we need a clear idea about the same.

scmmishra: Sounds good

PR #7104, opened 2023-05-17
chore: Check assignee exists or not before rendering the assignee avatar

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-1778/typeerror-cannot-read-properties-of-null-reading-assignee


scmmishra (approved):

scmmishra:
Tested locally! LGTM

PR #7107, opened 2023-05-17
chore: Add delay for slack job when message has attachments

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-1717/bug-slack-attachments-not-working-on-cloud


lib/integrations/slack/send_on_slack_service.rb
```diff
@@ -79,6 +79,7 @@ def avatar_url(sender)
 
   def send_message
     post_message if message_content.present?
+    sleep(3.seconds)
```
sojan-official, line 82: This will block a worker thread for 3 seconds, which isn't efficient. lets handle this in `HookJob`

ref: https://github.com/chatwoot/chatwoot/blob/b988a01df367e725d30c0870850247a66e1a0077/app/jobs/hook_job.rb#L24


let's follow the same design which we use for sendreplyjob
https://github.com/chatwoot/chatwoot/blob/b988a01df367e725d30c0870850247a66e1a0077/app/models/message.rb#L279

- move the send_on_slack_service call to a send_on_slack job
ie check whether message has attachment. if there is attachment schedule the send on slack job after for 2 seconds later. in other cases right away. 

also specs for the same for hook job.

sojan-official (approved):

PR #7120, opened 2023-05-18
feat(perf): improve performance of conversation filter API [CW-1605]

scmmishra:
## Description

This PR fixes some N+1 queries for the conversation filter service. This caches the labels and precomputes the total message count. 

## Type of change

### Baseline

Total Queries: 251

CleanShot 2023-05-18 at 18 07 44@2x

### Optimized

Total Queries: 213

CleanShot 2023-05-18 at 18 08 15@2x

### DIFF
--- a/app/views/api/v1/conversations/partials/_conversation.json.jbuilder
+++ b/app/views/api/v1/conversations/partials/_conversation.json.jbuilder
@@ -17,7 +17,7 @@ json.meta do
 end
 
 json.id conversation.display_id
-if conversation.messages.count.zero?
+if !conversation.messages.first.present?
   json.messages []
 elsif conversation.unread_incoming_messages.count.zero?
   json.messages [conversation.messages.includes([{ attachments: [{ file_attachment: [:blob] }] }]).last.try(:push_event_data)]
--- a/app/views/api/v1/conversations/partials/_conversation.json.jbuilder
+++ b/app/views/api/v1/conversations/partials/_conversation.json.jbuilder
@@ -17,7 +17,7 @@ json.meta do
 end
 
 json.id conversation.display_id
-if !conversation.messages.first.present?
+if conversation.messages.first.blank?
   json.messages []
 elsif conversation.unread_incoming_messages.count.zero?
   json.messages [conversation.messages.includes([{ attachments: [{ file_attachment: [:blob] }] }]).last.try(:push_event_data)]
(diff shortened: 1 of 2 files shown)


scmmishra:
@sojan-official @pranavrajs If you see the following lines https://github.com/chatwoot/chatwoot/blob/590ce788b99c7634936af07bf999c05abadb0416/app/views/api/v1/conversations/partials/_conversation.json.jbuilder#L23-L25

We request the `push_event_data` from the messages, this includes attachments too.

https://github.com/chatwoot/chatwoot/blob/590ce788b99c7634936af07bf999c05abadb0416/app/models/message.rb#L134-L148

We can remove the attachments from the query with an optional parameter `include_attachments` and use it in the jbuilder, this will reduce the number of queries further

scmmishra:
@sojan-official @pranavrajs found another opportunity,

https://github.com/chatwoot/chatwoot/blob/590ce788b99c7634936af07bf999c05abadb0416/app/models/contact.rb#L109-L121

The `avatar_url` is causing a N+1 query, I tried adding `{ messages: { sender: { avatar_attachment: [:blob] }]}}` to the `include` in `filter_service.rb` but it didn't help. LMK if you have any ideas.

Regardless, the current PR should also provide some improvements

app/views/api/v1/conversations/partials/_conversation.json.jbuilder
```diff
@@ -17,7 +17,7 @@ json.meta do
 end
 
 json.id conversation.display_id
-if conversation.messages.count.zero?
+if conversation.messages.first.blank?
   json.messages []
 elsif conversation.unread_incoming_messages.count.zero?
```
sojan-official, line 22: we can avoid the count query here.  do first.present?

scmmishra: I did `first.present?` earlier, then rubocop suggested to make it `first.blank?`Are the two different in their query execution?

sojan-official: `first.blank?` should be good . lets do the same for  line:22 as well.  the count query seems overkill there as well

sojan-official (commented):
LGTM. let's take the  other suggestions in subsequent PRs

app/views/api/v1/conversations/partials/_conversation.json.jbuilder
scmmishra: We could do that but then `conversation.unread_incoming_messages.count` is fetched again on line 43, unconditionally. So I figured it better be fetched on 22 as well

sojan-official (approved):

PR #7151, opened 2023-05-22
feat: Add an option to use darkMode only in the widget

pranavrajs:
- Added an option to use dark Mode only in the widget. 
- Accept `dark` in the sdk options.
- Change the Tailwind Dark Mode style strategy to `class` to override color changes based on the OS preference.

https://www.loom.com/share/dc16ac938c6f454e882e32b2c07b30f1


pranavrajs:
@scmmishra @iamsivin Please take a look.

scmmishra (approved):

PR #7164, opened 2023-05-22
fix: set custom filter count in redis

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-1610/show-the-count-of-conversations-in-the-view-in-the-sidebar

### DIFF
--- a/app/jobs/custom_filter_records_count_update_job.rb
+++ b/app/jobs/custom_filter_records_count_update_job.rb
@@ -1,0 +1,11 @@
+class CustomFilterRecordsCountUpdateJob < ApplicationJob
+  queue_as :low
+
+  def perform
+    CustomFilter.find_each(batch_size: 25) do |filter|
+      next if filter.fetch_record_count_from_redis.nil?
+
+      filter.set_records_count
+    end
+  end
+end
--- a/app/models/custom_filter.rb
+++ b/app/models/custom_filter.rb
@@ -21,4 +21,26 @@ class CustomFilter < ApplicationRecord
   belongs_to :account
 
   enum filter_type: { conversation: 0, contact: 1, report: 2 }
+
+  def records_count
+    get_record_count_from_redis || set_record_count_in_redis
+  end
+
+  def filter_records
+    Conversations::FilterService.new(query.with_indifferent_access, Current.user).perform
+  end
+
+  def set_record_count_in_redis
+    records = filter_records
+    Redis::Alfred.set(filter_count_key, records[:count][:all_count])
+    get_record_count_from_redis
+  end
+
+  def fetch_record_count_from_redis
+    Redis::Alfred.get(filter_count_key)
+  end
+
+  def filter_count_key
+    format(::Redis::Alfred::CUSTOM_FILTER_RECORDS_COUNT_KEY, account_id: account_id, filter_id: id)
+  end
 end
--- a/app/views/api/v1/models/_custom_filter.json.jbuilder
+++ b/app/views/api/v1/models/_custom_filter.json.jbuilder
@@ -4,3 +4,4 @@ json.filter_type resource.filter_type
 json.query resource.query
 json.created_at resource.created_at
 json.updated_at resource.updated_at
+json.count resource.records_count
--- a/lib/redis/redis_keys.rb
+++ b/lib/redis/redis_keys.rb
@@ -31,4 +31,5 @@ module Redis::RedisKeys
   LATEST_CHATWOOT_VERSION = 'LATEST_CHATWOOT_VERSION'.freeze
   # Check if a message create with same source-id is in progress?
   MESSAGE_SOURCE_KEY = 'MESSAGE_SOURCE_KEY::%<id>s'.freeze
+  CUSTOM_FILTER_RECORDS_COUNT_KEY = 'CUSTOM_FILTER::%<account_id>d::%<filter_id>d'.freeze
 end


app/jobs/custom_filter_records_count_update_job.rb
```diff
@@ -0,0 +1,14 @@
+class CustomFilterRecordsCountUpdateJob < ApplicationJob
+  queue_as :low
```
tejaswinichile, line 2: @sojan-official We wouldn't need this anymore; this line would take care of that, https://github.com/chatwoot/chatwoot/pull/7164/files#diff-8211fe4df2fdd38b2e0e66edd5f4b6bddc99907b9e5da03aeb006e4bb05b7a25R28

And this would happen only once until the filtered conversation count is not set in Redis, else it would fetch the count from Redis.

Let me know if this is okay then I will remove this job.

spec/controllers/api/v1/accounts/custom_filters_controller_spec.rb
```diff
@@ -65,23 +95,18 @@
...
 
         expect(response).to have_http_status(:success)
         json_response = response.parsed_body
         expect(json_response['name']).to eq 'vip-customers'
+        expect(json_response['count']).to eq '1'
       end
 
       it 'gives the error for 51st record' do
-        CustomFilter::MAX_FILTER_PER_USER.times do
+        (CustomFilter::MAX_FILTER_PER_USER - 1).times do
```
sojan-official, line 109: Rather than doing the ` -1`. let's destroy all the filters above this or do an account.custom_filters count as a query, which will be more readable  and won't let the reader wonder about the logic behind `-1`

app/jobs/custom_filter_records_count_update_job.rb
sojan-official: @tejaswinichile, let's decouple the logic into two jobs. 
1. iterate over all the custom filters and  schedule sync jobs for each filter
2. sync job for a custom filter that calculates its value. 

Doing this in one single job would create a super long-running job.

app/jobs/custom_filter_records_count_update_job.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def perform
+    CustomFilter.find_each(batch_size: 25) do |filter|
+      Current.account = filter.account
+
+      next if filter.fetch_record_count_from_redis.nil?
+
+      filter.set_records_count
+      Current.reset
```
sojan-official, line 11: Setting Current here needs improvement.  if the filter fails it might not be reset. rather make the account an argument for filter service that passing it on via current. 

also, move the filter calculation logic to a separate job as mentioned in above comment

app/models/custom_filter.rb
```diff
@@ -24,6 +24,28 @@ class CustomFilter < ApplicationRecord
   enum filter_type: { conversation: 0, contact: 1, report: 2 }
   validate :validate_number_of_filters
 
+  def records_count
+    fetch_record_count_from_redis || set_record_count_in_redis
```
sojan-official, line 28: Currently, some of the filters take quite a long while. so calculating it on the fly for each filter might time out the request.

If the `key` is nil lets return `nil` and enqueue the job to update the key rather. 

So that the value will be available in subsequent requests.

app/models/custom_filter.rb
```diff
@@ -24,6 +24,28 @@ class CustomFilter < ApplicationRecord
...
+  end
+
+  def filter_records
```
sojan-official, line 31: Should we use the filter.user_id here?
also lets pass filter.account as an argument that relying on current account in the filter service

let's try our best to keep usage of Current into controllers alone.

app/policies/custom_filter_policy.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def index?
+    @account_user.administrator? || @account_user.agent?
+  end
+
+  def update?
+    @account_user.administrator? || @account_user.agent?
+  end
+
```
sojan-official, line 17: we could move the logic  that `agent x` is not able to view filters created by `agent y`  etc here in future. the policy file seems like the right place for that

sojan-official (commented):
- added some comments

app/models/custom_filter.rb
tejaswinichile: `Current.account` usage has been removed.

app/jobs/sync_custom_filter_count_job.rb
```diff
@@ -0,0 +1,9 @@
+class SyncCustomFilterCountJob < ApplicationJob
+  queue_as :low
+
+  def perform(filter)
+    return if filter.fetch_record_count_from_redis.nil?
```
scmmishra, line 5: What if this data is stale? We don't set the data with expiry right?

app/models/custom_filter.rb
scmmishra: Any reason we need the user in the first place?

app/models/custom_filter.rb
```diff
@@ -24,6 +24,29 @@ class CustomFilter < ApplicationRecord
...
+  def records_count
+    fetch_record_count_from_redis
+  end
+
+  def filter_records
+    user = account.users.try(:first)
+    Conversations::FilterService.new(query.with_indifferent_access, user, account).perform
+  end
+
+  def set_record_count_in_redis
```
scmmishra, line 36: Is updating the filter possible? We should invalidate the record count if the filter is updated, we can be a bit smart about it by ensuring that the filter conditions have changed or not.

app/models/custom_filter.rb
```diff
@@ -24,6 +24,29 @@ class CustomFilter < ApplicationRecord
   enum filter_type: { conversation: 0, contact: 1, report: 2 }
   validate :validate_number_of_filters
 
+  def records_count
+    fetch_record_count_from_redis
```
scmmishra, line 28: I think this should have a fallback, in case the record count does not exist in redis, it should calculate, save it in redis and return it. 
CC: @sojan-official what do you think?

sojan-official: yep . ref: https://github.com/chatwoot/chatwoot/pull/7164#discussion_r1205545166 

- lets requeue the job in case count is nil

app/jobs/sync_custom_filter_count_job.rb
tejaswinichile: No. We will update the data only when a new conversation is created, considering that will be the only trigger to update the filter records count.

But if we can have a job to update the count too, after TTL.

app/models/custom_filter.rb
tejaswinichile: The main filter query involves the user in `FilterService.` for assigned conversations. We can update the query, but it affects automation as well as custom filters, so we kept it for another ticket.

app/models/custom_filter.rb
tejaswinichile: Yes, we have a ticket to update it on the conversation update. https://linear.app/chatwoot/issue/CW-1611/refresh-the-count-without-reload

app/models/custom_filter.rb
tejaswinichile: @sojan-official @scmmishra I have added these changes please have a look.

app/models/custom_filter.rb
```diff
@@ -24,6 +24,31 @@ class CustomFilter < ApplicationRecord
...
+  end
+
+  def filter_records
+    user = account.users.try(:first)
```
sojan-official, line 32: @tejaswinichile I am not sure we can do this.  The conversation permission for various agents will be based on their inbox access, right?  we need to get the Current user here to ensure we are only picking conversations which the agent has access to ?

sojan-official (commented):
@tejaswinichile Added a comment on fetching the user

app/models/custom_filter.rb
tejaswinichile: Isn't that for mine-records, should we do that for custom-filter as well?

sojan-official: Yep. even in all tabs. agent a might not have access to inbox A., so conversations from inbox A shouldn't be shown on their feed.

tejaswinichile: Note: We have the custom-filter and user association. We only display `user.custom_filters` in the sidebar.

sojan-official:
@tejaswinichile  code LGTM overall. But didn't get a chance to test this in local.
@scmmishra Could you take another pass and QA this?

app/models/custom_filter.rb
```diff
@@ -24,6 +24,29 @@ class CustomFilter < ApplicationRecord
...
+  end
+
+  def set_record_count_in_redis
+    records = filter_records
+    Redis::Alfred.set(filter_count_key, records[:count][:all_count])
+  end
+
+  def fetch_record_count_from_redis
+    number_of_records = Redis::Alfred.get(filter_count_key)
+    SyncCustomFilterCountJob.perform_later(self) if number_of_records.nil?
```
scmmishra, line 42: Should we keep this job synchronous? If the records are nil or outdated, we calculate on the fly and return. This ensures we always share the correct values

app/jobs/sync_custom_filter_count_job.rb
```diff
@@ -0,0 +1,9 @@
+class SyncCustomFilterCountJob < ApplicationJob
+  queue_as :low
+
+  def perform(filter)
+    Redis::Alfred.set(filter.filter_count_key, 0) if filter.filter_records.nil?
```
scmmishra, line 5: Do we need this explicitly, won't it be handled in `filter.set_record_count_in_redis`?

scmmishra (changes requested):
LGTM, just two small comments

scmmishra (commented):
Another thing I noticed, the count is returned as a string, can be parse it to number before sending it across?

app/models/custom_filter.rb
sojan-official: It would significantly overload the filter endpoint, as the calculation would be pretty time-consuming. + we would be iterating over each filter and will be calculating for each right?

tejaswinichile:
> Another thing I noticed, the count is returned as a string, can be parse it to number before sending it across?

@scmmishra Yes, let me make that change

app/models/custom_filter.rb
tejaswinichile: > We would be iterating over each filter and will be calculating for each, right?

Yes, that's why we didn't keep it in the synchronous job, it's getting updated on every new conversation update.

app/jobs/sync_custom_filter_count_job.rb
tejaswinichile: @scmmishra  As in it will be the same, there we would need to add 0 for nil records values.

app/models/custom_filter.rb
scmmishra: Noted, in that case this looks good

scmmishra:
@tejaswinichile are the open comments still pending? Can you close them if they are not relevant or are completed?

scmmishra (approved):

tejaswinichile:
@sojan-official @scmmishra This is working well.
https://www.loom.com/share/ca051c02ef574a9492106012cf1f5b8d?sid=c3799ad9-289b-4092-aafb-92fea4be79c9

PR #7179, opened 2023-05-24
fix: issue with slack job

tejaswinichile:
Screenshot 2023-05-24 at 4 48 26 PM

@sojan-official @pranavrajs This will resolve the issue tested it on production with exact changes :|


app/jobs/send_on_slack_job.rb
```diff
@@ -1,8 +1,7 @@
 class SendOnSlackJob < ApplicationJob
   queue_as :medium
-  pattr_initialize [:message!, :hook!]
```
tejaswinichile, line 3: This mapping never worked for `SendOnSlackJob` in the previous changes.

PR #7180, opened 2023-05-24
feat: allow superadmins to reset cache keys for IndexedDB

scmmishra:
This PR allows super admins to reset cache for an account. This will force the frontend to refetch the data again on next load

CleanShot 2023-05-24 at 12 21 25@2x

## Type of change

## Checklist:


app/models/concerns/cache_keys.rb
```diff
@@ -4,29 +4,41 @@ module CacheKeys
...
+    keys = {}
+    self.class.cacheable_models.each do |model|
+      keys[model.name.underscore.to_sym] = fetch_value_for_key(id, model.name.underscore)
+    end
+
+    keys
   end
 
   def invalidate_cache_key_for(key)
     prefixed_cache_key = get_prefixed_cache_key(id, key)
```
sojan-official, line 22:
```
 def get_prefixed_cache_key(account_id, key)
    "idb-cache-key-account-#{account_id}-#{key}"
  end
```

@scmmishra  We are expecting account_id as the argument for the cache keys right ?
but it seems like we are passing get_prefixed_cache_key(id, key)  -> id here ? which is the object id ? 

could that be the reason for original issue ?

scmmishra: Let me explore this

scmmishra: @sojan-official just confirmed, the ID passed here is the ID for the account

sojan-official: @scmmishra  Can we add a spec to ensure that it's the correct cache that gets cleared?

scmmishra: Sure, I'll do that

scmmishra: @sojan-official done

sojan-official (approved):
- Tested locally with the console to ensure the cache is cleared. 
- Added additional controller specs to test end to end

PR #7190, opened 2023-05-25
feat: [CW-1611] refresh the custom views count without reload

tejaswinichile:
fixes: https://linear.app/chatwoot/issue/CW-1611/refresh-the-count-without-reload

https://www.loom.com/share/b41ef8b195fd4d13821003f74d5e0d3b?sid=618a3eaf-67a8-4b32-9fa6-9eef6049f1a3


tejaswinichile:
@sojan-official I need to rebase this one with https://github.com/chatwoot/chatwoot/pull/7438 for the test review.

tejaswinichile:
Tested on staging as well https://www.loom.com/share/21e1bc7f001c483cb514d7db2d515e29

app/controllers/api/v1/accounts/custom_filters_controller.rb
```diff
@@ -13,10 +15,13 @@ def create
       permitted_payload.merge(account_id: Current.account.id)
     )
     render json: { error: @custom_filter.errors.messages }, status: :unprocessable_entity and return unless @custom_filter.valid?
+
+    update_filter_conversation_count(@custom_filter)
   end
 
   def update
     @custom_filter.update!(permitted_payload)
+    update_filter_conversation_count(@custom_filter)
```
sojan-official, line 24: This logic looks more suited for the controller than the model ( based on the pattern which we follow )
- the `update_filter_conversation_count` can happen in an after-save callback. 
-  `Events::Types` is already included in application record

app/controllers/api/v1/accounts/custom_filters_controller.rb
```diff
@@ -26,6 +31,14 @@ def destroy
...
   private
 
+  def update_filter_conversation_count(filter)
+    records = filter.filter_records
+
+    Redis::Alfred.set(filter.filter_count_key, records[:count][:all_count])
+
+    Rails.configuration.dispatcher.dispatch(CUSTOM_FILTER_UPDATED, Time.zone.now, custom_filter: filter)
+  end
+
```
sojan-official, line 41: We already have this logic in `custom_filter.rb` no need to duplicate it here again.

app/listeners/custom_views_listener.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def dispatch_custom_filter_event(account)
+    account.custom_filters.each do |filter|
+      records = filter.filter_records
+
+      next if (records[:count][:all_count]).zero? || records[:count][:all_count] == Redis::Alfred.get(filter.filter_count_key).to_i
+
+      Redis::Alfred.set(filter.filter_count_key, records[:count][:all_count])
+
+      Rails.configuration.dispatcher.dispatch(CUSTOM_FILTER_UPDATED, Time.zone.now, custom_filter: filter)
```
sojan-official, line 28: we don't need to duplicate this logic in multiple places 

the following logic can happen in methods in the `custom_filter` model  
- setting the count in Redis 
- dispatching the update event

app/listeners/action_cable_listener.rb
```diff
@@ -156,6 +156,14 @@ def conversation_mentioned(event)
...
   end
 
+  def custom_filter_updated(event)
+    custom_filter = event.data[:custom_filter]
+    account = custom_filter.account
+    tokens = user_tokens(account, account.agents)
+
+    broadcast(account, tokens, CUSTOM_FILTER_UPDATED, custom_filter.push_event_data)
+  end
+
```
sojan-official, line 166: specs for the `custom_filter_update` event in action cable listener
Also, since the filters are private you only need to broadcast the event to the filter owner right?

app/services/conversations/filter_service.rb
```diff
@@ -22,14 +22,18 @@ def perform
...
       current_filter = conversation_filters[query_hash['attribute_key']]
       @query_string += conversation_query_string(current_filter, query_hash, current_index)
     end
 
-    base_relation.where(@query_string, @filter_values.with_indifferent_access)
+    if records.present?
+      records.where(@query_string, @filter_values.with_indifferent_access)
+    else
+      base_relation.where(@query_string, @filter_values.with_indifferent_access)
+    end
```
sojan-official, line 36: - specs to cover this change

sojan-official (commented):
@tejaswinichile  Added a few comments.

app/listeners/action_cable_listener.rb
tejaswinichile: > Also, since the filters are private you only need to broadcast the event to the filter owner right?

Yes

tejaswinichile: > Also, since the filters are private you only need to broadcast the event to the filter owner right?

@sojan-official Sending user access token, wouldn't help here?

tejaswinichile:
@sojan-official Resolved all the comments, PTAL.

tejaswinichile:
https://www.loom.com/share/820591798c954ee5bccc824a071c2538

The issue that we resolved in the last commit
Screenshot 2023-07-17 at 1 15 05 PM

app/javascript/dashboard/store/modules/customViews.js
```diff
@@ -61,6 +61,13 @@ export const actions = {
       commit(types.SET_CUSTOM_VIEW_UI_FLAG, { isCreating: false });
     }
   },
+  setCustomViewCount: async ({ commit }, { filter }) => {
+    try {
+      commit(types.SET_CUSTOM_VIEW_COUNT, filter);
+    } catch (error) {
+      throw new Error(error);
+    }
```
scmmishra, line 69: The try catch does not really do anything, let's log it to sentry and let it fail silently

app/javascript/dashboard/store/modules/customViews.js
```diff
@@ -86,6 +93,7 @@ export const mutations = {
   [types.SET_CUSTOM_VIEW]: MutationHelpers.set,
   [types.UPDATE_CUSTOM_VIEW]: MutationHelpers.update,
   [types.DELETE_CUSTOM_VIEW]: MutationHelpers.destroy,
+  [types.SET_CUSTOM_VIEW_COUNT]: MutationHelpers.setSingleRecord,
```
scmmishra, line 96: Should we use `updateAttributes` instead?

scmmishra (changes requested):

app/javascript/dashboard/store/modules/customViews.js
tejaswinichile: @scmmishra That fetches the entire record, I have checked in other parts of the code we are using the single record update.

cc @iamsivin

scmmishra: Noted

pranavrajs:
We will revisit the PR when this feature is prioritized again.

PR #7200, opened 2023-05-28
chore: Show chatwoot_edition in Instance details

memorycancel:
To check the edition anytime in admin dashboard, Showing as below:

image

### DIFF
--- a/app/controllers/super_admin/instance_statuses_controller.rb
+++ b/app/controllers/super_admin/instance_statuses_controller.rb
@@ -5,6 +5,17 @@ class SuperAdmin::InstanceStatusesController < SuperAdmin::ApplicationController
     sha
     postgres_status
     redis_metrics
+    chatwoot_edition
+  end
+
+  def chatwoot_edition
+    @metrics['Chatwoot edition'] = if ChatwootApp.enterprise?
+                                     'Enterprise'
+                                   elsif ChatwootApp.custom?
+                                     'Custom'
+                                   else
+                                     'Community'
+                                   end
   end
 
   def chatwoot_version


app/controllers/super_admin/instance_statuses_controller.rb
```diff
@@ -5,6 +5,17 @@ def show
...
+
+  def chatwoot_edition
+    is_enterprise = if ChatwootApp.enterprise?
+                      'Enterprise'
+                    else
+                      'Community'
+                    end
+    is_custom = ChatwootApp.custom? ? 'Custom' : ''
+    @metrics['Chatwoot edition'] = "#{is_enterprise} #{is_custom}"
   end
```
sojan-official, line 20:
```suggestion
  def chatwoot_edition
    @metrics['Chatwoot edition'] = case
    when ChatwootApp.enterprise?
      'Enterprise'
    when ChatwootApp.custom?
      'Custom'
    else
      'Community'
    end
  end

```

sojan-official (commented):
Minor refactoring suggestions 
since we are only considering 3 cases
- community
- enterprise 
- custom

app/controllers/super_admin/instance_statuses_controller.rb
memorycancel: Make sense, Enterprise custom and Community custom are both custom. It's more simple, thanks! suggestion applied, PTAL.

memorycancel:
> Minor refactoring suggestions since we are only considering 3 cases
> 
>     * community
> 
>     * enterprise
> 
>     * custom

Make sense, Enterprise custom and Community custom are both custom. It's more simple, thanks! suggestion applied, PTAL.

app/controllers/super_admin/instance_statuses_controller.rb
```diff
@@ -5,6 +5,17 @@ def show
     sha
     postgres_status
     redis_metrics
+    chatwoot_edition
+  end
+
+  def chatwoot_edition
+    @metrics['Chatwoot edition'] = if ChatwootApp.enterprise?
+                                     'Enterprise'
+                                   elsif ChatwootApp.custom?
```
pranavrajs, line 14: What is custom here? @vishnu-narayanan @sojan-official

PR #7205, opened 2023-05-29
fix: do not show agent name if the settings are not enabled in inbox

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-1629/feat-ability-to-customise-the-name-of-sender-in-emails-sent-from

Screenshot 2023-05-29 at 1 06 46 PM


app/models/inbox.rb
```diff
@@ -5,6 +5,7 @@
 # Table name: inboxes
 #
 #  id                            :integer          not null, primary key
+#  agent_name_enabled            :boolean          default(FALSE)
```
sojan-official, line 8: This change is specific to email channel alone right ? should this be an attribute of `email_channel` rather than on inbox ? 

cc: @pranavrajs

tejaswinichile: For now, yes, we could use it further for other channels as well if this becomes main stream requirement.

tejaswinichile: @sojan-official Let me just move it to Channel::Email to get these changes moving.

pranavrajs: @sojan-official Answered in the Linear ticket. I re-wrote the spec to make it clear.

PR #7140, opened 2023-05-20
feat: add footer into vueapp

memorycancel:
Closes #7139 

New feat: Add footer `partitial` into `vueapp` layout

To feel free to add any changes into footer  without resolving the conflict which hard change in vueapp layout

For example:

https://github.com/TommyTechWorkshop/chatwoot/commit/968642da40fe82636e576ac1e63112decb3d073d
https://github.com/TommyTechWorkshop/chatwoot/commit/4a5db26593e1c485449c11947d65a15cabaf501f

In addition, is there a plan to decoupling the these codes in to separate folder which is like `GitLab`, so the forkers can customize their business without resolving the conflicts when merge the latest code from upstream. 
Could you please have a look at this, thanks @sojan-official  @pranavrajs

----
Besides, can we implement a mechanism to override the same name file which like 

`custom/app/views/dashboard/_footer.html.erb` 
 override 
`app/views/dashboard/_footer.html.erb`

PR #7213, opened 2023-05-29
feat: Introduce view changes for loading Custom extension

### LINKED ISSUE #7212: Implement a mechanism to override the same name file in views
can we implement a mechanism to override the same name file which like 

`custom/app/views/dashboard/_footer.html.erb` 
 override 
`app/views/dashboard/_footer.html.erb`

@sojan-official @pranavrajs

_Originally posted by @memorycancel in https://github.com/chatwoot/chatwoot/issues/7140#issuecomment-1566700986_

memorycancel:
Introduce view changes for loading Custom extension

This PR is split from https://github.com/TommyTechWorkshop/chatwoot/commit/9c349a36b000ba32e549d1393220d14c445f234d

Related to #7140

Closes #7212

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -3,6 +3,7 @@
 require_relative 'boot'
 
 require 'rails/all'
+require './lib/chatwoot_app'
 
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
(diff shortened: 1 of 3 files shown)


sojan-official:
@memorycancel  Sorry, we cannot take this in at the moment. It feels like a custom use-case and doesn't add much value to the product. We can re-open the PR in the future if there is a strong community request.

Maintaining a fork is the way to go if it is still needed. 👍

Issue #7098, opened 2023-05-16
[CW-1785] API Inbox with POST Webhook and Parameters


hugodeco:
### Is your feature or enhancement related to a problem? Please describe.

After update for 2.17, there is a limitation of 250 caracters at webhook from api channels.
We need to pass some parameters, variables and fixed ones, to the external system, so our whatsapp api (quepasa) stops to work after that. 

### Describe the solution you'd like

We need a post webhook with parameters or no limitation for the webhook url.
I Think that a POST Webhook will be better and more integrated.

### Describe alternatives you've considered

For now we are changing some names of parameters trying to reach these number.

### Additional context
Would be nice if we can avoid SSL Checks (CheckBox) for use in localhost.
A Checkbox to IGNORE SSL issues.

_No response_

<sub>[CW-1785](https://linear.app/chatwoot/issue/CW-1785/api-inbox-with-post-webhook-and-parameters)</sub>

pedroherpeto:
getting same here!

bejaranoaqp:
Hola, si a roto la integracion con esta nueva actualizacion

pedroherpeto:
@vishnu-narayanan some update about that?

sojan-official:
We will have the limit for webhook URLs updated

sojan-official:
auth for webhook will be handled in https://github.com/chatwoot/chatwoot/issues/4732

PR #7230, opened 2023-05-31
chore: Increase character limit for external url fields

sojan-official:
- Increase the external URL field validation to 2048 characters

fixes: https://github.com/chatwoot/chatwoot/issues/7098


hugodeco:
would be nice if we can pass body parameters

sojan-official:
@hugodeco  will take it up as a separate issue.

PR #7231, opened 2023-05-31
fix: check content disposition, for inline messages in mail

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-1718/chore-inline-images-are-shown-as-attachments-in-email-channel

### DIFF
--- a/app/mailboxes/mailbox_helper.rb
+++ b/app/mailboxes/mailbox_helper.rb
@@ -24,13 +24,45 @@ module MailboxHelper
     return if @message.blank?
 
     processed_mail.attachments.last(Message::NUMBER_OF_PERMITTED_ATTACHMENTS).each do |mail_attachment|
-      attachment = @message.attachments.new(
-        account_id: @conversation.account_id,
-        file_type: 'file'
-      )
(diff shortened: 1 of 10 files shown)


app/mailboxes/mailbox_helper.rb
```diff
@@ -24,15 +24,47 @@ def add_attachments_to_message
...
+      "cid:#{content_id}", inline_image_url(mail_attachment[:blob]).to_s
+    )
+    @message.save!
+  end
+
+  def inline_attachment?(mail_attachment)
+    mail_attachment[:original].inline?
+  end
+
+  def embed_plain_text_email_with_inline_image(mail_attachment)
```
tejaswinichile, line 55: Very rare case: if the user has no HTML content, not signature, no font changes, just plain text, and an image then we need to embed the images for the text_part of the email.

pranavrajs:
@tejaswinichile Excited to see this getting a proper fix. Let me know when it is ready for testing.

tejaswinichile:
@pranavrajs This changes are ready to test and deployed to staging. Please have a look.

PR #7232, opened 2023-05-31
fix: reply mailer with sender name

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-1925/use-the-message-sender-name-instead-of-assigned-agent-name-in-the

Screenshot 2023-05-31 at 11 44 03 PM
Screenshot 2023-05-31 at 11 44 21 PM


spec/mailers/conversation_reply_mailer_spec.rb
```diff
@@ -171,7 +171,7 @@
 
       it 'renders assignee name in the from address' do
         mail = described_class.email_reply(message)
-        expect(mail['from'].value).to eq "#{conversation.assignee.available_name} from #{smtp_email_channel.inbox.name} <#{smtp_email_channel.email}>"
+        expect(mail['from'].value).to eq "#{message.sender.name} from #{smtp_email_channel.inbox.name} <#{smtp_email_channel.email}>"
```
pranavrajs, line 174: @tejaswinichile Can you please add a case for the message sender name being unavailable as well as assigned agent name also unavailable?

pranavrajs (dismissed):
Looks good. Please add some more specs to cover the cases mentioned in the comment.

spec/mailers/conversation_reply_mailer_spec.rb
tejaswinichile: Yes.

tejaswinichile: @pranavrajs 
1. We have a DB constraint that the name can't be blank, so saving an empty name for the sender is impossible - we can not test if the sender is present with a blank name.
2. We can have the empty sender for the message; [need to do something for this association] adding the specs to handle the empty sender.

app/mailers/conversation_reply_mailer.rb
```diff
@@ -80,7 +80,8 @@ def last_outgoing_message
   end
 
   def assignee_name
-    @assignee_name ||= @agent&.available_name || 'Notifications'
+    sender_name = @message&.sender&.name || @agent&.available_name
+    @assignee_name ||= sender_name || 'Notifications'
   end
```
sojan-official, line 85: Should we change the function to `sender_name` instead?

sojan-official (commented):
The code looks good to me. Minor comment about the function naming.

app/mailers/conversation_reply_mailer.rb
tejaswinichile: Sure, Adding this as a new commit. It will have more impact.

tejaswinichile: @sojan-official I have updated the method's name to sender_name

app/mailers/conversation_reply_mailer.rb
```diff
@@ -79,8 +79,8 @@ def last_outgoing_message
     @conversation.messages.chat.where.not(message_type: :incoming)&.last
   end
 
-  def assignee_name
-    @assignee_name ||= @agent&.available_name || 'Notifications'
+  def sender_name
+    @sender_name ||= @message&.sender&.name || @agent&.available_name || 'Notifications'
```
vishnu-narayanan, line 83:
```   suggestion
@sender_name ||= @message&.sender&.name || @agent&.available_name || I18n.t('conversations.reply.email.sender.default')
```

@tejaswinichile  Should `Notifications` be read from i18n to better support localization?

vishnu-narayanan (approved):
minor comment

app/mailers/conversation_reply_mailer.rb
tejaswinichile: @vishnu-narayanan `Notifications` case never happens until someone changes the DB validation on not to save the name. I have kept it because it was always there from the start.

And we have the next PR to handle the whole sender name customization.

pranavrajs:
@tejaswinichile Can you fix the specs?

tejaswinichile:
@pranavrajs Specs, has been fixes, Ruby Ci is failing for other slow files.

PR #7243, opened 2023-06-02
fix: Wrap references string into array

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-1996/nomethoderror-undefined-method-each-for

### DIFF
--- a/app/mailboxes/imap/imap_mailbox.rb
+++ b/app/mailboxes/imap/imap_mailbox.rb
@@ -54,7 +54,7 @@ class Imap::ImapMailbox
 
     return if @inbound_mail.references.blank?
 
-    references = @inbound_mail.references
+    references = Array.wrap(@inbound_mail.references)
 
     references.each do |message_id|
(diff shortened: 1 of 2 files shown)


scmmishra (approved):

app/mailboxes/imap/imap_mailbox.rb
```diff
@@ -54,7 +54,7 @@ def find_message_by_references
 
     return if @inbound_mail.references.blank?
 
-    references = @inbound_mail.references
+    references = Array.wrap(@inbound_mail.references)
```
scmmishra, line 57: Approving this assuming that `@inbound_mail.references` will always be a string

scmmishra: @tejaswinichile is it possible for references to be of any other form, say a string of comma separated IDs?

vishnu-narayanan (approved):

PR #7257, opened 2023-06-06
feat: Ability to choose time for snooze conversation

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-1733/feat-ability-to-snooze-a-conversation-until-a-specific-time

This PR supports choosing a custom time for the snooze option.

**Preview**

CleanShot 2023-06-19 at 17 14 24


app/javascript/dashboard/helper/snoozeHelpers.js
```diff
@@ -0,0 +1,13 @@
+import { getUnixTime, format } from 'date-fns';
+import * as chrono from 'chrono-node';
+
+export const findSnoozeTime = snoozeType => {
```
scmmishra, line 4: I noticed we don't let users type arbritrary strings and limit them to the options from `SNOOZE_CONVERSATION_ACTIONS`, in that case we may not need `chrono` library, instead we can just define the method ourselves for each of the available options

scmmishra (commented):
Some thoughts on the modal UI

CleanShot 2023-06-12 at 18 38 56@2x

app/javascript/dashboard/helper/snoozeHelpers.js
```diff
@@ -1,11 +1,36 @@
...
+    return new Date(
+      currentDate.getFullYear(),
+      currentDate.getMonth(),
+      currentDate.getDate() + time,
+      9,
+      0,
+      0,
+      0
+    );
+  };
```
scmmishra, line 20: We can use `add` method from `date-fns`  for this instead of this implementation

app/javascript/dashboard/helper/snoozeHelpers.js
```diff
@@ -1,11 +1,36 @@
...
+
+  if (snoozeType === SNOOZE_OPTIONS.AN_HOUR_FROM_NOW) {
+    parsedDate = add(currentDate, { hours: 1 });
+  } else if (snoozeType === SNOOZE_OPTIONS.UNTIL_TOMORROW) {
+    parsedDate = addTime(1);
+  } else if (snoozeType === SNOOZE_OPTIONS.UNTIL_NEXT_WEEK) {
+    parsedDate = addTime(7);
+  } else if (snoozeType === SNOOZE_OPTIONS.UNTIL_NEXT_MONTH) {
+    parsedDate = addTime(30);
```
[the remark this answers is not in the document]
scmmishra: Another thought I had, what should "Next month" mean, should it be exactly one month from now, or the start of the next month?

GMail has options like these

CleanShot 2023-06-14 at 15 10 04@2x

Here next week starts on Monday, also they have a Later this week option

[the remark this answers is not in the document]
muhsin-k: Yes. I agree. We can do something like,

Next week -> Next week's first working day, Monday at 9.00 AM.
Next month -> Next month, first working day, first Monday at 9.00 AM.

I have also considered adding expected time as an option, but it may require customization of the command bar component.

app/javascript/dashboard/helper/snoozeHelpers.js
muhsin-k: @scmmishra 
- Regarding the alignment of the title and close button, I am using the `woot-modal-header` and some modifications are necessary. It would be best to treat this as a separate task since it is utilized on multiple screens.
- Calendar style issues. Do you mean something like this?

CleanShot 2023-06-14 at 15 26 40

muhsin-k: @scmmishra @pranavrajs @nithindavid Any feedback on the custom snooze modal?

CleanShot 2023-06-14 at 22 08 40

app/javascript/dashboard/components/buttons/ResolveAction.vue
```diff
@@ -59,72 +59,45 @@
...
   </div>
 </template>
 
 <script>
+import { getUnixTime } from 'date-fns';
 import { mapGetters } from 'vuex';
 import { mixin as clickaway } from 'vue-clickaway';
 import alertMixin from 'shared/mixins/alertMixin';
-import snoozeTimesMixin from 'dashboard/mixins/conversation/snoozeTimesMixin.js';
+import CustomSnoozeModal from 'dashboard/components/CustomSnoozeModal.vue';
```
iamsivin, line 91:
```suggestion
import CustomSnoozeModal from 'dashboard/components/CustomSnoozeModal';
```

app/javascript/dashboard/components/buttons/ResolveAction.vue
```diff
@@ -270,5 +261,9 @@ export default {
   right: 0;
   max-width: 20rem;
   min-width: 15.6rem;
+
+  .dropdown-menu__item {
+    margin-bottom: var(--space-zero);
```
iamsivin, line 266:
```suggestion
    margin-bottom: 0;
```

nithindavid:
@muhsin-k Can you add these custom styles to the date and time picker? I have added the styles for date selector only, would be great if we could do the same styles for time picker as well

1. Disabled cells 

```

color: var(--s-200);
background-color: var(--s-25);
```    

Screenshot 2023-06-15 at 4 49 32 PM

2. Today 

```
font-weight: var(--font-weight-bold);
```

app/javascript/dashboard/components/CustomSnoozeModal.vue
```diff
@@ -0,0 +1,75 @@
...
+        days: ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'],
+        yearFormat: 'YYYY',
+        monthFormat: 'MMMM',
+      },
+    };
+  },
+
+  methods: {
+    onClose() {
+      this.$emit('on-close');
```
nithindavid, line 47:
```suggestion
      this.$emit('close');
```

So we could follow `@close` style like we use in other components

nithindavid (commented):
Looks good overall, can you address the comments I have added?

pranavrajs (changes requested):
Nice work. Shared some feedback [here](https://linear.app/chatwoot/issue/CW-1733#comment-9d466282). Please take a look.

iamsivin (approved):
LGTM

PR #7260, opened 2023-06-06
fix: Cannot read properties of undefined (always_play_audio_alert)

pranavrajs:
Fixes https://linear.app/chatwoot/issue/CW-2029/typeerror-cannot-read-properties-of-undefined-reading-always-play

### DIFF
--- a/app/views/api/v1/accounts/create.json.jbuilder
+++ b/app/views/api/v1/accounts/create.json.jbuilder
@@ -6,6 +6,7 @@ json.data do
   json.display_name resource.display_name
   json.email resource.email
   json.account_id @account.id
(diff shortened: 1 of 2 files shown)


app/views/api/v1/accounts/create.json.jbuilder
```diff
@@ -6,6 +6,7 @@ json.data do
   json.display_name resource.display_name
   json.email resource.email
   json.account_id @account.id
+  json.created_at resource.created_at
```
pranavrajs, line 9: This could be the reason for https://linear.app/chatwoot/issue/CW-1315/the-contacts-are-not-synchronized-to-chatwoot-contacts

PR #7273, opened 2023-06-08
fix: Update mail check for html_part

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-2034/nomethoderror-undefined-method-gsub-for-nilnilclass-nomethoderror


sojan-official (commented):
Why didn't our specs catch this case? should we add specs to cover this ?

tejaswinichile:
> Why didn't our specs catch this case? 

This is a very weird mail that we received says the HTML part is there, but there is no content in it message_id: 924, channel: 6

> specs to cover this?

It seems like a very edge case; let me add that.

tejaswinichile:
@sojan-official I have tested this on production, but I was not able to create such an email. This seems to be from Zoho,  can you provide me that email to add it in the spec?

sojan-official (approved):

PR #7286, opened 2023-06-09
fix: search for nil in-reply-to messages

tejaswinichile:
Conversation#4915 

The message was appended to the wrong conversation due to recent conversation references changes for the email channel.

https://github.com/chatwoot/chatwoot/pull/7220/


sojan-official (approved):
LGTM.

PR #7293, opened 2023-06-12
fix: calculation for resolution count

scmmishra:
The current method that gets the resolved conversation count is the following, along with it's SQL query

```rb
def resolutions_count
  (get_grouped_values scope.conversations.where(account_id: account.id).resolved).count
end
```

<details>
<summary>Resultant Query</summary>

```sql
SELECT 
  COUNT(*) AS "count_all", 
  DATE_TRUNC(
    'day', "conversations"."created_at" :: timestamptz AT TIME ZONE 'Asia/Kolkata'
  ):: date AS "date_trunc_day_conversations_created_at_timestamptz_at_time_zon" 
FROM 
  "conversations" 
WHERE 
  "conversations"."account_id" = $1 
  AND "conversations"."account_id" = $2 
  AND "conversations"."status" = $3 
  AND (
    "conversations"."created_at" >= '2023-06-05 18:30:00' 
    AND "conversations"."created_at" < '2023-06-12 18:29:59'
  ) 
GROUP BY 
  DATE_TRUNC(
    'day', "conversations"."created_at" :: timestamptz AT TIME ZONE 'Asia/Kolkata'
  ):: date;
```

</details>

The problem is that it ranges with conversation created, so only the conversation created during that time period and resolved in that time period will show up

The solution in this PR is to rely on reporting events. We get a list of **distinct** `conversation_id` for all events named `conversation_resolved` (from the `reporting_events` table), scoped by current account and filtered to ensure that the conversation is currently actually resolved: 

```rb
def resolutions_count
  object_scope = scope.reporting_events
                      .joins(:conversation).select(:conversation_id)
                      .where(account_id: account.id, name: :conversation_resolved, conversations: { status: :resolved })
                      .distinct
  (get_grouped_values object_scope).count
end
```

<details>
<summary>Resultant Query</summary>

```sql
SELECT
  COUNT(DISTINCT "reporting_events"."conversation_id") AS "count_conversation_id",
  DATE_TRUNC(
    'day',
    "reporting_events"."created_at" :: timestamptz AT TIME ZONE 'Asia/Kolkata'
  ) :: date AS "date_trunc_day_reporting_events_created_at_timestamptz_at_time_"
FROM
  "reporting_events"
  INNER JOIN "conversations" ON "conversations"."id" = "reporting_events"."conversation_id"
WHERE
  "reporting_events"."account_id" = $ 1
  AND "reporting_events"."account_id" = $ 2
  AND "reporting_events"."name" = $ 3
  AND "conversations"."status" = $ 4
  AND (
    "reporting_events"."created_at" >= '2023-06-05 18:30:00'
    AND "reporting_events"."created_at" < '2023-06-12 18:29:59'
  )
GROUP BY
  DATE_TRUNC(
    'day',
    "reporting_events"."created_at" :: timestamptz AT TIME ZONE 'Asia/Kolkata'
  ) :: date;
```

</details>

## Type of change

## Checklist:

### DIFF
--- a/app/helpers/report_helper.rb
+++ b/app/helpers/report_helper.rb
@@ -29,7 +29,8 @@ module ReportHelper
   end
 
   def resolutions_count
-    (get_grouped_values scope.conversations.where(account_id: account.id).resolved).count
+    object_scope = scope.reporting_events.select(:conversation_id).where(account_id: account.id, name: :conversation_resolved).distinct
+    (get_grouped_values object_scope).count
   end
 
   def avg_first_response_time
--- a/app/helpers/report_helper.rb
+++ b/app/helpers/report_helper.rb
@@ -29,7 +29,8 @@ module ReportHelper
   end
 
   def resolutions_count
-    object_scope = scope.reporting_events.select(:conversation_id).where(account_id: account.id, name: :conversation_resolved).distinct
+    object_scope = scope.reporting_events.joins(:conversation).select(:conversation_id).where(account_id: account.id, name: :conversation_resolved,
+                                                                                              conversations: { status: :resolved }).distinct
     (get_grouped_values object_scope).count
   end
 
--- a/spec/builders/v2/report_builder_spec.rb
+++ b/spec/builders/v2/report_builder_spec.rb
@@ -107,13 +107,18 @@ describe V2::ReportBuilder do
 
         conversations = account.conversations.where('created_at < ?', 1.day.ago)
         perform_enqueued_jobs do
+          # Resolve all 5 conversations
           conversations.each(&:resolved!)
+
+          # Reopen 1 conversation
+          conversation.first.open!
         end
 
         builder = described_class.new(account, params)
         metrics = builder.timeseries
 
-        expect(metrics[Time.zone.today]).to be 5
+        # 4 conversations are resolved
+        expect(metrics[Time.zone.today]).to be 4
         expect(metrics[Time.zone.today - 2.days]).to be 0
       end
 
@@ -221,13 +226,18 @@ describe V2::ReportBuilder do
         conversations = account.conversations.where('created_at < ?', 1.day.ago)
 
         perform_enqueued_jobs do
+          # ensure 5 reporting events are created
           conversations.each(&:resolved!)
+
+          # open one of the conversations to check if it is not counted
+          conversations.last.open!
         end
 
         builder = described_class.new(account, params)
         metrics = builder.timeseries
 
-        expect(metrics[Time.zone.today]).to be 5
+        # this should count only 4 since the last conversation was reopened
+        expect(metrics[Time.zone.today]).to be 4
         expect(metrics[Time.zone.today - 2.days]).to be 0
       end
 
--- a/spec/builders/v2/report_builder_spec.rb
+++ b/spec/builders/v2/report_builder_spec.rb
@@ -111,7 +111,7 @@ describe V2::ReportBuilder do
           conversations.each(&:resolved!)
 
           # Reopen 1 conversation
-          conversation.first.open!
+          conversations.first.open!
         end
 
         builder = described_class.new(account, params)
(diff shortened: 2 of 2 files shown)


sojan-official (commented):
Can we also ensure that our tests cover this case?

scmmishra:
> Can we also ensure that our tests cover this case?

Yeah adding that. I had a doubt. 

Suppose we count a conversation as resolved since a value for it exists. However if the conversation status at present is still `Open`, should be include it or exclude it?

sojan-official:
> > Can we also ensure that our tests cover this case?
> 
> Yeah adding that. I had a doubt,
> 
> Suppose we count a conversation as resolved since a value for it exists. However, if the conversation status at present is still `Open`, should be include it or exclude it? This would mean the resolution count for a day in the past can actually reduce.
> 
> This can be done with the following condition
> 
> ```ruby
> scope.reporting_events.joins(:conversation).select(:conversation_id).where(account_id: account.id, name: :conversation_resolved, conversations: { status: :resolved }).distinct
> ```
> 
> CC: @pranavrajs @tejaswinichile @sojan-official
> 
> The problem here is with the approach, there are two choices
> 
> 1. **We count every instance a conversation was marked resolved** and show it in the counts. This would mean showing conversations that are currently open, as well as counting conversations twice if they were resolved multiple times in that time period
> 2. We **count only the conversations that are currently resolved** while counting the events in the given timeframe. This means, if Chatwoot clients are allowed to re-open conversations, this can reduce the count historically. That makes history mutable
> 
> The first approach is simpler, and makes the history permanent. But I think the **second approach is much more meaningful** as a metric even though it keeps the history mutable
> 
> > P.S. I have implemented the second approach

Approach 2 makes sense to me.  but will let @pranavrajs  take the final call

sojan-official (approved):

pranavrajs:
@scmmishra @sojan-official 

To determine whether we consider resolutions as unique events, we must first define the meanings of "resolution" and "re-open event."

If a conversation is re-opened and continues on the same thread, I believe that we should modify the history accordingly. However, if the conversation is re-opened and a different thread of conversation is initiated, it is not ideal to update the resolution event to a more recent date.

Let's consider the example of a website channel. The admin has the option to force a new conversation and disallow messages after resolution. This means that the user can start a new thread of conversation each time. Even if we add a buffer time (such as "conversations can be reopened within X minutes/hours"), having a single resolution event makes sense.

However, if we consider the WhatsApp channel, there are no multiple threads of conversation. The user might be discussing entirely different issues in the same conversation. In this case, having a single resolution event doesn't make sense.

I propose the following path forward:

- Let's stick with having a single event that will be updated if there is a new resolution event in the conversation. This seems to be the easiest way to handle things.
- We should count the number of re-open events (which we don’t right now) and show the list of conversations that have been re-opened.
- Admins should be able to configure multiple threads of conversations in WhatsApp, Facebook, and Instagram. For the first implementation, we can use a similar approach to what we have now with the website channel. If there is a new message and there are no open conversations, we will create a new conversation. The second iteration can be the option to add X minutes/hours as a buffer time to re-open the previous conversation. If a conversation is resolved and a new message from the customer comes within X minutes, we will re-open the conversation.

Let me know what you think.

scmmishra:
> * Let's stick with having a single event that will be updated if there is a new resolution event in the conversation. This seems to be the easiest way to handle things.

"single event that will be updated if there is a new resolution event in the conversation" isn't this the same as the current `conversation_resolved` event?

Correct me if I am wrong here, if you mean that we only consider the latest `conversation_resolved` event, this PR does the same. 

> * We should count the number of re-open events (which we don’t right now) and show the list of conversations that have been re-opened.
> * Admins should be able to configure multiple threads of conversations in WhatsApp, Facebook, and Instagram. For the first implementation, we can use a similar approach to what we have now with the website channel. If there is a new message and there are no open conversations, we will create a new conversation. The second iteration can be the option to add X minutes/hours as a buffer time to re-open the previous conversation. If a conversation is resolved and a new message from the customer comes within X minutes, we will re-open the conversation.


We can pick these up as a separate enhancement for sure

PR #7341, opened 2023-06-18
fix: Label Duplication in Bulk Actions

civilcoder55:
This PR fixes the issue of labels bulk assign

### To reproduce the issue
1- you must have two conversations
2- assign for example 4 labels on one of them (label-1, label-2, label-3, label4)
3- select the other conversation and assign (label-1) to it

### Expected Result
1- conversation one still be assigned to (label-1, label-2, label-3, label4)
2- conversation two will be assigned to (label-1)
### Actual Result
both conversations have the 4 labels 

### why this happens
when looping the conversations in bulk action to assign them to new labels, accessing labels somehow references the 
wrong model 

```rb
def add_labels(new_labels = nil)
    new_labels << labels
    update!(label_list: new_labels)
end
```

@Peteraymansalib discovered this issue and provided detailed steps to reproduce the issue. 

## Type of change

## How Has This Been Tested?
https://www.loom.com/share/9c5a173d1c9c42c289e14587d2923649?sid=42a90e9f-f9d3-4380-b6ec-628a8c4eb641


sojan-official:
@civilcoder55  Thanks for the fix and loom video. We will review this internally and get this merged. Would also be helpful if we could create an respec test case replicating this behaviour so that we can ensure the current fix fixes it.

PR #7345, opened 2023-06-19
feat: Ability to customise the email sender name [CW-1629] 

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-1629/feat-ability-to-customise-the-name-of-sender-in-emails-sent-from

Screenshot 2023-06-26 at 5 33 58 PM

https://www.loom.com/share/0e417b9b01624045bace2cb733ffabac?sid=cfb42d15-2895-4987-aba2-f10655a83ca1


sojan-official (commented):
Backend Code LGTM 
Let's ensure that we QA all the cases before merging.

app/javascript/dashboard/routes/dashboard/settings/inbox/components/SenderNameExamplePreview.vue
```diff
@@ -0,0 +1,144 @@
...
+            </div>
+          </div>
+        </div>
+      </preview-card>
+    </button>
+  </div>
+</template>
+
+<script>
+import PreviewCard from 'dashboard/components/ui/PreviewCard.vue';
```
iamsivin, line 38:
```suggestion
import PreviewCard from 'dashboard/components/ui/PreviewCard';
```

muhsin-k (approved):
Front end code LGTM.

db/schema.rb
```diff
@@ -584,6 +584,8 @@
     t.jsonb "auto_assignment_config", default: {}
     t.boolean "lock_to_single_conversation", default: false, null: false
     t.bigint "portal_id"
+    t.string "business_name"
+    t.boolean "custom_sender_name_enabled", default: true, null: false
```
pranavrajs, line 588: @sojan-official @tejaswinichile I would suggest to keep this an enum rather than a boolean. Something like sender_name_type as friendly:0 or professional:1

pranavrajs (changes requested):
Minor comment on the implementation, otherwise looks fine.

db/schema.rb
tejaswinichile: Okay. 👍🏼

scmmishra (commented):
CleanShot 2023-07-03 at 12 57 40@2x

this UI has two update buttons, both do the same thing, and both are primary. @nithindavid need your thoughts here. @tejaswinichile you can ignore this and merge it if you want, but let's solve the issue in an immidiate PR

CleanShot 2023-07-03 at 12 58 35@2x

The text overflow can be handled better for long names.

scmmishra:
P.S. I tested for RTL languages, it looks good

scmmishra (commented):
Another UX issue is that once I switch the "Sender name" option, it looks like it gets saved or selected. There is no indication that the changes are unsaved. It's a minor issue and can be dealt with in another PR. @nithindavid let me know if you have ideas to solve this.

https://github.com/chatwoot/chatwoot/assets/18097732/d7b6e217-037d-442f-b5fa-d1d652b1c499

I would suggest every time the user clicks one of those, we update it directly, since it's a reversible action, it does not need a second step.

iamsivin:
@scmmishra Nice suggestions. I have resolved the text-overflow issue. @pranavrajs @nithindavid, Can you take a look at the comments by Shivam?

tejaswinichile:
Screenshot 2023-07-04 at 4 55 35 PM


I tested the functionality again one final time before merging.
cc @scmmishra @sojan-official

scmmishra (approved):

scmmishra:
Left minor comments, you can do them in the UX PR, rest LGTM! @tejaswinichile

PR #7364, opened 2023-06-21
feat: Add a sort option for conversations waiting for a reply from an agent

pranavrajs:
Screenshot 2023-06-20 at 7 07 21 PM

This PR adds an additional sort option called `Longest waiting`.

**The problem:** If there are a large number of conversations in the queue, it is easier to prioritize (unless you use snooze). Sort by Created doesn't solve the issue of finding the conversations pending a reply. Unattended wouldn't give the option to see the conversations that had a first reply but are missing a followup reply.

**Alternatives considered:**
 
Using user_messaged_at: It doesn't solve the problem. I needed a list of conversations that are missing a followup reply and sort them based on the waiting time. 

**What is the PR?**

- Adds new field for the conversations model called `waiting_since.`
  - It is set to created_at time during the conversation creation phase.
  - It would be set to nil if an agent replies to the conversation (messages from bot, automation, campaigns etc would be ignored)
  - It would be set to the current time if a new message from the user is created. (It would not override the previous timestamp if there is already one since we are looking for the waiting time)
- I have not added migration for the existing data (Since this is not a constant value, I believe it is okay to ignore the migration, as and when there are new messages this would value would be adjusted). Also, there are 2M open conversations in our production system. If we need to do it, we have to run it via jobs.
- Add the sort option in the frontend.


app/models/message.rb
```diff
@@ -201,7 +201,17 @@ def email_notifiable_message?
...
+
+    return false if conversation.first_reply_created_at.present?
+
+    return false unless human_response?
+    return false if conversation.messages.outgoing
+                                .where.not(sender_type: 'AgentBot')
+                                .where.not(private: true)
+                                .where("(additional_attributes->'campaign_id') is null").count > 1
+
+    true
```
scmmishra, line 214:
```rb
def valid_first_reply?
  return false unless outgoing? && human_response? && !private?
  return false if conversation.first_reply_created_at.present?
  return false if conversation.messages.outgoing
                              .where.not(sender_type: 'AgentBot')
                              .where.not(private: true)
                              .where("(additional_attributes->'campaign_id') is null").count > 1

  true
end
```

This implementation can reduce the cognitive complexity error from codeclimate

scmmishra:
CleanShot 2023-06-21 at 14 14 31@2x

This error can be safely ignored, most likely is in-correct. Tried running the ESLint complexity plugin for the function, it seems to be well under 6

app/javascript/dashboard/i18n/locale/en/chatlist.json
```diff
@@ -50,6 +50,9 @@
       },
       "sort_on_priority": {
         "TEXT": "Priority"
+      },
+      "sort_on_waiting_since": {
+        "TEXT": "Longest waiting"
```
scmmishra, line 55:
```suggestion
        "TEXT": "Oldest Unanswered"
```

A suggestion (feel free to ignore it) We can perhaps use a better copy. Longest waiting may not be very clear.

app/javascript/dashboard/store/modules/conversations/getters.js
```diff
@@ -22,6 +22,21 @@ const getters = {
           CONVERSATION_PRIORITY_ORDER[b.priority]
         );
       },
+      sort_on_waiting_since: (a, b) => {
+        if (!a.waiting_since && !b.waiting_since) {
```
scmmishra, line 26:
```suggestion
        if (!a.waiting_since && !b.waiting_since) {
```

Is there a possibility `waiting_since` can be `0` and can affect this sort method? If not, can we check for null directly?

scmmishra (approved):
Approving it, minor comments which are mostly cosmetic

muhsin-k:
I have a suggestion that is not within the scope of this pull request. Can we add a hint about which sort order option is selected, similar to the conversation status, so the user can understand which sort option is currently being applied?

CleanShot 2023-06-21 at 14 42 40

@pranavrajs @scmmishra @nithindavid

app/javascript/dashboard/store/modules/conversations/getters.js
muhsin-k: I believe the test case covers the value of 0.

app/javascript/dashboard/i18n/locale/en/chatlist.json
muhsin-k: Same suggestion. What about the text `Outstanding Replies`?

muhsin-k (approved):
The PR looks good, except few minor comments.

app/javascript/dashboard/i18n/locale/en/chatlist.json
sojan-official: "pending response" ? :D

sojan-official (commented):
LGTM overall, barring the build issues.

spec/models/message_spec.rb
```diff
@@ -138,6 +138,37 @@
...
+
+    it 'sets the waiting_since if there is an incoming message' do
+      conversation.update(waiting_since: nil)
+      message.message_type = :incoming
+      message.save!
+
+      expect(conversation.waiting_since).not_to be_nil
+    end
+
+    it 'does not overwrite the previous value if there are newer messages' do
```
tejaswinichile, line 162: 💯

app/javascript/dashboard/i18n/locale/en/chatlist.json
pranavrajs: Let us go with "pending response" for now.

app/javascript/dashboard/store/modules/conversations/getters.js
pranavrajs: @scmmishra It is 0 if the value is nil (since we do waiting_since.to_i everywhere). I have covered this.

pranavrajs:
> I have a suggestion that is not within the scope of this pull request. Can we add a hint about which sort order option is selected, similar to the conversation status, so the user can understand which sort option is currently being applied?

Make sense, let us take it up as the enhancement.

pranavrajs:
> This error can be safely ignored, most likely is in-correct. Tried running the ESLint complexity plugin for the function, it seems to be well under 6

@scmmishra Time to update the tools/rules?

app/models/message.rb
pranavrajs: Not a fan of unless and ! in the same line. However, to keep the linter happy, I'll update this.

PR #7370, opened 2023-06-21
chore: Send FRONTEND_URL to AgentBots

jordan-brough:
Would it be reasonable to send the frontend url to bots? It seems like that would remove the requirement that bots be pre-aware of the URL they need to make callbacks to.

I'm opening this PR just for the discussion at the moment. I'm not sure if this is the correct place/way to put this.  And I haven't added tests. And if this seemed OK we might want to update WebhookListener the same way?


sojan-official:
@jordan-brough  Thanks for bringing this up. I think adding instance metadata to these webhooks will be a good idea, allowing developers to build integrations that can work across installations. Let me discuss this with the team and take this forward.

jordan-brough:
Thanks!

sojan-official:
@jordan-brough Since we haven't been able to prioritise this PR for a while, I am closing this for the moment. Until then, please go ahead with the alternative that you mentioned.

```
Alternatively I suppose I could just add a callback_url URL parameter to the AgentBot's outgoing_url
```

We will reopen this again once the team has more bandwidth to work on it.

PR #7379, opened 2023-06-23
feat: Allow Twilio::REST::TwilioError errors to raise and prevent Inbox creation

jordan-brough:
This update will mean that errors will roll back the current transaction and the error will be sent back to the frontend and the user will know that the Inbox did not finish setting up successfully. We already do this if the credentials are invalid (when `authenticate_twilio` fails in `TwilioChannelsController`).

The current code seems a bit unexpected to me.  When `Twilio::WebhookSetupService.new(...).perform` fails it doesn't return or raise anything that indicates it failed, besides having sent a message to the logs.

Is there a common use case here where failure is expected?  Is there a way to tell when failure is expected and only squelch the error in that case?

### DIFF
--- a/app/services/twilio/webhook_setup_service.rb
+++ b/app/services/twilio/webhook_setup_service.rb
@@ -9,8 +9,6 @@ class Twilio::WebhookSetupService
     else
       update_phone_number
     end
-  rescue Twilio::REST::TwilioError => e
-    Rails.logger.error "TWILIO_FAILURE: #{e.message}"
   end
 
   private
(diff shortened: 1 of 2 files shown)


sojan-official:
@jordan-brough, thanks for the catch, This seems to be present since the initial implementation of this feature,
ref: https://github.com/chatwoot/chatwoot/commit/a1a81e37996619798bb44f23ac2f2c0ed67d5f66 

I'm not currently aware of any specific errors that this code is handling. 
I will consult with @pranavrajs to ensure we haven't overlooked any scenarios. If there aren't any, I will proceed with merging the proposed changes. Thanks for your contribution.

PR #7397, opened 2023-06-26
Include email in the searched result

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-1651/question-how-is-this-search-result-returned


scmmishra (approved):
The code is good. @nithindavid can you take a look at the design and the copy

nithindavid:
Changed the component to show both name and email on a single line.
image

PR #7442, opened 2023-06-30
perf: Resolved the green badge issue

Aditay2002:
CHANGES DONE:-

1)In the template section:

a)Inside the <div> with the class font-medium text-base leading-4 flex items-center, the following changes were made:
b)Removed the conditional class binding for the Green badge's background color (bg-green-500).
c)Instead, added a v-if="isOnline" directive to conditionally render the Green badge.
d)Inside the conditional rendering, added the class h-2 w-2 rounded-full bg-green-500 to apply the background color when the user is online.
2)In the computed property isOnline:

a)Added const anyAgentOnline = this.availableAgents.length > 0; to check if there are any available agents.
b)Updated the return statement inside the if (workingHoursEnabled) block to return this.isInBetweenTheWorkingHours && anyAgentOnline;.
c)This change ensures that the Green badge is displayed only when the user is within the working hours and there is at least one available agent.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,6 @@ jobs:
     <<: *defaults
     steps:
       - checkout
-
       - run:
           name: Configure Bundler
           command: |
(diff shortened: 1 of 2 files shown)


pranavrajs:
@Aditay2002 Thanks for the PR. I'm not sure if I understood the changes completely. Here are the conditions that we have right now.

**_1/ The inbox does not have a business hour configured._**

If there are online agent, we would show the online status.

**_2/ The inbox has a business hour configured._**

In this case, irrespective of whether the agent is online or not, we expect the business to be shown as online. I'm not sure why there is a need for agent availability check.

Let me know.

sojan-official:
Closing this since there hasn't been any activity on this for a while. Let's reopen if required.

PR #7446, opened 2023-07-02
fix: contacts count on filter

civilcoder55:
This PR fixes the issue of contacts count on filtering

### To reproduce the issue
1- create a contact with a **unique** name and assign **two** labels for him
2- filter the contacts with this unique name

### Expected Result
- got a page with 1 contact = the contact you created before

### Actual Result
- I got the contact but the total count is wrong (total count == the number of labels assigned to the contact)
https://www.loom.com/share/601d7cd0ab284417ba87f71f95eff6ae?sid=bb115b99-ce7a-41f5-ba92-e2a68ad2895e

### why this happens
because of joining with labels resulting in the same contact repeated once every label (one to many)

### possible solution
- use distinct on contact id column

### changes 
- spec/services/contacts/filter_service_spec.rb
1- rename user_1 to first_user to solve rubocop warning  image
2- change additional_attributes case to filter on additional_attributes only 
3- create a case to filter on name
4- fix the expected count in each case


sojan-official:
@civilcoder55  Thanks a lot for the catch. We will have this tested and merged.

Issue #5598, opened 2022-10-11
feat: Internal Inbox for the team to communicate


sojan-official:
> Yes, the goal would be to introduce an internal chat communication like Slack or Teams so that Chatwoot becomes the central hub of all text-based internal and external communication. Private notes and @mentions work fine for any communication that is about a ticket or conversation however it does not work well when you want to ask a simple question unrelated to any current conversations.

SirRodney:
+1

AndersonOuverney:
+1

clairton:
👍

NLSARCOS:
+1

pranavrajs:
Closing this here as we will not build / accept PRs on this issue near future.

PR #7448, opened 2023-07-02
WIP: create channel internal

clairton:
refs #5598


zhiru:
needs to change message_builder.rb to 

class Messages::Internal::MessageBuilder < Messages::Messenger::MessageBuilder
  def perform
    super

    @message
  end
end


for some reason when testing the channel_internal is nto being created

pranavrajs:
@clairton Sorry, we cannot take this in. At the moment, the focus for Chatwoot is to manage external communication (communication with customers). Internal communication is already solved by Slack or Mattermost or Rocket.Chat. We don't intend to spread too thin before we build a good product in support and communication use-cases. 

I hope you understand.

PR #7475, opened 2023-07-06
feat: Sentiment Analysis

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-2134/introduce-a-gem-informers-to-check-the-sentiment-of-the-text


app/models/concerns/sentiment_analysis_helper.rb
scmmishra, line 1: Probably should move this to enterprise as well

app/models/concerns/sentiment_analysis_helper.rb
```diff
@@ -0,0 +1,40 @@
+module SentimentAnalysisHelper
+  extend ActiveSupport::Concern
+
+  def opening_sentiments
+    records = incoming_messages.first(average_message_count)
+    average_sentiment(records)
+  end
+
+  def closing_sentiments
```
scmmishra, line 9: We should be showing closing sentiments only if the conversation is resolved. Should we check that too

app/models/concerns/sentiment_analysis_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  private
+
+  def average_sentiment_label(records)
+    value = records.pluck(:sentiment).sum { |a| a['value'].to_i }
+    value.negative? ? 'negative' : 'positive'
+  end
+
+  def average_sentiment_score(records)
```
scmmishra, line 28: This looks good for now, lets see how this works out with real conversations. If it turns out this needs tweaking, maybe we can have a weighted average, say we have 5 conversations, for the opening sentiment the first message will have more weight, and subsequent messages will have lesser and lesser weight

We can use an exponential decay algorithm

```rb
decay_factor = 0.9  # This value should be between 0 and 1
weights = Array.new(n) { |i| (decay_factor ** (n - i - 1)) }
```

For now we can let this implementation stay

app/models/concerns/sentiment_analysis_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+    total = records.pluck(:sentiment).sum { |a| a['score'].to_f }
+    total / average_message_count
+  end
+
+  def average_message_count
+    incoming_messages.count >= 10 ? 5 : ((incoming_messages.count / 2) - 1)
+  end
```
scmmishra, line 35: We can just have a fixed number, say 5 messages, it's fine if the messages overlap. Right now, we need to test this with real-world conversations and then tweak the scores

 @sojan-official @pranavrajs WDYT?

scmmishra (changes requested):

app/models/concerns/sentiment_analysis_helper.rb
pranavrajs: Sounds good.

app/models/concerns/sentiment_analysis_helper.rb
tejaswinichile: @scmmishra Yes, I have not added the calls from the controller yet or in the `push_event_data` for the conversation. I was thinking, I will add the condition there itself.

app/models/concerns/sentiment_analysis_helper.rb
tejaswinichile: > We can use an exponential decay algorithm

Yes, this sounds good. This is for the cloud customer now to check how it turns out. Then we can think about this.

app/models/message.rb
```diff
@@ -365,4 +367,10 @@ def set_conversation_activity
     conversation.update_columns(last_activity_at: created_at)
     # rubocop:enable Rails/SkipsModelValidations
   end
+
+  def update_message_sentiments
+    # override in the enterprise ::Enterprise::SentimentAnalysisJob.perform_later(self)
+  end
```
scmmishra, line 373: @tejaswinichile don't have much context on this, but is it possible to not have this method here at all? CC: @sojan-official

scmmishra (approved):
One minor comment. Approving the PR anyway, if there's nothing to change feel free to go ahead and merge it.

PR #7476, opened 2023-07-06
feat: Add draft messages to the store

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-2194/store-draft-messages-in-the-vuex-store
Currently, all the draft messages are stored in local storage, which is not reactive. In order to implement more AI options, we need draft messages in the command bar and other AI assistance-related components. This pull request adds the ability to store draft messages in the store, which can be used to generate AI content from the draft message.


nithindavid (approved):
Basic cases tested here - https://www.loom.com/share/37afd89eb1cf47a6ad5516036d884015

Also verified the markdown cases which is not mentioned in the video

PR #7493, opened 2023-07-10
feat: Add more options for AI reply suggestions

muhsin-k:
Currently, we offer three options: rephrase, reply to suggestions, and summarise, with two tones: professional and friendly.

We are constantly striving to improve our service, and we plan to add more options, such as:

- Fixing spelling and grammar.
- Shortening text.
- Expanding text.
- Make it more friendly.
- Make it more formal.
- Simplifying the content.


spec/lib/integrations/openai/processor_service_spec.rb
```diff
@@ -97,5 +97,193 @@
         expect(result).to eq('This is a reply from openai.')
       end
     end
+
+    context 'when event name is label_suggestion with labels' do
```
sojan-official, line 101: spec for `label_Suggestions`  should be in enterprise spec ? -> think its already covered

sojan-official (commented):
small change

spec/lib/integrations/openai/processor_service_spec.rb
muhsin-k: @sojan-official Removed.

sojan-official (approved):

PR #7495, opened 2023-07-10
feat: add message condition

scmmishra:
Fetch label suggestions only if there are at least 3 incoming messages


enterprise/lib/enterprise/integrations/openai_processor_service.rb
```diff
@@ -22,6 +22,10 @@ def labels_with_messages
 
     character_count = labels.length
     conversation = find_conversation
+
+    # return nil if conversation has less than 3 incoming messages
+    return nil if conversation.messages.incoming.count < 3
```
sojan-official, line 27: lets add spec to capture this case as well

scmmishra: @sojan-official updated this, sorry missed this comment earlier

sojan-official (approved):

PR #7523, opened 2023-07-14
feat: Add support for API key authentication in Twilio

pranavrajs:
We use Account SID and Account Auth token for Twilio SMS authentication. One user had their Account Auth token leaked (conversation [here](https://app.chatwoot.com/app/accounts/1/conversations/26054)), which prompted them to switch to API Key authentication. API Key authentication is a better option as it reduces the privileges of the key and enables revocation whenever needed.

The changes are straightforward.

- Added a migration to include api_key_sid 
- Use auth_token for storing both api_key_secret and account_auth_token
- Update the UI to include the API Key SID changes.

Screenshot 2023-07-13 at 11 10 13 PM

I've tested this locally. It seems to be working without issues.


iamsivin (approved):
CLGTM

sojan-official (approved):
Code LGTM

PR #7524, opened 2023-07-14
fix: idb is not available in firefox private mode [CW-2217]

scmmishra:
This PR resolves the issue where the cache enabled API would fail to fetch when IDB was not available

## Type of change


muhsin-k (approved):
I have tested this PR locally and it is working fine on Firefox private windows.

https://github.com/chatwoot/chatwoot/assets/12408980/40fafd78-9dce-4dfe-976f-f856a5209bae

PR #7530, opened 2023-07-16
feat: Compute average response time of replies

pranavrajs:
Calculate the average response time of replies. 

**Conditions:** 

- Added a `reply_time` event.
- Added a reporting event listener to compute time based on the REPLY_CREATED event.


pranavrajs:
Getting this out anyways since the changes are additions, doesn't impact the existing features.

PR #7533, opened 2023-07-17
feat: add summary to conversations model

scmmishra:
> ~🚨 Merging this to a feature branch, coz the summary format update is a breaking change, and will break in the UI if it is not merged without the frontend changes~
> updated the prompt, this can be used in the main branch

This PR updates conversations model, to save summary and its timestamp. This PR has the following changes

1. Add the `summary` and `summary_generated_at` fields to summary model
2. Save summary once generated in `openai/processor_service.rb`
3. Add the fields to `partials/_conversation.json.jbuilder`
4. Cleanup summary markdown before rendering it

## Type of change


enterprise/lib/enterprise/integrations/openai_prompts/summary.txt
scmmishra, line 1: Converting this to markdown to better parse it

db/schema.rb
```diff
@@ -449,8 +449,10 @@
     t.datetime "assignee_last_seen_at", precision: nil
     t.datetime "first_reply_created_at", precision: nil
     t.integer "priority"
-    t.bigint "sla_policy_id"
     t.datetime "waiting_since"
+    t.bigint "sla_policy_id"
```
sojan-official, line 453: Order change of 'sla_policy_id' intended?

db/schema.rb
```diff
@@ -615,6 +617,19 @@
...
+    t.bigint "labelable_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["label_id"], name: "index_label_associations_on_label_id"
+    t.index ["labelable_type", "labelable_id"], name: "index_label_associations_on_contact", where: "((labelable_type)::text = 'Contact'::text)"
+    t.index ["labelable_type", "labelable_id"], name: "index_label_associations_on_conversation", where: "((labelable_type)::text = 'Conversation'::text)"
+    t.index ["labelable_type", "labelable_id"], name: "index_label_associations_on_labelable"
+    t.index ["labelable_type", "labelable_id"], name: "index_label_associations_on_labelable_type_and_labelable_id"
+  end
+
```
sojan-official, line 632: Are the changes in schema.rb from the migrations in this PR alone?

sojan-official (commented):
The code looks good to me. 
Please double-check the schema.rb

db/schema.rb
scmmishra: Oh these shouldn't be here, not sure how these came in, fixing it

scmmishra:
@sojan-official Done

scmmishra:
@sojan-official this PR is good to go on my end, PTAL

sojan-official:
@scmmishra  Let's reopen this once we have a clearer direction on how we will take the `summary` feature forward. Let's keep the model attributes trim so that it's easier to document and work with.  In the cases where the attribute is optional  / doesn't require indexing, let's leverage the content attributes jsonb field as well.

scmmishra:
> @scmmishra  Let's reopen this once we have a clearer direction on how we will take the `summary` feature forward. Let's keep the model attributes trim so that it's easier to document and work with.  In the cases where the attribute is optional  / doesn't require indexing, let's leverage the content attributes jsonb field as well. 


Yeah sounds good to me

PR #7538, opened 2023-07-18
feat: Add CTAs for AI features

muhsin-k:
This PR adds CTAs to improve the discoverability of AI features.

**If the integration is not enabled and the user is an admin:**

- Display the "Improve with AI" button with an animation.
- When the button is clicked, show a CTA modal with the following options:
    - Option to add an API key
    - Dismiss: Hiding the option and storing it in the database. The next time the CTA button will not show.

**If the integration is not enabled and the user is an agent:**

- Display the "Improve with AI" button with an animation.
- When the button is clicked, it should behave as it does normally. Subsequent loads will display the button as usual.


app/javascript/dashboard/components/widgets/AICTAModal.vue
```diff
@@ -0,0 +1,120 @@
...
+      } catch (error) {
+        const errorMessage = error?.response?.data?.message;
+        this.alertMessage =
+          errorMessage || this.$t('INTEGRATION_APPS.ADD.API.ERROR_MESSAGE');
+      } finally {
+        this.showAlert(this.alertMessage);
+      }
+    },
+    openOpenAIDoc() {
+      window.open('https://www.chatwoot.com/blog/v2-17', '_blank');
```
muhsin-k, line 104: Currently, we do not have detailed documentation about OpenAI. I will update this after creating one.

app/javascript/dashboard/components/widgets/AIAssistanceButton.vue
```diff
@@ -1,48 +1,85 @@
...
+        v-else
+        v-tooltip.top-end="$t('INTEGRATION_SETTINGS.OPEN_AI.AI_ASSIST')"
+        icon="wand"
+        color-scheme="secondary"
+        variant="smooth"
+        size="small"
+        @click="openAIAssist"
       />
-    </woot-modal>
+      <woot-modal
```
nithindavid, line 17: can we improve the condition check `<div v-if="!isFetchingAppIntegrations">` to consider error case as well, if not should we put the model, outside of the condition for `<div v-if="!isFetchingAppIntegrations">`

app/javascript/dashboard/components/widgets/AICTAModal.vue
```diff
@@ -0,0 +1,120 @@
+<template>
```
nithindavid, line 1: Since we are moving away from foundation, can we use tailwind styles here?

nithindavid (dismissed):

nithindavid:
The first time after clicking the CTAssist with ripple effect is giving me this, any idea @muhsin-k 
image

app/javascript/dashboard/components/widgets/AIAssistanceButton.vue
muhsin-k: I added this condition to hide the CTA button until the integrations are fetched.

muhsin-k:
@nithindavid @iamsivin All the review comments are address. Please take a look.

sojan-official:
@muhsin-k  ping

PR #7566, opened 2023-07-21
fix: no method error when conversation is nil

scmmishra:
Fixes: https://chatwoot-p3.sentry.io/issues/4314511928/

## Type of change


scmmishra:
@pranavrajs I tried finding the reason why conversation ID was not passed, couldn't find it. I've added a guard on the frontend for now. Regardless, the cases of a failing request are handled, so should not affect user experience as of now

sojan-official (approved):

PR #7567, opened 2023-07-21
fix: undefined method `first` for nil:NilClass

scmmishra:
Fixes: https://chatwoot-p3.sentry.io/issues/4313886754/?referrer=Linear

## Type of change

### DIFF
--- a/lib/integrations/openai_base_service.rb
+++ b/lib/integrations/openai_base_service.rb
@@ -75,6 +75,8 @@ class Integrations::OpenaiBaseService
     }
 
     response = HTTParty.post(API_URL, headers: headers, body: body)
-    JSON.parse(response.body)['choices'].first['message']['content']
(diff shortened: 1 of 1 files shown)


lib/integrations/openai_base_service.rb
```diff
@@ -75,6 +75,8 @@ def make_api_call(body)
     }
 
     response = HTTParty.post(API_URL, headers: headers, body: body)
-    JSON.parse(response.body)['choices'].first['message']['content']
+    choices = JSON.parse(response.body)['choices']
+
+    choices ? choices.first['message']['content'] : nil
```
sojan-official, line 80:
```suggestion
    choices.present? ? choices.first['message']['content'] : nil
```

will cover more cases like an empty string

sojan-official (approved):
minor suggestion

PR #7568, opened 2023-07-21
fix: undefined method `update_cache_key` for nil:NilClass

scmmishra:
Fixes: https://chatwoot-p3.sentry.io/issues/4312154361/?referrer=Linear

## Type of change

### DIFF
--- a/app/models/concerns/account_cache_revalidator.rb
+++ b/app/models/concerns/account_cache_revalidator.rb
@@ -2,10 +2,10 @@ module AccountCacheRevalidator
   extend ActiveSupport::Concern
 
   included do
-    after_commit :update_account_cache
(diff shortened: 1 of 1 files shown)


sojan-official:
@scmmishra  better approach will be to skip the callback during delete, which should be the case where the account ends up being empty.

```
module AccountCacheRevalidator
  extend ActiveSupport::Concern

  included do
    after_commit :update_account_cache, on: [:create, :update]
  end

  def update_account_cache
    account.update_cache_key(self.class.name.underscore)
  end
end
```

scmmishra:
> @scmmishra better approach will be to skip the callback during delete, which should be the case where the account ends up being empty.


@sojan-official done

sojan-official (approved):

PR #7578, opened 2023-07-21
feat: Add email forward-to option

tejaswinichile:
Fixes: https://linear.app/chatwoot/issue/CW-2147/re-design-the-email-channel-ui-if-possible

https://www.loom.com/share/25f1dba1c90a44259026b8cb33a27a25


sojan-official:
Let's revisit this later, as the implementation breaks the regular replies

PR #7588, opened 2023-07-24
fix: Improve reply prompt

muhsin-k:
- Improved the reply prompt for generating reply suggestion.
- Moved reply and summary prompts to a separate folder.


lib/integrations/openai/processor_service.rb
```diff
@@ -10,63 +10,44 @@ def summarize_message
...
-    build_api_call_body("#{AGENT_INSTRUCTION} Please make the following response more friendly. #{LANGUAGE_INSTRUCTION}")
-  end
-
-  def make_formal_body
-    build_api_call_body("#{AGENT_INSTRUCTION} Please make the following response more formal. #{LANGUAGE_INSTRUCTION}")
-  end
-
-  def simplify_body
-    build_api_call_body("#{AGENT_INSTRUCTION} Please simplify the following response. #{LANGUAGE_INSTRUCTION}")
+  def prompts_from_file(file_name)
```
scmmishra, line 49:
```suggestion
  def prompts_from_file(file_name)
```

This name is too close to the one used in enterprise, can be confusing, maybe we can let this be `prompt_from_file` and for the other one, we can override that

```rb
def prompt_from_file(file_name, enterprise = False)
  path = enterprise ? 'enterprise/lib/enterprise/integrations/openai_prompts' : 'lib/integrations/openai_prompts'
  Rails.root.join(path, "#{file_name}.txt").read
end
```

CC: @sojan-official

sojan-official: Sounds good. 
- Also, is there a case where we have separate prompts for open source/enterprise?
- Do we need to keep prompts under enterprise? 

if not we can have a unified folder for all the prompts.

muhsin-k: > Also, is there a case where we have separate prompts for open source/enterprise?

Yes. We use different prompts for the summary in EE and OS.

scmmishra (approved):

PR #7601, opened 2023-07-25
chore: Add June event for AI reply suggest dismissal

muhsin-k:
Added a June event for dismissal when the user dismisses the AI-generated content.


app/javascript/dashboard/components/widgets/AIAssistanceButton.vue
```diff
@@ -61,6 +61,9 @@ export default {
       }
     },
     hideAIAssistanceModal() {
+      this.recordAnalytics('DISMISS_AI_SUGGESTION', {
+        aiOption: this.aiOption,
```
scmmishra, line 65: @pranavrajs should we record the generated suggestion as well? 
Another option is that we implement a survey which is triggered on dismiss, we can record it to june. Ofcourse can be implemented separately after the v3 release

muhsin-k: Makes sense. That makes sense. When adding the survey, we should ask the user for permission to share the generated content.
Another suggestion is to include hint text that says "Results may not be accurate.”

pranavrajs: > Another option is that we implement a survey which is triggered on dismiss, we can record it to june. Ofcourse can be implemented separately after the v3 release

This makes sense.

scmmishra (approved):

Issue #4299, opened 2022-03-27
[CW-2299] Display name and full name both are showing on widget


ammaraslam:
**Describe the bug**

I've set Display name as **Sarah** and full name as test test ; considering the display name is Sarah, it should show Sarah everywhere on the widget - this seems like related to active agents where widget only shows the name of the active agent but when clicked on the widget, it switches to display name. 

**Expected behavior**

It should show the display name if it's set by user,  everywhere. 

**Recording**

https://user-images.githubusercontent.com/99071637/160279124-3541fb68-02cd-4202-869d-f869955822b3.mp4

OS : Windows 10
Browser : Chrome

<sub>[CW-2299](https://linear.app/chatwoot/issue/CW-2299/display-name-and-full-name-both-are-showing-on-widget)</sub>

manovasanth-b:
Is it fixed ? 
or shall i take this and fix.

ammaraslam:
> Is it fixed ? 
> or shall i take this and fix.

I doubt it was fixed, see if you're able to reproduce this.

cc : @sojan-official @pranavrajs

LiamAshdown:
I cannot replicate this on develop branch.

Possibly fixed?

sojan-official:
@LiamAshdown  Thanks for looking into this. I was able to replicate this in develop for campaigns
added a fix: #7631

LiamAshdown:
> @LiamAshdown Thanks for looking into this. I was able to replicate this in develop for campaigns added a fix: #7631

Ahhhh! I see, I was looking at the conversation message instead of the campaign.

Nice one :)

PR #7631, opened 2023-07-27
fix: Use Display name for campaigns

sojan-official:
fixes: #4299


iamsivin (approved):
I tested locally and LGTM.

PR #7647, opened 2023-07-31
fix: change insert chat

s00d:
# Pull Request Template

## Description

Currently, when creating a chat via sockets, it is added to the end, which can cause it to be missed. This fix will resolve the issue with chat creation. However, unfortunately, something else needs to be done regarding sorting, as sorting by priority still won't work.

## Type of change

- [✓] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

I just looked at the logic of chat creation.

## Checklist:

- [✓] My code follows the style guidelines of this project
- [✓] I have performed a self-review of my code
- [✓] I have commented on my code, particularly in hard-to-understand areas
- [✓] I have made corresponding changes to the documentation
- [✓] My changes generate no new warnings
- [✓] I have added tests that prove my fix is effective or that my feature works
- [✓] New and existing unit tests pass locally with my changes
- [✓] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/app/javascript/dashboard/store/modules/conversations/index.js
+++ b/app/javascript/dashboard/store/modules/conversations/index.js
@@ -29,7 +29,7 @@ export const mutations = {
         c => c.id === conversation.id
       );
       if (indexInCurrentList < 0) {
-        newAllConversations.push(conversation);
+        newAllConversations.unshift(conversation);
       }
     });
     _state.allConversations = newAllConversations;
@@ -186,7 +186,7 @@ export const mutations = {
   },
 
   [types.ADD_CONVERSATION](_state, conversation) {
-    _state.allConversations.push(conversation);
+    _state.allConversations.unshift(conversation);
   },
 
   [types.UPDATE_CONVERSATION](_state, conversation) {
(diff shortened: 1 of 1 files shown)


pranavrajs (changes requested):
@s00d Thanks for the PR. However, I'm not sure if I understand the issue properly. 

If the sort is `last_activity_at`, irrespective of the place where the object is inserted, the conversation would be shown at the top. We are sorting it based on the key before displaying it on the UI.

s00d:
> @s00d Thanks for the PR. However, I'm not sure if I understand the issue properly.
> 
> If the sort is `last_activity_at`, irrespective of the place where the object is inserted, the conversation would be shown at the top. We are sorting it based on the key before displaying it on the UI.

The problem is that you don't have sorting when creating a chat through sockets. For example, in the ADD_CONVERSATION function, you simply add a new dialogue to the end of the allConversations array. In the ChatList component, there are only filters based on the user, so the dialogue added through sockets remains at the bottom. Ideally, after pushing, you would sort the conversations. However, I don't have time to study how your filtering works, so my fix will simply add chats to the top regardless of the sorting.

PR #7648, opened 2023-07-31
feat: Add message.created_at to conversation_transcript

### LINKED ISSUE #2999: [CW-1866] Add date and timestamps to transcripts sent via email
Using a website Inbox, users or admins receive a chat transcript without date or timestamps, making it difficult to gauge the inter-message response times and start/end times for a chat session 

The date and timestamp are available on every message within the Chatwoot admin interface.  For contacts, the chat messages are divided by a date header.

When a chat transcript is sent, it would be great to have a global option or per-inbox option to allow the sending of precise date and timestamps.

**Additional context**
I'm happy for the date and timestamps to be sent to user & admins, but it might make sense to divide the option between admins, whom get the precise detail, and end users / contacts whom might only get the date without time precision, in case organisations want to hide that level of detail from end users.

Thanks!

<sub>[CW-1866](https://linear.app/chatwoot/issue/CW-1866/add-date-and-timestamps-to-transcripts-sent-via-email)</sub>

CristianDuta:
# Pull Request Template

## Description

Fixes #2999

I added timestamps to the email template in the most minimal way possible. This value of the timestamps come from the message and it is formatted and converted based on the inbox timezone and the format "%b %d, %Y, %I:%M %p"

## Type of change

## How Has This Been Tested?

Manually only. Example email:
Screenshot 2023-07-31 at 16 06 12

## Checklist:

- [ ] ~I have added tests that prove my fix is effective or that my feature works~


app/views/mailers/conversation_reply_mailer/conversation_transcript.html.erb
```diff
@@ -14,6 +14,9 @@
           Attachment [<a href="<%= attachment.file_url %>" _target="blank">Click here to view</a>]
         <% end %>
       <% end %>
+      <p style="font-size: 90%; color: #808080;">
+        <%= message.created_at.in_time_zone(@inbox.timezone).strftime('%b %d, %Y, %I:%M %p') if message.created_at %>
```
pranavrajs, line 18: @CristianDuta inbox.timezone is not a mandatory field. Any idea how to navigate that?

pranavrajs (changes requested):
Thanks for the PR. One small clarification.

app/views/mailers/conversation_reply_mailer/conversation_transcript.html.erb
CristianDuta: @pranavrajs Thank you for your question. 

That is indeed true and I thought of it. Although the timezone is not mandatory, a default value of UTC is set for each inbox channel type.

I also believe it is the duty of the administrator to set a timezone for the inboxes.

That being said, we could also add the TZ at the end to be more precise. What do you think?

CristianDuta: @pranavrajs would you like to see the timezone at the end of the string or do you think it would be too much? 

e.g.: `Aug 2, 2023, 07:26 AM UTC` or `Aug 2, 2023, 09:26 AM Europe/Berlin`

pranavrajs: Sounds good. I've updated the logic. However, I've removed the year from the timestamp. It doesn't seem necessary, but I can add it again if needed after feedback from customers.

pranavrajs (approved):

app/views/mailers/conversation_reply_mailer/conversation_transcript.html.erb
CristianDuta: @pranavrajs Thank you! Cheers!

PR #7652, opened 2023-08-01
fix: Update the default status of the hooks to enabled

pranavrajs:
In the recent Slack integration update, we made changes to how hooks are handled. Now, hooks require an additional check to be enabled. By default, new hooks are created in a disabled state, which might lead to issues with the hooks not being executed as expected. 

This migration updates the status attribute of hooks to have 'enabled' as the default value, ensuring that new hooks are enabled by default and can function properly.

### DIFF
--- a/app/models/integrations/hook.rb
+++ b/app/models/integrations/hook.rb
@@ -6,7 +6,7 @@
 #  access_token :string
 #  hook_type    :integer          default("account")
 #  settings     :jsonb
-#  status       :integer          default("disabled")
+#  status       :integer          default("enabled")
 #  created_at   :datetime         not null
 #  updated_at   :datetime         not null
 #  account_id   :integer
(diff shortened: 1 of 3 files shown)


sojan-official (approved):

sojan-official:
Note: We introduced this attribute assuming that objects would be created as 'disabled' by default, and then 'enabled' after verification. This approach caters to flows involving multiple steps, preventing integration callbacks from executing if the user has only completed part of the flow. For single-step integrations, 'enabled' will be passed explicitly upon creation if we're confident the credentials are valid. 

However, most of our integration code wasn't written considering this scenario, hence our current direction.

An alternate strategy could have been to introduce an 'after_create' callback that sets the value to 'enabled', audit each integration individually, make updates per the cases above, and then remove the callback once all integrations have been verified.

PR #7686, opened 2023-08-07
chore: fix flaky spec

sojan-official:
- fix flaky reporting specs
- fix flaky export spec


sojan-official:
New Flaky Specs 

rspec ./spec/controllers/api/v2/accounts/report_controller_spec.rb:49 # Reports API GET /api/v2/accounts/:account_id/reports/account when it is an authenticated user return timeseries metrics
rspec ./spec/controllers/api/v2/accounts/report_controller_spec.rb:180 # Reports API GET /api/v2/accounts/:account_id/reports/summary when it is an authenticated user returns summary metrics

sojan-official:
- To be merged after we execute this in various time points to ensure this is correct.

PR #7267, opened 2023-06-07
fix: Rename `unread_since` scope

nhippenmeyer:
## Description

This renames the `unread_since` scope to `created_since`, which more accurately describes what the scope returns.

The `EXTRACT(EPOCH FROM created_at) > (?)` clause was also simplified and rewritten as `created_at > ?`, which is equivalent:

image (1)

PR #7690, opened 2023-08-07
fix: Update unread_count being 0 if agent has not seen the conversation

pranavrajs:
Fixes https://linear.app/chatwoot/issue/CW-2360/new-conversations-are-created-with-unread-count-=-0

This PR https://github.com/chatwoot/chatwoot/pull/7267 introduced a bug. When a conversation is created, the agent_last_seen_at would not be available. Querying created_at > nil returned empty results. A case where agent_last_seen_at being empty should have been added. The spec was also missing to validate this case.

This PR would check for the existence of agent_last_seen_at attribute. If present, it would return all the message since the time, otherwise it would return all the messages.

### DIFF
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -166,11 +166,11 @@ class Conversation < ApplicationRecord
   end
 
   def unread_messages
-    messages.created_since(agent_last_seen_at)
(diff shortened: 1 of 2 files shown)


pranavrajs:
Merging this as the specs are passing.

Issue #5433, opened 2022-09-14
WhatsApp Cloud repeated messages issue


harismateen:
*Describe the bug*

Repeated message

*To Reproduce*

1. Go to 'Whatsapp cloud inbox'
2. User send picture
5. See error in the inbox

*Expected behavior*

It does not repeat the same message.

*Screenshots*
as attached
![image](https://user-images.githubusercontent.com/38554929/190058423-ad057abb-7709-455a-bdcf-9b2478a0b6e3.png)
![image](https://user-images.githubusercontent.com/38554929/190058437-5da7c4e2-a0f4-4e45-a216-415dafccf8b8.png)

sojan-official:
@harismateen, Could you check your server logs for any errors? We can't seem to reproduce this issue in our installation.

sojan-official:
- We recently observed a case where the installation received duplicate webhook events from meta for Whatsapp messages. So will be adding a fix for this.

PR #7701, opened 2023-08-10
feat: locking and retry in FB message parsing

scmmishra:
This PR adds a Redis based lock for Facebook message processing to avoid race conditions/concurrent processing of messages from the same sender.

This PR adds the `Redis::LockManager`, it provides a simple mechanism to handle distributed locks using Redis. example usage

```ruby
lock_manager = Redis::LockManager.new

if lock_manager.lock("some_key")
  # Critical code that should not be run concurrently
  lock_manager.unlock("some_key")
end
```

This also adds a new namespace pool called `$sherlock`, our key solver. This is added to ensure that locks don't collide anywhere.

The LockManager is used in `Webhooks::FacebookEventsJob`, which uses the [retry_on](# https://edgeapi.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on) method to retry the message creation at least 5 times, with a 10 second backoff.

## TODO

- [x] Tests for LockMangager
- [x] Tests for Webhooks::FacebookEventsJob retry mechanism
- [x] Figure out better timeouts for retry and lock duration


config/initializers/01_redis.rb
```diff
@@ -12,3 +12,10 @@
   config = Rails.env.test? ? MockRedis.new : Redis.new(Redis::Config.app)
   Redis::Namespace.new('velma', redis: config, warning: true)
 end
+
+# Sherlock: Key solver
+# Used for Redis-backed locks
+$sherlock = ConnectionPool.new(size: 5, timeout: 1) do
```
sojan-official, line 18: Let's not introduce a new connection pool since some of the self-hosted installations might be running on redis providers with less number of connections available.   lets reuse alfred

app/jobs/webhooks/facebook_events_job.rb
```diff
@@ -1,8 +1,28 @@
...
+
   queue_as :default
+  # https://edgeapi.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on
+  retry_on LockAcquisitionError, wait: 10.seconds, attempts: 5
 
   def perform(message)
     response = ::Integrations::Facebook::MessageParser.new(message)
-    ::Integrations::Facebook::MessageCreator.new(response).perform
+
+    lock_key = "#{response.sender_id}_#{response.recipient_id}"
```
sojan-official, line 11: let's use the `lib/redis/redis_keys.rb`  and define a specific key pattern for the case to avoid key collision

lib/redis/lock_manager.rb
```diff
@@ -0,0 +1,81 @@
...
+  #
+  # === Returns
+  # * +true+ if the lock was successfully acquired.
+  # * +false+ if the lock was not acquired.
+  def lock(key, timeout = LOCK_TIMEOUT)
+    value = Time.now.to_f.to_s
+    with_redis do |conn|
+      # nx: true means set the key only if it does not exist
+      conn.set(key, value, nx: true, ex: timeout) ? true : false
+    end
```
sojan-official, line 37: alfred should already have the method for the set.  we can reuse that 
might have to add support for the `nx:` argument there

sojan-official (commented):
LGTM overall. some minor comments on code organisation.

config/initializers/01_redis.rb
scmmishra: Sounds good, I thought of the same, but makes sense to re-use alfred

app/jobs/webhooks/facebook_events_job.rb
scmmishra: Noted

scmmishra: Is the key appropriate? Will it be unique for that sender conversation

lib/redis/lock_manager.rb
scmmishra: Done

scmmishra:
@sojan-official LMK your thoughts on the key used to acquire the lock? Also about the lock timings, I was thinking we can increase the lock timeout to about 30 seconds

.rubocop.yml
```diff
@@ -38,6 +38,7 @@ Style/GlobalVars:
     - 'config/initializers/01_redis.rb'
     - 'config/initializers/rack_attack.rb'
     - 'lib/redis/alfred.rb'
+    - 'lib/redis/lock_manager.rb'
```
sojan-official, line 41: this won't be required now right ?

app/jobs/webhooks/facebook_events_job.rb
```diff
@@ -1,8 +1,28 @@
 class Webhooks::FacebookEventsJob < ApplicationJob
+  class LockAcquisitionError < StandardError; end
+
   queue_as :default
+  # https://edgeapi.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on
+  retry_on LockAcquisitionError, wait: 10.seconds, attempts: 5
```
sojan-official, line 6: The timing looks okay. We could even reduce the timing to, say `2` seconds.  The process message job should be done within a sub-second.  the more time here, it would feel like subsequent messages to be lagging for the end user

lib/redis/alfred.rb
```diff
@@ -7,8 +7,8 @@ class << self
     # key operations
 
     # set a value in redis
-    def set(key, value)
-      $alfred.with { |conn| conn.set(key, value) }
+    def set(key, value, nx: false, ex: false) # rubocop:disable Naming/MethodParameterName
+      $alfred.with { |conn| conn.set(key, value, nx: nx, ex: ex) }
     end
```
sojan-official, line 13: lets add a comment over setex  with a  TODO: deprecation and recommending usage of `set`

lib/redis/lock_manager.rb
```diff
@@ -0,0 +1,63 @@
...
+# if lock_manager.lock("some_key")
+#   # Critical code that should not be run concurrently
+#   lock_manager.unlock("some_key")
+# end
+#
+class Redis::LockManager
+  # Default lock timeout set to 15 seconds. This means that if the lock isn't released
+  # within 15 seconds, it will automatically expire.
+  # This helps to avoid deadlocks in case the process holding the lock crashes or fails to release it.
+  LOCK_TIMEOUT = 15.seconds
```
sojan-official, line 18: lets reduce the LOCK_TIMEOUT to something like 5 seconds.. 
^ ref prev comment

lib/redis/redis_keys.rb
```diff
@@ -33,4 +33,7 @@ module Redis::RedisKeys
   MESSAGE_SOURCE_KEY = 'MESSAGE_SOURCE_KEY::%<id>s'.freeze
   CUSTOM_FILTER_RECORDS_COUNT_KEY = 'CUSTOM_FILTER::%<account_id>d::%<user_id>d::%<filter_id>d'.freeze
   OPENAI_CONVERSATION_KEY = 'OPEN_AI_CONVERSATION_KEY::%<event_name>s::%<conversation_id>d::%<updated_at>d'.freeze
+
+  ## Sempahores / Locks
+  FACEBOOK_MESSAGE_CREATE_LOCK = 'FB_MESSAGE_CREATE_LOCK::%<sender_id>s::%<recipient_id>s'.freeze
```
sojan-official, line 38: let's also add a comment here, something like  ` We don't want to process messages from the same sender concurrently  to prevent creating double conversations `
FACEBOOK_MESSAGE_MUTEX ?

sojan-official (commented):
LGTM overall .
Some minor comments.  let's have this added for whatsapp messages as well as a follow up 

ref: https://github.com/chatwoot/chatwoot/issues/5433

.rubocop.yml
scmmishra: Yep

app/jobs/webhooks/facebook_events_job.rb
sojan-official: this combination should be good. 👍 
even processing messages from `sender_id` alone sequentially should be okay.

scmmishra:
@sojan-official tests are all done, just need to take a look at the key used for locking and we're good to go

sojan-official:
The code looks good to me. Let's test this on staging and then get it merged.

scmmishra:
@sojan-official tested this on staging, working fine. We're good to merge this

scmmishra:
@sojan-official the breaking test is because of a security audit. This PR solves it https://github.com/chatwoot/chatwoot/pull/7765

sojan-official (approved):

PR #7712, opened 2023-08-11
feat: show label suggestions as toast when conversation is resolved [WIP]

scmmishra:
Preview: https://github.com/chatwoot/chatwoot/assets/18097732/046b81ac-1912-4c58-be0d-f30bea29e434

image


nithindavid:
@shivam Should we shrink the content?  Something like "Forgot these labels?"

scmmishra:
> Forgot these labels?

Updated the copy and the width, @nithindavid PTAL

https://github.com/chatwoot/chatwoot/assets/18097732/2fbd6988-357d-487b-aa65-e21d1e6f46a4

scmmishra:
@nithindavid I have implemented this, but the more I look at it, the more it feels too loud from a UX point of view. As a user I would wanna resolve and jump on the next conversation, the amount of visual affordance this popup has it a lot. WDYT? CC: @pranavrajs

scmmishra:
Closing this PR, will create a new one with a better implementation

PR #7729, opened 2023-08-16
chore: Ability to disable rack attack on widget endpoints

sojan-official:
Ability to disable rack attack specifically for widget endpoints


sojan-official:
- Verified in local machine

pranavrajs (approved):

PR #7746, opened 2023-08-17
fix: Fix restart

s00d:
Hello, if a client switches from one account to another, the chatwood stops working. We need to wait for the restart to complete everywhere. I added a promise, and that's enough for me, but I think you might need to modify the code on your end that uses the restart.

## Type of change

## How Has This Been Tested?

Received message from customer support. We have reviewed the code and noticed that the restart doesn't have enough time to execute.

## Checklist:


pranavrajs:
@s00d Thanks for the PR. Can you please share the error and a screencast of the behavior before this change? It is not very evident from the PR.

s00d:
> @s00d Thanks for the PR. Can you please share the error and a screencast of the behavior before this change? It is not very evident from the PR.

There are no errors; the issue lies in the logic. When changing the user, it is necessary to clear the old session and then send the new user's data. However, in your code, the reset method is being completed after the set_user method, causing the chat to break and stop receiving messages when sending data for another user.

I added a promise to understand when the reset will be executed.

PR #7773, opened 2023-08-21
feat: add mutex for slack send job

scmmishra:
DEPENDS ON: https://github.com/chatwoot/chatwoot/pull/7701


scmmishra:
> @sojan-official @pranavrajs we can make common implementation of this mutex, something like the following
> 
> ```ruby
> class MutexApplicationJob < ApplicationJob
>   class LockAcquisitionError < StandardError; end
> 
>   def with_lock(key_format, *args)
>     lock_key = format(key_format, *args)
>     lock_manager = Redis::LockManager.new
> 
>     if lock_manager.locked?(lock_key)
>       Rails.logger.error "[#{self.class.name}] Failed to acquire lock on attempt #{executions + 1}: #{lock_key}"
>       raise LockAcquisitionError, "Failed to acquire lock for key: #{lock_key}"
>     end
> 
>     begin
>       lock_manager.lock(lock_key)
>       Rails.logger.info "[#{self.class.name}] Acquired lock for: #{lock_key}"
>       yield
>     ensure
>       # Ensure that the lock is released even if there's an error in processing
>       lock_manager.unlock(lock_key)
>     end
>   end
> end
> ```
> 
> LMK what do you folks think about it

P.S. I've implemented this in the PR, PTAL

PR #7783, opened 2023-08-23
feat: implement mutex for `SlackSendJob`

scmmishra:
This PR implements redis based mutex for `SlackSendJob`

---

1. Implement mutex for Slack send job.
2. Create `MutexApplicationJob` that exposes `with_lock`
3. Refactor `facebook_events_job` to use the new `MutexApplicationJob`


sojan-official (approved):
Went through the code, and it looks good to me. 
Let's merge it after testing.

scmmishra:
@sojan-official tested this on staging. To replicate, try sending multiple messages really one after the other really fast.

<details>
<summary>
Before mutex
</summary>
CleanShot 2023-08-24 at 17 26 28@2x

CleanShot 2023-08-24 at 17 26 35@2x
</details>

After the deploy, the mutex "partially" works, multiple messages are not triggered on slack. But there is still some problem with setting the context, replies on slack don't show up. Will test that and see if it's a config issue or a bug

PR #7785, opened 2023-08-23
feat: update lock timeout and retry configuration

scmmishra:
The PR has the following changes

1. The retry will happen with exponential rather than fixed delays.
2. Update the default lock timeout to a more conservative number
3. Fix `execution` count that is logged

Blocked on: https://github.com/chatwoot/chatwoot/pull/7786


app/jobs/webhooks/facebook_events_job.rb
```diff
@@ -3,7 +3,7 @@ class LockAcquisitionError < StandardError; end
 
   queue_as :default
   # https://edgeapi.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on
-  retry_on LockAcquisitionError, wait: 2.seconds, attempts: 5
+  retry_on LockAcquisitionError, wait: proc { |attempt| 2**attempt }, attempts: 5
```
scmmishra, line 6: @sojan-official one problem with an exponential backoff is that the sequence in which the events are processed might get affected. 

Should I change it to a fixed time, say `4 seconds` instead of the original `2 seconds`?

```suggestion
  retry_on LockAcquisitionError, wait: 4.seconds, attempts: 5
```

sojan-official (approved):

PR #7792, opened 2023-08-24
chore: Revert gem update

sojan-official:
- updating redis client caused deployment issues in Heroku, hence reverting to prev version until its resolved


sojan-official:
- merging this to unblock develop deploys. we will get back to this later

PR #7800, opened 2023-08-24
fix: XSS in custom attributes

iamsivin:
# Pull Request Template

## Description

This PR includes a fix for XSS in the custom attributes link type.

Fixes https://linear.app/chatwoot/issue/CW-2423/xss-in-custom-attributes

## Type of change

## How Has This Been Tested?

## Checklist:


app/javascript/dashboard/components/CustomAttribute.vue
```diff
@@ -188,6 +188,9 @@ export default {
     urlValue() {
       return isValidURL(this.value) ? this.value : '---';
     },
+    hrefURL() {
+      return isValidURL(this.value) ? this.value : '';
```
scmmishra, line 192: Should we sanitise the input on the server where it is committed to the model instead of having a front-end only solution? 
CC: @sojan-official

sojan-official: My thoughts align with this article: https://benhoyt.com/writings/dont-sanitize-do-escape/
We could have URL validations for the backend. But it will also be safe to ensure the output is escaped.

scmmishra: Yep, we can have both. I suggested this because folks may access this data via an API to display in their apps, and would assume that the data here is safe, we can perhaps escape it while sending it in the json

scmmishra: @sojan-official what do we do here? I can approve this PR for now, should we make the backend changes too?

pranavrajs:
@iamsivin @sojan-official Please take a look at this and resolve the comments.

PR #7808, opened 2023-08-25
fix: Attachment sent from slack doesn't reach to chatwoot

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-2366/files-sent-from-slack-doesnt-reach-the-customer

https://www.loom.com/share/a606c17136284ec68e51a059508615e7


config/integration/apps.yml
```diff
@@ -13,7 +13,7 @@ slack:
   id: slack
   logo: slack.png
   i18n_key: slack
-  action: https://slack.com/oauth/v2/authorize?scope=commands,chat:write,channels:read,channels:manage,channels:join,groups:read,groups:write,im:write,mpim:write,users:read,users:read.email,chat:write.customize,channels:history,groups:history,mpim:history,im:history
+  action: https://slack.com/oauth/v2/authorize?scope=commands,chat:write,channels:read,channels:manage,channels:join,groups:read,groups:write,im:write,mpim:write,users:read,users:read.email,chat:write.customize,channels:history,groups:history,mpim:history,im:history,files:read
```
sojan-official, line 16: should we add https://api.slack.com/scopes/files:write permission as well ?

app/services/integrations/slack/incoming_message_builder.rb
```diff
@@ -93,7 +97,18 @@ def private_note?
   def create_message
     return unless conversation
```
sojan-official, line 98: There seems to be a code climate issue here. could refactor this method into smaller chunks

config/integration/apps.yml
muhsin-k: Yes.

app/services/integrations/slack/incoming_message_builder.rb
muhsin-k: Fixed.

pranavrajs:
Please break this down to smaller PR fixing only the change that is required to fix it. Rest should be taken as a separate chore.

muhsin-k:
> Please break this down to smaller PR fixing only the change that is required to fix it. Rest should be taken as a separate chore.

Created new PR for the fix, https://github.com/chatwoot/chatwoot/pull/7852, Other changes will address in a separate PR.

Issue #7809, opened 2023-08-26
Error Encountered When Attempting to Delete User-Associated Accounts via Super Admin Console


smtbos:
### Describe the bug

When accessing the super admin console and following the steps below, an error is encountered while attempting to delete user-associated accounts:

1. Log in to the super admin console.
2. Navigate to the user section.
3. Click on a specific user.
4. Within the user's profile, there is a list of accounts associated with them.
5. When attempting to delete (destroy) one of these associated accounts, an error is displayed.

This issue prevents the successful deletion of user-associated accounts, impeding the expected functionality within the admin console. Further investigation is needed to identify the root cause of this error and implement a resolution to ensure smooth management of user accounts and associated data.

### To Reproduce

Not any.

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

None

smtbos:
Hello @sojan-official,

Please check the issue and let me know if the solution I have done is ok or not.

Thanks

sojan-official:
Thank you so much for your contribution. We truly appreciate your efforts. 🙏 

Our team will prioritize reviewing and testing your changes according to our existing schedule and workload. 
Should there be a need for any further actions or information from your side, we'll follow up here. 📨

Once again, we're grateful for your involvement and contribution to our community. 🚀🌟 🙌

PR #7810, opened 2023-08-26
fix: Error Encountered When Attempting to Delete User-Associated Accounts via Super Admin Console

smtbos:
## Description

This pull request addresses the issue where an error is encountered while attempting to delete user-associated accounts through the Super Admin Console. The problem arises from an error that occurs when the "destroy" action is triggered on an associated account.
Fixes: #7809

## Type of change

[User collection view](app/views/super_admin/users/_collection.html.erb)


smtbos:
@sojan-official

Please check this PR.

sojan-official (approved):
verified in local machine

PR #7828, opened 2023-08-31
feat: mutex for `InstagramEventsJob` [CW-2447]

scmmishra:
This PR adds a mutex for `InstagramEventsJob`; the key used is the ID of the users Instagram Professional account.

### DIFF
--- a/app/jobs/webhooks/facebook_events_job.rb
+++ b/app/jobs/webhooks/facebook_events_job.rb
@@ -1,6 +1,6 @@
 class Webhooks::FacebookEventsJob < MutexApplicationJob
   queue_as :default
-  retry_on LockAcquisitionError, wait: 1.second, attempts: 6
+  retry_on LockAcquisitionError, wait: 1.second, attempts: 8
 
   def perform(message)
     response = ::Integrations::Facebook::MessageParser.new(message)
--- a/app/jobs/webhooks/instagram_events_job.rb
+++ b/app/jobs/webhooks/instagram_events_job.rb
@@ -30,14 +30,11 @@ class Webhooks::InstagramEventsJob < MutexApplicationJob
   private
 
   def ig_account_id
-    @entries.first[:id]
+    @entries&.first&.dig(:id)
   end
 
   def sender_id
-    return unless @entries.first[:messaging].presence
-
-    messaging = @entries.first[:messaging]
-    messaging.first[:sender][:id]
+    @entries&.dig(0, :messaging, 0, :sender, :id)
   end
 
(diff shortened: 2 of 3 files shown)


scmmishra:
@sojan-official @pranavrajs the message entries we receive for processing looks something like the following

```json
[
  {
    "time": 1693460815900,
    "id": "17841425205523412",
    "messaging": [
      {
        "sender": { "id": "17841425205523412" },
        "recipient": { "id": "XXXXXXXXXXXXXXXX" },
        "timestamp": 1693460815485,
        "message": {
          "mid": "aWdfZAG1faXRlbToxOklHTWVzc2FnZAUlEOjE3ODQxNDI1MjA1NTIzNDEyOjM0MDI4MjM2Njg0MTcxMDMwMTI0NDI1OTA0OTQ1MTY3NDQ0NTcwMDozMTIzODgzODI2MjExNTU4NzcxMDI0MjMwMjU4MTUzODgxNgZDZD",
          "text": "https://biostore.uz/product/hhs-a1-l-carnitine-lepidium-bitkisel-30-kaps",
          "is_echo": true
        }
      }
    ]
  }
]
```

The code loops over the root array and then on the `messaging` array and creates a message for each entry. The problem is that I have added a lock using the `id` from the first root item as the key for the lock. This would mean only one incoming message will be processed per inbox, regardless of who sent it. **This has the possibility to choke the system if a single sender starts bombarding messages.** Need your thoughts here

ref: [FB docs](https://developers.facebook.com/docs/messenger-platform/instagram/features/webhook)

sojan-official (approved):

scmmishra:
> This has the possibility to choke the system if a single sender starts bombarding messages. Need your thoughts here.

@sojan-official what about this?

sojan-official:
@scmmishra I have yet to see a case in Facebook APIs where the messaging node has more than one entry.  

> Does the root array always have only one item?
we can go ahead with this assumption and use the sender id as the key


> A possible solution is that we refactor the Webhooks::InstagramEventsJob and let it process a single valid message only, while Webhooks::InstagramController handles the loop

this also sounds like a good solution, which is future-proof. We can take it up if you think the required refactoring is relatively simple and the testing isn't complex. other wise we can go ahead with the assumption of having 1 item in the array and sender id to be present always in it

scmmishra:
@sojan-official 

> we can go ahead with this assumption and use the sender id as the key

Let's merge this with the sender_id. The refactoring changes, I can do this week in a follow-up PR. For now, we can merge and deploy this to ensure that the bug is fixed

PR #7848, opened 2023-09-04
fix: Fixes height shift issue with resizable textarea [cw-2438]

nithindavid:
## Description
https://www.loom.com/share/198054f28bd44fa6887b2371e99cf2a9

Fixes # (issue)
https://linear.app/chatwoot/issue/CW-2438/messenger-editor-break-on-chatwoot-cloud

## Type of change

## How Has This Been Tested?

## Checklist:


app/javascript/widget/components/ChatInputWrap.vue
```diff
@@ -207,12 +209,18 @@ export default {
...
-  margin-top: $space-small;
-  margin-bottom: $space-small;
+  padding: 0;
+  display: flex;
+  box-sizing: content-box;
+  position: relative;
+  padding-top: $space-small;
+  top: $space-smaller;
+  align-self: flex-start;
+  background: transparent;
```
iamsivin, line 224: @nithindavid Can we use tailwind since we are moving?

app/javascript/shared/components/ResizableTextArea.vue
```diff
@@ -34,18 +35,19 @@ export default {
...
     },
   },
   mounted() {
     this.$nextTick(() => {
-      if (this.value) {
-        this.resizeTextarea();
-      }
+      this.resizeTextarea();
     });
   },
```
scmmishra, line 47: Let's make the watch immediate, that way we won't need the mounted call

scmmishra (changes requested):

app/javascript/shared/components/ResizableTextArea.vue
```diff
@@ -34,18 +35,14 @@ export default {
...
+      this.$nextTick(() => {
         this.resizeTextarea();
-      }
-    });
+      });
+    },
   },
   methods: {
     resizeTextarea() {
+      this.$el.style.height = 'inherit';
```
pranavrajs, line 45: Updating the DOM directly doesn't seem like a good practice. @scmmishra Thoughts?

pranavrajs (changes requested):
Please see the comment.

app/javascript/shared/components/ResizableTextArea.vue
scmmishra: Let's use a computed property that returns the `height` and the style prop can be set directly

nithindavid: I tried making this a computed property. But it does not get updated as we need. Any other ways you can think? Or we need to stick with this implementation 
image

nithindavid: @pranavrajs @scmmishra

scmmishra: Can we just return the value and see?

scmmishra: Another thing is that the height: 'inherit' will always be overriden here

nithindavid: @scmmishra Tried returning the value and keeping the inherit before it as the default. This isn't working either. 
image

tried debounce function on the method also, but it isn't working as expected.

scmmishra: Hey @nithindavid I tried rewriting the computed method as follows

```js
textAreaHeight() {
  if (!this.value) {
    return `${this.minHeight}rem`;
  }
  return this.$refs.textarea.scrollHeight + 'px';
}
```

and added the style prop as follows

```
:style="{
      height: textAreaHeight,
    }"
```

Seems to be working fine after that, can you give it a shot in case I missed something

<details>

<summary>
Full Code
</summary>

```vue
<template>
  <textarea
    ref="textarea"
    :placeholder="placeholder"
    :value="value"
    :style="{
      height: textAreaHeight,
    }"
    rows="1"
    @input="onInput"
    @focus="onFocus"
    @keyup="onKeyup"
    @blur="onBlur"
  />
</template>

<script>
const TYPING_INDICATOR_IDLE_TIME = 4000;
export default {
  props: {
    placeholder: {
      type: String,
      default: '',
    },
    value: {
      type: String,
      default: '',
    },
    minHeight: {
      type: Number,
      default: 2,
    },
  },
  data() {
    return {
      idleTimer: null,
    };
  },
  computed: {
    textAreaHeight() {
      if (!this.value) {
        return `${this.minHeight}rem`;
      }
      return this.$refs.textarea.scrollHeight + 'px';
    },
  },
  methods: {
    onInput(event) {
      this.$emit('input', event.target.value);
    },
    resetTyping() {
      this.$emit('typing-off');
      this.idleTimer = null;
    },
    turnOffIdleTimer() {
      if (this.idleTimer) {
        clearTimeout(this.idleTimer);
      }
    },
    onKeyup() {
      if (!this.idleTimer) {
        this.$emit('typing-on');
      }
      this.turnOffIdleTimer();
      this.idleTimer = setTimeout(
        () => this.resetTyping(),
        TYPING_INDICATOR_IDLE_TIME
      );
    },
    onBlur() {
      this.turnOffIdleTimer();
      this.resetTyping();
      this.$emit('blur');
    },
    onFocus() {
      this.$emit('focus');
    },
    focus() {
      this.$refs.textarea.focus();
    },
  },
};
</script>

```
</details>

nithindavid: @scmmishra @pranavrajs Here's the case where the editor isn't set with the correct height. This is with the suggested solution. 

https://www.loom.com/share/83dd7a25a198464e83b2b611fd72923d

nithindavid: @pranavrajs, with the current implementation the `resizeTextarea()` is called once for reloads and value changes. Also, not using tailwindcss as we have to use `h-[inherit]` which is getting overridden by lots of our existing styles.

app/javascript/shared/components/ResizableTextArea.vue
```diff
@@ -162,3 +170,8 @@ export default {
   },
 };
 </script>
+<style scoped>
+textarea {
+  height: inherit;
```
nithindavid, line 175: I'm not using tailwindcss as we have to use h-[inherit] which is getting overridden by lots of our existing styles.

pranavrajs:
Closing as discussed, we will take an alternate route.

PR #7852, opened 2023-09-04
fix: Attachment sent from slack doesn't reach to chatwoot

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-2366/files-sent-from-slack-doesnt-reach-the-customer

https://www.loom.com/share/a606c17136284ec68e51a059508615e7

### DIFF
--- a/app/models/concerns/slack_message_creation.rb
+++ b/app/models/concerns/slack_message_creation.rb
@@ -10,11 +10,11 @@ module SlackMessageCreation
     @message.save!
     { status: 'success' }
   rescue Slack::Web::Api::Errors::MissingScope => e
-    handle_missing_scope_error(e)
+    ChatwootExceptionTracker.new(e, account: conversation.account).capture_exception
+    disable_and_reauthorize
   end
 
-  def handle_missing_scope_error(_error)
-    Rails.logger.error e
+  def disable_and_reauthorize
     integration_hook.prompt_reauthorization!
     integration_hook.disable
   end
--- a/app/models/concerns/slack_message_creation.rb
+++ b/app/models/concerns/slack_message_creation.rb
@@ -1,0 +1,69 @@
+module SlackMessageCreation
+  extend ActiveSupport::Concern
+
+  private
+
+  def create_message
+    return unless conversation
+
+    build_message
+    @message.save!
+    { status: 'success' }
+  rescue Slack::Web::Api::Errors::MissingScope => e
+    handle_missing_scope_error(e)
+  end
+
+  def handle_missing_scope_error(_error)
+    Rails.logger.error e
+    integration_hook.prompt_reauthorization!
+    integration_hook.disable
+  end
+
+  def build_message
+    @message = conversation.messages.build(
+      message_type: :outgoing,
+      account_id: conversation.account_id,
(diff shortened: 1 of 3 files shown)


app/models/concerns/slack_message_creation.rb
```diff
@@ -0,0 +1,69 @@
...
+    { status: 'success' }
+  rescue Slack::Web::Api::Errors::MissingScope => e
+    handle_missing_scope_error(e)
+  end
+
+  def handle_missing_scope_error(_error)
+    Rails.logger.error e
+    integration_hook.prompt_reauthorization!
+    integration_hook.disable
+  end
```
scmmishra, line 20: This shouldn't be handled like this, instead. Change this function to handle only disabling the hook

```suggestion
  def disable_and_reauthorize
    integration_hook.prompt_reauthorization!
    integration_hook.disable
  end
```
and

```
  def create_message
    return unless conversation

    build_message
    @message.save!
    { status: 'success' }
  rescue Slack::Web::Api::Errors::MissingScope => e
    Rails.logger.error e
    disable_and_reauthorize
  end
```

scmmishra: Also, we can use `ChatwootExceptionTracker`

muhsin-k: @scmmishra Done

app/models/concerns/slack_message_creation.rb
```diff
@@ -0,0 +1,69 @@
+module SlackMessageCreation
+  extend ActiveSupport::Concern
+
+  private
+
+  def create_message
```
scmmishra, line 6: @sojan-official is it okay for a method that is supposed to be an endpoint be in a concern?

pranavrajs: @scmmishra This is not an endpoint. It is used in the background job.

pranavrajs: However, I believe this concern is not written is great way. I will merge this so that the bug is fixed. Please re-consider the implmentation @muhsin-k

pranavrajs (approved):

app/models/concerns/slack_message_creation.rb
muhsin-k: Sure. I will pair with @sojan-official  and create separate PR.

sojan-official: @muhsin-k  models/concerns is suited for placing methods we share across different models. 
Example: `avatarable`, `reauthorizable` etc. Since we can share this code across multiple models. 

In this particular case, the logic placed here is tied very tightly to slack message builder. 
We can place this module alongside incoming_message_builde itself in the same folder .. as we are moving this method into a helper module specifically to satisfy the code climate requirement.  also we don't need to extend `activesupportconcern`

sojan-official: * another option is app/helpers . but since i don't forsee us reusing this anywhere. lets keep it alondside the incoming message builder .. under integrations/slack namespace

muhsin-k: https://github.com/chatwoot/chatwoot/pull/7912

PR #7861, opened 2023-09-05
fix: processing inline images does not save all images

scmmishra:


scmmishra:
@pranavrajs added some tests for the inline images

pranavrajs (approved):

PR #7863, opened 2023-09-05
chore: Improve the behaviour of lock to single conversation

sojan-official:
When `lock to single conversation` is disabled. Incoming messages will start a new conversation instead of reopening the existing resolved ones.  This seems more inline with the behaviour of outgoing conversations. 

Fixes: https://linear.app/chatwoot/issue/CW-2442/bug-lock-to-single-conversation-for-whatsapp-is-not-wokring


sojan-official:
@pranavrajs  Just wanted to ensure the behaviour is inline with the expectations.

pranavrajs (approved):

PR #7866, opened 2023-09-06
feat: update rack attack ip limit

scmmishra:
This PR updates rack attach per-ip limit. Also, allows fetching the value from the ENV


scmmishra:
@pranavrajs @sojan-official should I implement a redis based safelist too?

config/initializers/rack_attack.rb
```diff
@@ -46,7 +46,7 @@ def path_without_extentions
   #
   # Key: "rack::attack:#{Time.now.to_i/:period}:req/ip:#{req.ip}"
 
-  throttle('req/ip', limit: 300, period: 1.minute, &:ip)
+  throttle('req/ip', limit: ENV.fetch('RACK_ATTACK_LIMIT', 300), period: 1.minute, &:ip)
```
pranavrajs, line 49: Let us use 3000 here

pranavrajs (changes requested):
One minor comment

pranavrajs:
> @pranavrajs @sojan-official should I implement a redis based safelist too?

Let us start with this. If it becomes a problem, let us implement it.

pranavrajs (approved):

PR #7875, opened 2023-09-07
fix: Integration page responsiveness issues

muhsin-k:
**Before**

CleanShot 2023-09-07 at 13 32 46@2x

**After**

CleanShot 2023-09-07 at 13 32 08@2x


iamsivin (approved):
LGTM

PR #7899, opened 2023-09-12
chore: improve the behaviour of lock to single conversation

sojan-official:
the previous PR #7863 was reverted in #7886 because of the cases where the conversation could be in status like `pending` and `snoozed`. This PR covers those conditions and adds more tests to future-proof the changes.


pranavrajs:
Before merging this PR, we need to consider that lock_to_single_conversation is false by default. This means that new threads will be created in every account, which is not the previous behavior. Have we taken this into account? Should we handle this?

sojan-official:
@pranavrajs  lets sync up on this and move this forward

pranavrajs (approved):

PR #7901, opened 2023-09-12
fix: Erase null bytes from incoming Twilio messages

jordan-brough:
We've had some messages come in from a few different phone numbers that had null bytes in them. I don't know how this happens. They don't seem to be malicious.

They currently cause the Postgres gem to raise an error when Chatwoot attempts to save the message body to the database:
> ArgumentError (string contains null byte)

Related Rails GitHub issue: https://github.com/rails/rails/issues/26891

### DIFF
--- a/app/services/twilio/incoming_message_service.rb
+++ b/app/services/twilio/incoming_message_service.rb
@@ -9,7 +9,7 @@ class Twilio::IncomingMessageService
     set_contact
     set_conversation
     @message = @conversation.messages.create!(
-      content: params[:Body],
+      content: message_body,
       account_id: @inbox.account_id,
       inbox_id: @inbox.id,
       message_type: :incoming,
@@ -44,6 +44,10 @@ class Twilio::IncomingMessageService
 
   def formatted_phone_number
     TelephoneNumber.parse(phone_number).international_number
+  end
+
+  def message_body
+    params[:Body].delete("\u0000")
   end
 
   def set_contact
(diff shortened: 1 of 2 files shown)


pranavrajs:
@jordan-brough Thanks for the PR. Can you please share one such examples? Any specific idea that this message is received?

jordan-brough:
Hi Pranav, sure. Here's one example we got (with some info redacted). This was a customer replying with some information requested:

```ruby
>> twilio_client.messages('SMXXX').fetch.body
=> "Danny Xxx bar code XXXXX. Phone #XXX-XXX-XXXX  email xx.yyy\u0000live.com. Thank you"
```
In this case, it kind of looks like the null byte somehow got in there instead of an `@` sign.  But that has not always been the case.

Another example:

```ruby
>>  twilio_client.messages('SMXXX').fetch.body
=> "MLA format and in-text citations \u0000 Quizlet\nhttps://quizlet.com/4..."
```
in that case I'm not sure what it's doing in there.

I'm not sure how these get into the messages from people, but it's definitely coming that way from Twilio and in Chatwoot it raises an exception and causes the message to be missed.

app/services/twilio/incoming_message_service.rb
```diff
@@ -46,6 +46,10 @@ def formatted_phone_number
     TelephoneNumber.parse(phone_number).international_number
   end
 
+  def message_body
+    params[:Body].delete("\u0000")
```
muhsin-k, line 50: @jordan-brough This will throw an error if the body doesn't exist in the param. This [PR](https://github.com/chatwoot/chatwoot/pull/7910) will fix the issue.

PR #7902, opened 2023-09-12
feat: update facebook api version [feature/cw-2482]

scmmishra:
Update facebook API version


scmmishra:
@sojan-official in the codebase, I found whatsapp uses v14 and v13 in different place. Do they have any significance?

sojan-official:
@scmmishra There was a payload change at some point. so we need to see whether Chatwoot is compatible with the change

sojan-official (approved):

scmmishra:
@pranavrajs did we change the python version on prod again? In that case, do you think we should merge this?

PR #7910, opened 2023-09-13
fix: Check body exists before deleting null bytes in Twilio

muhsin-k:
This PR will add the fix for checking body exist or not before deleting the null bytes in Twilio payload.

Reference https://github.com/chatwoot/chatwoot/pull/7901


sojan-official (approved):

jordan-brough:
@muhsin-k thanks!

PR #7914, opened 2023-09-13
feat: add upload under account scope

scmmishra:
This PR makes the following changes

1. Move upload endpoint under account scope
2. Update the frontend helper to use account scope, and fetch account ID from the route
3. Update specs


scmmishra:
@sojan-official this is good to go, PTAL

sojan-official (approved):

PR #7921, opened 2023-09-15
feat: Slack link unfurling

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-2369/render-private-information-in-the-metadata-if-slack-app-is-connected

## Configuring your app 

To support custom app unfurling, your app needs to have the permissions to listen to links posted into Slack and unfurl those links.

### Scopes and token types 

Navigate to your [app configuration](https://api.slack.com/apps) and choose the OAuth & Permissions sidebar to select scopes.

Add the [links:read](https://api.slack.com/scopes/links:read) & [links:write](https://api.slack.com/scopes/links:write) scopes under **Bot Token Scopes*.

### Events 

Navigate to the Event Subscriptions sidebar.

Under Subscribe to bot events, add the [link_shared event](https://api.slack.com/events/link_shared) to receive information about links posted into Slack that are associated with your app.

The [link_shared](https://api.slack.com/events/link_shared) event will dispatch for all channels in each workspace it is installed in.

### Register your domain(s) 

On the same Event Subscriptions page, click App unfurl domains near the bottom of the page. Your app can register up to five domains, and will only receive [link_shared](https://api.slack.com/events/link_shared) events for URLs that match your registered domains.

More details can be found [here](https://api.slack.com/reference/messaging/link-unfurling#setup).


config/initializers/monkey_patches/chat.rb
```diff
@@ -0,0 +1,21 @@
+module Slack
```
muhsin-k, line 1: There is an issue with link [unfurling](https://api.slack.com/reference/messaging/link-unfurling#attaching_content) that requires a combination of `ts` and `channel`, or `unfurl_id` and `source`. The slack-ruby-client library has added validation to ensure that 'channel' and 'ts' are mandatory for link unfurling, as seen [here](https://github.com/slack-ruby/slack-ruby-client/blob/506376c8868561a9b2d936d12d6533da39985fdb/lib/slack/web/api/endpoints/chat.rb#L239): . However, we need to update the validation in the gem itself to make it work properly. The current version of the Ruby client is 2.0.0, but the latest version is 2.1.0. We will fork it, update it, and raise an upstream PR to make it live on the actual gem after the upgrade.

PR #7938, opened 2023-09-19
chore: Migration status in super admin

sojan-official:
Pending migrations are often cause for issues in Chatwoot self-hosted installation. This feature should help to identify such issues easily

Screenshot 2023-09-18 at 8 32 00 PM


vishnu-narayanan (approved):
This will be good to have 🙇‍♂️ 💯

PR #7940, opened 2023-09-19
feat: Slack link unfurling

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-2369/render-private-information-in-the-metadata-if-slack-app-is-connected

## Configuring your app 

To support custom app unfurling, your app needs to have the permissions to listen to links posted into Slack and unfurl those links.

### Scopes and token types 

Navigate to your [app configuration](https://api.slack.com/apps) and choose the OAuth & Permissions sidebar to select scopes.

Add the [links:read](https://api.slack.com/scopes/links:read) & [links:write](https://api.slack.com/scopes/links:write) scopes under **Bot Token Scopes*.

### Events 

Navigate to the Event Subscriptions sidebar.

Under Subscribe to bot events, add the [link_shared event](https://api.slack.com/events/link_shared) to receive information about links posted into Slack that are associated with your app.

The [link_shared](https://api.slack.com/events/link_shared) event will dispatch for all channels in each workspace it is installed in.

### Register your domain(s) 

On the same Event Subscriptions page, click App unfurl domains near the bottom of the page. Your app can register up to five domains, and will only receive [link_shared](https://api.slack.com/events/link_shared) events for URLs that match your registered domains.

More details can be found [here](https://api.slack.com/reference/messaging/link-unfurling#setup).

### DIFF
--- a/config/initializers/monkey_patches/chat.rb
+++ b/config/initializers/monkey_patches/chat.rb
@@ -1,0 +1,21 @@
+module Slack
+  module Web
+    module Api
+      module Endpoints
+        module Chat
+          # TODO: Remove this monkey patch when PR for this issue https://github.com/slack-ruby/slack-ruby-client/issues/388 is merged
+          def chat_unfurl(options = {})
+            if (options[:channel].nil? || options[:ts].nil?) && (options[:unfurl_id].nil? || options[:source].nil?)
+              raise ArgumentError, 'Either a combination of :channel and :ts or :unfurl_id and :source is required'
+            end
+
+            raise ArgumentError, 'Required arguments :unfurls missing' if options[:unfurls].nil?
+
+            options = options.merge(channel: conversations_id(options)['channel']['id']) if options[:channel]
+            post('chat.unfurl', options)
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/integrations/slack/incoming_message_builder.rb
+++ b/lib/integrations/slack/incoming_message_builder.rb
@@ -1,9 +1,11 @@
 class Integrations::Slack::IncomingMessageBuilder
   include Integrations::Slack::SlackMessageHelper
+  include Integrations::Slack::LinkUnfurlHelper
+
   attr_reader :params
 
   SUPPORTED_EVENT_TYPES = %w[event_callback url_verification].freeze
-  SUPPORTED_EVENTS = %w[message].freeze
+  SUPPORTED_EVENTS = %w[message link_shared].freeze
   SUPPORTED_MESSAGE_TYPES = %w[rich_text].freeze
 
   def initialize(params)
@@ -17,6 +19,8 @@ class Integrations::Slack::IncomingMessageBuilder
       verify_hook
     elsif create_message?
       create_message
+    elsif link_shared?
+      create_link_shared_message
     end
   end
 
@@ -68,6 +72,55 @@ class Integrations::Slack::IncomingMessageBuilder
     thread_timestamp_available? && supported_message? && integration_hook
   end
 
+  def link_shared?
+    params[:event][:type] == 'link_shared'
+  end
+
+  def create_link_shared_message
+    conversation = conversation_from_params
+    return unless conversation
+
+    user_info = conntact_attributes(conversation).slice(:user_name, :email, :phone_number, :company_name)
+    unfurls = generate_unfurls(conversation_url, user_info)
+
+    send_unfurls(unfurls)
+  end
+
+  def conntact_attributes(conversation)
+    contact = conversation.contact
+    user_name = contact&.name.presence || '---'
+    email = contact&.email.presence || '---'
+    phone_number = contact&.phone_number.presence || '---'
+    company_name = contact&.additional_attributes&.dig('company_name').presence || '---'
+
+    {
+      user_name: user_name,
+      email: email,
+      phone_number: phone_number,
+      company_name: company_name
+    }
+  end
+
+  def conversation_from_params
+    conversation_id = extract_conversation_id(conversation_url)
+    return unless conversation_id
+
+    find_conversation_by_id(conversation_id)
+  end
+
+  def conversation_url
+    params.dig(:event, :links, 0, :url)
+  end
+
+  def send_unfurls(unfurls)
+    slack_service = Integrations::Slack::SendOnSlackService.new(message: nil, hook: integration_hook)
+    slack_service.link_unfurl(
+      unfurl_id: params.dig(:event, :unfurl_id),
+      source: params.dig(:event, :source),
+      unfurls: JSON.generate(unfurls)
+    )
+  end
+
   def message
     params[:event][:blocks]&.first
   end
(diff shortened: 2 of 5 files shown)


config/initializers/monkey_patches/chat.rb
```diff
@@ -0,0 +1,21 @@
+module Slack
```
muhsin-k, line 1: There is an issue with link [unfurling](https://api.slack.com/reference/messaging/link-unfurling#attaching_content) that requires a combination of `ts` and `channel`, or `unfurl_id` and `source`. The slack-ruby-client library has added validation to ensure that 'channel' and 'ts' are mandatory for link unfurling, as seen [here](https://github.com/slack-ruby/slack-ruby-client/blob/506376c8868561a9b2d936d12d6533da39985fdb/lib/slack/web/api/endpoints/chat.rb#L239): . However, we need to update the validation in the gem itself to make it work properly. The current version of the Ruby client is 2.0.0, but the latest version is 2.1.0. We will fork it, update it, and raise an upstream PR to make it live on the actual gem after the upgrade.

lib/integrations/slack/link_unfurl_helper.rb
```diff
@@ -0,0 +1,53 @@
+module Integrations::Slack::LinkUnfurlHelper
+  def generate_unfurls(url, user_info)
```
muhsin-k, line 2: This method will generate blocks like [this](https://app.slack.com/block-kit-builder/T7NN675QT#%7B%22blocks%22:%5B%7B%22type%22:%22section%22,%22fields%22:%5B%7B%22type%22:%22mrkdwn%22,%22text%22:%22*Name:*%5CnCandice%20Matherson%22%7D,%7B%22type%22:%22mrkdwn%22,%22text%22:%22*Email:*%5Cncandice@example.com%22%7D,%7B%22type%22:%22mrkdwn%22,%22text%22:%22*Phone:*%5Cn974237862323%22%7D,%7B%22type%22:%22mrkdwn%22,%22text%22:%22*Company:*%5CnChatwoot%22%7D%5D%7D,%7B%22type%22:%22actions%22,%22elements%22:%5B%7B%22type%22:%22button%22,%22text%22:%7B%22type%22:%22plain_text%22,%22text%22:%22Open%20conversation%22,%22emoji%22:true%7D,%22value%22:%22https://d738-2406-7400-81-3754-8520-3004-f0a1-6c23.ngrok-free.app/app/accounts/1/conversations/100%22,%22action_id%22:%22button-action%22%7D%5D%7D%5D%7D).
CleanShot 2023-09-19 at 12 24 03@2x

pranavrajs:
@muhsin-k

1/ Does it affect any of the existing apps? Does it break anything or is the permission just an add-on? 

2/ Can we standardize the preview and the slack initialization message? Both have the same content, however it seems to be using a different layout. Having the same layout would be ideal.

muhsin-k:
> 1/ Does it affect any of the existing apps? Does it break anything or is the permission just an add-on?
>
Yes. The bot permission is just add-on. The Unfurling will not work if the scopes and URL are not added. 

> 2/ Can we standardize the preview and the slack initialization message? Both have the same content, however it seems to be using a different layout. Having the same layout would be ideal.

Yes. We can standardize like this.

CleanShot 2023-09-27 at 20 19 30@2x

config/initializers/monkey_patches/chat.rb
```diff
@@ -0,0 +1,21 @@
...
+  module Web
+    module Api
+      module Endpoints
+        module Chat
+          # TODO: Remove this monkey patch when PR for this issue https://github.com/slack-ruby/slack-ruby-client/issues/388 is merged
```
pranavrajs, line 6: Can we raise a PR to the original repo?

lib/integrations/slack/slack_link_unfurl_service.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def extract_conversation_id(url)
+    conversation_id_regex = %r{/conversations/(\d+)}
+    match_data = url.match(conversation_id_regex)
+    match_data[1] if match_data
+  end
+
+  def find_conversation_by_id(conversation_id)
+    Conversation.find_by(display_id: conversation_id)
```
pranavrajs, line 66: The display_id is scoped to an account. This query would return random results. You would have to find the account from the URL and find the conversation based on the URL.

pranavrajs: Additional a check has to be added to make sure that the account has a Slack integration and the workspace ID is same as the callback URL to prevent data leaks.

pranavrajs (changes requested):
Please see the individual specs.

lib/integrations/slack/incoming_message_builder.rb
```diff
@@ -17,6 +18,8 @@ def perform
       verify_hook
     elsif create_message?
       create_message
+    elsif link_shared?
+      Integrations::Slack::SlackLinkUnfurlService.new(params: params, integration_hook: integration_hook).perform
```
pranavrajs, line 22: A suggestion here would be to make it a job rather than running it in synchronously in the webhook callback itself.

lib/integrations/slack/link_unfurl_formatter.rb
```diff
@@ -0,0 +1,59 @@
...
+  def open_conversation_button(url)
+    [
+      {
+        'type' => 'actions',
+        'elements' => [
+          {
+            'type' => 'button',
+            'text' => {
+              'type' => 'plain_text',
+              'text' => 'Open conversation',
```
pranavrajs, line 49: This can be added as an I18n text. Not a blocker though.

lib/integrations/slack/slack_link_unfurl_service.rb
pranavrajs: Account 1 -> Slack Hook 1 

If I share a link from Account 2, it shouldn't render the details as the workspace doesn't have access to this data.

config/initializers/monkey_patches/chat.rb
muhsin-k: Current version of the Ruby slack client in chatwoot is 2.0.0, but the latest version is 2.1.0. I will do this after the upgrade.

lib/integrations/slack/slack_link_unfurl_service.rb
muhsin-k: Fixed.

app/controllers/api/v1/integrations/webhooks_controller.rb
```diff
@@ -1,7 +1,15 @@
...
   def create
-    builder = Integrations::Slack::IncomingMessageBuilder.new(params)
+    builder = Integrations::Slack::IncomingMessageBuilder.new(permitted_params)
     response = builder.perform
     render json: response
   end
+
+  private
+
+  # TODO: Fix me later, this is a temporary solution to permit all params for slack unfurling job.
```
pranavrajs, line 10:
```suggestion
  # TODO: This is a temporary solution to permit all params for slack unfurling job.
```

app/controllers/api/v1/integrations/webhooks_controller.rb
```diff
@@ -1,7 +1,15 @@
...
-    builder = Integrations::Slack::IncomingMessageBuilder.new(params)
+    builder = Integrations::Slack::IncomingMessageBuilder.new(permitted_params)
     response = builder.perform
     render json: response
   end
+
+  private
+
+  # TODO: This is a temporary solution to permit all params for slack unfurling job.
+  # We should only permit the params that we need  Handle all the params based on events and send it to the respective services
```
pranavrajs, line 11:
```suggestion
  # We should only permit the params that we use. Handle all the params based on events and send it to the respective services.
```

pranavrajs (approved):

PR #7944, opened 2023-09-20
fix: handling markdown before replacing or appending signature [CW-2532]

scmmishra:
This PR fixes an issue where a signature would not be replaced properly in the editor.  The problem is that when a signature reached our editor, which looks like the following `- Jane Doe`, it is fed to the editor, which converts it to commonmark, which looks like the following `* Jane Doe` 

The problem arises when we try to look for the signature, since the output and input don't match, the program concludes that there is no signature in the body, and we simply append it again. On every mount this happens and here we have it, infinite signature glitch

## Type of change

### DIFF
--- a/app/javascript/dashboard/components/widgets/WootWriter/FullEditor.vue
+++ b/app/javascript/dashboard/components/widgets/WootWriter/FullEditor.vue
@@ -218,7 +218,7 @@ export default {
 </script>
 
 <style lang="scss">
-@import '@chatwoot/prosemirror-schema/src/styles/article.scss';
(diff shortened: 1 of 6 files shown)


app/javascript/dashboard/components/widgets/WootWriter/FullEditor.vue
```diff
@@ -218,7 +218,7 @@ export default {
 </script>
 
 <style lang="scss">
-@import '~@chatwoot/prosemirror-schema/src/styles/article.scss';
+@import '@chatwoot/prosemirror-schema/src/styles/article.scss';
```
pranavrajs, line 221: Is this intentional?

PR #7995, opened 2023-09-27
feat: Response management in super admin

sojan-official:
- Introduce dashboards for response management in super admin for quality evaluation


pranavrajs:
@sojan-official Added a fix for buttons overlapping in the small screen.
Screenshot 2023-09-27 at 6 15 23 PM

PR #8002, opened 2023-09-27
fix: Use installation locale if configured

### LINKED ISSUE #7894: [CW-2487] DEFAULT_LOCALE=pt_BR in the environment variables.
@jtayrone This is for the account. If you want to change the installation language, please use set DEFAULT_LOCALE=pt_BR in the environment variables.

_Originally posted by @pranavrajs in https://github.com/chatwoot/chatwoot/issues/7892#issuecomment-1713781331_
            

<sub>[CW-2487](https://linear.app/chatwoot/issue/CW-2487/default-locale=pt-br-in-the-environment-variables)</sub>

pranavrajs:
Fixes https://linear.app/chatwoot/issue/CW-2487/default-locale=pt-br-in-the-environment-variables 
Fixes #7894 

--- 

To test:

- Set DEFAULT_LOCALE = pt_BR in .env file
- Restart the application and see if the login / signup / reset password / set new password pages have the locale changed.

### DIFF
--- a/app/javascript/v3/App.vue
+++ b/app/javascript/v3/App.vue
@@ -15,6 +15,7 @@ export default {
   mounted() {
     this.setColorTheme();
     this.listenToThemeChanges();
+    this.setLocale(window.chatwootConfig.selectedLocale);
   },
   methods: {
     setColorTheme() {
@@ -34,6 +35,9 @@ export default {
           this.theme = 'light';
         }
       };
+    },
+    setLocale(locale) {
+      this.$root.$i18n.locale = locale;
     },
   },
 };


app/javascript/v3/App.vue
```diff
@@ -15,6 +15,7 @@ export default {
   mounted() {
     this.setColorTheme();
     this.listenToThemeChanges();
+    this.setLocale(window.chatwootConfig.selectedLocale);
```
scmmishra, line 18: What happens if the `selectedLocale` is `undefined`? Do we have to worry about that? Also should we be worried about `chatwootConfig` being `undefined`

pranavrajs: It is never undefined. We take it from I18n.locale. default is en.

PR #8003, opened 2023-09-27
fix: Token type `hr` not supported by Markdown parser

scmmishra:
For the message schema we provide, `---` which is a horizontal rule element is not handled. This results in an error. This PR adds a exception catch to ensure that the signatures continues to work

To test, update the signature from the DB console to the following

```
---

Signature
```

Try using the editor after that, the dashes should be gone without error

### DIFF
--- a/app/javascript/dashboard/helper/editorHelper.js
+++ b/app/javascript/dashboard/helper/editorHelper.js
@@ -3,6 +3,7 @@ import {
   MessageMarkdownTransformer,
   MessageMarkdownSerializer,
 } from '@chatwoot/prosemirror-schema';
+import * as Sentry from '@sentry/browser';
 
 window.messageSchema = messageSchema;
 window.MessageMarkdownTransformer = MessageMarkdownTransformer;
@@ -27,6 +28,7 @@ export function cleanSignature(signature) {
   } catch (e) {
     // eslint-disable-next-line no-console
     console.error(e);
+    Sentry.captureException(e);
     // The parser can break on some cases
     // for example, Token type `hr` not supported by Markdown parser
     return signature;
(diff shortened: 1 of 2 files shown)


scmmishra:
@nithindavid need your thoughts here, on what schema should I use for the Markdown parser

nithindavid:
@scmmishra Do we need the conversion  to common mark if we add serialiser which would remove/support the hr?

scmmishra:
> @scmmishra Do we need the conversion to common mark if we add serialiser which would remove/support the hr?

We had to do this in this PR: https://github.com/chatwoot/chatwoot/pull/7944
The problem was that signatures like `- Jack` would get converted to `* Jack` by the editor automatically after pushing it. So any search replace methods used to manage signatures stopped working. So we thought it's best to Sanitze it by the parsers & serializer used internally by the editor to convert it to a compliant common mark format

scmmishra:
@nithindavid this is ready for review

scmmishra:
@iamsivin @nithindavid let me know if this is good to go

iamsivin:
> @iamsivin @nithindavid let me know if this is good to go

@scmmishra Is it possible to add test cases for this PR?

scmmishra:
> > @iamsivin @nithindavid let me know if this is good to go
> 
> 
> 
> @scmmishra Is it possible to add test cases for this PR?

Added it to editorHelper.spec.js, anything in particular to have tested?

iamsivin:
> > > @iamsivin @nithindavid let me know if this is good to go
> > 
> > 
> > @scmmishra Is it possible to add test cases for this PR?
> 
> Added it to editorHelper.spec.js, anything in particular to have tested?

Not that, I mean the test case for testing the PR locally.

iamsivin (approved):
Code LGTM

scmmishra:
> > > > @iamsivin @nithindavid let me know if this is good to go
> 
> > > 
> 
> > > 
> 
> > > @scmmishra Is it possible to add test cases for this PR?
> 
> > 
> 
> > Added it to editorHelper.spec.js, anything in particular to have tested?
> 
> 
> 
> Not that, I mean the test case for testing the PR locally.

Done

Issue #6890, opened 2023-04-12
[CW-1931] Draft through API both for cross device support but also for automated drafting of conversation


VSinghhh:
**Is your enhancement request related to a problem? Please describe.**
With a high volume of conversations and a fairly high complexity of some of our questions we want our bots to 'draft' replies rather than actually directly sending the replies.

Chatwoot has suport for drafts but only through[ local storage:](https://github.com/orgs/chatwoot/discussions/5662)
https://www.chatwoot.com/blog/v2-10-0/
https://www.youtube.com/watch?v=NAtxmt2lKQY
https://github.com/chatwoot/chatwoot/issues/4876

**Describe the solution you'd like**
We want the following flow

1. Our platform analyses the email - and while we have a high ability to confidently identify intents of the incoming emails, it is not 100% right. For this reason we would like the incoming email to be analyzed by our backend and assign a label to the conversation. We understand that the Chatwoot API works for this
2. We would like our backend to instruct the reply (draft?) to be written- but that our agent is the one who actually clicks "send" to send it. Essentially an auto-canned /drafted reply. I didnt see any API for drafting a message, and @pranavrajs in the YouTube clip says that drafts dont work between devices (the draft is in the browser local store rather than saved in the BE/system) at the time of launch of the draft feature - Jan 2023; but that it was coming.

The purpose is that our platform based on the analysis of incoming email does most of the work. If it is wrong in identifying the intent, the agent can relabel the email. (over time we will retrain our platform to identify the incoming email correctly in future (based on new label))

Td;lr: We want our platform to 'suggest'/autofil a reply - and if it is correct, the agent just clicks "send"

**Describe alternatives you've considered**
None available as far as we've seen

**Additional context**
Add any other context or screenshots about the enhancement request here.

<sub>[CW-1931](https://linear.app/chatwoot/issue/CW-1931/draft-through-api-both-for-cross-device-support-but-also-for-automated)</sub>

pranavrajs:
@VSinghhh This seems similar to the reply suggestions feature that we are working on. I will add more details here later.

VSinghhh:
Hi,
Just want to check in how its going with this?
A difference could be that the 'reply suggestions' that you guys might be building (LLM or similar?) might be different from our own BE-logic where we have our own setup which we can handle using drafts: 

Should we built API/BE for drafts ourselves or wait to test your reply suggestions?

sojan-official:
partial work done in #8018

PR #8018, opened 2023-09-29
feat: Draft through API for Cross-Device Support

jderecho:
# Pull Request Template

## Description

Fixes https://github.com/chatwoot/chatwoot/issues/6890

## Type of change

## Checklist:


app/controllers/api/v1/accounts/conversations_controller.rb
```diff
@@ -80,6 +80,12 @@ def toggle_typing_status
     head :ok
   end
 
+  def update_draft_message
+    @conversation.draft_message_content = params.permit(:message)[:message]
```
pranavrajs, line 84: Saving it as an attribute to conversation is not ideal. There are multiple people looking at the same conversation. Showing a draft saved by one of the users to another user is not correct.

pranavrajs, line 84: 1/ Saving the drafts in a conversation is not ideal. As there are multiple users looking at the same conversation, exposing a draft written by one user to another user is not a great experience. 

2/ The draft message is short-living item, it shouldn't be stored in the database instead it can be saved in Redis storage with key to identify the user, the account and the conversation. The key value pair can be deleted once the message is sent.

pranavrajs, line 84: 1/ Saving the drafts in a conversation is not ideal. As there are multiple users looking at the same conversation, exposing a draft written by one user to another user is not a great experience.

2/ The draft message is short-living item, it shouldn't be stored in the database instead it can be saved in Redis storage with key to identify the user, the account and the conversation. The key value pair can be deleted once the message is sent.

app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
```diff
@@ -629,7 +630,7 @@ export default {
       if (this.conversationIdByRoute) {
         const key = `draft-${this.conversationIdByRoute}-${this.replyType}`;
         const messageFromStore =
-          this.$store.getters['draftMessages/get'](key) || '';
+          this.$store.getters['draftMessages/get'](key) || this.currentChat.draft_message_content || '';
```
pranavrajs, line 633: Same comment as above. We should not be polluting the conversation namespace rather consider the draftMessage as an entity by itself.

app/controllers/api/v1/accounts/conversations_controller.rb
pranavrajs: 3/ Let us consider draftMessages as a separate entity rather than tying it to conversation. This would give them a separate namespace for APIS. `/accounts/<account-id>/conversations/<conversation-id>/draft-messages` or something similar would work.

app/javascript/dashboard/api/inbox/conversation.js
```diff
@@ -112,6 +112,12 @@ class ConversationApi extends ApiClient {
     return axios.post(`${this.url}/${conversationId}/transcript`, { email });
   }
 
+  updateDraftMessage({ conversationId, message }) {
```
pranavrajs, line 115: As mentioned above, let us move draftMessages from the conversation namespace to its own namespace.

app/javascript/dashboard/store/modules/conversations/actions.js
```diff
@@ -405,6 +405,21 @@ const actions = {
     }
   },
 
+  updateDraftMessage: async (
```
pranavrajs, line 408: We have a store called draftMessages. Let us use that. 

https://github.com/chatwoot/chatwoot/blob/develop/app/javascript/dashboard/store/modules/draftMessages.js

pranavrajs (changes requested):
Thanks for the PR. The implementation needs a bit of work. Let me know if you have any questions.

jderecho:
hi @pranavrajs I refactored the code to use Redis caching, let me know if I implemented it correctly and if there are other things to consider. Thanks

pranavrajs:
@jderecho Thanks for addressing the comments. On a quick look, this looks fine. Let me test it out properly and complete the review.

sojan-official:
Note: We will be decoupling the frontend changes in this PR separately.

sojan-official (approved):

sojan-official:
Note: Follow-up enhancement -> add user ids to the drafts to differentiate drafts between users

PR #8048, opened 2023-10-04
feat: Ability to assign to self in macros

sojan-official:
Add options 'self' to assign agent drop-down in macros, which allows creating macros which will assign the conversation to the executing user.


app/javascript/dashboard/mixins/macrosMixin.js
```diff
@@ -7,7 +7,7 @@ export default {
         case 'send_email_to_team':
           return this.teams;
         case 'assign_agent':
-          return this.agents;
+          return [{ id: 'self', name: 'Self' }, ...this.agents];
```
muhsin-k, line 10: i18n

muhsin-k (approved):

PR #8050, opened 2023-10-04
feat: setup store for "Reply To" feature

scmmishra:
This PR sets up stores for the "Reply to" feature. It is very similar to how we save draft messages


scmmishra:
Closing this, gonna try a new approach

PR #8076, opened 2023-10-10
feat: support reply to for incoming messages on facebook

scmmishra:
This PR enables support for "Reply To" on incoming messages on Facebook. This has the following changes

1. Add `ensure_in_reply_to` to ensure that both `in_reply_to_external_id` and `in_reply_to` are set, if one of the values is present
2. Update facebook message parser, and message builder to handle `in_reply_to`

CleanShot 2023-10-13 at 10 16 50

P.S. Facebook only supports reply-to for incoming messages, not for outgoing messages

### How to test

1. Send a message "A" from Facebook to Chatwoot 
2. Send message "B" from Chatwoot to Facebook on that conversation
3. Reply to message "B" with a new message from Facebook, the UI should show up on Chatwoot
4. Reply to the original message "A" with a new message from Facebook again, the UI should show up on Chatwoot


app/models/message.rb
```diff
@@ -233,6 +234,19 @@ def ensure_processed_message_content
     self.processed_message_content = message_content&.truncate(150_000)
   end
 
+  # fetch the in_reply_to message and set the external id
+  def ensure_in_reply_to_external_id
+    in_reply_to = content_attributes['in_reply_to']
```
pranavrajs, line 239:
```suggestion
    in_reply_to = content_attributes&.['in_reply_to']
```

scmmishra: @pranavrajs had to revert this, the linter was complaining

CleanShot 2023-10-12 at 22 04 16

app/models/message.rb
```diff
@@ -233,6 +234,20 @@ def ensure_processed_message_content
     self.processed_message_content = message_content&.truncate(150_000)
   end
 
+  # fetch the in_reply_to message and set the external id
+  def ensure_in_reply_to
```
pranavrajs, line 238: Can we make this a service and use it everywhere else? The attributes and the actions are common, right? 

1/ Get message by source_id
2/ If present, populate in_reply_to and in_reply_to_external_id

scmmishra: You mean, moving this out as a concern?

muhsin-k:
@scmmishra Can we display thumbnails for attachment replies? It can be confusing to identify which attachment the user is referring to when there are multiple attachments. 
CleanShot 2023-10-13 at 11 41 47@2x

muhsin-k (approved):
Verified both text and attachment messages for incoming message and it is working fine.

scmmishra:
@pranavrajs can you do a code review

app/models/message.rb
scmmishra: Saw the changes https://github.com/chatwoot/chatwoot/issues/6964, please let me know if you want to pair up on externalizing this?

PR #8077, opened 2023-10-10
fix: Update duplicate ids for dashboard app frame

pranavrajs:
Fixes https://linear.app/chatwoot/issue/CW-2639/dashboard-app-issues

The dashboard app IDs were generated incorrectly. The query getElementById returns the first object all the time, irrespective of the element from which it was queried which results in the missing postMessage calls.


scmmishra (approved):
LGTM!

PR #8085, opened 2023-10-11
chore: Ensure null validation for private attribute in messages

sojan-official:
- Ensure null validation for private attribute in messages


pranavrajs (approved):

sojan-official:
Note: ran and marked migration on prod manually.

PR #8089, opened 2023-10-11
feat: Updated public portal header design

iamsivin:
# Pull Request Template

## Description

Fixes https://linear.app/chatwoot/issue/CW-2615/implement-the-header-navbar-for-the-public-portal

## Type of change

## How Has This Been Tested?

**Loom video**
https://www.loom.com/share/175a965279d6485c9b5d0d38401fd701?sid=078a95c4-e2d1-4037-b526-1fbfd68778e8

## Checklist:


app/views/public/api/v1/portals/_header.html.erb
```diff
@@ -8,34 +7,161 @@
...
-            <option <%= locale == params[:locale] ? 'selected': '' %> value="<%= locale %>"><%= locale %></option>
-          <% end %>
-        </select>
-        <div class="pointer-events-none absolute inset-y-0 right-0 flex items-center px-2 text-slate-700 dark:text-slate-200">
-          <svg class="fill-current h-4 w-4" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20"><path d="M9.293 12.95l.707.707L15.657 8l-1.414-1.414L10 10.828 5.757 6.586 4.343 8z"/></svg>
+
+      <%# Appearance toggle section %>
+      <div class="flex-grow flex-shrink-0 relative cursor-pointer px-1 py-2">
+        <button class="toggle-appearance flex flex-row items-center stroke-slate-700 dark:stroke-slate-200 hover:stroke-woot-500 dark:hover:stroke-woot-500 text-slate-800 dark:text-slate-100 hover:text-woot-500 dark:hover:text-woot-500 gap-1" type="button" onclick="toggleAppearanceDropdown()">
+          <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
pranavrajs, line 32: Can we figure out something else here? The usage of SVG like this is a bit off-putting. Let me know.

iamsivin: @pranavrajs Done https://github.com/chatwoot/chatwoot/pull/8089/commits/2eff4e9b470579eb83e08d8466dccb7422179df3

app/helpers/portal_helper.rb
```diff
@@ -8,4 +8,19 @@ def generate_portal_bg(portal_color, theme)
     bg_image = theme == 'dark' ? 'grid_dark.svg' : 'grid.svg'
     "background: url(/assets/images/hc/#{bg_image}) #{generate_portal_bg_color(portal_color, theme)}"
   end
+
+  def language_name(locale)
+    language_map = YAML.load_file(Rails.root.join('config/languages/language_map.yml'))
```
scmmishra, line 13: This loads the yaml file with every function call, this could be in-efficient. 

You can declare a module variable at the top level

```rb
module PortalHelper
    @@language_map = nil
  
    # rest of the code
end
```

and update this method

```rb
def language_name(locale)
  @@language_map ||= YAML.load_file(Rails.root.join('config/languages/language_map.yml'))
  @@language_map[locale] || locale
end
```

This should effectively memoize this, as far as my understanding goes, this should persist across network requests in production. Not sure if this is a good pattern or not. CC: @sojan-official @pranavrajs

app/views/public/api/v1/portals/_header.html.erb
```diff
@@ -8,34 +7,111 @@
...
   </nav>
 </header>
+<script>
+  const toggleAppearanceDropdown = () => {
+    const dropdown = document.getElementById('appearance-dropdown');
+    dropdown.style.display = (dropdown.style.display === 'none' || !dropdown.style.display) ? 'flex' : 'none';
+  };
+
+  const switchTheme = (theme) => {
+    const bodyDiv = document.querySelector('.antialiased');
```
scmmishra, line 95: If this is the body we are targeting, we can simply use `document.body`

```suggestion
    const bodyDiv = document.body;
```

app/views/public/api/v1/portals/_header.html.erb
```diff
@@ -8,34 +7,111 @@
...
+    bodyDiv.classList.remove('light', 'dark');
+    if (theme !== 'system') {
+      bodyDiv.classList.add(theme);
+      window.location.href = updateURLParameter(window.location.href, 'theme', theme);
```
scmmishra, line 99: Any reason to update the URL? Do we want a reload to happen when the theme changes? Can we do it without that?

config/languages/language_map.yml
scmmishra, line 1: Let's see if there is a Rails Gem for this?

scmmishra (changes requested):

app/views/public/api/v1/portals/_header.html.erb
iamsivin: Yes, we need to reload the header section with SVG is not getting updated when we change the class.

config/languages/language_map.yml
iamsivin: I couldn't find one with all the languages we have in our locale constant file.

app/views/public/api/v1/portals/_header.html.erb
iamsivin: Renamed the bodyDiv to parentDiv. We are adding the theme class inside the body.

app/helpers/portal_helper.rb
iamsivin: @scmmishra Any idea 
image

scmmishra: 🥲 @sojan-official @pranavrajs need your Ruby skills here.

@iamsivin feel free to revert that for now.

app/views/public/api/v1/portals/_header.html.erb
scmmishra: In that case, let's use a `id` or a better class name, which is more specific instead of `antialiased`.

config/languages/language_map.yml
scmmishra: No worries, resolving this

app/views/public/api/v1/portals/_header.html.erb
iamsivin: DOne

app/helpers/portal_helper.rb
iamsivin: Reverted it

muhsin-k (approved):

PR #8092, opened 2023-10-12
feat: Add delivery reports for live chat

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-2646/delivery-reports-for-live-chat


app/javascript/dashboard/components/widgets/conversation/bubble/Actions.vue
```diff
@@ -204,6 +204,9 @@ export default {
       if (this.isAWhatsAppChannel || this.isATwilioChannel) {
         return this.sourceId && this.isSent;
       }
+      if (this.isAWebWidgetInbox) {
```
pranavrajs, line 207: This will never be executed, please revert.

pranavrajs (changes requested):
Nice work! Please find the inline comments.

pranavrajs (approved):

PR #8105, opened 2023-10-13
feat: support reply to for Telegram

scmmishra:
This PR enables reply to for Telegram

## Changes

1. Update `message_builder.rb` and `message.js` to accept `content_attributes` json in the FormData
2. Update telegram `incoming_message_service.rb` to process `reply_to_message` from the webhook
3. Update `app/models/channel/telegram.rb` to include `reply_to_message_id` in the payload that is sent to Telegram

## Preview

|  Dashboard  |  Telegram  |
|:---:|:---:|
| CleanShot 2023-10-18 at 18 16 06@2x | CleanShot 2023-10-18 at 18 17 09@2x |
|:---:|:---:|


sojan-official (commented):
Looks good overall. Can we have a test case where a valid reply to message-id is present for both message request and attachment message request

scmmishra:
@sojan-official Added tests to ensure that the `content_attributes` are parsed correctly regardless of it's format. 

> Can we have a test case where a valid reply to message-id is present for both message request and attachment message request

You mean a frontend test for this?

scmmishra:
@sojan-official let me know if this is good go

muhsin-k (approved):
Verified in local environment.

CleanShot 2023-10-19 at 12 58 19

IMG_0216

sojan-official (approved):

PR #8107, opened 2023-10-13
feat: support reply to for outgoing message in WhatsApp

scmmishra:
This PR enables reply to for WhatsApp

## Changes

1. Move the `in_reply_to` and `in_reply_to_external_id` to a separate service `Messages::InReplyToMessageBuilder` and use it in `models/message.rb`
2. Remove the code that sets the `in_reply_to` for incoming WhatsApp messages, since it is handled in the message model itself
3. Update message payload for WhatsApp to include reply to

## Preview

CleanShot 2023-10-16 at 18 12 50@2x


app/models/message.rb
```diff
@@ -233,6 +234,20 @@ def ensure_processed_message_content
...
+    in_reply_to_external_id = content_attributes[:in_reply_to_external_id]
+
+    if in_reply_to.present? && in_reply_to_external_id.blank?
+      message = conversation.messages.find_by(id: in_reply_to)
+      content_attributes[:in_reply_to_external_id] = message.try(:source_id)
+    elsif in_reply_to_external_id.present? && in_reply_to.blank?
+      message = conversation.messages.find_by(source_id: in_reply_to_external_id)
+      content_attributes[:in_reply_to] = message.try(:id)
+    end
+  end
```
pranavrajs, line 249: This is what I was suggesting in the previous PR. This code is repeated everywhere. It could be rewritten as the following. 

```rb
class Messages::InReplyToMessageBuilder
  pattr_initialize [:message!, :in_reply_to!, :in_reply_to_external_id!]

  delegate :conversation, to: :message

  def perform
    if @in_reply_to.present? && @in_reply_to_external_id.blank?
      message.content_attributes[:in_reply_to_external_id] = in_reply_to_message.try(:source_id)
    elsif @in_reply_to_external_id.present? && @in_reply_to.blank?
     message.content_attributes[:in_reply_to] = in_reply_to_message.try(:id)
    end
    message.save!
  end

  private

  def in_reply_to_message
    return @conversation.messages.find_by(id: @in_reply_to) if @in_reply_to.present?

    return @conversation.messages.find_by(source_id: @in_reply_to_external_id) if @in_reply_to_external_id

    nil
  end
end

# ....

def ensure_in_reply_to
  Messages::InReplyToMessageBuilder.new(message: message, in_reply_to: in_reply_to, in_reply_to_external_id:  in_reply_to_external_id).perform
end

```

pranavrajs: It can be simplified as follows. @scmmishra Make sure you reset both the attributes if in_reply_to message is not present.

```rb
class Messages::InReplyToMessageBuilder
  pattr_initialize [:message!, :in_reply_to!, :in_reply_to_external_id!]

  delegate :conversation, to: :message

  def perform
    set_in_reply_to_attribute if @in_reply_to.present? || @in_reply_to_external_id.present?
  end

  private

  def set_in_reply_to_attribute
    @message.content_attributes[:in_reply_to_external_id] = in_reply_to_message.try(:source_id)
    @message.content_attributes[:in_reply_to] = in_reply_to_message.try(:id)
    @message.save!
  end

  def in_reply_to_message
    return @conversation.messages.find_by(id: @in_reply_to) if @in_reply_to.present?

    return @conversation.messages.find_by(source_id: @in_reply_to_external_id) if @in_reply_to_external_id

    nil
  end
end
```

scmmishra: @pranavrajs this is done

spec/services/whatsapp/providers/whatsapp_cloud_service_spec.rb
```diff
@@ -19,6 +19,7 @@
           .with(
             body: {
               messaging_product: 'whatsapp',
+              context: nil,
```
sojan-official, line 22: Can we have a spec where the context id is populated with a valid value?

PR #8111, opened 2023-10-13
feat: Adds the ability to set an emoji for category

iamsivin:
# Pull Request Template

## Description

The PR adds the ability to save an emoji for the help center category.

Fixes https://linear.app/chatwoot/issue/CW-2618/add-an-attribute-to-select-an-iconemoji-for-the-help-center-category

## Type of change

## How Has This Been Tested?

**Loom video**
https://www.loom.com/share/a5fe89bc7d6843e1b9560c84a3f75b6a?sid=dd7c59aa-e8f5-477f-9bf8-a8e2b2041309

## Checklist:


nithindavid (commented):
Please see the minor comment left

app/javascript/dashboard/routes/dashboard/helpcenter/pages/categories/AddCategory.vue
```diff
@@ -21,15 +21,14 @@
...
-          class="w-full"
-          :error="nameError"
+        <category-name-icon-input
           :label="$t('HELP_CENTER.CATEGORY.ADD.NAME.LABEL')"
           :placeholder="$t('HELP_CENTER.CATEGORY.ADD.NAME.PLACEHOLDER')"
           :help-text="$t('HELP_CENTER.CATEGORY.ADD.NAME.HELP_TEXT')"
-          @input="onNameChange"
+          :has-error="$v.name.$error"
+          :error-message="$t('HELP_CENTER.CATEGORY.ADD.NAME.ERROR')"
+          @on-name-change="onNameChange"
```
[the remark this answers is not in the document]
iamsivin: Done

muhsin-k (approved):

app/javascript/dashboard/routes/dashboard/helpcenter/components/HelpCenterLayout.vue
```diff
@@ -224,7 +224,9 @@ export default {
           key: 'category',
           children: this.categories.map(category => ({
             id: category.id,
-            label: category.name,
+            label: category.icon
```
scmmishra, line 227: Instead of this, it would be better to have the JSON response from the backend to include it as a `label` field

Updating the `app/views/public/api/v1/models/hc/_portal.json.jbuilder` should work

app/javascript/dashboard/i18n/locale/en/helpCenter.json
```diff
@@ -363,9 +363,9 @@
         "PORTAL": "Portal",
         "LOCALE": "Locale",
         "NAME": {
-          "LABEL": "Name",
+          "LABEL": "Icon & Name",
```
scmmishra, line 366: We don't need to say "Icon & Name" explicitly, we can simply say "Name" the icon part is implied given that the picker is available

CC: @Hricha-Shandily @pranavrajs

app/javascript/dashboard/routes/dashboard/helpcenter/pages/categories/NameEmojiInput.vue
```diff
@@ -0,0 +1,134 @@
+<template>
+  <div class="flex items-center relative">
+    <div
+      v-if="icon"
+      class="absolute cursor-pointer z-20 p-0.5 top-7 ltr:left-[38px] rtl:right-[38px] bg-red-100 hover:bg-red-200/50 dark:bg-red-800/30 dark:hover:bg-red-800/60 text-red-400 dark:text-red-500 hover:text-red-700 dark:hover:text-red-300 rounded-xl"
+      @click="onClickInsertEmoji('')"
+    >
+      <fluent-icon size="12" icon="delete" type="outline" />
+    </div>
```
scmmishra, line 9: instead of having this wrapped inside the input visually, we can have it separate, something like this

#### Default

image

#### on hover

image


perhaps speak to Diya if she can suggest a better design.

scmmishra (commented):
Some suggestions for now, will do a detailed review again. Also do we need a migration to replace existing titles with emojis to the new format?

app/javascript/dashboard/routes/dashboard/helpcenter/pages/categories/NameEmojiInput.vue
iamsivin: @scmmishra Updated with the suggestion. When we hover the background, a remove button will appear.
image

app/javascript/dashboard/i18n/locale/en/helpCenter.json
iamsivin: Moved back to Name

pranavrajs: Name is fine.

app/javascript/dashboard/routes/dashboard/helpcenter/components/HelpCenterLayout.vue
iamsivin: @pranavrajs What do you think about this? Do we need a separate icon key? We are using it in the public portal.

iamsivin:
> Some suggestions for now, will do a detailed review again. Also do we need a migration to replace existing titles with emojis to the new format?

As per the conversation with Pranav, adding a separate field in the portal category setting was suggested. CC: @pranavrajs

app/javascript/dashboard/routes/dashboard/helpcenter/pages/categories/NameEmojiInput.vue
scmmishra: Can we adjust the x spacing of that to make it even, should look like a square. Also, the icon, we probably need a background for it. It would be worth making that a button with `rounded-full bg-red-100` and the `x` svg in the center

app/javascript/dashboard/routes/dashboard/helpcenter/components/HelpCenterLayout.vue
scmmishra: We can keep the icon key, just have a label from the backend, makes it instantly usable across the app

app/javascript/dashboard/routes/dashboard/helpcenter/pages/categories/NameEmojiInput.vue
scmmishra, line 1: @iamsivin this is good for now, for future components lets use the composition syntax

app/javascript/dashboard/routes/dashboard/helpcenter/components/HelpCenterLayout.vue
iamsivin: Okay, do you mean to add a new key named label with value icon + name?

app/javascript/dashboard/routes/dashboard/helpcenter/pages/categories/NameEmojiInput.vue
```diff
@@ -0,0 +1,144 @@
+<template>
+  <div
+    class="flex items-center relative"
+    @mouseover="showRemoveEmojiButton = true"
+    @mouseleave="showRemoveEmojiButton = false"
```
scmmishra, line 5: You could use `group-hover` from tailwind to handle this.

ref: https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state

Just add `group` class to the parent, and `hidden group-hover:block` to the icon

app/javascript/dashboard/routes/dashboard/helpcenter/pages/categories/NameEmojiInput.vue
```diff
@@ -0,0 +1,144 @@
...
+  >
+    <div
```
scmmishra, line 7: This should be a button if it is clickable

app/javascript/dashboard/routes/dashboard/helpcenter/pages/categories/NameEmojiInput.vue
```diff
@@ -0,0 +1,144 @@
...
+    >
+      <fluent-icon
+        size="18"
+        icon="dismiss-circle"
+        type="outline"
+        class="text-red-400 dark:text-red-500 hover:text-red-700 dark:hover:text-red-300"
+      />
+    </div>
+    <div
+      class="flex items-center justify-center absolute w-14 z-10 hover:bg-slate-50 dark:hover:bg-slate-700 cursor-pointer h-[2.4375rem] rounded-[4px] top-[25px] border border-solid border-slate-200 dark:border-slate-800"
```
scmmishra, line 20: `h-[2.4375rem]` Why such a specific number? 🥹

Also, `rounded-[4px]` I guess the standard tailwind `rounded` is 4px, but it's getting overridden. See if `rounded` works as is, else add it as `rounded-base`

app/javascript/dashboard/routes/dashboard/helpcenter/pages/categories/NameEmojiInput.vue
```diff
@@ -0,0 +1,144 @@
...
+      <span v-if="icon" v-dompurify-html="icon" class="text-lg" />
+      <fluent-icon
+        v-else
+        size="18"
+        icon="emoji"
+        type="outline"
+        class="text-slate-800 dark:text-slate-100"
+      />
+    </div>
+    <woot-input
```
scmmishra, line 32: I guess this input can be outside the emoji picker & input. That way you won't have to handle the `ml-16` and `mr-16` The parent can be a flex

```html
<div class="flex gap-4">
  <emoji-input ...>
  <input v-model="name" ...>
</div>
```

app/javascript/dashboard/routes/dashboard/helpcenter/pages/categories/NameEmojiInput.vue
```diff
@@ -0,0 +1,144 @@
...
+      v-model.trim="name"
```
scmmishra, line 33: Let's not use `.trim` here, instead let's trim it at the API. Having it here causes some weird issues when editing the text

app/javascript/dashboard/routes/dashboard/helpcenter/pages/categories/NameEmojiInput.vue
```diff
@@ -0,0 +1,144 @@
...
+      if (this.showEmojiPicker) {
+        this.showEmojiPicker = false;
+      }
+    },
+  },
+};
+</script>
+
+<style scoped lang="scss">
+.input-container::v-deep {
```
scmmishra, line 134: Is there a way to avoid `v-deep`, this syntax is deprecated in Vue 3. We can keep it though, not an issue to block this PR over

scmmishra (changes requested):

app/javascript/dashboard/routes/dashboard/helpcenter/pages/categories/NameEmojiInput.vue
absurdiya: - In the empty state we can keep the "add emoji" icon.
- We should make the box size square.
- The cross can have a background and should be in a neutral color. I've used slate-10 for strokes, and slate-3 for background
image
image

absurdiya: Empty state:
image

app/javascript/dashboard/routes/dashboard/helpcenter/pages/categories/NameEmojiInput.vue
iamsivin: > Why such a specific number? 🥹
Added it to match height with the input

app/javascript/dashboard/routes/dashboard/helpcenter/pages/categories/NameEmojiInput.vue
iamsivin: @scmmishra I tried it before. But it will look like this. Currently, we are using the woot-input component.
image

app/javascript/dashboard/routes/dashboard/helpcenter/pages/categories/NameEmojiInput.vue
iamsivin: Sure

app/javascript/dashboard/routes/dashboard/helpcenter/components/HelpCenterLayout.vue
iamsivin: Let's keep it as is for now, but we can make changes in the future. Thoughts? @pranavrajs

scmmishra: Sounds good!

scmmishra (approved):

PR #8125, opened 2023-10-17
feat: Instagram delivery reports

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-2571/delivery-report-for-instagram


app/javascript/dashboard/components/widgets/conversation/bubble/Actions.vue
```diff
@@ -231,7 +231,11 @@ export default {
         return contactLastSeenAt >= this.createdAt;
       }
 
-      if (this.isAWhatsAppChannel || this.isATwilioChannel) {
```
pranavrajs, line 234: Add the same condition for sent status as well.

pranavrajs (approved):

PR #8136, opened 2023-10-18
feat: Facebook delivery reports

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-2570/delivery-report-for-facebook


config/initializers/facebook_messenger.rb
```diff
@@ -29,14 +29,13 @@ def bot
...
-    # delivery.seq       # => 37
-    updater = Integrations::Facebook::DeliveryStatus.new(delivery)
-    updater.perform
-    Rails.logger.info "Human was online at #{delivery.at}"
+    Rails.logger.info "Recieved delivery status #{delivery.to_json}"
+    Webhooks::FacebookDeliveryJob.perform_later(delivery.to_json)
+  end
+
+  Facebook::Messenger::Bot.on :read do |read|
+    Rails.logger.info "Recieved read status  #{read.to_json}"
```
muhsin-k, line 37: Added these logs to verify if events are being received in production, as they are not appearing locally.

app/jobs/webhooks/facebook_delivery_job.rb
scmmishra, line 1: This seems fine, perhaps in the future we might need to add the mutex here, but got to ensure that the sequence of events stay the same before that. No need to change anything for now

lib/integrations/facebook/delivery_status.rb
```diff
@@ -1,32 +1,37 @@
...
+    process_delivery_status if params.delivery_watermark
+    process_read_status if params.read_watermark
   end
 
   private
 
-  def sender_id
-    @params.sender['id']
+  def process_delivery_status
+    timestamp = Time.zone.at(params.delivery_watermark.to_i).to_datetime
```
scmmishra, line 17: Can you confirm if this saves the time in UTC 

Cc: @sojan-official can you confirm if this is required

muhsin-k: I believe `Time.zone` defaults to UTC.  The payload is something like this,
CleanShot 2023-11-16 at 13 08 49@2x

sojan-official: https://chat.openai.com/share/781e05cf-afab-488d-afb3-4885e7df3d82

muhsin-k: @sojan-official @scmmishra Fixed.

scmmishra (approved):

PR #8147, opened 2023-10-19
refactor: UI flags for Reply To

scmmishra:
Facebook does not support `ReplyTo` via their API, so we have to disable that explicitly for them. This PR handles that


iamsivin (approved):
Code LGTM

nithindavid (approved):

PR #8154, opened 2023-10-20
feat: add gtm to vueapp

scmmishra:
Add GTAG for conversation tracking


pranavrajs (approved):

scmmishra:
> @scmmishra Let us move clarity and other scripts to GTM if possible.

At the moment, the MS clarity uses two separate token. 

Also, we might have to use two separate GTAGs since we don't want the marketing scripts to run on the app side, discussing this with Hricha and figuring this out. Will make a corresponding PR once figured out.

PR #8200, opened 2023-10-25
feat: Handle Line send message/attachments errors

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-2690/handle-error-status-for-line-channel

**Preview**

CleanShot 2023-10-25 at 23 44 10


app/services/line/send_on_line_service.rb
```diff
@@ -6,6 +6,24 @@ def channel_class
   end
 
   def perform_reply
-    channel.client.push_message(message.conversation.contact_inbox.source_id, [{ type: 'text', text: message.content }])
+    response = channel.client.push_message(message.conversation.contact_inbox.source_id, [{ type: 'text', text: message.content }])
+    return if response.blank?
+
+    parsed_json = JSON.parse(response.body)
+    # If the request not success, update the message status to failed and save the external error
+    message.update!(status: :failed, external_error: external_error(parsed_json)) unless response.code == '200'
```
sojan-official, line 14: should we update status to delivered if the response is 200 here as well ?

muhsin-k: @sojan-official Even if the response is 200, it doesn't guarantee that the message is delivered. We can mark it as sent, which we have already handled.

sojan-official (approved):

PR #8207, opened 2023-10-25
fix: Telegram model specs

muhsin-k:
Fixes the spec issues from the [PR](https://github.com/chatwoot/chatwoot/issues/8173).


pranavrajs:
My bad, thanks for covering this. I will get this merged, once the spec passes.

pranavrajs (approved):

PR #8237, opened 2023-10-27
feat: Save sort Conversations filter

iamsivin:
# Pull Request Template

## Description

This PR will save the last sort value to the UI setting and retain the previous selection after reloading.

Fixes https://linear.app/chatwoot/issue/CW-2662/sort-conversations-by-creation-date-for-fair-client-handling

## Type of change

## How Has This Been Tested?

**Loom video**
https://www.loom.com/share/932370d0a19e4f538fc2b8693f8d9b05?sid=f7f85fde-e81d-4d21-aa5e-fe4de9d91737

## Checklist:

### DIFF
--- a/app/javascript/dashboard/components/ChatList.vue
+++ b/app/javascript/dashboard/components/ChatList.vue
@@ -207,6 +207,7 @@ import DeleteCustomViews from 'dashboard/routes/dashboard/customviews/DeleteCust
 import ConversationBulkActions from './widgets/conversation/conversationBulkActions/Index.vue';
 import alertMixin from 'shared/mixins/alertMixin';
 import filterMixin from 'shared/mixins/filterMixin';
+import uiSettingsMixin from 'dashboard/mixins/uiSettings';
 import languages from 'dashboard/components/widgets/conversation/advancedFilterItems/languages';
 import countries from 'shared/constants/countries';
 import { generateValuesForEditCustomViews } from 'dashboard/helper/customViewsHelper';
@@ -238,6 +239,7 @@ export default {
     eventListenerMixins,
     alertMixin,
     filterMixin,
+    uiSettingsMixin,
   ],
   props: {
     conversationInbox: {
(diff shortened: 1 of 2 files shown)


app/javascript/dashboard/components/ChatList.vue
```diff
@@ -544,6 +552,11 @@ export default {
       this.$store.dispatch('customViews/update', payloadData);
       this.closeAdvanceFiltersModal();
     },
+    setSavedSortByFilter() {
+      const { status, order_by: orderBy } = this.savedSortByFilter || {};
```
scmmishra, line 556: Instead of handling this here, let's ensure that `savedSortBytFilter` always returns an object

```suggestion
      const { status, order_by: orderBy } = this.savedSortByFilter;
```

app/javascript/dashboard/components/ChatList.vue
```diff
@@ -544,6 +552,11 @@ export default {
       this.$store.dispatch('customViews/update', payloadData);
       this.closeAdvanceFiltersModal();
     },
+    setSavedSortByFilter() {
```
scmmishra, line 555: The function name can improve, we also update the status, so let's rename it to something like `setFiltersFromUISettings`

app/javascript/dashboard/components/ChatList.vue
```diff
@@ -321,6 +323,9 @@ export default {
     hasAppliedFiltersOrActiveFolders() {
       return this.hasAppliedFilters || this.hasActiveFolders;
     },
+    savedSortByFilter() {
+      return this.uiSettings?.conversations_filter_by;
+    },
```
scmmishra, line 328: If this computed value is not used in other places, also that `setSavedSortByFilter` is only triggered at `created`, we can get rid of this method and access it in `setSavedSortByFilter` directly

app/javascript/dashboard/components/widgets/conversation/ConversationBasicFilter.vue
```diff
@@ -85,6 +89,15 @@ export default {
     },
     onChangeFilter(type, value) {
       this.$emit('changeFilter', type, value);
+      this.saveSelectedFilter(type, value);
+    },
+    saveSelectedFilter(type, value) {
```
scmmishra, line 94: The `type` and `value` seems a bit odd here, seems like `type` is holding the actual value. Let's fix it upstream if required

scmmishra (changes requested):

scmmishra (approved):

PR #8277, opened 2023-11-02
feat: Mark the messages as failed if the API channel webhooks fail for any reason.

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-2709/ability-to-see-delivery-reports-for-my-api-channel-inbox

CleanShot 2023-11-02 at 07 58 39@2x


lib/webhooks/trigger.rb
```diff
@@ -1,13 +1,67 @@
...
+
+  def should_handle_error?
+    @webhook_type == :api_inbox_webhook && SUPPORTED_ERROR_HANDLE_EVENTS.include?(@payload[:event])
+  end
+
+  def update_message_status(error)
+    message.update!(status: :failed, external_error: error.message)
+  end
+
+  def conversation
```
sojan-official, line 48: do we need to find the conversation here ? we have the message id right ? why not just look up for message directly

muhsin-k: Makes sense. Updated the PR.

sojan-official (approved):

PR #8280, opened 2023-11-02
feat: allow inbox specific flags for signature toggle

scmmishra:


scmmishra:
@pranavrajs can you check if this is a good implementation. @nithindavid can do the code review

nithindavid (approved):
Tested on local and working as expected.

PR #8298, opened 2023-11-04
Chore(Translation): Label from "login" to "Login" in Login page

Cauen:
1 letter change in PT_BR translation.
Its seems like that translation are not present in https://chatwoot.crowdin.com/

From:
image
To:
image


pranavrajs:
Thanks! 🙏

PR #8304, opened 2023-11-07
fix: Opens foreign links from article page in new tab [cw-2725]

nithindavid:
**Description**
This PR goes through every link in an article and if they are external links, `target=_blank` will be added to these links to make them open in a new tab.

**Loom Video**
https://www.loom.com/share/ca58fc179eba40b4bd31b22a708d087d?sid=c29bc167-eea6-4617-abc0-8aaa43bf10b1

Fixes # (issue)
https://linear.app/chatwoot/issue/CW-2725/external-links-from-helpdesk-being-opened-in-livechat-widget

## Type of change

## How Has This Been Tested?

## Checklist:


app/javascript/portal/portalHelpers.js
```diff
@@ -115,6 +115,32 @@ export const InitializationHelpers = {
...
+      // Modify external links only on articles page
+      const isOnArticlePage =
+        isSameHost && window.location.pathname.includes('/articles/');
+
+      if (isExternalLink && isOnArticlePage) {
+        link.target = '_blank';
+      }
+    });
+  },
+
```
iamsivin, line 143: @nithindavid Can we add `isOnArticlePage` check before `document.querySelectorAll`. This can avoid unnecessary iterations if the user is not on an article page. Thoughts?

For example:
```
    if (isOnArticlePage) {
      document.querySelectorAll('a').forEach(link => {
    
        if (isExternalLink) {
          //
        }
      });
    }
```

iamsivin (commented):
I tested it locally, and the PR is working as expected I have a small comment. Please take a look.

app/javascript/portal/portalHelpers.js
nithindavid: Thanks for the suggestion. I have added this.

nithindavid:
@iamsivin I have added specs and changes. can you take a look?

iamsivin (approved):
I tested it locally and the PR is working as expected.
https://www.loom.com/share/60995a63be604d8fb04c083732ed70c7?sid=2be2d16c-c9de-419c-8b8f-1642a926b11d

app/javascript/portal/portalHelpers.js
```diff
@@ -68,6 +68,30 @@ export const updateThemeStyles = theme => {
   setPortalClass(theme);
 };
 
+export const openExternalLinksInNewTab = () => {
```
scmmishra, line 71: Is there a way to do this on the backend? I think it should be handled there. I can imagine this being a performance issue for large pages with a lot of links, since this needs to run on initial page load. 

Even if we have to do this in JS, let's check this for a large page with a lot of links and see if it performs well, you can simulate a throttled CPU in Chrome, if it's slow, an alternative approach can be the following

```js
document.addEventListener('click', function(event) {
  if (event.target.tagName === 'A') {
    const link = event.target;
    const isInternalLink = link.hostname === window.location.hostname || link.href.includes(window.portalConfig.customDomain) || link.href.includes(window.portalConfig.hostURL);
    
    if (!isInternalLink) {
      link.target = '_blank';
      link.rel = 'noopener noreferrer'; // Security and performance benefits
      // Prevent default if you want to stop the link from opening in the current tab
      event.preventDefault(); 
    }
  }
});
```
This only checks links when they are clicked, rather than iterating over all links on the page. Also, reduces the amount of work done during page load, which can improve perceived performance for large pages

nithindavid: This is better :) Thanks for the suggestion.

scmmishra (approved):

PR #8328, opened 2023-11-09
fix: Change the route change logic based on bus events, after creating a new conversation in the widget.

muhsin-k:
Fixes https://github.com/chatwoot/chatwoot-react-native-widget/issues/35#issuecomment-1803242934


app/javascript/widget/views/PreChatForm.vue
```diff
@@ -4,28 +4,22 @@
...
     PreChatForm,
   },
   mixins: [configMixin, routerMixin],
-  computed: {
-    ...mapGetters({
-      conversationSize: 'conversation/getConversationSize',
-    }),
-  },
-  watch: {
-    conversationSize(newSize, oldSize) {
```
scmmishra, line 24: What did this change do? Also, is the functionality preserved even after removing this?

muhsin-k: Yes. In the past, we change the route from the pre-chat form based on the message size. Now, the route changes once a conversation is successfully created.

scmmishra (approved):

app/javascript/widget/views/PreChatForm.vue
scmmishra: Noted, approving the changes

BusDar:
Dear Chatwoot Support Team,
 
We updated Chatwoot installation from v3.2.0 to v3.3.1 but unfortunately, the Chat widget is still not working correctly. Could you please confirm that the previously mentioned fix is included in the v3.3.1? If it’s not, is it possible to apply the fix manually and what must be done to make it work? Alternatively, could you please consider issuing a hotfix with the previously mentioned fix? This is mission-critical for us. 

Thank you in advance for your patience, understanding and hope to hear from you soon

Regards,
Darius

muhsin-k:
@BusDar The fix is available in the latest release. Could you please describe the problem you're currently experiencing?

PR #8329, opened 2023-11-09
feat: Changes to include delivery reports for all channels

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-2574/[dashboard]-frontend-changes-to-include-delivery-reports-for-all

CleanShot 2023-11-11 at 14 34 1 (1)


app/javascript/dashboard/components/widgets/conversation/bubble/Actions.vue
```diff
@@ -252,6 +247,10 @@ export default {
         return this.sourceId && this.isRead;
       }
 
+      if (this.isAWebWidgetInbox || this.isAPIInbox) {
+        return this.isRead;
```
scmmishra, line 251: The condition here and the condition this replaces does not match, the computed property checks the message status on the other hand the previous condition compared timestamps. Is this intentional

muhsin-k: Previously, we find the message read status based on the contact's last seen time. Now, we're updating this flow to mark all messages as read while updating the contact's last seen time. So, every message have the status `read` once the user has viewed it. https://github.com/chatwoot/chatwoot/pull/8092/files#diff-77f099f036eeb74868f3b7f1e330b962f032baabbb4341fbbe7bbce42c6d5eacR5

scmmishra (approved):

PR #8381, opened 2023-11-18
fix: Breakage of message pane when Openai integration isn't enabled

clairton:
error when not exist openai integration

`Cannot read properties of undefined (reading 'hooks')`

Captura de Tela 2023-11-18 às 10 08 12


sojan-official:
@clairton, thanks a lot for the fix

PR #8400, opened 2023-11-22
fix: Update campaign routes to fix the rendering issue

pranavrajs:
As a part of https://github.com/chatwoot/chatwoot/pull/8394, the rendering of campaigns has been updated. However, this resulted in the breakage in production. This PR fixes the rendering. 

- Fix the rendering of campaign pages.
- Update the macro editor to work on mobile/tablet layout.

Fixes https://linear.app/chatwoot/issue/CW-2798/campaigns-page-on-production-is-not-loading


nithindavid (approved):
Tested and working as expected.

PR #8402, opened 2023-11-22
feat: Adds dropdown to switch locales from articles list page

nithindavid:
## Description
This PR adds a locale switcher to the articles page. This is the first of many tiny changes to improve the flows in portal UX. 

image

**Video**
https://www.loom.com/share/136e09d17fa44a859ebbba7517ff0148?sid=68d75af9-a591-4d99-9fbc-2c430a5aacda

## Type of change

## How Has This Been Tested?

- Switching language selected and verifying the articles
- Switching language selected and verifying the categories


pranavrajs (changes requested):
Looks good, some minor comments. 

1/ The dropdown doesn't close if I click on the same item again. 
2/ It would be better to show the actual name of the locale rather than the en_US or locale param.
|Suggested | Current implementation|
| -- | -- |
| Screenshot 2023-11-22 at 12 55 42 PM | Screenshot 2023-11-22 at 12 54 37 PM |

app/javascript/dashboard/routes/dashboard/helpcenter/components/Header/ArticleHeader.vue
```diff
@@ -146,16 +203,32 @@ export default {
       this.$emit('close');
       this.showSortByDropdown = false;
     },
+    openLocaleDropdown() {
+      this.$emit('open');
+      this.showLocaleDropdown = true;
+    },
+    closeLocaleDropdown() {
+      this.$emit('close');
```
scmmishra, line 211: Why do we need to emit open and close events?

app/javascript/dashboard/routes/dashboard/helpcenter/components/HelpCenterLayout.vue
```diff
@@ -262,6 +262,9 @@ export default {
         this.fetchPortalAndItsCategories();
       }
     },
+    '$route.params.locale'() {
```
scmmishra, line 265: Can we do this with a bus event instead of watching a deep object? WDYT? I do not have strong opinions on either approach here, but a bus event seems more explicit

app/javascript/dashboard/routes/dashboard/helpcenter/components/Header/ArticleHeader.vue
nithindavid: This was added by mistake, fixing this.

scmmishra (approved):

PR #8488, opened 2023-12-05
feat: Add sticker support in Line channel

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-2826/feat-support-incoming-sticker-images-in-line-channel

Currently, Chatwoot doesn't support line stickers. As a temporary solution, we will display stickers as images using the sticker ID in markdown format. This is subject to change in the future. We've chosen not to download and display the sticker as an image because the sticker's information and images are the property of the creator or legal owner. We aim to avoid storing it on our server without their consent. If there are any permission or rendering issues, the URL may break, and we'll display the sticker ID as text instead.

**Preview**

CleanShot 2023-12-05 at 12 57 13@2x

### DIFF
--- a/app/javascript/dashboard/components/widgets/conversation/MessagePreview.vue
+++ b/app/javascript/dashboard/components/widgets/conversation/MessagePreview.vue
@@ -20,7 +20,15 @@
         icon="info"
       />
     </template>
-    <span v-if="message.content">
+    <span v-if="message.content && isMessageSticker">
+      <fluent-icon
+        size="16"
+        class="-mt-0.5 align-middle inline-block text-slate-600 dark:text-slate-300"
+        icon="image"
+      />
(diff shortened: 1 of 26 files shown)


app/javascript/dashboard/components/widgets/conversation/MessagePreview.vue
```diff
@@ -20,7 +20,15 @@
         icon="info"
       />
     </template>
-    <span v-if="message.content">
+    <span v-if="message.content && isMessageSticker">
+      <fluent-icon
+        size="16"
+        class="-mt-0.5 align-middle inline-block text-slate-600 dark:text-slate-300"
+        icon="image"
```
muhsin-k, line 27: Used the image icon because we are rendering the sticker in image format.

pranavrajs (approved):

PR #8509, opened 2023-12-07
feat: Updates the design of the network connection warning snack bar

nithindavid:
## Description
Reducing the size and position of network disconnected warning component. Also, delaying the socket disconnected event by 4 seconds as it's very frequently.

image

image


scmmishra (commented):
A tooltip on a toast doesn't seem great. If we just write "Offline" we can avoid the tooltip. 

CleanShot 2023-12-11 at 15 53 42@2x

---

Nitpick, would be great to incorporate the dismiss button within the toast

app/javascript/dashboard/components/NetworkNotification.vue
```diff
@@ -47,7 +57,12 @@ export default {
   mounted() {
     window.addEventListener('offline', this.updateOnlineStatus);
     window.bus.$on(BUS_EVENTS.WEBSOCKET_DISCONNECT, () => {
-      this.updateOnlineStatus({ type: 'offline' });
+      // TODO: Remove this after completing the conversation list refetching
+      // TODO: DIRTY FIX : CLEAN UP THIS WITH PROPER FIX, DELAYING THE RECONNECT FOR NOW
```
scmmishra, line 61: Didn't understand what is happening here, let me know if you wanna pair on this

app/javascript/dashboard/i18n/locale/en/generalSettings.json
```diff
@@ -92,7 +92,8 @@
   },
   "NETWORK": {
     "NOTIFICATION": {
-      "TEXT": "Disconnected from Chatwoot"
+      "TEXT": "Disconnected from Chatwoot",
```
iamsivin, line 95: @nithindavid Do we need this i18n here? Are we using it anywhere?

nithindavid: @iamsivin I have removed this. Can you take a look again?

iamsivin (commented):
@nithindavid Is it possible to tweak the color of the banner and the close button to better suit the dark mode interface?
image

nithindavid:
@iamsivin I have added support for dark mode as well.

image

image

app/javascript/dashboard/components/NetworkNotification.vue
```diff
@@ -47,7 +57,12 @@ export default {
...
+      // THE CABLE IS FIRING IS VERY COMMON AND THUS INTERFERING USER EXPERIENCE
+      setTimeout(() => {
+        this.updateOnlineStatus({ type: 'offline' });
+      }, 4000);
```
iamsivin, line 65: @nithindavid, Can you check if it's working correctly? The set timeout is set to run 4 seconds after the WEBSOCKET_DISCONNECT event is detected, but the snack bar triggers immediately when I turn offline from the network tab. I think it's because of offline eventListener.
https://www.loom.com/share/0cffc02bfbb24360b74ac27a704c2f43?sid=3d3322d9-d9c2-4e94-986d-3a1f1a8193af

nithindavid: Yes, we also have network state change. This 4 second delay is only for the action cable events.

PR #8527, opened 2023-12-11
feat(perf): Cache labels on the conversation model

pranavrajs:
I began by #8525, which increases the conversation page size to 100. However, this can't be reliably sent to all self-hosted instances because the response time is higher than the 25 items. (It took 6 seconds on my machine). 

This PR could decrease one query in the conversation jbuilder. Currently, we fetch those during runtime, which increases the API response time. In this PR, 

- Add an attribute `cached_label_list` on the conversation table to store the cached list of labels. 
- Run a migration to touch all the conversations to generate the cache for the first time.

### DIFF
--- a/app/controllers/api/v1/accounts/conversations/assignments_controller.rb
+++ b/app/controllers/api/v1/accounts/conversations/assignments_controller.rb
@@ -14,7 +14,8 @@ class Api::V1::Accounts::Conversations::AssignmentsController < Api::V1::Account
 
   def set_agent
     @agent = Current.account.users.find_by(id: params[:assignee_id])
-    @conversation.update_assignee(@agent)
+    @conversation.assignee = @agent
+    @conversation.save!
     render_agent
   end
 
--- a/app/controllers/api/v1/accounts/conversations_controller.rb
+++ b/app/controllers/api/v1/accounts/conversations_controller.rb
@@ -110,8 +110,8 @@ class Api::V1::Accounts::ConversationsController < Api::V1::Accounts::BaseContro
   end
 
   def assign_conversation
-    @agent = Current.account.users.find(current_user.id)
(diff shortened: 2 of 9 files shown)


app/controllers/api/v1/accounts/conversations/assignments_controller.rb
```diff
@@ -14,7 +14,8 @@ def create
 
   def set_agent
     @agent = Current.account.users.find_by(id: params[:assignee_id])
-    @conversation.update_assignee(@agent)
+    @conversation.assignee = @agent
+    @conversation.save!
```
pranavrajs, line 18: Updating this just because I removed the update_assignee method in conversation.

app/controllers/api/v1/accounts/conversations_controller.rb
```diff
@@ -110,8 +110,8 @@ def set_conversation_status
   end
 
   def assign_conversation
-    @agent = Current.account.users.find(current_user.id)
-    @conversation.update_assignee(@agent)
+    @conversation.assignee = current_user
+    @conversation.save!
```
pranavrajs, line 114: Updating this just because I removed the update_assignee method in conversation. Also, I don't think we need to find the user again, we can use the current_user directly. @muhsin-k

app/models/conversation.rb
```diff
@@ -185,6 +182,10 @@ def webhook_data
     Conversations::EventDataPresenter.new(self).push_data
   end
 
+  def cached_label_list_array
+    (cached_label_list || '').split(',')
+  end
```
pranavrajs, line 187: Used in 2 places. If there is a better way to do this let me know @sojan-official

sojan-official (approved):

PR #8543, opened 2023-12-13
feat: Settings screen for custom branding

sojan-official:
- Ability to configure Chatwoot custom branding from UI when using paid plans

Screenshot 2023-12-12 at 4 24 01 PM


spec/controllers/super_admin/app_config_controller_spec.rb
```diff
@@ -34,13 +34,13 @@
     context 'when it is an aunthenticated super admin' do
```
vishnu-narayanan, line 34: Should we include an additional specification to verify that the allowed_config does not contain customization variables intended for the 'enterprise' plan?"

vishnu-narayanan (approved):

spec/controllers/super_admin/app_config_controller_spec.rb
sojan-official: Not too worried about this at the moment since the config reset job which we are planning will reset it back to the old values.  also, there is always the option to update directly from the database/console so 🤷‍♂️

vishnu-narayanan: 😅

PR #8554, opened 2023-12-14
feat: Conversation sort old activity

clairton:
# Conversation sort old activity


pranavrajs:
@clairton Thanks for the PR. We merged a PR yesterday, fixing the same problem. Please see it here. https://github.com/chatwoot/chatwoot/pull/8532

Screenshot 2023-12-14 at 12 14 45 PM

PR #8585, opened 2023-12-19
feat: Add `last_activity_at` field to Notification model

muhsin-k:
This PR will introduce a `last_activity_at` date-time field to the Notification model for ordering notifications by their snooze time. This means if a notification is snoozed, the last_activity_at field will update, bringing the notification to the top.

### DIFF
--- a/db/migrate/20231219073832_add_last_activity_at_to_notifications.rb
+++ b/db/migrate/20231219073832_add_last_activity_at_to_notifications.rb
@@ -1,0 +1,6 @@
+class AddLastActivityAtToNotifications < ActiveRecord::Migration[7.0]
+  def change
+    add_column :notifications, :last_activity_at, :datetime, default: nil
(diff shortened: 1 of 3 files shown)


db/migrate/20231219073832_add_last_activity_at_to_notifications.rb
```diff
@@ -0,0 +1,6 @@
+class AddLastActivityAtToNotifications < ActiveRecord::Migration[7.0]
+  def change
+    add_column :notifications, :last_activity_at, :datetime, default: nil
```
sojan-official, line 3: we can do a not null to ensure consistency

sojan-official (commented):
Looks good. 

- minor comment . lets add a not null constraint to last activity at to ensure that the value doesn't get saved as nil.

sojan-official (approved):

PR #8592, opened 2023-12-21
chore: Temp support for 360DialogV2

sojan-official:
Before sunsetting the 360Dialog Provider, we will temporarily support the V2 URL in response to requests from some enterprise clients. However, we advise against creating new inboxes with this configuration due to its impending deprecation. We recommend using or transitioning to the WhatsApp Cloud Provider for the WhatsApp channel.

### DIFF
--- a/app/jobs/webhooks/whatsapp_events_job.rb
+++ b/app/jobs/webhooks/whatsapp_events_job.rb
@@ -44,7 +44,7 @@ class Webhooks::WhatsappEventsJob < ApplicationJob
     channel = Channel::Whatsapp.find_by(phone_number: phone_number)
 
     # TODO: Remove this in future when we deprecate the 360Dialog Provider
-    return channel unless channel.provider == 'whatsapp_cloud'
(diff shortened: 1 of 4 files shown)


sojan-official:
@pranavrajs We are Not adding additional tests since we will be removing this going forward.

pranavrajs (approved):

anirudhsama:
Hey we are using 360Dialog provider with the On Prem option. We had earlier used WhatsApp Cloud API, but moved away from it due to billing issues with Meta. It's been more than 2 months now and Meta hasn't gotten back with any resolution and that business account is still locked. With a third party provider, it is much easier to resolve these issues so it is important to continue supporting them IMO.

Another option might be to provide support for any WhatsApp API provider which exposes the standard WhatsApp Cloud API. It might be something like configuring the base URL or something like that. That way it should work with multiple providers.

sojan-official:
@anirudhsama Please see https://github.com/orgs/chatwoot/discussions/2759#discussion-3504024

PR #8602, opened 2023-12-23
feat: sla 1 - refactor sla_policies model and add applied_sla model

vishnu-narayanan:
# Pull Request Template

## Description

`PR 1` for the `SLA` feature, which includes data models for SLA 

Fixes https://linear.app/chatwoot/issue/CW-2981/add-data-models

### Changelog

For `sla_policies` model,

- rename `frt_threshold` column to `first_response_time_threshold`
- rename `rt_threshold` column to `next_response_time_threshold` 
- add column `description`
- add column `resolution_time_threshold`

Add `applied_slas` model 
- belongs to account, conversation and  and sla_policy
- status field to hold `hit`, `missed`, `active`, `cancelled`

## Type of change

## How Has This Been Tested?

- Ran locally
- Added specs

## Checklist:


enterprise/app/models/sla_policy.rb
```diff
@@ -3,10 +3,12 @@
 # Table name: sla_policies
 #
 #  id                         :bigint           not null, primary key
-#  frt_threshold              :float
+#  first_response_time        :float
```
pranavrajs, line 6: I believe the naming would okay because SLA models are supposed to save only the threshold. Nothing else. However, would it be easier to have threshold in the name for a quick search. 

The `first_response_time` search would return 507 results from 123 files.
Screenshot 2024-01-19 at 10 05 43 AM

pranavrajs (commented):
A minor comment on naming, otherwise looks fine.

enterprise/app/models/sla_policy.rb
vishnu-narayanan: True. I will make changes accordingly.

db/migrate/20231223040257_create_applied_slas.rb
```diff
@@ -0,0 +1,13 @@
+class CreateAppliedSlas < ActiveRecord::Migration[7.0]
+  def change
+    create_table :applied_slas do |t|
+      t.references :account, null: false, foreign_key: true
```
sojan-official, line 4: lets remove the foreign keys 

ref: https://stackoverflow.com/a/2373517/939299 

> TLDR:  Rails doesn't prevent you from using foreign key constraints in your database, it just doesn't give them to you by default. The "Rails Way" is to depend on Rails to manage your database, cascade your deletes, enforce referential integrity, etc. 

We follow the rails way mostly as the foreign keys make it a little bit more tricky for deletes etc.

vishnu-narayanan: cool cool ✅

sojan-official (approved):

Issue #8622, opened 2023-12-29
Bug: formData.append Adds 'null' String When replyTo is null


Maklu:
### Describe the bug

https://github.com/chatwoot/chatwoot/blob/78fab7897dfd7df05339b1154f7e115b3e86675e/app/javascript/widget/api/endPoints.js#L55

When the `replyTo` variable is `null`, the `formData.append` method incorrectly appends it as the string `"null"` to the form data. This behavior leads to unexpected outcomes when processing the form data, as the expectation is that `replyTo` should either be a valid string or not included at all.

### To Reproduce

1. Initialize a variable `replyTo` as `null`.
2. Use `formData.append('message[reply_to]', replyTo)` to add `replyTo` to `formData`.
3. Inspect the resulting `formData` and observe that `replyTo` is appended as the string `"null"`.

### Expected behavior

The `formData.append` method should not append `replyTo` to the form data if `replyTo` is `null`. Ideally, it should either omit the entry entirely or handle it in a way that does not result in the string representation of `null`.

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

None

PR #8623, opened 2023-12-29
fix: Prevent formData.append from Adding 'null' String for null replyTo

Maklu:
## Description

**Change Summary**: Implemented a conditional check in `formData.append` to prevent appending 'null' as a string when `replyTo` is `null`. This change ensures the integrity of the form data sent to the server.

**Issue Fixed**: Fixes an issue where `formData.append` was incorrectly appending a string 'null' when `replyTo` was actually `null`.  #8622 

**Motivation and Context**: 
- To prevent sending invalid data in form submissions.
- To align the form data handling with expected standards.

**Dependencies**: This change does not require any new dependencies.

## Type of change

## How Has This Been Tested?

**Test Configuration**:
- Tested in a local development environment using JavaScript and HTML forms.
- Verified that `formData` does not include 'reply_to' when `replyTo` is `null`.

**Steps to Reproduce**:
1. Set `replyTo` to `null`.
2. Execute the function containing `formData.append('message[reply_to]', replyTo)`.
3. Check `formData` to ensure it doesn't append 'reply_to' field.

## Checklist:


JoseLourenc0 (commented):
I would suggest on working with a validation for falsy elements (1), not just nullable or at least using strict compare (2).
(1)
```javascript
 if (![null, undefined, '', 0, false].includes(replyTo)) {
    formData.append('message[reply_to]', replyTo);
  }
```
(2)
```javascript
 if (replyTo !== null) {
    formData.append('message[reply_to]', replyTo);
  }
```
There are also some other ways to verify that so you can apply DRY concept since there are already simillar kinds of validation in other places of chatwoot.

scmmishra (approved):

pranavrajs (approved):

PR #8624, opened 2023-12-29
feat: scroll to replied message

clairton:
Scroll to replied message on click


clairton:
https://github.com/chatwoot/chatwoot/assets/471685/e12e1b48-27a0-48b5-9d3f-39ddd5df7bd9

scmmishra:
@clairton can you test if this works for a longer conversation where the message is probably not loaded and needs to be fetched?

clairton:
@scmmishra its true, what code would be use to load this messages?

pranavrajs:
@scmmishra Let us merge this and consider the case where the message is unavailable as a separate one.

pranavrajs (approved):

pranavrajs:
@clairton Thanks for the PR. This is helpful.

PR #8639, opened 2024-01-04
feat: trigger handoff when agent bot is the actor

scmmishra:
This PR adds a feature to auto-trigger handoff events when an Agent bot toggles a conversation status from `Pending` to `Open`

## Type of change

## How Has This Been Tested?

Unit tests, covers enough cases

## Checklist:

### DIFF
--- a/app/controllers/concerns/access_token_auth_helper.rb
+++ b/app/controllers/concerns/access_token_auth_helper.rb
@@ -15,6 +15,7 @@ module AccessTokenAuthHelper
 
     @resource = @access_token.owner
     Current.user = @resource if current_user.is_a?(User)
+    Current.executed_by = @resource
   end
 
   def validate_bot_access_token!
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/access_token_auth_helper.rb
```diff
@@ -15,6 +15,7 @@ def authenticate_access_token!
 
     @resource = @access_token.owner
     Current.user = @resource if current_user.is_a?(User)
+    Current.executed_by = @resource
```
scmmishra, line 18: Had to do this since the `Current.user` is only set when the authorized entity is an actual `User`. This could have additional implications, for instance in `status_change_activity` in `activity_message_handler.rb`. 

An alternative is to add `Current.actor` or `Current.bot` as a new attribute and use that? But if we want to keep the code `DRY` we can use `Current.executed_by` and handle the case everywhere it is not handled across the code

CC: @pranavrajs @sojan-official

sojan-official: @scmmishra  I think we can go ahead with just the check whether `Current.user.nil?`.  since the only case where the `toggle_status` method will be called is when the request is authenticated ( either a user/bot).  I think for now, let's rely solely on that check. 

Let's circle back later and look into setting Current.user = bot and verify all the activity message generation at a later point so that this will be much cleaner and more predictable.

sojan-official (approved):

PR #8645, opened 2024-01-05
feat: Persist all the notifications

muhsin-k:
Right now, we don't create notification objects if the user doesn't subscribe to the notifications from his profile settings. 

For the inbox view on the web/mobile to work, we would need to create a notification object for all the following supported notifications:

- conversation_assignment
- assigned_conversation_new_message
- conversation_mention
- participating_conversation_new_message


app/models/notification.rb
```diff
@@ -122,17 +122,28 @@ def content
...
 
   def process_notification_delivery
-    Notification::PushNotificationJob.perform_later(self)
+    Notification::PushNotificationJob.perform_later(self) if user_subscribed_to_notification?
 
     # Should we do something about the case where user subscribed to both push and email ?
     # In future, we could probably add condition here to enqueue the job for 30 seconds later
     # when push enabled and then check in email job whether notification has been read already.
-    Notification::EmailNotificationJob.perform_later(self)
+    Notification::EmailNotificationJob.perform_later(self) if user_subscribed_to_notification?
```
sojan-official, line 130: This should only be sent email notifications that are subscribed.

app/models/notification.rb
```diff
@@ -122,17 +122,28 @@ def content
...
+  def user_subscribed_to_notification?
+    notification_setting = user.notification_settings.find_by(account_id: account.id)
+    # added for the case where an assignee might be removed from the account but remains in conversation
+    return if notification_setting.blank?
+
+    return true if notification_setting.public_send("email_#{notification_type}?")
+    return true if notification_setting.public_send("push_#{notification_type}?")
+
+    false
+  end
```
sojan-official, line 145: Possibly accept the type of notification delivery as an argument for this method.

sojan-official (changes requested):
Couple of minor comments.

app/models/notification.rb
muhsin-k: Fixed.

app/models/notification.rb
muhsin-k: Fixed.

sojan-official (approved):

PR #8660, opened 2024-01-08
feat: UI for unsupported message

scmmishra:
localhost_3000_app_accounts_1_conversations_8504 (1)
localhost_3000_app_accounts_1_conversations_8504

## Type of change

### DIFF
--- a/app/javascript/dashboard/i18n/locale/en/conversation.json
+++ b/app/javascript/dashboard/i18n/locale/en/conversation.json
@@ -41,6 +41,7 @@
     "SAVE_CONTACT": "Save",
     "UPLOADING_ATTACHMENTS": "Uploading attachments...",
     "REPLIED_TO_STORY": "Replied to your story",
+    "UNSUPPORTED_MESSAGE": "This message is unsupported. You can view this message on your phone.",
     "SUCCESS_DELETE_MESSAGE": "Message deleted successfully",
     "FAIL_DELETE_MESSSAGE": "Couldn't delete message! Try again",
     "NO_RESPONSE": "No response",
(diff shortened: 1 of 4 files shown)


scmmishra:
@pranavrajs need your thoughts on the UI and copy
CC: @nithindavid @absurdiya

absurdiya:
> @pranavrajs need your thoughts on the UI and copy CC: @nithindavid @absurdiya

We should use a warning/error color palette, this seems very similar to private note (which is also yellow)

app/javascript/dashboard/i18n/locale/en/conversation.json
```diff
@@ -41,6 +41,7 @@
     "SAVE_CONTACT": "Save",
     "UPLOADING_ATTACHMENTS": "Uploading attachments...",
     "REPLIED_TO_STORY": "Replied to your story",
+    "UNSUPPORTED_MESSAGE": "This message is unsupported. You can view this message on the mobile app.",
```
sojan-official, line 44: lets also tweak this text little bit.  view this on mobile app must be confusing to the user as they would imagine the mobile app to be chatwoot. lets change this to reflect that they need to view this message on facbook/instagram

scmmishra:
> > @pranavrajs need your thoughts on the UI and copy CC: @nithindavid @absurdiya
> 
> We should use a warning/error color palette, this seems very similar to private note (which is also yellow)

@absurdiya updated, PTAL

absurdiya:
> > > @pranavrajs need your thoughts on the UI and copy CC: @nithindavid @absurdiya
> > 
> > 
> > We should use a warning/error color palette, this seems very similar to private note (which is also yellow)
> 
> @absurdiya updated, PTAL

Looks good to me

pranavrajs (approved):

PR #8661, opened 2024-01-08
feat: set lock timeout to 1 second

scmmishra:
We notice a lot of `MutexApplicationJob::LockAcquisitionError` especially for high velocity instances. This PR attempts to address that

- Reducing the default timeout to 1 second instead 2. Theoretically this should release the lock faster, given that the wait time in all cases is 1 second, having a lock timeout more than that does not make sense. Beside for us, the Mutex providing a 1 second delay is sufficient to handle concurrency issues
- Increase the attempt count for slack to 8, previously this has shown to improve delivery for FB and Insta


scmmishra:

scmmishra:
@sojan-official PTAL

sojan-official (approved):

PR #8668, opened 2024-01-09
chore: Handle stripe events without plan data

sojan-official:
`plan` isn't a guaranteed object in stripe `customer.subscription.updated` events.  It can be null for cases like `send_invoice` for `past_due` event as seen in the payload shown in sentry error.

fixes: https://linear.app/chatwoot/issue/CW-2925/nomethoderror-undefined-method-[]-for-nilnilclass-nomethoderror


pranavrajs (approved):

pranavrajs:
@sojan-official Can you please add the context here in the following format for future reference? 

1/ The research you did -- where did it impact, how much did it impact?
2/ Why is the change you made the right solution? Explain it in plain english.
3/ What are the alternatives you considered?

PR #8671, opened 2024-01-09
fix: TypeError cannot read properties of undefined (reading 'queryselector')

iamsivin:
# Pull Request Template

## Description

Fixes https://linear.app/chatwoot/issue/CW-2947/typeerror-cannot-read-properties-of-undefined-reading-queryselector

**Loom video**
https://www.loom.com/share/65dec11313354f15a378e9ba557b8919?sid=60f675a4-f31f-42ae-86f3-b167029d20ba

## Type of change

## How Has This Been Tested?

**Steps to reproduce -**
1. Open the canned response popup from the editor by /
2. search for some random responses that are not on the list
3. Press the arrow keys now, you can check the console if there is any error.

## Checklist:


app/javascript/dashboard/components/widgets/mentions/MentionBox.vue
```diff
@@ -68,6 +68,7 @@ export default {
     },
     selectedIndex() {
       const container = this.$refs.mentionsListContainer;
+      if (!container) return;
```
scmmishra, line 71: If `container` is not available, perhaps we can just search the ID on the `$root` or `$parent`

Also, did you manage to reproduce this? Add a watcher and see it's value, where in the lifecycle is it missing?

iamsivin: @scmmishra I just wanted to clarify that the mentionsListContainer is the wrapper holding the list items. I've reproduced this issue. When we open canned responses in the editor using the `/` key and search for canned responses that are not available in the list then if we accidentally click the arrow keys in that case the container section is missing and throws an error.

iamsivin: @scmmishra Done

nithindavid (approved):

scmmishra (approved):

PR #8673, opened 2024-01-09
fix: SQL error when rules with missing attributes is triggered

scmmishra:
Suppose we have a condition in an automation rule with a custom attribute, if that attribute it deleted, there can arise a condition when the generated query string can be malformed.

Suppose we have a filter on `conversation.created` event. And the rule looks like the following 

```json
[
  {
    "values": [
      "open"
    ],
    "attribute_key": "status",
    "query_operator": "and",
    "filter_operator": "equal_to",
    "custom_attribute_type": ""
  },
  {
    "values": [
      "true"
    ],
    "attribute_key": "missing_attribute",
    "filter_operator": "equal_to",
    "custom_attribute_type": "conversation_attribute"
  }
]
```

- When the first condition is parsed, and `apply_filter` appends to an empty query string, since there is a `query_operator` an `AND` is appended too. So for the first pass, the query string looks like `conversations.status IN (0) and`
- Now when the `apply_filter` is run for the second condition, [this line](https://github.com/chatwoot/chatwoot/blob/046ce68a45063436a3a4c7e82aeb91e0fa612eaf/app/services/automation_rules/conditions_filter_service.rb#L56C6-L56C6) evaluates to nil, causing the condition to break

This PR runs adds a sanitation step that ensures any rules without the said custom attribute is removed.

---

P.S. I tried a hack to fix this, Spoiler alert, it failed hilariously. But I'll document it anyway.

We can update the `apply_filter` method to add a final else condition `1 = 1`, so instead of appending nothing, it appends a blank condition. But then this only works for `AND` condition. 

```rb
  def apply_filter(query_hash, current_index)
    conversation_filter = @conversation_filters[query_hash['attribute_key']]
    contact_filter = @contact_filters[query_hash['attribute_key']]
    message_filter = @message_filters[query_hash['attribute_key']]

    @query_string += if conversation_filter
                       conversation_query_string('conversations', conversation_filter, query_hash.with_indifferent_access, current_index)
                     elsif contact_filter
                       contact_query_string(contact_filter, query_hash.with_indifferent_access, current_index)
                     elsif message_filter
                       message_query_string(message_filter, query_hash.with_indifferent_access, current_index)
                     elsif custom_attribute(query_hash['attribute_key'], @account, query_hash['custom_attribute_type'])
                       # send table name according to attribute key right now we are supporting contact based custom attribute filter
                       custom_attribute_query(query_hash.with_indifferent_access, query_hash['custom_attribute_type'], current_index)
                     else
                       '1 = 1'
                     end
  end
```

### DIFF
--- a/app/services/automation_rules/conditions_filter_service.rb
+++ b/app/services/automation_rules/conditions_filter_service.rb
@@ -20,7 +20,7 @@ class AutomationRules::ConditionsFilterService < FilterService
     @message_filters = @filters['messages']
     @attribute_changed_query_filter = []
 
-    @rule.conditions.each_with_index do |query_hash, current_index|
+    valid_rule_conditions.each_with_index do |query_hash, current_index|
       @attribute_changed_query_filter << query_hash and next if query_hash['filter_operator'] == 'attribute_changed'
 
       apply_filter(query_hash, current_index)
@@ -31,6 +31,28 @@ class AutomationRules::ConditionsFilterService < FilterService
     records = perform_attribute_changed_filter(records) if @attribute_changed_query_filter.any?
 
     records.any?
+  end
+
+  def valid_rule_conditions
+    filtered_rules = []
+    @rule.conditions.each do |query_hash|
+      filtered_rules << query_hash if valid_filter?(query_hash)
+    end
+
+    # [{"values"=>["open"], "attribute_key"=>"status", "query_operator"=>"and", "filter_operator"=>"equal_to", "custom_attribute_type"=>""}]
+    # go to the last condition and remove the `query_operator` key
+    filtered_rules.last.delete('query_operator')
+
+    filtered_rules
+  end
+
+  def valid_filter?(query_hash)
+    conversation_filter = @conversation_filters[query_hash['attribute_key']]
+    contact_filter = @contact_filters[query_hash['attribute_key']]
+    message_filter = @message_filters[query_hash['attribute_key']]
+
+    conversation_filter || contact_filter || message_filter ||
+      custom_attribute(query_hash['attribute_key'], @account, query_hash['custom_attribute_type'])
   end
 
   def filter_operation(query_hash, current_index)
--- a/app/services/automation_rules/conditions_filter_service.rb
+++ b/app/services/automation_rules/conditions_filter_service.rb
@@ -39,7 +39,6 @@ class AutomationRules::ConditionsFilterService < FilterService
       filtered_rules << query_hash if valid_filter?(query_hash)
     end
 
-    # [{"values"=>["open"], "attribute_key"=>"status", "query_operator"=>"and", "filter_operator"=>"equal_to", "custom_attribute_type"=>""}]
     # go to the last condition and remove the `query_operator` key
     filtered_rules.last.delete('query_operator')
 


scmmishra:
@pranavrajs @sojan-official this is only half a fix, also needs some good testing, I'll add the unit tests too. For now can you folks please take a look and see if this is a good solution?

**Also, this is only half of the solution.** We need to add validations in the backend to do either one of the two, update the rule conditions ourself when an automation rule is deleted or check if a custom attribute is used in an automation rule and don't let the user delete the attribute unless they edit the rule

scmmishra:
@pranavrajs @sojan-official PTAL whenever you can

pranavrajs (changes requested):
As I understand it, there are two problems here:

1. When custom attributes are removed, the associated automation where the attributes are used is not being removed.
2. The automation service is generating an invalid query.

There are different options to address the first issue. 
One option is to add validation for custom attribute deletion to ensure all the associated automation is deleted. Another option is to automatically delete the automation and notify the customer via email. We can consider this as a separate task.

Let us consider the second problem only for now. We have two potential options:

1. Perform query validation before executing it.
2. Modify the query generation process to include the query operator only if the second condition is valid.


The first approach is correct, which seems to be what you were trying to do here. However, it appears that you have unintentionally implemented the second option. Let us dissect and see.

For example, take the following scenario:

```rb
[
  {
    "values": ["open"],
    "attribute_key": "status",
    "query_operator": "and",
    "filter_operator": "equal_to"
  },
  {
    "values": ["true"],
    "attribute_key": "cloud_customer", // Deleted from the account
    "filter_operator": "equal_to",
    "custom_attribute_type": "conversation_attribute"
  }
]
```

The expectation is that the automation should only run when the status is set to "Open" and the "cloud_customer" attribute is turned on. If we delete the custom attribute `cloud_customer,` then the `valid_rule_conditions` method would only return the first object. This would generate a query using `records.where(Status=open)` instead of `records.where(Status=open and cloud_customer=true)`. 

Note that these conditions are inside an automation rule. Consequently, the implementation in this PR would run the automation for every conversation where the status is set to "Open". The implementation should have prevented running the rule altogether if there is at least one invalid condition.

I would suggest the following changes:

1. Move the validations to a service called `ConditionsValidationService`. This service should perform the following tasks:

- Check if the contact, conversation, or message attribute is a supported filter by consulting the `filter_keys.json` file.
- Check if the custom attribute is still present in the account.

The output of the validation service can be true or false depending on whether the condition is valid or not. 

2. Update the filter service to something like the one below.

```rb
class AutomationRules::ConditionsFilterService < FilterService
  def perform
    return false if AutomationRules::ConditionValidationService.new(rule: rule).perform

     # Do the rest and fetch the matching records
  end
end
```

In addition to this, I would suggest disabling the query if we find any invalid conditions. That would avoid future runs as well.

I hope this helps.

scmmishra:
@pranavrajs @sojan-official made the changes, now if any of the rule is invalid, the automation rule won't run at all. 

I figured, it's not a good idea to patch the conditions ourself in the code, since this can cause automation to trigger for scenarios the user never intended to

sojan-official:
Note for the follow-up task: we could mark the automation as disabled if the validation fails and render the status as inactive in the UI for better indication for the user. 

The code looks good to me. Could you also take care of the code climate issue?

scmmishra:
> we could mark the automation as disabled if the validation fails and render the status as inactive in the UI for better indication for the user.

I was about to implement this but figured there is no way for the user to know that the automation is disabled, what do you suggest? We can automatically send an email to the admin about this

sojan-official:
@scmmishra  Yep. sending an email to the admin sounds good 👍

sojan-official (approved):

PR #8677, opened 2024-01-10
chore: Fix sentry errors in email processing for bounce notifications

sojan-official:
This case occurs for bounce notification emails where the from address is `From: "" (Mail Delivery System)` . We will be discarding these emails for now.

Fixes: https://linear.app/chatwoot/issue/CW-2793/activerecordrecordinvalid-validation-failed-email-invalid-email

### DIFF
--- a/app/mailboxes/support_mailbox.rb
+++ b/app/mailboxes/support_mailbox.rb
@@ -7,6 +7,8 @@ class SupportMailbox < ApplicationMailbox
                     :decorate_mail
 
   def process
+    Rails.logger.info "Processing email #{mail.message_id} from #{original_sender_email} to #{mail.to} with subject #{mail.subject}"
+
     # to turn off spam conversation creation
     return unless @account.active?
     # prevent loop from chatwoot notification emails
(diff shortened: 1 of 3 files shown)


app/mailboxes/support_mailbox.rb
```diff
@@ -12,6 +12,14 @@ def process
     # prevent loop from chatwoot notification emails
     return if notification_email_from_chatwoot?
 
+    # return if email doesn't have a valid sender
+    # This can happen in cases like bounce emails for invalid contact email address
+    # TODO: Handle the bounce seperately and mark the contact as invalid
+    # we are checking for @ since the returned value could be "\"\"" for some email clients
+    return unless original_sender_email.include?('@')
+
+    Rails.logger.info "Processing email from #{original_sender_email} to #{mail.to} with subject #{mail.subject}"
```
pranavrajs, line 21: Can we add this log above?

pranavrajs: With this implementation, we would miss the emails with invalid to. Also I would suggest logging the email id.

pranavrajs (approved):

PR #8679, opened 2024-01-10
fix: Check the file attachment URL exists before downloading attachments in Telegram.

muhsin-k:
**Where did it impact, how much did it impact?**

There are 413 events reported in the sentry.

We use the [getFile](https://core.telegram.org/bots/api#getfile) method provided by Telegram to locate the attachment path. By using HTTParty, we retrieve the attachment path using this getFile method, and subsequently download the file using the *Down* gem and the get the file. 

However, there is a [condition](https://github.com/chatwoot/chatwoot/blob/develop/app/models/channel/telegram.rb#L52) that if the request fails for any reason, nil should be returned. In such cases, passing nil as an argument to the download method will result in the following error.

Untitled

**Solution**

I have added a check to `get_telegram_file_path` is blank or not.


app/services/telegram/incoming_message_service.rb
```diff
@@ -100,6 +100,8 @@ def file_content_type
   def attach_files
     return unless file
 
+    return if inbox.channel.get_telegram_file_path(file[:file_id]).blank?
```
sojan-official, line 103: lets assign the value to to a variable 

```
file_download_path = inbox.channel.get_telegram_file_path(file[:file_id])
return if file_download_path.blank?
```

let's also add a spec for the same case .. stubing the API call to get download path to return error

sojan-official (commented):
few minor comments

app/services/telegram/incoming_message_service.rb
```diff
@@ -100,6 +100,9 @@ def file_content_type
   def attach_files
     return unless file
 
+    file_download_path = inbox.channel.get_telegram_file_path(file[:file_id])
```
scmmishra, line 103: WDYT of using a function to check if the URL is valid?

```rb
def valid_url?(url)
  return false if url.nil? || url.strip.empty?

  uri = URI.parse(url)
  %w[http https].include?(uri.scheme) && !uri.host.nil?
rescue URI::InvalidURIError
  false
end
```

muhsin-k: The download URL is always valid here. We are handling the case when the download response is empty for some reason.

sojan-official (approved):

PR #8681, opened 2024-01-10
fix: Twilo attachment download fallback to the step where authorization headers are not passed.

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-2911/downclienterror-400-bad-request-downclienterror

**Where did it impact, how much did it impact?**

In the previous release, we enabled "HTTP Basic Authentication" to ensure that all attachments requiring HTTP authentication are secured. This is particularly important for media files that may contain sensitive data, as recommended by Twilio. However, some users experienced issues because they did not enable this option, despite our alerts prompting them to do so. (There are 2.6K events in sentry).

**Solution**

If the authenticated attachment download call fails, add another call to download the attachment without authentication.

### DIFF
--- a/app/services/twilio/incoming_message_service.rb
+++ b/app/services/twilio/incoming_message_service.rb
@@ -137,7 +137,9 @@ class Twilio::IncomingMessageService
     )
   end
 
+  # This is just a temporary workaround since some users have not yet enabled media protection. We will remove this in the future.
   def handle_download_attachment_error(error)
+    logger.info "Error downloading attachment from Twilio: #{error.message}"
     if error.message.include?('401 Unauthorized')
       Down.download(params[:MediaUrl0])
     else
--- a/app/services/twilio/incoming_message_service.rb
+++ b/app/services/twilio/incoming_message_service.rb
@@ -141,7 +141,7 @@ class Twilio::IncomingMessageService
     if error.message.include?('401 Unauthorized')
       Down.download(params[:MediaUrl0])
     else
-      ChatwootExceptionTracker.new(e, account: @inbox.account).capture_exception
+      ChatwootExceptionTracker.new(error, account: @inbox.account).capture_exception
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/services/twilio/incoming_message_service.rb
```diff
@@ -126,4 +122,26 @@ def attach_files
...
+    Down.download(
+      params[:MediaUrl0],
+      # https://support.twilio.com/hc/en-us/articles/223183748-Protect-Media-Access-with-HTTP-Basic-Authentication-for-Programmable-Messaging
+      http_basic_authentication: [twilio_channel.account_sid, twilio_channel.auth_token || twilio_channel.api_key_sid]
+    )
+  end
+
+  def handle_download_attachment_error(error)
+    if error.message.include?('401 Unauthorized')
+      Down.download(params[:MediaUrl0])
```
scmmishra, line 142: What if this has an error? That is not handled right?

scmmishra, line 142: We probably should log something to track the retry

app/services/twilio/incoming_message_service.rb
```diff
@@ -126,4 +122,26 @@ def attach_files
...
+  def download_with_auth
+    Down.download(
+      params[:MediaUrl0],
+      # https://support.twilio.com/hc/en-us/articles/223183748-Protect-Media-Access-with-HTTP-Basic-Authentication-for-Programmable-Messaging
+      http_basic_authentication: [twilio_channel.account_sid, twilio_channel.auth_token || twilio_channel.api_key_sid]
+    )
+  end
+
+  def handle_download_attachment_error(error)
+    if error.message.include?('401 Unauthorized')
```
scmmishra, line 141: Is there a way we can get an HTTP error code instead of checking against a string?

scmmishra (changes requested):

app/services/twilio/incoming_message_service.rb
muhsin-k: We are encountering the 401 Unauthorized error for users who have not yet enabled media protection. We will exclude all other codes.
@sojan-official thoughts?

muhsin-k: Done

PR #8682, opened 2024-01-10
feat: fallback to DB localStorage for idb names

scmmishra:
The `window.indexedDB.databases()` function is not available in all browsers. It's a non-standard method that's currently only supported in Chrome and Opera. This method returns a promise that resolves to a list of all database names associated with the origin of the calling script.

On Firefox, we won't be able to use this method as it's not supported. Instead, we will need to keep track of your database names in another way. One common approach is to store the names of your databases in a separate IndexedDB database or LocalStorage. This PR implements exactly that!

### DIFF
--- a/app/javascript/dashboard/helper/CacheHelper/DataManager.js
+++ b/app/javascript/dashboard/helper/CacheHelper/DataManager.js
@@ -10,6 +10,7 @@ export class DataManager {
 
   async initDb() {
     if (this.db) return this.db;
+    const dbName = `cw-store-${this.accountId}`;
     this.db = await openDB(`cw-store-${this.accountId}`, DATA_VERSION, {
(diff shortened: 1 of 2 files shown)


scmmishra:
@muhsin-k can you review this whenever you get the time

pranavrajs:
@scmmishra How is this working now in Firefox if this is the case?

scmmishra:
> @scmmishra How is this working now in Firefox if this is the case?

@pranavrajs we store the IDB names with the key `cw-idb-names` in the localstorage, in case the `databases()` call fails, we fallback to the localstorage for names. The fallback will work for Firefox and other browsers, for Chrome based browsers the fallback won't be triggered

muhsin-k (approved):
LGTM

PR #8706, opened 2024-01-15
fix: Handle Contact import `MalformedCSVError`

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-2926/csvmalformedcsverror-any-value-after-quoted-field-isnt-allowed-in-line

**Where and how much impact did it have?**

So far, 56 incidents have been reported in the sentry.

The primary cause of this issue is when someone tries to upload an invalid CSV file.
I confirmed the reason for the issue by checking a couple of CSV files from the database.

**Why is the change you made the right solution?**

I have added a rescue condition to handle errors raised when the CSV is malformed, and an email report is sent to the customer.

CleanShot 2024-01-15 at 16 22 40@2x


sojan-official (approved):
This looks good

PR #8723, opened 2024-01-17
fix:  Notification count is incorrect when the number of notifications exceeds the page size.

muhsin-k:
The notification count is incorrect when the number of notifications exceeds the page size.


app/javascript/dashboard/routes/dashboard/notifications/components/NotificationsView.vue
```diff
@@ -11,6 +11,7 @@
       <table-footer
         :current-page="Number(meta.currentPage)"
         :total-count="meta.count"
+        :page-size="15"
```
muhsin-k, line 14: The page size is https://github.com/chatwoot/chatwoot/blob/develop/app/finders/notification_finder.rb#L4

sojan-official (approved):

PR #8733, opened 2024-01-18
feat: use `short_summary` for downloading reports [CW-2962]

scmmishra:
This PR removes redundant queries that are made when downloading reports.

The `generate_readable_report_metrics` method only returned three metrics, so there is no point in calculating the rest. This should significantly speed up report generation for downloads

```rb
  def generate_readable_report_metrics(report_metric)
    [
      report_metric[:conversations_count],
      time_to_minutes(report_metric[:avg_first_response_time]),
      time_to_minutes(report_metric[:avg_resolution_time])
    ]
  end
```

#### Before Fix

CleanShot 2024-01-18 at 13 08 42@2x

#### After Fix

CleanShot 2024-01-18 at 13 08 39@2x

### DIFF
--- a/app/builders/v2/report_builder.rb
+++ b/app/builders/v2/report_builder.rb
@@ -46,6 +46,14 @@ class V2::ReportBuilder
     }
   end
 
+  def short_summary
+    {
+      conversations_count: conversations.count,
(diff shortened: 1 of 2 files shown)


vishnu-narayanan (approved):
Good catch 🚀

PR #8743, opened 2024-01-18
fix: Update the message-id check to use messageId parsing from `mail` gem

pranavrajs:
- Update the message ID parsing logic to use mail gem
- Update the code to improve readability

Followup: I will refactor this to use service object and a cleaner code :)


pranavrajs:
@sojan-official To test:

- Completely new inbox. It should populate the mails from yesterday.
- Email sending and receiving actions should be working properly.

sojan-official:
## Testings 
- Verified that emails are fetched correctly 
- Verified email sending and receiving are working correctly

sojan-official (approved):

PR #8744, opened 2024-01-19
fix: raise_lock_acquisition_error if the job cannot set the lock

pranavrajs:
Consider a scenario where two jobs are concurrently accessing a job with a lock mechanism. If the second job accesses `lock_manager.locked?` before the first job called `lock_manager.lock(lock_key)`, it would return a false positive. At this point, both jobs can execute freely.

To address this issue, the following change ensures that only the current thread that has set the lock can execute, otherwise a LockAcquisitionError will be thrown.

### DIFF
--- a/app/jobs/mutex_application_job.rb
+++ b/app/jobs/mutex_application_job.rb
@@ -18,18 +18,22 @@ class MutexApplicationJob < ApplicationJob
     lock_key = format(key_format, *args)
     lock_manager = Redis::LockManager.new
 
-    if lock_manager.locked?(lock_key)
-      Rails.logger.warn "[#{self.class.name}] Failed to acquire lock on attempt #{executions}: #{lock_key}"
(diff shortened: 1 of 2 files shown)


scmmishra (commented):
```rb
begin
  if lock_manager.lock(lock_key)
    Rails.logger.info "[#{self.class.name}] Acquired lock for: #{lock_key} on attempt #{executions}"
    yield
  else
    Rails.logger.warn "[#{self.class.name}] Failed to acquire lock on attempt #{executions}: #{lock_key}"
    raise LockAcquisitionError, "Failed to acquire lock for key: #{lock_key}"
  end
ensure
  lock_manager.unlock(lock_key)
end
```

If we simply update the condition to this and don't implicitly check locked, it will work. The check for locked is the one that introduces the additional race condition

sojan-official (approved):

PR #8763, opened 2024-01-23
feat(perf): update query to do a simpler search [CW-2997]

scmmishra:
Message search would frequently timeout. The reason was that query would join conversation too, the new query searches the message table directly

#### Old Query

Explain: https://explain.dalibo.com/plan/ba6d2g39c441edca
Duration: `365ms`

```sql
SELECT
  messages.*
FROM
  messages
  INNER JOIN conversations ON messages.conversation_id = conversations.id
WHERE
  conversations.inbox_id = 123
  AND messages.source_id = 'conversation/some-id'
ORDER BY
  messages.created_at ASC
LIMIT 1;
```

#### New Query

Explain: https://explain.dalibo.com/plan/34c9423c61515g73
Duration: `1.7ms`

```sql
SELECT
  messages.*
FROM
  messages
WHERE
  messages.inbox_id = 123
  AND messages.source_id = 'conversation/some-id'
ORDER BY
  messages.created_at ASC
LIMIT 1;
```

For previous instance, the messages was accessed via the conversation association, the new one skips conversation all-together. 

### Notes

We could further speed this up significantly with a index on `created_at`, `inbox_id` and `source_id` on the messages table. However it's best to add it only if this occurs again, any new index added an overhead our insert and update performance as well.

> The fastest query is the one that's not executed
> ~ Shivam Mishra

### DIFF
--- a/app/jobs/inboxes/fetch_imap_emails_job.rb
+++ b/app/jobs/inboxes/fetch_imap_emails_job.rb
@@ -137,7 +137,7 @@ class Inboxes::FetchImapEmailsJob < MutexApplicationJob
   end
 
   def email_already_present?(channel, message_id)
-    Message.find_by(source_id: message_id, inbox_id: channel.inbox).present?
+    channel.inbox.messages.find_by(source_id: message_id).present?
   end
 
   def build_mail_from_string(raw_email_content)
--- a/app/models/inbox.rb
+++ b/app/models/inbox.rb
@@ -66,7 +66,7 @@ class Inbox < ApplicationRecord
   has_many :inbox_members, dependent: :destroy_async
   has_many :members, through: :inbox_members, source: :user
   has_many :conversations, dependent: :destroy_async
-  has_many :messages, through: :conversations
+  has_many :messages, dependent: :destroy_async
 
   has_one :agent_bot_inbox, dependent: :destroy_async
   has_one :agent_bot, through: :agent_bot_inbox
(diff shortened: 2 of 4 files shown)


scmmishra:
The commit https://github.com/chatwoot/chatwoot/pull/8763/commits/9a1af32899acd40f09c25ccb55177362535ff329 introduces a change that in theory should fix this issue across the app. Instead, it surfaces a really interesting quirk (which can look like a bug too)

The test `imap_mailbo_spec.rb` has the following case

https://github.com/chatwoot/chatwoot/blob/d0cd1c8887ccd3bf5685fdc88431c72e485322e4/spec/mailboxes/imap/imap_mailbox_spec.rb#L45-L80

This fails the moment we change the relation between `message` and `inbox` models to this `has_many :messages, dependent: :destroy_async` from `has_many :messages, through: :conversations`

Why does this happen? Well, if you see at the start of the test we make the channel and inbox

https://github.com/chatwoot/chatwoot/blob/d0cd1c8887ccd3bf5685fdc88431c72e485322e4/spec/mailboxes/imap/imap_mailbox_spec.rb#L9-L14

Here is what happens when the test is run

- When the channel is created, a inbox `inbox#1` is automatically created with it, the next line simply creates another inbox `inbox#2`.
- On line `46` when the conversation is created, the inbox value is set to `channel.inbox` which means `inbox#1`
- On line`56` and `64` the messages are created with `inbox`, which means `inbox#2`
- Now when the message is processed, since there is no association between inbox-conversation-message, the message created is not properly associated to the correct inbox.

When the association existed previously, the correct `inbox_id` was automatically set from the `conversation`

sojan-official:
- note from the discussion -> We will fix the spec since it isn't ideal case in the system to have a single channel belonging to two inboxes.  'one channel can belong only to one inbox'

sojan-official (approved):

PR #8766, opened 2024-01-24
fix(refactor): Cleanup the specs and the logic for Fetch IMAP job

pranavrajs:
- Created a fetch_email_service with a base class and one for Microsoft. 
- Rewrote the specs to cover the correct use cases.

Fixes https://linear.app/chatwoot/issue/CW-3011/imap-issues
Fixes https://linear.app/chatwoot/issue/CW-2997/activerecordquerycanceled-pgquerycanceled-error-canceling-statement


sojan-official (approved):
Code. LGTM .

- tested and verified with google accounts in my local environment. 

Let merge it once we verify with Microsoft account as well . @pranavrajs

PR #8770, opened 2024-01-24
fix: mutex timeout and error handling

scmmishra:
This PR fixes two issues

1. The `ensure` block released the lock even on `LockAcquisitionError`
2. Custom timeout was not allowed

This also refactored the `with_lock` method, now the key has to be constructed in the parent function itself


app/jobs/inboxes/fetch_imap_emails_job.rb
```diff
@@ -6,7 +6,8 @@ class Inboxes::FetchImapEmailsJob < MutexApplicationJob
   def perform(channel)
     return unless should_fetch_email?(channel)
 
-    with_lock(::Redis::Alfred::EMAIL_MESSAGE_MUTEX, inbox_id: channel.inbox.id) do
+    key = format(::Redis::Alfred::EMAIL_MESSAGE_MUTEX, inbox_id: channel.inbox.id)
+    with_lock(key) do
```
sojan-official, line 10: this job can be long running. so sticking to safe side

```suggestion
    with_lock(key, 5.minutes) do
```

sojan-official (commented):
minor comments

sojan-official (approved):

sojan-official (approved):

PR #8779, opened 2024-01-24
fix: message created_at receive timestamp field in whatsapp cloud api

clairton:
A simple fix to use field timestamp from webhooks payloads in whatsapp cloud api

https://developers.facebook.com/docs/whatsapp/cloud-api/webhooks/payload-examples

### DIFF
--- a/app/services/whatsapp/incoming_message_base_service.rb
+++ b/app/services/whatsapp/incoming_message_base_service.rb
@@ -144,6 +144,7 @@ class Whatsapp::IncomingMessageBaseService
       message_type: :incoming,
       sender: @contact,
       source_id: message[:id].to_s,
+      created_at: Time.at(message[:timestamp], in: 'UTC'),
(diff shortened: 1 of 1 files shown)


pranavrajs:
@clairton Thanks for bringing this up. We will discuss this internally before moving forward. I believe it is necessary to preserve the timestamp from the source. We need to support this in other channels, so we will investigate this for all other channels and proceed once we have a proper product spec.

sojan-official:
Note: We can populate this on external_created_at

sojan-official:
Hey @clariton, created_at is a Rails-managed field and shouldn’t be modified directly. In Chatwoot, parts of the application depend on created_at being in sync with the primary key. Changing it can introduce unexpected issues.

 Except in edge cases there shouldn't be any delays in Chatwoot on processing the meta web-hooks so the created_at should be really close to the created_at value from meta, so lets reopen this later if its still required.

PR #8782, opened 2024-01-25
feat: disable upload after one file for IG and FB messenger

scmmishra:
It seems like it's not possible to send multiple files in a single message via the instagram API. This PR disables the file upload once a single attachment is added.

The actual fix to this would be to update `SendOnInstagramService` and `SendOnFacebookService` to send these attachments as multiple different messages, but this may take some time, the UI change is a stopgap solution for now


scmmishra:
Closing this since https://github.com/chatwoot/chatwoot/pull/8783 is close to completion and is tested

PR #8804, opened 2024-01-29
feat: new accounts controller for signup+onboarding

scmmishra:
This PR adds controllers for the new signup flow, this flow only requires email and password.

- Add a new `V2::AccountsController` and update routes for v2 namespace
- Update `Account Builder` to handle missing `account_name` and `user_full_name`
- Remove model validations in `Account` and `User` to allow not having `name` param
- Add validation in `V1::AccountsController`, to ensure `account_name` and `user_full_name` is present, to ensure backward compatibility
- Update tests for `V1::AccountsController`
- Add tests for `V2::AccountsController`

## Type of change

## How Has This Been Tested?

- [x] Unit Tests
- [x] Manually on staging


app/controllers/api/v2/accounts_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    raise ActionController::InvalidAuthenticityToken, 'Invalid Captcha' unless ChatwootCaptcha.new(params[:h_captcha_client_response]).valid?
+  end
+
+  # def pundit_user
+  #   {
+  #     user: current_user,
+  #     account: @account,
+  #     account_user: @current_account_user
+  #   }
+  # end
```
sojan-official, line 55: we can remove this right ?

config/locales/en.yml
```diff
@@ -30,7 +30,7 @@
 # available at https://guides.rubyonrails.org/i18n.html.
 
 en:
-  hello: "Hello world"
```
sojan-official, line 33: We can revert the change en.yml, which is not relevant to this PR. not to pollute the history

config/routes.rb
```diff
@@ -287,16 +287,18 @@
...
-          collection do
-            get :summary
-            get :agents
-            get :inboxes
-            get :labels
-            get :teams
-            get :conversations
-            get :conversation_traffic
+      resources :accounts, only: [:create] do
+        scope module: :accounts do
```
sojan-official, line 291:
```
resources :accounts, only: [:create], module: :accounts do
end
```

won't this work ?  again not to impact the routes which are not relevant to this PR

sojan-official (commented):
LGTM mostly. 
couple of minor comments

sojan-official (approved):

PR #8805, opened 2024-01-29
chore: Move EE OpenAI spec to correct folder

sojan-official:
- we previously had this spec inside enterprise folder which wouldn't be picked up by our build process, so moving to the correct folder instead.


sojan-official:
@scmmishra  could you take a look at the failing specs ?

scmmishra:
> @scmmishra  could you take a look at the failing specs ?

Sure

scmmishra (approved):

PR #8808, opened 2024-01-29
feat: show banner if email is unverified

scmmishra:
This PR adds the following changes

1. Show verification banner if the user email is not confirmed
2. Add endpoint to resend confirmation, along with throttling for the endpoint
3. Update `banner` component to allow custom icon

CleanShot 2024-01-29 at 18 14 24@2x


app/controllers/api/v1/profiles_controller.rb
```diff
@@ -31,6 +31,12 @@ def set_active_account
     head :ok
   end
 
+  def resend_confirmation
+    return @user.send_confirmation_instructions if @user
```
pranavrajs, line 35: Should we check if the user is confirmed already?

app/javascript/dashboard/App.vue
```diff
@@ -2,12 +2,13 @@
...
-    class="app-wrapper h-full flex-grow-0 min-h-0 w-full"
+    class="flex-grow-0 w-full h-full min-h-0 app-wrapper"
     :class="{ 'app-rtl--wrapper': isRTLView }"
     :dir="isRTLView ? 'rtl' : 'ltr'"
   >
     <update-banner :latest-chatwoot-version="latestChatwootVersion" />
     <template v-if="currentAccountId">
+      <pending-email-verification-banner />
       <payment-pending-banner />
       <upgrade-banner />
```
pranavrajs, line 13: Shall we combine these and switch message and style?

app/controllers/api/v1/profiles_controller.rb
scmmishra: Yep we can, I'll update the PR

app/javascript/dashboard/App.vue
scmmishra: They use a single `<banner>` component in the markup, the separate component is just to encapsulate the visibility logic and actions. I can merge the three but then the component will become too complicated. WDYT?

spec/controllers/api/v1/profiles_controller_spec.rb
```diff
@@ -242,4 +242,26 @@
...
+    end
+
+    context 'when it is an authenticated user' do
+      let(:agent) { create(:user, password: 'Test123!', account: account, role: :agent) }
+
+      it 'resends the confirmation email' do
+        post '/api/v1/profile/resend_confirmation',
+             headers: agent.create_new_auth_token,
+             as: :json
+
```
sojan-official, line 262: let's add an expectation here that email is being sent as well.

sojan-official (approved):

app/javascript/dashboard/components/app/PendingEmailVerificationBanner.vue
```diff
@@ -0,0 +1,43 @@
...
+      return this.$t('APP_GLOBAL.RESEND_VERIFICATION_MAIL');
+    },
+    shouldShowBanner() {
+      return !this.currentUser.confirmed;
+    },
+  },
+  methods: {
+    resendVerificationEmail() {
+      this.$store.dispatch('resendConfirmation');
+      this.showAlert(this.$t('APP_GLOBAL.EMAIL_VERIFICATION_SENT'));
```
pranavrajs, line 39: @scmmishra There is a mixin you can use to show the alerts

pranavrajs (approved):

PR #8820, opened 2024-01-31
chore: Refactor RefreshOauthTokenService to improve readability

pranavrajs:
- Added a trait called `microsoft_email` for the Channel::Email factory.
- Rewrote the logic to make it simple to understand
- Rewrote the specs for readability

### DIFF
--- a/spec/jobs/inboxes/fetch_imap_emails_job_spec.rb
+++ b/spec/jobs/inboxes/fetch_imap_emails_job_spec.rb
@@ -8,11 +8,8 @@ RSpec.describe Inboxes::FetchImapEmailsJob do
     create(:channel_email, imap_enabled: true, imap_address: 'imap.gmail.com', imap_port: 993, imap_login: 'imap@gmail.com',
                            imap_password: 'password', imap_inbox_synced_at: Time.now.utc, account: account)
   end
-  let(:microsoft_imap_email_channel) do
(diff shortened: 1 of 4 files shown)


spec/jobs/inboxes/fetch_imap_emails_job_spec.rb
```diff
@@ -8,11 +8,8 @@
     create(:channel_email, imap_enabled: true, imap_address: 'imap.gmail.com', imap_port: 993, imap_login: 'imap@gmail.com',
                            imap_password: 'password', imap_inbox_synced_at: Time.now.utc, account: account)
   end
-  let(:microsoft_imap_email_channel) do
-    create(:channel_email, provider: 'microsoft', imap_enabled: true, imap_address: 'outlook.office365.com',
-                           imap_port: 993, imap_login: 'imap@outlook.com', imap_password: 'password', account: account,
-                           provider_config: { access_token: 'access_token' })
-  end
+
+  let(:microsoft_imap_email_channel) { create(:channel_email, :microsoft_email) }
```
pranavrajs, line 12: This change is done because the spec `when Imap fetch new emails for Microsoft mailer` broke with this change. 

Earlier, if expires_on was empty, the check on the access token validity was done against the current time, which would always return true. The refresh service was never called. This change would force a refresh service call and update the credentials. Since there was not stub for the external service calls, the spec broke. Instead of adding the case where token is invalid, I've added a channel with a valid token.

sojan-official (approved):

PR #8833, opened 2024-01-31
fix: Use BODY.PEEK[HEADER] to avoid parsing issues with mail providers

pranavrajs:
- Use `BODY.PEEK[HEADER]` instead of `BODY[HEADER.FIELDS (MESSAGE-ID)]`
- Remove `BEFORE` param from the `FETCH` operation.


sojan-official (approved):

app/jobs/inboxes/fetch_imap_emails_job.rb
```diff
@@ -85,7 +90,7 @@ def fetch_message_ids_with_sequence(imap_client, channel)
       end
 
       batch_message_ids.each do |data|
-        message_id = build_mail_from_string(data.attr['BODY[HEADER.FIELDS (MESSAGE-ID)]']).message_id
+        message_id = build_mail_from_string(data.attr['BODY.PEEK[HEADER]']).message_id
```
sojan-official, line 93:
```suggestion
        message_id = build_mail_from_string(data.attr['BODY.[HEADER]']).message_id
```
While testing, emails weren't getting created on my local machine.  I believe this is different from the current deployment.

sojan-official:
@pranavrajs  should we look into adding specs for this case, which the current specs didn't catch or take up separately?

spec/jobs/inboxes/fetch_imap_emails_job_spec.rb
```diff
@@ -53,25 +52,19 @@
...
       imap = double
-
       allow(Net::IMAP).to receive(:new).and_return(imap)
       allow(imap).to receive(:authenticate)
       allow(imap).to receive(:select)
       allow(imap).to receive(:search).and_return([1])
-      allow(imap).to receive(:fetch).and_return([imap_fetch_mail])
-
-      allow(Mail).to receive(:read_from_string).and_return(inbound_mail_with_no_date.mail)
+      allow(imap).to receive(:fetch).with([1], 'BODY.PEEK[HEADER]').and_return([email_header])
```
pranavrajs, line 66:
```suggestion
      allow(imap).to receive(:fetch).with([1], 'BODY[HEADER]').and_return([email_header])
```

pranavrajs:
@sojan-official, I will clean it up here (https://github.com/chatwoot/chatwoot/pull/8766).

sojan-official (approved):

PR #8850, opened 2024-02-05
feat: Form input component to use with onboarding form

nithindavid:
**Description**
Adds form input component that needs to be used across onboarding wizard.

image

image


app/javascript/v3/components/Form/Input.vue
```diff
@@ -1,44 +1,42 @@
...
-    </div>
-  </div>
+  <with-label
+    :label="label"
+    :icon="icon"
+    :name="name"
+    :has-error="hasError"
+    :error-message="errorMessage"
+  >
+    <template #label> {{ label }} <slot /> </template>
```
scmmishra, line 9: We don't need the slot usage here since the label is already passed as a prop

nithindavid:
@scmmishra Added the changes, please take a look

scmmishra (approved):

PR #8854, opened 2024-02-05
feat: Wizard step component to use with onboarding

nithindavid:
**Description**

Wizard step tracking component to show progress in onboarding flow.

image


app/javascript/v3/views/onboarding/OnboardingStep.vue
```diff
@@ -0,0 +1,60 @@
...
+      required: true,
+    },
+    isActive: {
+      type: Boolean,
+      default: false,
+    },
+    hidden: {
+      type: Boolean,
+      default: false,
+    },
```
scmmishra, line 48:
```suggestion
```

We can remove the hidden prop and just use a v-if in the component usage

app/javascript/v3/views/onboarding/OnboardingStep.vue
```diff
@@ -0,0 +1,60 @@
+<template>
+  <div
+    v-if="!hidden"
+    class="flex items-center gap-2 p-2 text-lg font-bold mb-6 relative before:absolute before:h-10 before:w-[1px] before:bg-slate-200 before:-bottom-8 before:left-[24px] hide-before-of-last"
+    :class="{
+      'text-woot-500 ': isActive,
+      'text-slate-400': (!isActive && !isComplete) || isComplete,
```
scmmishra, line 7:
```suggestion
      'text-slate-400': !isActive || isComplete,
```

We can simplify the condition,  since if `isComplete` is true, the condition will always be true, regardless of `isActive`.

app/javascript/v3/views/onboarding/OnboardingStep.vue
```diff
@@ -0,0 +1,60 @@
...
+      'before:bg-woot-500': !isActive && isComplete,
+      hidden: hidden,
+    }"
+  >
+    <div
+      class="grid w-8 h-8 border border-solid rounded-full place-content-center"
+      :class="{
+        'border-woot-500': (isActive && !isComplete) || isComplete,
```
scmmishra, line 15:
```suggestion
      'border-woot-500': !isActive || isComplete,
```

We can simplify the condition,  since if `isComplete` is true, the condition will always be true, regardless of `isActive`.

scmmishra (changes requested):

nithindavid:
@scmmishra I have added the changes suggested. Please take a look

PR #8864, opened 2024-02-06
chore: Apply fixes for items in rubocop_todo [CW-1806]

sojan-official:
This PR addresses several items listed in our rubocop_todo by implementing the necessary corrections and enhancements. As a result, we are now able to remove the rubocop_todo file entirely, streamlining our codebase and ensuring adherence to our coding standards.

fixes: https://linear.app/chatwoot/issue/CW-1806/chore-rubocop-audit


app/models/macro.rb
```diff
@@ -21,7 +21,7 @@ class Macro < ApplicationRecord
 
   belongs_to :account
   belongs_to :created_by,
-             class_name: :User, optional: true
+             class_name: :User, optional: true, inverse_of: :macros
```
pranavrajs, line 24: I'm seeing the usage of inverse_of multiple places. Why do we need these?

lib/events/base.rb
```diff
@@ -9,6 +9,6 @@ def initialize(name, timestamp, data)
   end
 
   def method_name
-    name.to_s.gsub('.', '_')
+    name.to_s.tr('.', '_')
```
pranavrajs, line 12: Any reason why this is needed?

spec/mailers/confirmation_instructions_spec.rb
```diff
@@ -2,8 +2,8 @@
 
 require 'rails_helper'
 
-RSpec.describe 'Confirmation Instructions' do
-  describe :notify do
+RSpec.describe 'Devise::Mailer' do
```
pranavrajs, line 5: Are we testing Devise::Mailer here, doesn't seem appropriate,

spec/models/contact_spec.rb
```diff
@@ -5,11 +5,11 @@
 require Rails.root.join 'spec/models/concerns/avatarable_shared.rb'
 
 RSpec.describe Contact do
-  context 'validations' do
+  context 'when validations' do
```
pranavrajs, line 8:
```suggestion
  context 'with validations' do
```

spec/models/contact_spec.rb
```diff
@@ -5,11 +5,11 @@
...
     it { is_expected.to validate_presence_of(:account_id) }
   end
 
-  context 'associations' do
+  context 'when associations' do
```
pranavrajs, line 12:
```suggestion
  context 'with associations' do
```

spec/models/conversation_spec.rb
```diff
@@ -193,7 +193,7 @@
     it 'creates conversation activities' do
       conversation.update(
         status: :resolved,
-        contact_last_seen_at: Time.now,
+        contact_last_seen_at: Time.zone.now,
```
pranavrajs, line 196: Do we need to use `travel_to` to avoid flaky tests.

spec/models/user_spec.rb
```diff
@@ -7,11 +7,11 @@
 RSpec.describe User do
   let!(:user) { create(:user) }
 
-  context 'validations' do
+  context 'when validations' do
```
pranavrajs, line 10: I guess with validations is a better option than when validations as mentioned in a comment above.

pranavrajs (approved):
Minor comments, otherwise looks good.

spec/mailers/confirmation_instructions_spec.rb
sojan-official: the underlying class being tested here is Devise::Mailer , there is no confirmation mailer available in the code base. 

```
confirmation_mail = Devise::Mailer.confirmation_instructions(confirmable_user.reload, nil, {})
```

lib/events/base.rb
sojan-official: its an automatic rubocop substitution for better performance 

```
gsub is used for replacing substrings with new substrings, while tr is used for replacing individual characters with other characters.
```

app/models/macro.rb
sojan-official: https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html

```
The automatic guessing of the inverse association uses a heuristic based on the name of the class, so it may not work for all associations, especially the ones with non-standard names
```

ref: https://stackoverflow.com/a/39670478/939299

```
 It's a hint to ActiveRecord to use already loaded data instead of fetching it again through a relationship.
 ```

spec/models/conversation_spec.rb
sojan-official: this test is not dependant on the time value we provide here

PR #8894, opened 2024-02-08
feat: Interface to validate response_source

sojan-official:
This PR adds a UI to validate the response source quality quickly.  It also helps to test with sample questions and update responses in the database when missing.

https://www.loom.com/share/19a6f15db6d14548a5a3c8f65a82f2a6


enterprise/app/controllers/super_admin/response_sources_controller.rb
```diff
@@ -41,4 +41,52 @@ class SuperAdmin::ResponseSourcesController < SuperAdmin::ApplicationController
...
+  def render_vue_component(component_name, props = {})
+    html_options = {
+      id: 'app',
+      data: {
+        component_name: component_name,
+        props: props.to_json
+      }
+    }
+    content_tag(:div, '', html_options)
+  end
```
pranavrajs, line 59: @sojan-official Can you please check if you can move this to base controller? If so, the render method can be used anywhere in the code.

pranavrajs (approved):

PR #8907, opened 2024-02-12
fix: Inbox view Read/Snoozed display filters

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-3065/read-snoozed-display-options-are-not-working-properly

Summary of all the cases.

- When "Read" and "Snooze" are selected-> Show all notifications including read, unread, and snoozed.
- When "Read" is selected and "Snooze" is unselected -> Show only read and unread notifications and do not show snoozed notifications.
- When "Snooze" is selected and "Read" is unselected -> Show only snooze and unread notifications and do not show read notifications.
- When neither "Read" nor "Snooze" is selected -> Show only unread notifications.


app/finders/notification_finder.rb
```diff
@@ -26,20 +26,41 @@ def count
...
+
+  def exclude_read_notifications
+    @notifications = @notifications.where(read_at: nil)
   end
 
-  def filter_by_read_status
-    @notifications = @notifications.where.not(read_at: nil) if params[:type] == 'read'
+  def include_unread_notifications
+    @notifications = @notifications.where('snoozed_until IS NULL AND read_at IS NULL')
   end
```
pranavrajs, line 64: Can't we simplify this to the following? 

```rb

def set_up
  find_all_notifications
  filter_snoozed_notifications
  fitler_read_notifications
end

def find_all_notifications
  @notifications = current_user.notifications.where(account_id: @current_account.id)  
end

def filter_snoozed_notifications
  @notifications = @notifications.where.not(snoozed_until: nil) if params[:status].empty?
end

def fitler_read_notifications
  @notifications = @notifications.where.not(read_at: nil) if params[:type] != 'read'
end
```

pranavrajs: I would also suggest to change the params to something like the following.

Instead of 
```
/notifications?type=read&status=snoozed
```

let us change it to

```
/notifications?includes[]=snoozed&includes[]=read
```

This would make the filters easy to understand.

The finder would be simplified to

```
class NotificationFinder
  def set_up
    find_all_notifications
    filter_snoozed_notifications
    fitler_read_notifications
  end

  def find_all_notifications
    @notifications = current_user.notifications.where(account_id: @current_account.id)  
  end

  def filter_snoozed_notifications
    @notifications = @notifications.where(snoozed_until: nil) unless type_included?('snoozed')
  end

  def fitler_read_notifications
    @notifications = @notifications.where(read_at: nil) unless type_included?('read')
  end

  def type_included?(type)
    (params[:includes] || []).includes?(type)
  end
end
```

pranavrajs (changes requested):
Please check the comment about the implementation above.

pranavrajs (approved):

PR #8920, opened 2024-02-13
fix: throttle conversation search endpoint to prevent abuse

muhsin-k:
Throttle conversation search endpoint to prevent abuse/maintain db perf

ref: https://discord.com/channels/897869226579222540/1206919316402999326/1206953097444720711

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -154,6 +154,12 @@ class Rack::Attack
     match_data[:account_id] if match_data.present?
   end
 
+  # Prevent abuse of conversation search api
+  throttle('/api/v1/accounts/:account_id/conversations/search', limit: 5, period: 1.minute) do |req|
+    match_data = %r{/api/v1/accounts/(?<account_id>\d+)/conversations/search}.match(req.path)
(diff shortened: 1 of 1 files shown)


config/initializers/rack_attack.rb
```diff
@@ -154,6 +154,12 @@ def path_without_extentions
     match_data[:account_id] if match_data.present?
   end
 
+  # Prevent abuse of conversation search api
+  throttle('/api/v1/accounts/:account_id/conversations/search', limit: 5, period: 1.minute) do |req|
```
sojan-official, line 158: The search API which we use at the moment is `search/messages`, `search/conversations` etc

sojan-official (commented):
@muhsin-k let's keep this on hold and get to this if we see any further observance of abuse. Since we have autocomplete for search there is a high chance that we might hit this endpoint frequently.

muhsin-k:
> @muhsin-k let's keep this on hold and get to this if we see any further observance of abuse. Since we have autocomplete for search there is a high chance that we might hit this endpoint frequently.

Sure. Closing this PR now.

PR #8921, opened 2024-02-13
fix: downcase email before finding

scmmishra:
For `AgentBuilder` if the user is already present, we find it and use it to create an `AccountUser`. However the problem is that the `find_by` is not case sensitive, this means that if `ben@example.com` is already present, and the user tries to invite `Ben@example.com` the `find_by` returns nil, and an attempt is made to create a new user. This attempt fails at the DB validation stage, since email is supposed to be unique. 

This PR fixes this by updating the `find_by` query


scmmishra:
@pranavrajs there are other instances where `User.find_by` is called with email, should I update it there too?

pranavrajs:
@scmmishra Yes, please. Let us consider downcased email as the standard.

sojan-official (commented):
@scmmishra Can we look into adding a custom method for User and Contact models where we do email lookup which ensures that find by is always performed with param.downcase  ? We can also add a rubocop to ensure this stay consistent in the code base.  


ref: https://chat.openai.com/share/faf7d27b-6412-4349-b43c-854ffc974e28

PR #8934, opened 2024-02-14
chore: Hide banners on onboarding view

nithindavid:
## Description
This change will hide the payment pending and unverified email banners on onboarding page.


iamsivin (approved):

app/javascript/dashboard/components/app/PaymentPendingBanner.vue
```diff
@@ -34,6 +34,9 @@ export default {
     actionButtonMessage() {
       return this.$t('GENERAL_SETTINGS.OPEN_BILLING');
     },
+    isOnboardingView() {
```
pranavrajs, line 37: This shouldn't be handled here, it should be at the parent component. You can see the logic being repeated in two places.

app/javascript/dashboard/components/app/PaymentPendingBanner.vue
```diff
@@ -34,6 +34,9 @@ export default {
...
+      return this.$route.name.includes('onboarding_');
```
pranavrajs, line 38: Move this to a helper. Check actionsHelper.

pranavrajs (changes requested):
Please check the comments.

app/javascript/dashboard/components/app/PaymentPendingBanner.vue
nithindavid: @pranavrajs The parent doesn't handle any v-if conditions and all the v-if conditions are written here. If we move just this condition it would create inconsistencies IMHO.

pranavrajs (approved):

Issue #8888, opened 2024-02-08
Sockets are disconnected periodically


numb7:
### Describe the bug

Sockets are disconnected periodically and we can't see new messages in the admin without refreshing the page in the browser.
The problem is solved by overloading the self-hosted chatwoot server.

The monitoring system detected an error.
Label:
trigger = PodThrottlingGreaterThan100mCpuWithin10Min
Alert Description:
Container in service chatwoot-workers throttling greater than 0.1 CPU within 10 minutes: 0.16 CPU

Can you tell me what can cause 

Self-hosted chatwoot v 3.0.0

Translated with DeepL.com (free version)

### To Reproduce

you can't retrace the steps 

### Environment

Kubernetes

### Cloud Provider

None

### Platform

Browser

### Operating system

3.0.0

sojan-official:
ref: https://github.com/chatwoot/chatwoot/issues/6411

pranavrajs:
Duplicate https://github.com/chatwoot/chatwoot/issues/6411

pranavrajs:
This configuration would allow you to automatically resolve the conversation after a specific time period. Configure the time frame and personalized message to the user below.

When a disconnect occurs, the web-socket events intended for the UI are discarded and lost.

* Caches are present for Agents, Labels, Inboxes, and Teams. We need to verify whether these caches have been invalidated. If they have, we should use the APIs to refetch the models. This logic already exists in `account.cache_invalidated`.
* For the currently active conversation, fetch the messages were added / updated after the disconnection time.
* Retrieve the conversations that were updated after the disconnection time based on the active filters in the UI.

PR #8947, opened 2024-02-15
chore: Support for updated_within in ConverationFinder

sojan-official:
- `updated_within' accepts value in seconds and returns all conversations updated in the given period with out pagination. This API will assist in our refetch logic on socket disconnect  

ref: https://github.com/chatwoot/chatwoot/issues/8888

### DIFF
--- a/app/finders/conversation_finder.rb
+++ b/app/finders/conversation_finder.rb
@@ -168,11 +168,13 @@ class ConversationFinder
       :taggings, :inbox, { assignee: { avatar_attachment: [:blob] } }, { contact: { avatar_attachment: [:blob] } }, :team, :contact_inbox
     )
 
+    sort_by, sort_order = SORT_OPTIONS[params[:sort_by]] || SORT_OPTIONS['last_activity_at_desc']
+    @conversations = @conversations.send(sort_by, sort_order)
+
     if params[:updated_within].present?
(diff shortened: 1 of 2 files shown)


app/finders/conversation_finder.rb
```diff
@@ -168,7 +168,11 @@ def conversations
...
     )
 
-    sort_by, sort_order = SORT_OPTIONS[params[:sort_by]] || SORT_OPTIONS['last_activity_at_desc']
-    @conversations.send(sort_by, sort_order).page(current_page).per(ENV.fetch('CONVERSATION_RESULTS_PER_PAGE', '25').to_i)
+    if params[:updated_within].present?
+      @conversations = @conversations.where('conversations.updated_at > ?', Time.zone.now - params[:updated_within].to_i.seconds)
+    else
+      sort_by, sort_order = SORT_OPTIONS[params[:sort_by]] || SORT_OPTIONS['last_activity_at_desc']
+      @conversations.send(sort_by, sort_order).page(current_page).per(ENV.fetch('CONVERSATION_RESULTS_PER_PAGE', '25').to_i)
+    end
```
pranavrajs, line 176: Wouldn't this dismiss the sort order? I would suggest getting the sort order in place and dismissing only the page size.

pranavrajs (changes requested):
Minor comment, otherwise looks fine.

PR #8963, opened 2024-02-19
fix: `support_email` and `inbound_email_domain` returning empty string

sojan-official:
- While debugging a prod issue, we notified that if someone enabled `custom_reply_email' or `custom_reply_email` email features and updated their account. The account's values for attributes `domain` and `support_email` will change to "" instead of nil, which causes issues with email sending. Hence adding appropriate fixes for the same.


vishnu-narayanan (approved):
LGTM

PR #8973, opened 2024-02-20
feat: Conversation update API for sla_policy_id

sojan-official:
- Add an endpoint for updating conversation attributes (priority / sla_policy_id )
- Swagger spec 
- minor chores around the conversation API/spec

Fixes: https://linear.app/chatwoot/issue/CW-2100/feat-backend-api-to-update-the-sla-of-a-conversation


app/controllers/api/v1/accounts/conversations_controller.rb
```diff
@@ -36,6 +36,10 @@ def create
     end
   end
 
+  def update
+    @conversation.update!(permitted_update_params)
```
vishnu-narayanan, line 40: Would this allow updating a conversation with different sla_policies? We should restrict updating a `conversation` with another policy if SLA is already set.

spec/enterprise/models/conversation_spec.rb
```diff
@@ -34,4 +34,34 @@
...
+      it 'creates applied sla record if sla policy is present' do
+        conversation.sla_policy = sla_policy
+        conversation.save!
+        expect(conversation.applied_sla.sla_policy_id).to eq(sla_policy.id)
+      end
+
+      it 'deletes applied sla record if sla policy is removed' do
+        conversation.sla_policy = sla_policy
+        conversation.save!
+        conversation.sla_policy = nil
```
vishnu-narayanan, line 60: Removing `sla_policy` from a conversation should not delete `applied_sla` events. This will affect reporting. Associated events can be deleted if parent `sla_policy` is deleted.

app/controllers/api/v1/accounts/conversations_controller.rb
sojan-official: We can handle this validation in the model 👍

sojan-official:
Note: for now we will limit adding SLA only to conversations where SLA is not present.  other cases API will error

enterprise/app/models/applied_sla.rb
```diff
@@ -23,6 +23,13 @@ class AppliedSla < ApplicationRecord
   belongs_to :conversation
 
   validates :account_id, uniqueness: { scope: %i[sla_policy_id conversation_id] }
+  before_validation :ensure_account_id
 
   enum sla_status: { active: 0, hit: 1, missed: 2 }
```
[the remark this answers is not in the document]
sojan-official: will remove the case for removing SLAs in this PR

vishnu-narayanan (approved):

PR #8986, opened 2024-02-22
feat: use autoload_lib

scmmishra:


scmmishra:
This uses a option that is only available in rails 7.1

PR #8806, opened 2024-01-29
feat: Update the `first_name`, `last_name` and `middle_name` when the contact created

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-3013/update-the-last-name-and-middle-name-when-the-contact-updatedcreated

PR #8989, opened 2024-02-22
feat: Add contact helper

muhsin-k:
- Add contact helper.
- Add method to split full name in to parts (first_name, last_name, middle_name, prefix and suffix).

I have created this PR to reduce the complexity of the [PR](https://github.com/chatwoot/chatwoot/pull/8806).


app/helpers/contact_helper.rb
```diff
@@ -0,0 +1,18 @@
+module ContactHelper
+  def split_first_and_last_name(full_name)
```
pranavrajs, line 2: Found this https://gist.github.com/sukima/491187

muhsin-k: Nice one. Thanks for sharing. I have added more cases.

sojan-official (approved):

PR #9017, opened 2024-02-26
feat: disable automation rules if condition fails multiple times

scmmishra:
This PR adds a feature to disable automation rules if the conditions fail multiple times. It also sends an email to the administrators notifying them of this change.

## Type of change

## How Has This Been Tested?

- [x] Unit Tests
- [ ] Manual Test on Staging

## Checklist:


scmmishra:

app/models/automation_rule.rb
```diff
@@ -20,9 +20,13 @@
 class AutomationRule < ApplicationRecord
   include Rails.application.routes.url_helpers
 
+  MAX_ERROR_THRESHOLD = 3
```
sojan-official, line 23: can't we use reauthorizable concern for this case as well ?  seems like lot of logic is pretty common

sojan-official (commented):
@scmmishra  I think we can handle this via `reauthorizable` concern. we wouldn't need to include a lot of extra code.

app/models/automation_rule.rb
scmmishra: @sojan-official I've made the changes, let me test it out properly. I'll let you know once it's ready for review

app/models/concerns/error_trackable.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def threshold_breached?
+    error_count >= error_threshold
+  end
+
+  private
+
+  def error_count_key
+    format(::Redis::Alfred::ERROR_TRACKABLE_COUNT, obj_type: self.class.table_name.singularize, obj_id: id)
```
scmmishra, line 34: This PR changes the key to a standard one across the app, the Reauthorizable had a specific key. Is this fine? @sojan-official

sojan-official (approved):
@scmmishra made some slight changes to the code so that we can include the shared reauthorizable specs into the automation rule model specs as well. 

Please check if i missed something

PR #9024, opened 2024-02-26
fix: Fixes 8925 reply email to "To" address

### LINKED ISSUE #8925: "To" addresses are lost in Email Inboxes when replying via Chatwoot UI
# Description

If Chatwoot receives an email like this:
```
From: john@aaa.test
To: support@mycompany.example.com, jane@aaa.test
Cc: jill@aaa.test
```
then when an agent replies in Chatwoot, the response email will only be delivered to `john@aaa.test` and `jill@aaa.test`, and it won't be delivered to `jane@aaa.test`.

This is a common scenario for us, and it results in frustration when people get dropped off the email chain.

I believe this is due to this javascript code:
https://github.com/chatwoot/chatwoot/blob/608592db0c11db142c9a281d1d34b715ad33d6c0/app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue#L1122-L1148

The `to` variable in that code only gets set in a very particular case and is usually dropped.

# Potential Solutions

### Option A: Add "To" addresses to the CC field in the editor

UPDATE: @pranavrajs [confirmed that](https://github.com/chatwoot/chatwoot/issues/8925#issuecomment-1961997686) we should go with this option.

Update the above code to add each address in `emailAttributes.to` to the `cc` variable, **except** when the address is:
* The Inbox email address (e.g. `support@mycompany.example.com` in the example above)
* A reply email address (e.g. `reply+xxx@mycompany.example.com`)
* The current conversation contact's email

### ~Option B: Add a "To" field in the message editor and populate it.~

UPDATE: We want to go with the other option.

The message editor only has "CC" and "BCC" fields. Add another field to the editor for "To" addresses.
Evaluate the addresses in `emailAttributes.to` as described in Option A, but add any relevant addresses to `to` instead of `cc`.
We should ensure the current conversation contact's email is in either `to` or `cc`, or else I believe [this code](https://github.com/chatwoot/chatwoot/blob/cb2b75ca17db51f2eaefdafbb349b2ad31b87500/app/mailers/conversation_reply_mailer.rb#L186-L189) will cause the conversation contact to not receive the email.
We should also be mindful of [this code](https://github.com/chatwoot/chatwoot/blob/608592db0c11db142c9a281d1d34b715ad33d6c0/app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue#L1133-L1138) and make sure that scenario is still handled correctly.

# Other notes/requirements

We should add tests covering any new or updated behavior, and covering existing behavior if we don't have tests for that already.

### Environment

Linux VM

### Platform

Browser

ezhil56x:
## Description

- Each address in the to will be added to cc list if it follows specific conditions
- If the email is not a support or reply email, and if it is not the conversation contact's email

Fixes #8925 
/claim #8925 

## Type of change

## Checklist:

### DIFF
--- a/app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
+++ b/app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
@@ -1137,6 +1137,20 @@ export default {
         cc.push(conversationContact);
       }
 
+      // Add each address in emailAttributes.to to the "cc" list, expect for specific conditions
+      // If the email is not a support or reply email, and is not the conversation contact's email, then add it to the "cc" list
+      if (Array.isArray(emailAttributes.to)) {
+        emailAttributes.to.forEach(email => {
+          if (
+            !email.includes('support') &&
+            !email.includes('reply') &&
+            email !== conversationContact
+          ) {
+            cc.push(email);
+          }
+        });
+      }
+
       // Remove the conversation contact's email from the BCC list if present
       let bcc = (emailAttributes.bcc || []).filter(
         email => email !== conversationContact


ezhil56x:
@pranavrajs Any update?

app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
```diff
@@ -1137,6 +1137,20 @@ export default {
         cc.push(conversationContact);
       }
 
+      // Add each address in emailAttributes.to to the "cc" list, expect for specific conditions
+      // If the email is not a support or reply email, and is not the conversation contact's email, then add it to the "cc" list
+      if (Array.isArray(emailAttributes.to)) {
+        emailAttributes.to.forEach(email => {
+          if (
+            !email.includes('support') &&
```
jordan-brough, line 1145: This is not a robust test for the email address of an inbox

app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
```diff
@@ -1137,6 +1137,20 @@ export default {
...
+            !email.includes('reply') &&
```
jordan-brough, line 1146: This does not seem specific enough

app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
jordan-brough, line 1: One of the requirements for https://github.com/chatwoot/chatwoot/issues/8925 was to add tests:

> We should add tests covering any new or updated behavior, and covering existing behavior if we don't have tests for that already.

app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
ezhil56x: Do you have any suggestion on how this can be implemented?

app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
ezhil56x: Will add tests in the next commit

scmmishra (changes requested):
Hey @ezhil56x this is not a good fix. As pointed by @jordan-brough, Matching against a hard-coded list of values is not a robust test. You should take a look at all the emails in question from the attributes and then append them to the correct list.

scmmishra:
Hey @ezhil56x I am closing this PR, feel free to re-open it in case you're still working on it

PR #9026, opened 2024-02-27
chore: Reply suggestions using response_sources

sojan-official:
- When reply bot is enabled, the OpenAi integration will use response_sources to generate reply suggestions


pranavrajs:
| Before | After | 
| Screenshot 2024-03-03 at 8 19 24 PM | Screenshot 2024-03-03 at 8 21 53 PM |

pranavrajs (approved):

PR #9030, opened 2024-02-27
chore: Component changes for bot reports

sojan-official:
We want to reuse the existing reporting components for our upcoming bot reports. Hence, making the required changes and cleaning up existing components.

- `ReportContainer.vue` to accept report keys via prompts ( We will be using different report keys in bot reports )
-  `CsatMetricCard` to `ReportMetricCard`  for generic use since we are using it also for the bot reports.

----

Screenshot 2024-02-27 at 7 02 38 PM


pranavrajs (commented):
Made some changes to the component to use composition API and removed the custom CSS classes.

PR #9031, opened 2024-02-27
feat: Backend APIs for the bot metrics

sojan-official:
## Changes

- [x] pending count in live reports
- [x] Report builder should return bot_resolutions_count and bot_handoffs_count  
- [x] bot_metrics endpoint to return date for the ReportMetricCard
- [x] Spec Changes


scmmishra (approved):
LGTM!

app/builders/v2/reports/bot_metrics_builder.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  def bot_resolutions_count
+    account.reporting_events.joins(:conversation).select(:conversation_id).where(account_id: account.id, name: :conversation_bot_handoff,
+                                                                                 created_at: range).distinct.count
+  end
+
+  def bot_handoffs_count
+    account.reporting_events.joins(:conversation).select(:conversation_id).where(account_id: account.id, name: :conversation_bot_handoff,
+                                                                                 created_at: range).distinct.count
+  end
```
scmmishra, line 48: These queries are the same, is this intentional?

scmmishra (changes requested):

app/builders/v2/reports/bot_metrics_builder.rb
sojan-official: 😁 good catch. nope. should be resolution on the first one

sojan-official: updated the specs to captures missing case here

spec/controllers/api/v2/accounts/report_controller_spec.rb
```diff
@@ -399,4 +441,41 @@
...
+
+    context 'when it is an authenticated user' do
+      let(:params) do
+        super().merge(
+          since: 7.days.ago.to_i.to_s,
+          until: end_of_today.to_s
+        )
+      end
+
+      it 'returns unauthorized' do
```
pranavrajs, line 462:
```suggestion
      it 'returns unauthorized if the user is an agent' do
```

PR #9068, opened 2024-03-06
fix: inline attachments not handled if tag was missing

scmmishra:
When processing inline images, `MailboxHelper` would find the image that looks like `[image: image.png]` and replace it with an `` tag, however this search candidate is not present when the email is sent from Apple mail app (verified) or Yahoo mail (Yet to be verified).

This PR updates the method to handle both cases, in case the search candidate is not present, we simply append the image at the end


vishnu-narayanan (approved):
LGTM

PR #9107, opened 2024-03-13
feat: Update the `contact_type` when creating or updating the contact

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-3009/update-the-contact-type-when-the-contact-createdupdated


app/services/contacts/sync_attributes.rb
```diff
@@ -0,0 +1,37 @@
...
+    # If the contact has an email or phone number or social details( facebook_user_id, instagram_user_id, etc) then it is a lead
+    # If contact is from external channel like facebook, instagram, whatsapp, etc then it is a lead
+    return unless @contact.email.present? || @contact.phone_number.present? || social_details_present?
+
+    @contact.contact_type = 'lead'
+  end
+
+  def social_details_present?
+    @contact.additional_attributes.keys.any? do |key|
+      key.to_s.match?(/^social_\w+_user_(id|name)$/i) && @contact.additional_attributes[key].present?
```
muhsin-k, line 34: The expected format for unique channel information should be `social_{channel_name}_{user_id/user_name}`.
[Reference PR](https://github.com/chatwoot/chatwoot/pull/8802)

sojan-official: @muhsin-k, can we use `key.starts_with?('social_')` , which is easier to read and would work since we reserved social_ for this use case.

sojan-official (commented):
Minor comment. apart from that LGTM

app/services/contacts/sync_attributes.rb
muhsin-k: @sojan-official Makes sense, I have updated the PR.

sojan-official (approved):

PR #9110, opened 2024-03-14
fix: Locale not correct in root url when accessing help center with custom domain

scmmishra:
The portal is served on the subdomain via two controllers, the `portals_controller` and the `dashbaord_controller`.  The `SwitchLocale` concern in the dashboard controller does not check for portals locale, this PR fixes that


app/controllers/concerns/switch_locale.rb
```diff
@@ -16,6 +17,18 @@ def switch_locale_using_account_locale(&)
...
+    return if params[:locale]
+
+    domain = request.host
+    return if domain == URI.parse(ENV.fetch('FRONTEND_URL', '')).host
+
+    @portal = Portal.find_by(custom_domain: domain)
+    return unless @portal
+
+    @portal.default_locale
+  end
```
scmmishra, line 30: @sojan-official need your thoughts on this

sojan-official: Should also check for `HELPCENTER_URL` as well. But makes sense.

scmmishra: This is specifically for the root url page, there is appropriate validation for the help center route for this

pranavrajs:
@sojan-official Can you please take this to production as soon as possible?

PR #9124, opened 2024-03-18
feat: Add a migration to update all the `contact_type` from visitor to lead for all the contacts

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-3027/add-a-migration-to-update-the-contact-type-for-all-the-existing.
- Set default `contact_type` will be lead.
- Set `contact_type` visitor default for live chat contacts.


muhsin-k:
Closing this PR now. We will come back to this later.

PR #9126, opened 2024-03-19
feat: add sla events table

vishnu-narayanan:
# Pull Request Template

## Description

- Add `sla_events` table to keep track of sla misses

## Changelog

- Add a new table sla_events with the following attributes.
```
event_type: Represents the type of event, which can be 'frt', 'nrt', or 'rt'.
meta: Contains the message_id, but only used in the 'nrt' metric.

created_at: The time when the breach occurred.

conversation_id: The ID of the conversation where the breach occurred.
applied_sla_id: The ID of the Service Level Agreement (SLA) that was breached.

```

Fixes https://linear.app/chatwoot/issue/CW-3137/add-support-for-the-sla-events-table

## Type of change

## How Has This Been Tested?

- Tested locally
- Added specs

## Checklist:

### DIFF
--- a/db/migrate/20240319062553_create_sla_events.rb
+++ b/db/migrate/20240319062553_create_sla_events.rb
@@ -3,6 +3,9 @@ class CreateSlaEvents < ActiveRecord::Migration[7.0]
     create_table :sla_events do |t|
       t.references :applied_sla, null: false
       t.references :conversation, null: false
+      t.references :account, null: false
+      t.references :sla_policy, null: false
+      t.references :inbox, null: false
 
       t.integer :event_type
       t.jsonb :meta, default: {}
--- a/db/migrate/20240319062553_create_sla_events.rb
+++ b/db/migrate/20240319062553_create_sla_events.rb
@@ -4,7 +4,7 @@ class CreateSlaEvents < ActiveRecord::Migration[7.0]
       t.references :applied_sla, null: false
       t.references :conversation, null: false
 
-      t.string :event_type
+      t.integer :event_type
       t.jsonb :meta, default: {}
 
       t.timestamps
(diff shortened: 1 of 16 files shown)


enterprise/app/models/sla_event.rb
```diff
@@ -0,0 +1,19 @@
...
+#  updated_at      :datetime         not null
+#  applied_sla_id  :bigint           not null
+#  conversation_id :bigint           not null
+#
+# Indexes
+#
+#  index_sla_events_on_applied_sla_id   (applied_sla_id)
+#  index_sla_events_on_conversation_id  (conversation_id)
+#
+class SlaEvent < ApplicationRecord
```
muhsin-k, line 18: Please add the has many relation in `applied_slas` model,
```
has_many :sla_events, dependent: :destroy_async
```

db/schema.rb
```diff
@@ -842,6 +842,17 @@
     t.index ["user_id"], name: "index_reporting_events_on_user_id"
   end
 
+  create_table "sla_events", force: :cascade do |t|
+    t.bigint "applied_sla_id", null: false
+    t.bigint "conversation_id", null: false
+    t.string "event_type"
```
muhsin-k, line 848: Shouldn't it be of type enum since we only have three types (frt, nrt, or rt)?

db/migrate/20240319062553_create_sla_events.rb
```diff
@@ -0,0 +1,13 @@
+class CreateSlaEvents < ActiveRecord::Migration[7.0]
+  def change
+    create_table :sla_events do |t|
+      t.references :applied_sla, null: false
```
sojan-official, line 4: We could also add an `account_id`, 'inbox_id' and 'sla_id' as well to make it easier to query all the events related to one account as well as doing the filtering with out having to do a join

sojan-official: we can auto populate this values into the model in before validation using `ensure_*` method pattern which we are using

vishnu-narayanan: Makes sense. Done.

muhsin-k (approved):

sojan-official (approved):

PR #9132, opened 2024-03-20
feat: Add SLA metrics API

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-3129/[ee]-create-a-report-api-to-fetch-the-metrics-related-to-sla


enterprise/app/controllers/api/v1/accounts/applied_slas_controller.rb
```diff
@@ -0,0 +1,28 @@
+class Api::V1::Accounts::AppliedSlasController < Api::V1::Accounts::EnterpriseAccountsController
+  include Sift
+  include DateRangeHelper
+
+  before_action :set_sla_responses, only: [:metrics]
+  before_action :check_admin_authorization?
+
+  def metrics
+    @total_applied_slas = @sla_responses.count
+    @number_of_sla_breaches = @sla_responses.missed.count
```
muhsin-k, line 10: Calculate the hit percentage in the front end,

```
((total_applied_slas - number_of_sla_breaches) / total_applied_slas) * 100
```

muhsin-k: @pranavrajs @sojan-official Need your thoughts here?

enterprise/app/controllers/api/v1/accounts/applied_slas_controller.rb
```diff
@@ -0,0 +1,28 @@
+class Api::V1::Accounts::AppliedSlasController < Api::V1::Accounts::EnterpriseAccountsController
```
vishnu-narayanan, line 1: Should this be under reports namespace?

enterprise/app/controllers/api/v1/accounts/applied_slas_controller.rb
sojan-official: i think we can return the percentage as well along with the APIs . would be easier for the API consumers

enterprise/app/controllers/api/v1/accounts/applied_slas_controller.rb
sojan-official: This sounds good.

sojan-official (commented):
LGTM mostly. couple of minor comments

enterprise/app/controllers/api/v1/accounts/applied_slas_controller.rb
muhsin-k: @sojan-official Done.

enterprise/app/controllers/api/v1/accounts/applied_slas_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  sort_on :created_at, type: :datetime
+
+  def index; end
+
+  def metrics
+    @total_applied_slas = total_applied_slas
+    @number_of_sla_breaches = number_of_sla_breaches
+    @hit_rate = hit_rate
+  end
```
scmmishra, line 20: We need not assign them as instance variables, we can use the methods directly in `metrics.json.jbuilder`

Either way since, the methods do the assignment themselves, we need to assign it again

enterprise/app/controllers/api/v1/accounts/applied_slas_controller.rb
```diff
@@ -0,0 +1,61 @@
+class Api::V1::Accounts::AppliedSlasController < Api::V1::Accounts::EnterpriseAccountsController
+  include Sift
+  include DateRangeHelper
+
+  RESULTS_PER_PAGE = 25
+
+  before_action :set_applied_slas, only: [:index, :metrics]
+  before_action :set_current_page, only: [:index]
+  before_action :set_current_page_appiled_slas, only: [:index]
```
scmmishra, line 9: Probably a nitpick but `paginate_slas` is a better name for this

muhsin-k:
@scmmishra I've addressed all the comments. Could you please review?

scmmishra:
> @scmmishra I've addressed all the comments. Could you please review?

Sure, on it

scmmishra (approved):

Issue #9094, opened 2024-03-12
Tag bug


muriloorama:
### Describe the bug

When you add more than 14 tags, the contact freezes, you can't even add or remove the tags. the only way is to delete the contact and add less than 14 tags

### To Reproduce

When you add more than 14 tags, the contact freezes, you can't even add or remove the tags. the only way is to delete the contact and add less than 14 tags

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

Browser

### Browser and version

3.6.0

### Additional context

When you add more than 14 tags, the contact freezes, you can't even add or remove the tags. the only way is to delete the contact and add less than 14 tags

pranavrajs:
@muriloorama Interesting! Let us test it out

scmmishra:
@muriloorama we couldn't verify this exactly, which OS and browser are you using?

muriloorama:
I'm using Windows 11 and Chrome. Did you manage to add more than 14 tags and then remove them?

pranavrajs:
@sojan @shivam One potential issue could be that the cached_label_list is a string, limited to 255 characters, compared to the text field's 30,000 characters. The bug might occur if the field overflows due to an abundance of labels.

The number 14 may not be significant, but the total number of characters in the labels could be. Try adding more labels to see if the issue persists.

We are testing with labels of 5-10 characters each. The total combined character count should be around 140, which is still within the acceptable limits. Try adding more to see if you can reproduce it.

scmmishra:
Let me give this a shot

scmmishra:
> Member

@pranavrajs you were right! I've fixed this in https://github.com/chatwoot/chatwoot/pull/9143

PR #9143, opened 2024-03-22
feat: Convert `cached_label_list` to text

scmmishra:
We use `cached_label_list` column to hold the list of labels attached to a conversation, however the column type for it was `string` with a limit of 255 chars, this is not sufficient, since a conversation can have virtually unlimited labels. 

This caused the addition of any new labels to break once the char count reached 255. This PR fixes that by changing the column type to text

Fixes: https://github.com/chatwoot/chatwoot/issues/9094

## Type of change

## How Has This Been Tested?

Tested locally for up and down queries both

## Checklist:

### DIFF
--- a/db/migrate/20240322071629_convert_cached_label_list_to_text.rb
+++ b/db/migrate/20240322071629_convert_cached_label_list_to_text.rb
@@ -1,0 +1,32 @@
+class ConvertCachedLabelListToText < ActiveRecord::Migration[7.0]
+  def up
+    change_column :conversations, :cached_label_list, :text
+  end
+
+  def down
+    # This might cause data loss if the text is longer than 255 characters
+    # lets start by truncating the data to 255 characters
+    Conversation.where('LENGTH(cached_label_list) > 255').find_in_batches do |conversation_batch|
+      Conversation.transaction do
+        conversation_batch.each do |conversation|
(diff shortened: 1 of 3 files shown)


sojan-official (approved):

db/migrate/20240322071629_convert_cached_label_list_to_text.rb
```diff
@@ -0,0 +1,32 @@
+class ConvertCachedLabelListToText < ActiveRecord::Migration[7.0]
+  def up
+    change_column :conversations, :cached_label_list, :text
+  end
+
+  def down
+    # This might cause data loss if the text is longer than 255 characters
+    # lets start by truncating the data to 255 characters
+    Conversation.where('LENGTH(cached_label_list) > 255').find_in_batches do |conversation_batch|
```
pranavrajs, line 9: This would timeout.

scmmishra: In production we would just have to do this manually, but for any other self hosted instances this would work fine. Besides this is just the down query, I don't think we would need to do this a lot. Let me know if you want me to move this out as a script?

sojan-official: @scmmishra, this should be fine. lets go ahead with merging this.

PR #9144, opened 2024-03-22
feat: use better loading state

scmmishra:

1. Handle loading state in the `SettingsLayout` with default loader and message spinner
2. Create loading state using `BaseSettingsListItem`

https://github.com/chatwoot/chatwoot/assets/18097732/bc29f2de-036e-47ba-82b1-c4019e7852b7

### DIFF
--- a/app/javascript/dashboard/routes/dashboard/settings/SettingsLayout.vue
+++ b/app/javascript/dashboard/routes/dashboard/settings/SettingsLayout.vue
@@ -1,6 +1,23 @@
+<script setup>
+defineProps({
+  isLoading: {
+    type: Boolean,
+    default: false,
+  },
+  loadingMessage: {
+    type: String,
+    default: '',
+  },
+});
+</script>
 <template>
   <div class="flex flex-col w-full h-full gap-10 font-inter">
     <slot name="header" />
-    <slot name="body" />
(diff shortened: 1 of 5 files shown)


scmmishra:
@iamsivin @pranavrajs @absurdiya PTAL and see if this is good

absurdiya:
> @iamsivin @pranavrajs @absurdiya PTAL and see if this is good

Looks good to me, would just prefer the three square boxes to be more rectangular and have less space between them, or we can make it one wide rectangle as well to simplify things

scmmishra:
> Looks good to me, would just prefer the three square boxes to be more rectangular and have less space between them, or we can make it one wide rectangle as well to simplify things


https://github.com/chatwoot/chatwoot/assets/18097732/607ad6fe-9c30-4451-9c52-4af80d96697b

How does this look? @absurdiya

absurdiya:
> > Looks good to me, would just prefer the three square boxes to be more rectangular and have less space between them, or we can make it one wide rectangle as well to simplify things
> 
>  CleanShot.2024-03-22.at.16.28.32.mp4 
> How does this look? @absurdiya

Looks good!

pranavrajs:
Looks good.

PR #9174, opened 2024-03-29
feat: Conversation API to return applied_sla and sla_events

sojan-official:
- Include applied_sla and sla_events in conversation api endpoints
- Include applied_sla and sla_events in conversation push_data

fixes: https://linear.app/chatwoot/issue/CW-3143/update-the-conversation-api-to-return-the-sla-policy-breached-sla

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -25,3 +25,4 @@ class ApplicationController < ActionController::Base
     }
   end
 end
+ApplicationController.include_mod_with('Concerns::ApplicationControllerConcern')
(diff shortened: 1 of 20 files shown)


sojan-official:
@vishnu-narayanan  @muhsin-k  Could you guys take a look? Also, do we need to include the sla_events in these payloads? We could move it as a separate API if we aren't directly showing it everywhere. It will help us reduce queries in conversation endpoints.

muhsin-k:
> @vishnu-narayanan @muhsin-k Could you guys take a look? Also, do we need to include the sla_events in these payloads? We could move it as a separate API if we aren't directly showing it everywhere. It will help us reduce queries in conversation endpoints.
Makes sense. We show the SLA events in a popover when the user hovers over the SLA breach item.
cc @iamsivin

muhsin-k (approved):

muhsin-k (approved):

PR #9179, opened 2024-04-01
fix: add priority in filter keys [CW-3177]

scmmishra:
`AutomationRules::ConditionValidationService` would fail, since `filter_keys.yml` didn't have the priority attribute


scmmishra:
@pranavrajs @sojan-official the test was breaking previously, because the filter service would use the standard priority for filtering instead of the custom attribute. But then, the standard field `priority` is an integer, and the SQL query generated in the tests was using string values, this caused a SQL error

sojan-official (approved):

PR #9182, opened 2024-04-02
feat: update create SLA modal design

scmmishra:
> [!WARNING]
> This PR updates the default modal, that means the UI changes will reflect across the app.

1. Update the modal to match new UI
2. Update the SLA form to match the new UI
3. Create a new toggle component, inside SLA. It can be moved once we agree on a code-organization pattern

CleanShot 2024-04-02 at 08 28 45@2x


app/javascript/dashboard/routes/dashboard/settings/sla/SlaForm.vue
```diff
@@ -5,6 +5,11 @@
         v-model="name"
         :class="{ error: $v.name.$error }"
         class="w-full"
+        :styles="{
```
pranavrajs, line 8: Can't we just use the Tailwind Classes with !? For eg: `!rounded-lg`

app/javascript/dashboard/routes/dashboard/settings/sla/components/Toggle.vue
```diff
@@ -0,0 +1,59 @@
...
+      :checked="isChecked"
+      @click="toggleCheckbox"
+    />
+    <div
+      class="absolute inset-0 p-0.5 transition-all ease-in-out duration-200 rounded-full cursor-pointer slider bg-slate-100"
+    />
+  </label>
+</template>
+
+<style>
```
pranavrajs, line 34: I would suggest referring to https://github.com/shadcn-ui/ui/blob/main/apps/www/registry/default/ui/switch.tsx instead of writing the custom CSS here.

pranavrajs (commented):
I've some comments on the implementation, please take a look.

app/javascript/dashboard/routes/dashboard/settings/sla/components/Toggle.vue
scmmishra: I can move most of the styles to tailwind classes, I'll make the change!

app/javascript/dashboard/routes/dashboard/settings/sla/SlaForm.vue
scmmishra: This is not the default tailwind `style` attribute, WootInput supports this so that the styles can be applied to the input component which is a child element in the DOM tree. Applying the `class` directly will only put the class on the root element of the component which is the label

scmmishra: If need be, we can port the components from the onboarding here. We can add that without making the PR too big, lmk!

scmmishra: @pranavrajs let me know what do you think

scmmishra: @pranavrajs PTAL

muhsin-k (approved):

app/javascript/dashboard/routes/dashboard/settings/sla/components/Toggle.vue
```diff
@@ -0,0 +1,36 @@
...
+<template>
+  <label class="relative inline-block h-4 w-7">
+    <input
+      class="hidden peer"
+      type="checkbox"
+      :checked="isChecked"
+      @click="toggleCheckbox"
+    />
+    <div
+      :class="[
```
iamsivin, line 29: @scmmishra Can you add dark mode support for this toggle switch.
image

app/javascript/dashboard/routes/dashboard/settings/sla/components/Toggle.vue
iamsivin, line 1: There is an existing component called `app/javascript/dashboard/components/ui/Switch.vue`, which we used as `<woot-switch/>` in multiple places can we update the existing one instead of creating a new one?

iamsivin (commented):
@scmmishra Added some minor comments. Could you please take a look?

app/javascript/dashboard/routes/dashboard/settings/sla/SlaForm.vue
```diff
@@ -29,23 +39,29 @@
         @isInValid="handleIsInvalid(index, $event)"
       />
 
-      <div class="flex items-center w-full gap-2">
-        <input id="sla_bh" v-model="onlyDuringBusinessHours" type="checkbox" />
-        <label for="sla_bh">
+      <div
+        class="mt-3 flex h-10 items-center text-sm w-full gap-2 border border-solid border-slate-200 dark:border-slate-600 px-3 py-1.5 rounded-xl justify-between"
+      >
+        <span for="sla_bh" class="text-slate-700">
```
iamsivin, line 45: Dark mode color is missing.

iamsivin (approved):

PR #9187, opened 2024-04-03
chore: Handle duplicate Webhooks from slack

sojan-official:
Ref: https://linear.app/chatwoot/issue/CW-3176/duplicate-messages-are-created-when-a-duplicate-webhook-is-sent-from


muhsin-k (approved):
The code looks good to me and has been verified in a local environment.

PR #9213, opened 2024-04-09
fix: Remove background color from table footer

muhsin-k:
**Before**

CleanShot 2024-04-09 at 10 32 41@2x

**After**

CleanShot 2024-04-09 at 10 32 19@2x

### DIFF
--- a/app/javascript/dashboard/routes/dashboard/settings/reports/components/SLA/SLATable.vue
+++ b/app/javascript/dashboard/routes/dashboard/settings/reports/components/SLA/SLATable.vue
@@ -50,7 +50,6 @@
       :current-page="currentPage"
       :total-count="totalCount"
       :page-size="pageSize"
-      class="border-none"
       @page-change="onPageChange"
     />
   </div>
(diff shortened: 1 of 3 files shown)


app/javascript/dashboard/routes/dashboard/settings/reports/components/SLA/SLATable.vue
```diff
@@ -50,7 +50,7 @@
       :current-page="currentPage"
       :total-count="totalCount"
       :page-size="pageSize"
-      class="bg-slate-25 dark:bg-slate-900 sticky bottom-0 border-none mt-4"
+      class="border-none"
```
iamsivin, line 53: @muhsin-k Do we need `border-none` here? In the `table-footer` component, there is no border.
```suggestion
```
```
<footer
    v-if="isFooterVisible"	  
    class="h-12 flex items-center justify-between px-6"
/>	 
```

iamsivin (approved):

vishnu-narayanan (approved):

PR #9223, opened 2024-04-11
feat: do not send contact details to the widget

scmmishra:
Sending contact details to the client was unnecessary, this PR removes that

### DIFF
--- a/app/javascript/widget/components/ChatFooter.vue
+++ b/app/javascript/widget/components/ChatFooter.vue
@@ -87,7 +87,7 @@ export default {
       return !allowMessagesAfterResolved && status === 'resolved';
     },
     showEmailTranscriptButton() {
-      return this.currentUser && this.currentUser.email;
+      return this.currentUser && this.currentUser.has_email;
     },
     hasReplyTo() {
       return (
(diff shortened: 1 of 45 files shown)


app/javascript/widget/components/ChatFooter.vue
```diff
@@ -87,7 +87,7 @@ export default {
       return !allowMessagesAfterResolved && status === 'resolved';
     },
     showEmailTranscriptButton() {
-      return this.currentUser && this.currentUser.email;
+      return this.currentUser && this.currentUser.has_email;
```
muhsin-k, line 90: I would suggest to use computed property for `hasEmail`.
```
hasEmail() {
   return this.currentUser && this.currentUser.has_email;
},
```

app/javascript/widget/components/ChatFooter.vue
```diff
@@ -141,12 +141,10 @@ export default {
       this.inReplyTo = message;
     },
     async sendTranscript() {
-      const { email } = this.currentUser;
-      if (email) {
+      const { has_email: hasEmail } = this.currentUser;
+      if (hasEmail) {
```
muhsin-k, line 145:
```suggestion
      if (this.hasEmail) {
```

app/javascript/widget/components/ChatFooter.vue
muhsin-k, line 90:
```suggestion
      return this.hasEmail
```

muhsin-k (approved):

Issue #8968, opened 2024-02-19
IMAP connections not getting closed or re-used


af7567:
### Describe the bug

The IMAP mailbox is getting checked by chatwoot, but emails are not getting picked up because chatwoot has too many open connections to the IMAP server. New connections will only be allowed after the IMAP server has closed the inactive connection which can take a while.

IMAP server log:
```
Feb 19 17:17:30 mail dovecot: imap-login: Disconnected: Maximum number of connections from user+IP exceeded (mail_max_userip_connections=10): user=<hello>, method=PLAIN, rip=12.12.12.246, lip=12.12.12.242, TLS, session=<F6/ERL8RMO5SRor2>
Feb 19 17:18:02 mail dovecot: imap-login: Disconnected: Maximum number of connections from user+IP exceeded (mail_max_userip_connections=10): user=<hello>, method=PLAIN, rip=12.12.12.246, lip=12.12.12.242, TLS, session=<rk6rRr8RyNpSRor2>
Feb 19 17:18:18 mail dovecot: imap(hello)<25824><yt5Y3L4RYpVSRor2>: Disconnected: Inactivity - no input for 1800 secs in=125 out=22242 deleted=0 expunged=0 trashed=0 hdr_count=4 hdr_bytes=11113 body_count=1 body_bytes=9964
```
Output of `netstat | grep imap` in the chatwoot VM
```
tcp        0      0 chatwoot:34600          mail.example.:imaps ESTABLISHED
tcp        0      0 chatwoot:53066          mail.example.:imaps CLOSE_WAIT 
tcp        0      0 chatwoot:35058          mail.example.:imaps ESTABLISHED
tcp        0      0 chatwoot:35436          mail.example.:imaps CLOSE_WAIT 
tcp        0      0 chatwoot:35238          mail.example.:imaps ESTABLISHED
tcp        0      0 chatwoot:41402          mail.example.:imaps CLOSE_WAIT 
tcp        0      0 chatwoot:39630          mail.example.:imaps ESTABLISHED
tcp        0      0 chatwoot:58892          mail.example.:imaps ESTABLISHED
tcp        0      0 chatwoot:49162          mail.example.:imaps ESTABLISHED
tcp        0      0 chatwoot:43262          mail.example.:imaps ESTABLISHED
tcp        0      0 chatwoot:60308          mail.example.:imaps ESTABLISHED
tcp        0      0 chatwoot:45826          mail.example.:imaps ESTABLISHED
```
[...]

PR #9225, opened 2024-04-11
fix: Terminate IMAP connection after the job is complete

pranavrajs:
Fixes https://github.com/chatwoot/chatwoot/issues/8968

The existing IMAP connections were not immediately terminated after the job completion. This delay could lead to some providers throttling connections, ultimately causing the "Connection reset by peer - SSL_connect" error.

In this PR, I've introduced an option to log out from the IMAP server. If we fail to logout properly, a forced disconnect will occur. 

The underlying method that was overridden was `perform`, it has been changed to `fetch_emails`. The `perform` method remains the same—it accepts no input and returns the same list of inbound email, while also terminating the connection.

The suggestion [here](https://github.com/chatwoot/chatwoot/issues/8968#issuecomment-1977618980) made sense, however I've used `logout` in place of `close` and used the `disconnect` as a fallback.

### DIFF
--- a/app/services/imap/base_fetch_email_service.rb
+++ b/app/services/imap/base_fetch_email_service.rb
@@ -8,8 +8,10 @@ class Imap::BaseFetchEmailService
   end
 
   def perform
-    fetch_emails
+    inbound_emails = fetch_emails
     terminate_imap_connection
+
+    inbound_emails
   end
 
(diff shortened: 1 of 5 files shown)


sojan-official (approved):
LGTM

PR #9239, opened 2024-04-16
feat: Use the font `Inter Display` for help center articles

pranavrajs:
Use the Inter Display for better readability


scmmishra:
@pranavrajs should we just use Inter across the app? It's much more balanced, current font is a bit wider, great as a display font, not sure as a UI font?

pranavrajs:
@scmmishra We would use Inter for body and Inter Display for title going forward. This PR could be merged as it is.

scmmishra (approved):

PR #9240, opened 2024-04-16
feat: add upgrade banner for enterprise

scmmishra:


sojan-official:
note: we will handle features to update cases for the cloud separately based on billing plans. for now, we will do an upgrade manually.

PR #9258, opened 2024-04-18
feat: Add new colors

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-3219/update-palette-colors

Script to convert Tailwind colors to CSS Variables: https://gist.github.com/scmmishra/f37299625c25f2831de5c486ecf415b8


tailwind.config.js
```diff
@@ -129,6 +129,75 @@ module.exports = {
         800: violetDark.violet6,
         900: violet.violet12,
       },
+      primary: {
```
muhsin-k, line 132: Let's retain all the colors under a new name. We can eliminate the old colors once we transition every component to the new design.

scmmishra (approved):
LGTM!

PR #9260, opened 2024-04-19
fix: Audio attachment issues

iamsivin:
# Pull Request Template

## Description

This PR will fix the following issues,
1. **Fix:** Doesn't have the ability to remove audio attachments ( x button is not showed )
2. **Fix:** We already show the voice recorder feature. We don't need to show the attachment separately. In the case of voice recording, just the recorder should be enough. 
3. **Fix:** Recorder controls are not visible in light mode
4. **Fix:** Div inside the recorder overlays the screen and takes over the button clicks. So, the other `woot-buttons` in the editor must be clicked multiple times to do any action. (Only when the recorder is visible after recording)
5. **Fix:** When we clear the recorder button, it will reset all attachments and messages from the editor.
6. Moved the `attachment-preview` component to the composition API.

Fixes https://linear.app/chatwoot/issue/CW-3222/issues-audio-attachment-issues

## Type of change

## How Has This Been Tested?

**Loom video**
https://www.loom.com/share/e2ba7a51f1fa4f29b7f29707a0536390?sid=4798f093-2a6e-4a01-a73b-e13aed289447

## Checklist:


app/javascript/dashboard/components/widgets/AttachmentsPreview.vue
```diff
@@ -23,73 +23,59 @@
...
+});
+
+const onRemoveAttachment = itemIndex => {
+  const updatedAttachments = filteredAttachments.value.filter(
+    (item, index) => itemIndex !== index
+  );
+  const recordedAudios = props.attachments.filter(
+    attachment => attachment?.isRecordedAudio
+  );
+  emits('remove-attachment', [...updatedAttachments, ...recordedAudios]);
```
muhsin-k, line 65: Instead of `filteredAttachments` we can give a meaningful name like `nonAudioAttachments`.

```
const audioAttachments = computed(() =>
  props.attachments.filter(a => a.isRecordedAudio)
);
const nonAudioAttachments = computed(() =>
  props.attachments.filter(a => !a.isRecordedAudio)
);

const onRemoveAttachment = itemIndex => {
  emits(
    'remove-attachment',
    nonAudioAttachments.value
      .filter((_, index) => index !== itemIndex)
      .concat(audioAttachments.value)
  );
};
```

iamsivin: @muhsin-k Done

PR #9266, opened 2024-04-19
feat: Support document file uploads on telegram channel

sojan-official:
The previous implementation of file uploads in telegram channel only supported media group attachments. This cause breakages when attachments are mixed mode or contains document file types like txt, .avi etc.  This PR fixes that.

## Test cases

- [x] single attachments -> photo, video, audio , txt , pdf
- [x] grouped attachments -> photo, video, audio,  photo+video, txt
- [x] mixed group attachments -> photo (2) + video + audio (2) + txt
- [x] Ensure success status in all these cases


pranavrajs:
Tested multiple cases, seems to be working.

pranavrajs (approved):

PR #9276, opened 2024-04-22
feat: Add button component

muhsin-k:
| Light theme | Dark theme |
|-------------|--------------|
| CleanShot 2024-04-24 at 21 05 57@2x     | CleanShot 2024-04-24 at 21 05 47@2x       |


app/javascript/v3/components/Form/Button.vue
```diff
@@ -0,0 +1,142 @@
...
+    default: '',
+  },
+  colorScheme: {
+    type: String,
+    default: 'primary',
+  },
+  classNames: {
+    type: [String, Object],
+    default: '',
+  },
```
scmmishra, line 57: classnames should not be added as a prop, anyway a class applied on the component usage will show up here. We can use the `$attrs` to access the attributes provided to the component in case we need it

app/javascript/v3/components/Form/Button.vue
```diff
@@ -0,0 +1,142 @@
...
+    default: false,
+  },
+  rightIcon: {
+    type: Boolean,
+    default: false,
+  },
+  iconOnly: {
+    type: Boolean,
+    default: false,
+  },
```
scmmishra, line 69: Not sure if we need this, the check for `$slots.default` implicitly means that the button is icon only

app/javascript/v3/components/Form/Button.vue
```diff
@@ -0,0 +1,142 @@
...
+  }
+  return 'h-10 px-4 py-2.5 text';
+});
+
+const iconSize = computed(() => {
+  if (props.size === 'medium') {
+    return 14;
+  }
+  return 16;
+});
```
scmmishra, line 131: We can use an `em` based unit, that way the icon size will automatically adjust to the font size

app/javascript/v3/components/Form/Button.vue
```diff
@@ -0,0 +1,142 @@
+<template>
+  <button
+    class="inline-flex items-center gap-1 text-sm font-medium reset-base rounded-xl w-fit"
+    :type="type"
+    :class="buttonClasses"
+    :disabled="isDisabled"
```
scmmishra, line 6: We can use `attr` fallback for disabled check, same for the `type` attribute

app/javascript/v3/components/Form/Button.vue
```diff
@@ -0,0 +1,142 @@
...
+const buttonClasses = computed(() => [
+  colorClass.value,
+  sizeClass.value,
+  props.classNames,
+]);
+const handleClick = event => {
+  if (!props.isDisabled) {
+    emits('click', event);
+  }
+};
```
scmmishra, line 141: For this too, I think (need to verify) we can avoid handling this event and let the fallthrough attributes handle this

See: https://vuejs.org/guide/components/attrs.html#accessing-fallthrough-attributes-in-javascript

scmmishra:
@muhsin-k added some points, also can you ensure this handles RTL direction also

muhsin-k:
@scmmishra All the review comments are addressed. RTL direction is also working fine.

app/javascript/v3/components/Form/Button.vue
```diff
@@ -0,0 +1,112 @@
...
+  }
+  return 'h-10 px-4 py-2.5 text';
+});
+
+const iconSize = computed(() => {
+  if (props.size === 'medium') {
+    return '1em';
+  }
+  return '1.2em';
+});
```
scmmishra, line 103: What are the possible size options for this? `medium` and `large`? Lets add them specifically along with a validator for the props. 
Another thing, we can simply keep a fixed iconSize of `1.16em` So when the font-size this so when the font size is `12px` the icon size will be ~`14px` and when the icon font size is `14px` it will be ~`16px`

This will help us avoid computation of any sort, and in the future adding a new size means the icon size will automatically adjust. We can also add a max height and max width property to the icon later or add it as a class to the `fluent-icon` usage in this and set the value to `1cap` this will ensure that the icon will never be taller than the capital letter of the font size

ref: [cap unit](https://www.w3.org/TR/css-values-4/#cap)

app/javascript/v3/components/Form/Button.vue
```diff
@@ -0,0 +1,112 @@
...
+      default:
+        'bg-ruby-600 text-white hover:bg-ruby-700 active:bg-ruby-700 focus:ring focus:ring-offset-1 focus:ring-ruby-400',
+    },
+  };
+
+  const schemeStyles = styleMap[props.colorScheme] || {};
+  const variantStyle = schemeStyles[props.variant] || schemeStyles.default;
+
+  return variantStyle;
+});
```
scmmishra, line 89: For this function, we can extract the base classes to avoid duplication

```js
const baseClasses = {
  outline: 'outline outline-1 -outline-offset-1 focus:ring focus:ring-offset-1',
  ghost: 'hover:text-600 active:text-600 focus:ring focus:ring-offset-1',
  default: 'text-white hover:bg-700 active:bg-700 focus:ring focus:ring-offset-1',
};
```

scmmishra (changes requested):
Last few points, everything else looks good to go. Let's also add tests for this component

app/javascript/v3/components/Form/Button.vue
scmmishra: @muhsin-k did you get an chance to explore this suggestion?

app/javascript/v3/components/Form/Button.vue
muhsin-k: Makes sense.

scmmishra (approved):

PR #9296, opened 2024-04-25
feat: include resolution time and waiting time in team & agent report

scmmishra:


scmmishra:
@sojan-official need your review on this. A thing to note is that this might reverse the perf benefits we got in https://github.com/chatwoot/chatwoot/pull/8733

app/builders/v2/report_builder.rb
```diff
@@ -50,7 +50,9 @@ def short_summary
     {
       conversations_count: conversations.count,
       avg_first_response_time: avg_first_response_time_summary,
-      avg_resolution_time: avg_resolution_time_summary
+      avg_resolution_time: avg_resolution_time_summary,
+      resolutions_count: resolutions.count,
+      reply_time: reply_time_summary
```
pranavrajs, line 55: @scmmishra Can we run the query only on the download? These would add additional query and would fail already overloaded API.

pranavrajs (changes requested):
Please find the comment above.

app/builders/v2/report_builder.rb
scmmishra: The report for download is made using the `generate_report` function in report_helpers which uses the `short_summary` method internally. All these params are required to generate the new downloads. 

One thing I can do is use the `summary` method for agent and teams report and rest of the reports can use the short summary. You can see the changes in this commit https://github.com/chatwoot/chatwoot/pull/9296/commits/134e8865bc744e353652228a6ca077298881e3db

Is this what you were looking for?

pranavrajs (approved):

PR #9310, opened 2024-04-27
feat: Revamp basic profile, avatar and message signature

muhsin-k:
| Light mode | Dark mode |
|--------------------|---------------------|
| CleanShot 2024-04-27 at 11 57 11@2x| CleanShot 2024-04-27 at 11 57 38@2x


app/javascript/dashboard/routes/dashboard/settings/personal/BasePersonalItem.vue
scmmishra, line 1: I am assuming this wraps a section for a form, we can call it `FormSection` and move it to a more generic location

app/javascript/dashboard/routes/dashboard/settings/personal/Index.vue
```diff
@@ -0,0 +1,159 @@
...
+  },
+  mixins: [alertMixin, globalConfigMixin, uiSettingsMixin],
+  data() {
+    return {
+      avatarFile: '',
+      avatarUrl: '',
+      name: '',
+      displayName: '',
+      email: '',
+      alertMessage: '',
```
scmmishra, line 71: Alert message does not have to be a state managed by the component, the usage is limited to the function and can be defined there itself

app/javascript/dashboard/routes/dashboard/settings/personal/Index.vue
```diff
@@ -0,0 +1,159 @@
...
+      currentUserId: 'getCurrentUserID',
+      globalConfig: 'globalConfig/get',
+    }),
+  },
+  watch: {
+    currentUserId(newCurrentUserId, prevCurrentUserId) {
+      if (prevCurrentUserId !== newCurrentUserId) {
+        this.initializeUser();
+      }
+    },
```
scmmishra, line 87: Why do we need this? Since a user would be able to access their profile only, so there shouldn't be a reason that the `userId` changes

app/javascript/dashboard/routes/dashboard/settings/personal/Index.vue
```diff
@@ -0,0 +1,159 @@
...
+  methods: {
+    initializeUser() {
+      this.name = this.currentUser.name;
+      this.email = this.currentUser.email;
+      this.avatarUrl = this.currentUser.avatar_url;
+      this.displayName = this.currentUser.display_name;
+      this.messageSignature = this.currentUser.message_signature;
+    },
+    isEditorHotKeyEnabled,
+    async updateProfile(userAttributes, type = 'profile') {
```
scmmishra, line 103: This method handles both profile and signature update, while the same API endpoint is called eventually, it is a bit confusion, let's split these into two methods

app/javascript/dashboard/routes/dashboard/settings/personal/MessageSignature.vue
```diff
@@ -0,0 +1,52 @@
+<template>
+  <div class="flex flex-col gap-6">
+    <form class="flex flex-col gap-6" @submit.prevent="updateSignature()">
```
scmmishra, line 3: I don't think we need a `div` wrapping the `form`, since there's no other children

app/javascript/dashboard/routes/dashboard/settings/personal/MessageSignature.vue
```diff
@@ -0,0 +1,52 @@
...
+watch(
+  () => props.messageSignature,
+  newValue => {
+    signature.value = newValue;
+  }
+);
+
+const updateSignature = () => {
+  emit('update-signature', { signature: signature.value }, 'signature');
+};
```
scmmishra, line 51: Here, a computed getter and setter would work much better

```
const signatureComputed = computed({
  get: () => props.messageSignature,
  set: (newValue) => {
    emit('update-signature', { signature: newValue }, 'signature');
  },
});
```

The current approach introduces a mutable state `signature` while, `signatureComputed` is not. Also the reactivity in the previous case came from the `watch`, here the reactivity is handled by vue since `signatureComputed` is a computedRef.

app/javascript/dashboard/routes/dashboard/settings/personal/UserBasicDetails.vue
scmmishra, line 1: We can rewrite this in the new syntax

app/javascript/dashboard/routes/dashboard/settings/personal/UserBasicDetails.vue
```diff
@@ -0,0 +1,139 @@
...
+  },
+  computed: {
+    inputStyles() {
+      return {
+        borderRadius: '12px',
+        padding: '6px 12px',
+        fontSize: '14px',
+        marginBottom: '2px',
+      };
+    },
```
scmmishra, line 102: This should be in the `data`

app/javascript/dashboard/routes/dashboard/settings/personal/UserProfilePicture.vue
```diff
@@ -0,0 +1,43 @@
...
+      </span>
+
+      <profile-avatar
+        :src="src"
+        :name="userNameWithoutEmoji"
+        @change="updateProfilePicture"
+        @delete="deleteProfilePicture"
+      />
+    </div>
+  </div>
```
scmmishra, line 15:
```suggestion
  <div class="flex flex-col gap-2">
    <span class="text-sm font-medium text-ash-900">
      {{ $t('PROFILE_SETTINGS.FORM.PICTURE') }}
    </span>

    <profile-avatar
      :src="src"
      :name="userNameWithoutEmoji"
      @change="updateProfilePicture"
      @delete="deleteProfilePicture"
    />
  </div>
```

scmmishra (changes requested):
Suggested a few changes. Also, what happens to the old page for profile settings?

app/javascript/dashboard/routes/dashboard/settings/personal/MessageSignature.vue
muhsin-k: If we use computed the `update-signature` function is called on every input, which is not necessary. We only need to save the signature when the user clicks the "Save message signature" button.

muhsin-k:
> Also, what happens to the old page for profile settings?
@scmmishra This just a temporary page for testing. Once I merge all the components for profile section. I will rename this route and delete the old profile page.

app/javascript/dashboard/routes/dashboard/settings/personal/MessageSignature.vue
muhsin-k: @scmmishra

app/javascript/dashboard/routes/dashboard/settings/personal/MessageSignature.vue
```diff
@@ -0,0 +1,50 @@
...
+      color-scheme="primary"
+      variant="solid"
+      size="large"
+    >
+      {{ $t('PROFILE_SETTINGS.FORM.MESSAGE_SIGNATURE_SECTION.BTN_TEXT') }}
+    </form-button>
+  </form>
+</template>
+<script setup>
+import { defineProps, ref, watch } from 'vue';
```
iamsivin, line 24: We don't need to import the `defineProps`
```suggestion
import { ref, watch } from 'vue';
```

app/javascript/dashboard/routes/dashboard/settings/personal/personal.routes.js
```diff
@@ -0,0 +1,20 @@
+import { frontendURL } from '../../../../helper/URLHelper';
```
iamsivin, line 1:
```suggestion
import { frontendURL } from 'dashboard/helper/URLHelper';
```

app/javascript/dashboard/routes/dashboard/settings/personal/MessageSignature.vue
scmmishra: Makes sense, resolving this

app/javascript/dashboard/components/FormSection.vue
```diff
@@ -0,0 +1,31 @@
...
+        {{ header }}
+      </h4>
+      <div class="flex flex-row items-center justify-between">
+        <div class="flex-grow h-px bg-ash-200" />
+      </div>
+      <p v-if="description" class="mb-0 text-sm font-normal text-ash-900">
+        {{ description }}
+      </p>
+    </div>
+    <div v-if="$slots.default" class="flex flex-col w-full gap-6">
```
scmmishra, line 14:
```suggestion
    <div class="flex flex-col w-full gap-6">
```

We don't need this check, since this component should not be used without the default slot

app/javascript/dashboard/routes/dashboard/settings/personal/Index.vue
scmmishra, line 1: Let's rewrite this in the setup syntax once the composables are merged. Approving it for now

scmmishra (approved):

PR #9311, opened 2024-04-27
feat: Revamp hotkeys and change password in profile settings

muhsin-k:
| Light mode | Dark mode |
|--------------------|---------------------|
CleanShot 2024-04-27 at 12 09 58@2x| CleanShot 2024-04-27 at 12 10 11@2x


app/javascript/dashboard/routes/dashboard/settings/personal/HotKeyCard.vue
```diff
@@ -0,0 +1,64 @@
+<template>
+  <div
+    class="flex flex-col gap-4 w-full h-fit sm:max-h-[220px] p-4 sm:max-w-[350px] rounded-md border border-solid border-ash-200"
+    :class="{
+      'border border-solid border-primary-300 ': active,
```
iamsivin, line 5: @muhsin-k We do not need to add the `border border-solid` here again already added at the top. And I think we don't need to add the `border-solid` because the default value for the border will be the same.

iamsivin (approved):
Looks good

PR #9318, opened 2024-04-29
feat: allow feature plan map in super admin

scmmishra:
This PR adds the following changes

1. Add `CHATWOOT_CLOUD_PLAN_FEATURES` config
1. ~Show `CHATWOOT_CLOUD_PLANS` and `CHATWOOT_CLOUD_PLAN_FEATURES` in the dashboard, but keep them disabled~ (reverted)
1. Add `subscribed_features` method in `models/enterprise/account` and include it in the JSON response

CleanShot 2024-04-29 at 14 32 23@2x

CleanShot 2024-04-29 at 15 28 46@2x


app/views/super_admin/app_configs/show.html.erb
```diff
@@ -13,18 +13,25 @@
...
         </div>
-        <div class="field-unit__field -mt-2 ">
+        <div class="-mt-2 field-unit__field ">
           <% if @installation_configs[key]&.dig('type') == 'boolean' %>
             <%= form.select "app_config[#{key}]", 
               [["True", true], ["False", false]], 
               { selected: ActiveModel::Type::Boolean.new.cast(@app_config[key]) },
               class: "mt-2 border border-slate-100 p-1 rounded-md" 
             %>
+          <% elsif @installation_configs[key]&.dig('type') == 'json' %>
```
scmmishra, line 23: @sojan-official need your thoughts on this, it re-uses the `locked` attributed, not sure if it is intended for that

sojan-official: The locked config is for the page `http://localhost:3000/super_admin/installation_configs`, which we are slowly deprecating. We filter the locked values here and won't display it in the list. 

For this particular case, I think we only need to show this for the chatwoot cloud alone. We don't need to show these values in the UI for even the premium plans. we can add the check based on `DEPLOYMENT_ENV`

scmmishra: Makes sense, I guess we can hide it completely. I'll make the required changes

scmmishra (commented):
@sojan-official made the suggested changes

sojan-official (approved):

sojan-official:
@scmmishra code looks good to me. please check the failing spec

scmmishra:
@sojan-official fixed it!

PR #9341, opened 2024-05-02
feat: don't allow dropdown navigation on input focused

scmmishra:
Fixes https://linear.app/chatwoot/issue/CW-3266/i-cannot-click-up-or-down-arrow-buttons-in-the-editor


iamsivin (approved):
Tested the PR. And it's working fine.

PR #9347, opened 2024-05-03
feat: exporting contacts takes the filters into account

scmmishra:
This PR allows exporting contacts with the current filter.

### Behaviour

It matches the results from the UI

Page | Export Result
-- | --
All Contacts - Filter Applied | Exports filtered contacts
All Contacts - No Filters | Exports all contacts
Segment Page | Exports the active segment
Labels Page - No Filters | Exports all contacts with label
Labels Page - Filters Applied | Exports filtered contacts, without label filter

## Todo

- [x] Exporting Segments
- [x] Exporting Labels

<details><summary>Filtered Export</summary>
<p>

CleanShot 2024-05-03 at 16 08 40@2x

</p>
</details> 

<details><summary>Default Export</summary>
<p>

CleanShot 2024-05-03 at 16 08 55@2x

</p>
</details>


scmmishra:
@pranavrajs this is good to review PTAL

sojan-official:
Note: Added some minor changes 
- Reorder the parameters for the contact filter service and export job to stay consistent with the priority 
- Removed the dependency of the contact filter service on the Current
- Ensured that the exports return only the resolved contacts ( for all pages and labels )

> The filtered export still returns unresolved contacts; I left it since the behaviour is constant with the filter service. lets address in subsequent PR

sojan-official:
Verified the following cases

- All contacts export all resolved contacts
- Filtered contacts exports filtered contact ( including unresolved contacts )
- label page exports filtered contacts with the labels

sojan-official (approved):

sojan-official:
Note: further clean-ups 

- renaming some variables to void confusions
- removed the conditions added to the filter service since these changes shouldn't affect it.

PR #9349, opened 2024-05-06
refactor: use processed message content for slack

scmmishra:
Slack messages can get pretty long especially for email threads, this PR updates the payload to first look for `processed_message_content` and then fallback to `content`

The `processed_message_content` for emails ensures that only the main email body is used in the slack payload.


scmmishra:
Tested this locally, works fine. @pranavrajs @muhsin-k  PTAL

muhsin-k (approved):

PR #9436, opened 2024-05-08
fix: no fallback for created_by & updated_by

scmmishra:
Fixes https://chatwoot-p3.sentry.io/share/issue/29730a9df77b402b8ee384f86d80081c

Tried finding a root cause, but couldn't seems like the best solution for now


iamsivin (approved):
Look good

fayazara (approved):
LGTM

pranavrajs:
@scmmishra Is this happening when the user is deleted from the system?

pranavrajs:
Screenshot 2024-05-08 at 1 39 05 PM

Yep! This was the case. If the user is removed from the account, the UI would throw an error.

PR #9441, opened 2024-05-09
feat: add inbox reconnection banner

scmmishra:
CleanShot 2024-05-09 at 12 44 07@2x

CleanShot 2024-05-09 at 12 44 36@2x


fayazara (approved):
Looks good, tested locally

PR #9443, opened 2024-05-09
fix: widget does not load when navigating on pages with view transition [CW-3249]

scmmishra:
When visiting another page in turbo, the `body` is replaced completely, this makes the navigation more SPA like. However the new body does not contain any of the elements added by a client script. This causes the SDK to fail. 

This PR fixes this by restoring all the elements when rendering the new body. The alternative is re-running the script, however our SDK needs more changes to support resuming, besides it causes the widget to disappear and come back, making it the inferior choice in terms of UX

Something similar happens with Astro with view transitions, this PR fixes that too

## The fix

1. Adding a unique ID to the bubble, the styles as well as the widget holder
2. Adding a `restoreWidgetInDOM` method to `DOMHelpers` this does the work of replacing the elements as required.
3. Adding the event listener for `turbo:before-render` and `astro:before-swap`

## Things to note

While not necessary, we need to add `data-turbo-eval="false"` to the widget embed script, otherwise the sdk script tag will keep getting added with each navigation. It won't affect the widget.

This is not an issue with astro, since it's default behaviour is to no re-run scripts

---

P.S. We can remove the conditions that check if a site uses `turbo` or `astro`


app/javascript/packs/sdk.js
```diff
@@ -10,14 +10,33 @@ import {
...
   }
 
+  if (window.Turbo) {
+    // if this is a Rails Turbo app
+    document.addEventListener('turbo:before-render', event =>
+      restoreWidgetInDOM(event.detail.newBody)
+    );
+  }
+
+  if (document.querySelector('meta[name="generator"][content^="Astro v"]')) {
```
scmmishra, line 33: @pranavrajs this condition is not the most reliable, since one can remove the generator, I am removing this for now. This will add the event listener anyway, but it won't be triggered unless the page uses astro

scmmishra:
@pranavrajs @fayazara PTAL

pranavrajs (approved):

fayazara:
Quick question - is this intended? @scmmishra 
https://github.com/chatwoot/chatwoot/assets/15716057/ba26c44d-bb93-44af-a2fe-8a84a91f1d0e

scmmishra:
> Quick question - is this intended? @scmmishra
> 
>  CleanShot.2024-05-15.at.11.20.16.mp4

Yeah, once the widget remounts, the iframe is forced to render again, nothing we can do about it

PR #9448, opened 2024-05-09
fix: Update the voice note format to MP3 to fix the delivery issues 

pranavrajs:
Use MP3 as the default format to send voice notes recorded from Chatwoot. This change was made to fix the issue of Telegram voice notes not working with the error `WEBPAGE_CURL_FAILED` .

Telegram treats the mp3 recordings as audio attachments. Once we can identify a fix for the original issue, we will revisit the `ogg` implementation.


sojan-official:
note: tested and verified working on staging

PR #9457, opened 2024-05-13
fix: Widget phone number input country undefined in onSelectCountry

iamsivin:
# Pull Request Template

## Description

This PR will fix this sentry [issue](https://chatwoot-p3.sentry.io/issues/5291039795/)

**Issue**
The root cause of this issue is the usage of `mentionSelectionKeyboardMixin.` Because the key events are always active even if the dropdown is not visible. So, if we press the enter key, this error will be thrown into the console.

**Solution**
Get rid of the usage of `mentionSelectionKeyboardMixin` and handle it directly in the component, and always check if the dropdown is active or not.
And fix the selected is not resetting on search.

Fixes https://linear.app/chatwoot/issue/CW-3282/phonenumberinput-country-undefined-in-onselectcountry

## Type of change

## How Has This Been Tested?

**Steps**
1. Enable phone input from the pre-chat form.
2. And open the widget and start a conversation.
3. Then press the enter key, and you can see this error in the console.

## Checklist:


scmmishra (approved):
Good thinking using the default events! LGTM!

PR #9476, opened 2024-05-15
feat: Improve Report API performance

pranavrajs:
- Re-write the methods for clarity
- Remove the dependency on the ReportHelper class.
- Remove n+1 queries in the average metric time series data.


pranavrajs:
Code Climate issue can be ignored as we will remove the reports_helper in the subsequent PRs.

sojan-official (approved):
- verified all reports are loading in local

PR #9482, opened 2024-05-16
feat: move Google config to installation_config

scmmishra:

1. Add `GOOGLE_OAUTH_CLIENT_ID` and `GOOGLE_OAUTH_CLIENT_SECRET` to installation config
2. Add Google config to `super_admin/features.yml`
3. Replace usage of `ENV.fetch` with `GlobalConfigService.load` for fetch client ID and Secret


config/initializers/omniauth.rb
```diff
@@ -1,5 +1,6 @@
 Rails.application.config.middleware.use OmniAuth::Builder do
-  provider :google_oauth2, ENV.fetch('GOOGLE_OAUTH_CLIENT_ID', nil), ENV.fetch('GOOGLE_OAUTH_CLIENT_SECRET', nil), {
-    provider_ignores_state: true
-  }
+  client_id = GlobalConfigService.load('GOOGLE_OAUTH_CLIENT_ID', nil)
```
scmmishra, line 2: @sojan-official getting the values from `GlobalConfigService` in this initializer, fails the CI test. You got ideas on how to fix this?

scmmishra: P.S. Using this branch to test, https://github.com/chatwoot/chatwoot/pull/9504
I have removed the yarn step to speed up the CI

scmmishra: There's a possible solution here: https://github.com/chatwoot/chatwoot/pull/9504/files#diff-84ed1be2c135a8c628a3fb258e88c059ae24b3cb90abce9b5a5f74e96eb5c7ee

Not sure if it feels right. While the DB exists, the rescue case will not run. That means in production this will run as expected, while bypassing the issue in CI

LMK what you think @sojan-official

scmmishra: @pranavrajs @sojan-official here, the `OmniAuth::Builder` does a lot more than just setting up the client, it also creates a Rack middleware stack by using the `use` method inherited from `Rack::Builder`. [ref](https://github.com/omniauth/omniauth/blob/master/lib/omniauth/builder.rb) 

Removing this seems non-trivial

sojan-official: we can look into lazyload hooks ? 
ref: https://gorails.com/guides/rails-lazy-load-hooks-cheatsheet
https://edgeapi.rubyonrails.org/classes/ActiveSupport/LazyLoadHooks.html

scmmishra: Tried lazy loading, it fixes CI, but fails in production. Investigating it
Error for ref: https://chatwoot-p3.sentry.io/issues/5394916763

scmmishra: I might have figured this one out, testing on staging

scmmishra: @sojan-official this is resolved, PTAL

scmmishra: False alarm, still figuring this out

scmmishra: https://github.com/omniauth/omniauth/issues/1077

Seems like there's no way to do this, @sojan-official can you give this a shot?

sojan-official: will take this forward

PR #9510, opened 2024-05-21
feat: new re-authorization flow for Microsoft

scmmishra:
This PR adds a cleaner re-authorization flow to Microsoft. This PR has the following changes

1. Use `reauthorization_required` value for Microsoft Channel
2. Refactor `InboxReconnectionRequired` to reuse the `banner` component
3. Refactor `microsoft/Reauthorize.vue` to reuse `InboxReconnectionRequired` component
4. Update `reauthorizable.rb` to update cache keys if the model has an inbox
5. Update `microsoft/callbacks_controller.rb` to handle the reauthorization case with a redirect to the inbox settings page if the inbox already exists at the time of authorization. 

## How Has This Been Tested?

- [x] Local Instance
- [ ] Staging Instance
- [x] Unit tests

## Pending Tasks

- [ ] ~Success Toast~ will do this in a follow-up PR with the screen

## Demo

The following video shows the whole process of creation and re-authorization of the Microsoft channel

https://www.loom.com/share/e5cd9bd4439c4741b0dcfe66d67f88b3?sid=100f3642-43e4-46b3-8123-88a5dd9d8509


scmmishra:
@pranavrajs I figured the flow is simple enough that we don't need an additional screen for the reauthorization, let me know how it looks, I've attached a demo video. 

However I can see the value add of the screen, easier to provide more information also more detailed error messages. Let me know how this flow looks, I can create a PR on top of this to add the additional screens

pranavrajs:
@scmmishra, great job! We wanted a separate page to display the error or success messages resulting from the flow.

Currently, during our initial inbox setup process, we redirect users to the dashboard when an error occurs, which can be confusing. This has led to multiple reports from people who don't fully understand the error.

This flow is fine. However, if we can ensure that the follow-up includes the screen mentioned in the issue, we can proceed with merging the PRs.

config/routes.rb
```diff
@@ -20,6 +20,7 @@
     get '/app/accounts/:account_id/settings/inboxes/new/microsoft', to: 'dashboard#index', as: 'app_new_microsoft_inbox'
     get '/app/accounts/:account_id/settings/inboxes/new/:inbox_id/agents', to: 'dashboard#index', as: 'app_twitter_inbox_agents'
     get '/app/accounts/:account_id/settings/inboxes/new/:inbox_id/agents', to: 'dashboard#index', as: 'app_microsoft_inbox_agents'
+    get '/app/accounts/:account_id/settings/inboxes/:inbox_id', to: 'dashboard#index', as: 'app_microsoft_inbox_settings'
```
pranavrajs, line 23: Why do we need this?

pranavrajs (approved):
Minor comments, otherwise looks good.

config/routes.rb
scmmishra: The original callback redirected to the `app_microsoft_inbox_agents` this is the onboarding page where the admin is asked to add agents. The new route is for the inbox settings page, the user is redirected here when reauthorized, I can rename it to `app_inbox_settings` since it is not Microsoft specific

scmmishra:
@muhsin-k need your review on this PR, also if you can test it once

muhsin-k (approved):

PR #9511, opened 2024-05-21
fix: Cannot read properties of undefined (reading 'toLowerCase')

scmmishra:
Tried to replicate the issue, but Sentry didn't have enough information. `toggleMessageSignature` is a user triggered action in `ReplyBottomPanel.vue`, the value for `channelType` is provided from `inboxMixin`. The error will occur if either `inbox` is an empty object `{}` or `channel_type` in `inbox` object is undefined.

I couldn't find any instance where this could be the case. The PR has a stop gap solution that ensures that no action is triggered


iamsivin (approved):
Code LGTM

PR #9553, opened 2024-05-27
feat: add reauthorization banner for Facebook

scmmishra:

1. Use `reauthorization_required` as standard instead of inbox specific flags
1. Show alert on the sidebar along with a tooltip for Microsoft and Facebook both
1. Update  `facebook/Reauthorize.vue` component to use the banner UI instead of the form
1. Update alert icon layout on the sidebar to make it more prominent

#### Updated tooltip on the sidebar

CleanShot 2024-05-28 at 12 49 32@2x

#### Updated facebook page

CleanShot 2024-05-28 at 12 51 30@2x

### DIFF
--- a/app/javascript/dashboard/components/layout/sidebarComponents/SecondaryChildNavItem.vue
+++ b/app/javascript/dashboard/components/layout/sidebarComponents/SecondaryChildNavItem.vue
@@ -65,11 +65,11 @@
         </div>
         <span
           v-if="warningIcon"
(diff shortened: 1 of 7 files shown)


muhsin-k (approved):
Tested locally and reconnection is also working well.
CleanShot 2024-05-29 at 10 39 48@2x
CleanShot 2024-05-29 at 10 40 02@2x

PR #9560, opened 2024-05-29
fix: Handle OpenAI API errors

muhsin-k:
We have OpenAI integration  to enhance conversations, which requires users to input their OpenAI key. Currently, we lack validation to check whether the provided key is valid or if the associated account has sufficient quota for using the OpenAI integration. As a result, users may not receive proper error messages when attempting to use certain AI features with an invalid key. This PR addresses this issue by introducing a fix that will display meaningful error messages on the front end, indicating what went wrong.

**Preview**
CleanShot 2024-05-29 at 12 55 25@2x


sojan-official (approved):
Code LGTM. 
Looks good to merge after testing.

sojan-official:
@muhsin-k, could you take care of the spec failures? let's get this merged after that !!

PR #9605, opened 2024-06-07
fix: Use message sender ID in the analytics for first response time

pranavrajs:
`conversation.assignee_id` was taken into consideration for first response time calculation. There was one flaw in this approach. A lot of customer support agents respond to the conversation first before assigning it to themselves. This makes the analytics broken.

To fix this, we will take the message sender instead of assignee in calculations, which should solve the issue with metric.

Fixes https://linear.app/chatwoot/issue/CW-3375/first-response-reporting-events-use-assignee-id-instead-of-agent-who


sojan-official (approved):

sojan-official:
Verified locally that first response reports are getting generated correctly even if the agent is not the assignee

PR #9612, opened 2024-06-08
fix: clean cache to process message again when fail

clairton:
# clean cache to process message again when fail

When message fail delete key process cache


clairton:
@pranavrajs @sojan-official sorry by mentionate, but this PR is very important!

clairton:
without this fix, some messages is not processed @pranavrajs @sojan-official

pranavrajs:
@clairton Can you please share the case where this breaks? How do I reproduce it?

clairton:
when an error occurs in processing, and the job runs again, it finds the cache and discards the message

pranavrajs:
> when an error occurs in processing, and the job runs again, it finds the cache and discards the message

@clairton I understand this, but where is the error occurring? There are only DB calls in between. Can you share a case where this breaks?

PR #9629, opened 2024-06-11
feat: FCM HTTP v1 API changes

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-3210/legacy-firebase-changes


app/services/notification/fcm_service.rb
```diff
@@ -0,0 +1,45 @@
...
+    }
+  end
+
+  def credentials_path
+    file = Tempfile.new(['firebase', '.json'])
+    file.write(@credentials)
+    file.rewind
+    file.path
+  ensure
+    file.close
```
scmmishra, line 43:
```suggestion
    StringIO.new(@credentials)
```

let's do this instead? As suggested in their [README](https://github.com/decision-labs/fcm?tab=readme-ov-file#about-the-google_application_credentials_path). This will avoid the IO overhead of creating a Tempfile

scmmishra (changes requested):

scmmishra (approved):

Issue #1074, opened 2020-07-20
Email replies attach files as links, does not attach as attachment.


pranavrajs:
![image](https://github.com/chatwoot/chatwoot/assets/2246121/42f80511-9d7c-417f-9483-404bbe3a3024)

Instead of `attachment [Click here to view]` see the possibility to have images as inline to the message.

<sub>From [SyncLinear.com](https://synclinear.com) | [CW-2851](https://linear.app/chatwoot/issue/CW-2851/enhancement-email-replies-doesnot-attach-files)</sub>

sony-mathew:
I think it would be better to take this on once we have a rich text editor and we save the content of a message in a conversation as HTML or other rich text formats.

japps20:
@sony-mathew Another option is to use the format.

attachment - "Filename.pdf" [Click here to view]
attachment - "Another Filename.png" [Click here to view]
attachment - "A third Filename.pdf" [Click here to view]

Or

attachment - 
"[Filename.pdf](Filename.pdf)"
"[Another Filename.png](Another_Filename.png)"
"[A third Filename.pdf](A_third_Filename.pdf)"

Including the list and name of the attachments gives the recipients some comfort. It will also work both all type of attachment types.

sony-mathew:
@japps20 This makes more sense.  @sojan-official what do you think? Also, this could be a good first issue for hacktoberfest.

jean-philippeansel:
@sojan-official It would be even better if it was multilingual by design:

**English**
attachment - "[Filename.pdf](https://github.com/chatwoot/chatwoot/issues/Filename.pdf)" "[Another Filename.png](https://github.com/chatwoot/chatwoot/issues/Another_Filename.png)" "[A third Filename.pdf](https://github.com/chatwoot/chatwoot/issues/A_third_Filename.pdf)"

**French:**
Pièces jointes - "[Filename.pdf](https://github.com/chatwoot/chatwoot/issues/Filename.pdf)" "[Another Filename.png](https://github.com/chatwoot/chatwoot/issues/Another_Filename.png)" "[A third Filename.pdf](https://github.com/chatwoot/chatwoot/issues/A_third_Filename.pdf)"

Etc.

pranavrajs:
Things to consider here:

* We should configure a maximum limit for the attachments. Investigate the maximum size limit on the email clients (For eg: GMail attaches files more than 25MB as a drive link). All other attachments should be sent as a link. 
[...]

PR #9655, opened 2024-06-19
feat: Add attachments in email reply

marcomarinho:
# Add attachments in email reply

## Description

Add ability to send files as attachments instead of links

Fixes https://github.com/chatwoot/chatwoot/issues/1074 

## Type of change

## How Has This Been Tested?

Send email replies with attachments, notice that files are sent as attachments in the received email reply.

## Checklist:

### DIFF
--- a/app/mailers/conversation_reply_mailer_helper.rb
+++ b/app/mailers/conversation_reply_mailer_helper.rb
@@ -19,6 +19,33 @@ module ConversationReplyMailerHelper
 
     Rails.logger.info("Email sent from #{email_from} to #{to_emails} with subject #{mail_subject}")
 
+    if @message.attachments.present?
+      @options[:attachments] = []
+
+      @message.attachments.each do |attachment|
+        raw_data = attachment.file.download
+        attachment_name = attachment.file.filename.to_s
+        temp_dir = Rails.root.join('tmp/uploads')
+        FileUtils.mkdir_p(temp_dir)
+        temp_file_path = File.join(temp_dir, attachment_name)
+        File.write(temp_file_path, raw_data, mode: 'wb')
+        temp_file_path
+
+        # Get the size of the file before downloading
+        file_size = raw_data.bytesize
+
+        if file_size < 25.megabytes
+          # Store the temp file path and attachment name
+          @options[:attachments] << { name: attachment_name, path: temp_file_path }
+          mail.attachments[attachment_name] = File.read(temp_file_path)
+
+          Rails.logger.info("Attachment saved to #{temp_file_path}.")
+        else
+          Rails.logger.warn("Attachment #{attachment_name} is larger than 25MB and will be sent as a link")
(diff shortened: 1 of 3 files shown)


marcomarinho:
Still in WIP. 
TODO:
- Only trigger if attachments are less than 25MB
- Remove attachment link part in email_reply.html.erb with the condition above

marcomarinho:
Added 25MB condition

marcomarinho:
Some questions:
- transcript conversation should keep the current behavior with sending attachments as links? 
- reply with and without summary should keep the current behavior with sending attachments as links? I don't think I have such feature to test in my self-host, will need someone to test if current code breaks it.
- As of now, the 25Mb condition is per attachment, maybe we should also sum all attachments for that 25Mb condition instead? Like gmail does, what counts is the total size of attachments, not per attachment right?

Also a side note:
- I'm getting Net::ReadTimeout sometimes, had to increase the SMTP readout times, I can't be sure if it's due to this feature or not since I started using it with this feature already. But since it attaches files maybe takes longer to process the mail ? We should double check this to make sure doesn't happens to people or we have to increase the defaults SMTP readout times with this commit. Will need people to test further 

Cc @sojan-official

marcomarinho:
@sojan-official been using it for a month and no issues so far. Can we look into what's needed to merge?

sojan-official:
@marcomarinho, I apologise for the delay on this.  I just wanted to confirm that this is on our radar; we will be prioritising this after the current work on the Vite upgrade.

darker0n:
+1

darker0n:
Hello, I got this error when sending [Conversation Continuity](https://www.chatwoot.com/docs/self-hosted/configuration/features/email-channel/conversation-continuity) email.

Screenshot 2024-10-01 at 1 20 08 AM

With a standard email channel messages work fine.

app/mailers/conversation_reply_mailer_helper.rb
```diff
@@ -19,6 +19,33 @@ def prepare_mail(cc_bcc_enabled)
 
     Rails.logger.info("Email sent from #{email_from} to #{to_emails} with subject #{mail_subject}")
 
+    if @message.attachments.present?
```
darker0n, line 22: To fix [this](https://github.com/chatwoot/chatwoot/pull/9655#issuecomment-2384170806) just edit this line `if @message&.attachments.present?`

marcomarinho:
I can upstream it and fix the bug reported, although up to @sojan-official and his team to merge it after

pranavrajs:
@marcomarinho Is it possible to allow us to edit the PR? We would like to test this out.

sojan-official:
Taking this up over : https://github.com/chatwoot/chatwoot/pull/11304 
Since we didn't have the permissions to update the original PR 

@marcomarinho thanks again for the work.

PR #9668, opened 2024-06-23
feat: Add the ability to un-assign teams using automation

clairton:
# Pull Request Template

## Description

Add none to team list in automation


clairton:
Captura de Tela 2024-06-22 às 21 15 57

pranavrajs:
@clairton, thank you for raising the PR. Upon testing, I found that the action wasn't functioning correctly. This issue isn't a result of your changes, but it was a pre-existing bug. We didn't discover it earlier since we weren't using the `unassign_team ` method. So, I've added a fix for it. Now, it should be good to merge. 

I will wait for the specs to pass and will get it merged afterwards.

pranavrajs (approved):

PR #7156, opened 2023-05-22
feat: Attachments view

iamsivin:
# Pull Request Template

## Description

This PR includes,
1. Ability to see all attachments related to current chat (like- audio, video, image) by clicking arrow buttons
2. Adds the ability to control it by key shortcuts
   - Escape key closes the modal
   - Arrow the right key to move to the right if multiple attachments exist.
   - Arrow the left key to move to the left if multiple attachments exist.
3. Add the ability to close the attachment preview when we click outside the image

Fixes https://linear.app/chatwoot/issue/CW-1540/usability-feedback

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?
https://www.loom.com/share/c669af96739b4c48b55d67cf71e1feb7

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #9706, opened 2024-07-01
fix: add check for chat attachments

scmmishra:
There might be a case where the chat is initialised as an empty object without the `attachments` array present. This PR handles that

Ref: https://github.com/chatwoot/chatwoot/pull/7156/


iamsivin (approved):
Tested locally and LGTM

app/javascript/dashboard/store/modules/conversations/index.js
```diff
@@ -151,7 +151,7 @@ export const mutations = {
       selectedChatId: conversationId,
     });
 
-    if (!chat) return;
+    if (!chat?.attachments) return;
```
pranavrajs, line 154: Can you identify the case where the chat is initialized without attachments?

marcomarinho: I do have this same issue where chat is initialized without attachments, mainly with inline images.. will test this and give feedback aswell

pranavrajs: Thanks! Appreciate it.

marcomarinho: Doesn't work in my use case. Inline images still not attached.

scmmishra:
Fixed here: https://github.com/chatwoot/chatwoot/pull/9784

PR #9757, opened 2024-07-10
fix: Add a DISCONNECT_DELAY_THRESHOLD while fetching the conversations

pranavrajs:
The disconnect threshold is added to account for delays in identifying disconnections (for example, the websocket disconnection takes up to 3 seconds) while fetching the latest updated conversations or messages.

In this case, the cable disconnection event takes about 3 seconds to fire. If there was a conversation which was created in this 3 second, it would not be displayed in the UI until the refresh.

### DIFF
--- a/app/javascript/dashboard/helper/ReconnectService.js
+++ b/app/javascript/dashboard/helper/ReconnectService.js
@@ -9,6 +9,11 @@ import {
 
 const MAX_DISCONNECT_SECONDS = 10800;
 
+// The disconnect delay threshold is added to account for delays in identifying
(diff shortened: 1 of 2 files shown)


iamsivin (approved):
I have tested this PR locally, and it's working fine.

**Before**
The conversation created is missing when its reconnected.
https://www.loom.com/share/a7fad0e3f3d146f190b3ce2798a5f31f?sid=66003036-ddb7-47e7-9a4c-b135775b0b0f

**After**
https://www.loom.com/share/9fd8dc0681114802986eb9cc90aafa54?sid=d03358d9-44b2-4b6a-ac9f-1cd62f7d1877

PR #9765, opened 2024-07-11
fix: TypeError cannot read property 'focus' of undefined

iamsivin:
# Pull Request Template

## Description

**Issue**
The error `Cannot read property 'focus' of undefined`occurs because the `this.$refs.searchbar` is undefined at the time `focusSearchInput()` method is called. This can happen if the method is called before the component has fully mounted or if the `ref="searchbar"` is not correctly attached to any element. 

I tried multiple ways to reproduce this issue, but none worked.

**Solution**
To resolve this issue, I can add a check to ensure that `this.$refs.searchbar` is defined before calling `focus()` method.

Fixes https://linear.app/chatwoot/issue/CW-3405/typeerror-cannot-read-property-focus-of-undefined

## Type of change

## Checklist:

### DIFF
--- a/app/javascript/shared/components/emoji/EmojiInput.vue
+++ b/app/javascript/shared/components/emoji/EmojiInput.vue
@@ -4,12 +4,12 @@
     class="emoji-dialog bg-white shadow-lg dark:bg-slate-900 rounded-md border border-solid border-slate-75 dark:border-slate-800/50 box-content h-[300px] absolute right-0 -top-[95px] w-80 z-20"
   >
     <div class="flex flex-col">
-      <div class="emoji-search--wrap flex gap-2">
+      <div class="flex gap-2 emoji-search--wrap">
         <input
           ref="searchbar"
           v-model="search"
           type="text"
-          class="emoji-search--input focus:box-shadow-blue dark:focus:box-shadow-dark"
+          class="emoji-search--input focus:box-shadow-blue dark:focus:box-shadow-dark !mb-0"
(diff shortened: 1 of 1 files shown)


app/javascript/shared/components/emoji/EmojiInput.vue
```diff
@@ -174,7 +174,7 @@ export default {
       return categoryItem ? categoryItem.emojis[0].emoji : '';
     },
     focusSearchInput() {
-      this.$refs.searchbar.focus();
+      this.$refs.searchbar?.focus();
```
scmmishra, line 177: Can we use `autofocus` property on the input instead?

iamsivin: @scmmishra I tried adding the `autofocus` property for input, but it doesn't work.

scmmishra: Were you able to reproduce it?. Try adding a `$nextTick` see if that fixes the issue.

iamsivin: @scmmishra No, I couldn't able to reproduce it.

iamsivin: @scmmishra Updated the PR

scmmishra: Sounds good, if this occurs again, we will now. I will update the sentry rule

scmmishra (approved):

PR #9768, opened 2024-07-12
fix: return if user is not present

scmmishra:
Fixes https://chatwoot-p3.sentry.io/share/issue/aa6a3c0accb345e984ed80eaddc3efd5/

In the case where the user is redirected to the page from a redirect like google login, the analytics helper breaks. This PR fixes it by checking if the user is present.

This does not break the user email tracking since the user is redirected anyway and the `AnalyticsHelper` is reinitialised


iamsivin (approved):
LGTM

PR #9777, opened 2024-07-15
fix: TypeError cannot read properties of undefined (reading '$store')

iamsivin:
# Pull Request Template

## Description

**Cases**
https://linear.app/chatwoot/issue/CW-3396/typeerror-cannot-read-properties-of-undefined-reading-dollarstore#comment-f018b29c

Fixes https://linear.app/chatwoot/issue/CW-3396/typeerror-cannot-read-properties-of-undefined-reading-dollarstore

## Type of change

## Checklist:

### DIFF
--- a/app/javascript/dashboard/helper/AudioAlerts/DashboardAudioNotificationHelper.js
+++ b/app/javascript/dashboard/helper/AudioAlerts/DashboardAudioNotificationHelper.js
@@ -52,6 +52,10 @@ class DashboardAudioNotificationHelper {
   };
 
   executeRecurringNotification = () => {
+    if (!window.WOOT || !window.WOOT.$store) {
+      this.clearSetTimeout();
+      return;
+    }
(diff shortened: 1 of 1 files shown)


scmmishra:
This issue is an extremely narrow edge case that does not affect the app functionality. It occurs here

1. `executeRecurringNotification` is the method that sets a timeout of 30 seconds and pings as required
2. This method is triggered when one of the following events happen `click`, `touchstart`, `keypress`, `keydown`
3. The class in question is initalized before the Vue app is mounted in `application.js`, the vue app is mounted on `window.onload` event
4. In a case where the Vue app is not created (i.e. `window.onload` event is not fired) within 30 seconds of a user input, this error occurs.

So far we have only had 2 such events in the Sentry history, we can safely archive it until it happens again.

scmmishra (approved):

PR #9816, opened 2024-07-22
refactor: use `structuredClone` instead of `JSON.stringify` and `JSON.parse`

scmmishra:

1. Add a method that creates a deep clone of the provided object. This function attempts to use the `structuredClone` method if available. If `structuredClone` is not supported, it falls back to using `JSON.parse(JSON.stringify(obj))`.
2. Replace usage of `JSON.parse(JSON.stringify` across the app with `structuredClone`

### Example 

```js
const original = { a: 1, b: { c: 2 } };
const clone = cloneObject(original);
console.log(clone); // { a: 1, b: { c: 2 } }
console.log(original === clone); // false
```

### References

1. [MDN](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone)
2. [Can I Use](https://caniuse.com/mdn-api_structuredclone)


app/javascript/dashboard/helper/clone.js
```diff
@@ -0,0 +1,33 @@
...
+ *                     JSON methods.
+ *
+ * @example
+ * const original = { a: 1, b: { c: 2 } };
+ * const clone = cloneObject(original);
+ * console.log(clone); // { a: 1, b: { c: 2 } }
+ * console.log(original === clone); // false
+ */
+export function cloneObject(obj) {
+  if (typeof structuredClone === 'function') {
```
muhsin-k, line 22: I would suggest adding a try-catch block could make the function more robust.

```
try {
    return structuredClone(obj);
  } catch (error) {
    try {
      return JSON.parse(JSON.stringify(obj));
    } catch (error) {
      throw new TypeError('Failed to clone object.');
    }
}
```

muhsin-k, line 22: Should we add the case for type nil?
```
 if (obj === null) {
    return obj;
 }
```

scmmishra: `structuredClone(null)` and `JSON.parse(JSON.stringify(null))` both return `null` anyway, so we should be good

scmmishra: An error will be thrown anyway, seems futile to catch the error and then throw another one while handling that.

muhsin-k (approved):

pranavrajs:
@scmmishra 
- This is a utility function and should be moved to @chatwoot/utils.
- Use the spyOn function whenever possible instead of using `global.`
- Mock restore is set up globally—you don't need to restore it manually each time.

Weslenndo:
A é aki ento

Em dom, 8 de set de 2024 00:33, github-actions[bot] <
***@***.***> escreveu:

PR #9825, opened 2024-07-23
feat: Update the integration page

pranavrajs:
Combine integrations and applications page into one page. 

Screenshot 2024-07-23 at 3 30 51 PM

Major changes:

- The app enabled?, active? checks are all moved to backend.
- The dashboard_apps integration is also now part of the apps.yml file.
- Updated the header design for the new settings pages.
- Merged the folders integrationapps and integrations. 
- Updated the copy to match the size of the card and provide clear instruction. 
- Only the list page is updated in this PR, rest of the pages are yet to be migrated.

| Integration | Verified | 
| -- | -- |
| Dashboard Apps | ✅ | 
| Dyte | ✅ | 
| Slack | ✅ | 
| Webhooks | ✅ | 
| Dialogflow | ✅ | 
| Google Translate | ✅ | 
| OpenAI | ✅ | 
| Linear | ✅ |


app/javascript/dashboard/assets/scss/_woot.scss
```diff
@@ -3,6 +3,9 @@
 @import 'tailwindcss/utilities';
 
 @import 'shared/assets/fonts/plus-jakarta';
+@import 'shared/assets/fonts/InterDisplay/inter-display';
```
pranavrajs, line 6: Using Inter display for headings and Inter for the rest of the content.

app/javascript/dashboard/components/CustomBrandPolicyWrapper.vue
```diff
@@ -0,0 +1,28 @@
+<script setup>
```
pranavrajs, line 1: Added this component so that next if there is a custom brand check, we can use the component directly.

app/javascript/dashboard/components/CustomBrandPolicyWrapper.vue
```diff
@@ -0,0 +1,28 @@
...
+import { useStoreGetters } from 'dashboard/composables/store';
+import { computed } from 'vue';
+const props = defineProps({
+  showOnCustomBrandedInstance: {
```
muhsin-k, line 5: There's a mismatch between the prop name in the defineProps and where it's used in the computed property. The prop is defined as `showOnCustomBrandedInstance`, but it's referred to as `displayOnCustomBrandedInstance` in the computed property.

app/javascript/dashboard/components/CustomBrandPolicyWrapper.vue
```diff
@@ -0,0 +1,28 @@
...
+
+const isACustomBrandedInstance =
+  getters['globalConfig/isACustomBrandedInstance'];
+
+const shouldShowContent = computed(() => {
+  if (props.displayOnCustomBrandedInstance) {
+    return true;
+  }
+  return !isACustomBrandedInstance.value;
+});
```
muhsin-k, line 21:
```suggestion
const { globalConfig: { isACustomBrandedInstance } } = useStoreGetters();

const shouldShowContent = computed(() => 
  props.showOnCustomBrandedInstance || !isACustomBrandedInstance.value
);
```

app/javascript/dashboard/routes/dashboard/settings/integrations/IntegrationItem.vue
```diff
@@ -0,0 +1,90 @@
...
+import { useI18n } from 'dashboard/composables/useI18n';
+import { frontendURL } from 'dashboard/helper/URLHelper';
+import { useInstallationName } from 'shared/mixins/globalConfigMixin';
+
+const props = defineProps({
+  id: {
+    type: [String, Number],
+    required: true,
+  },
+  logo: {
```
muhsin-k, line 13: Unused prop.

muhsin-k (commented):
@pranavrajs Please check the inline comments.

app/javascript/dashboard/components/CustomBrandPolicyWrapper.vue
pranavrajs: I've enabled the eslint rules for the same.

pranavrajs: Good catch!

app/javascript/dashboard/components/CustomBrandPolicyWrapper.vue
pranavrajs: This suggestion didn't work. The output useGetters is of the format `<module>/<getter>`. You can't use destructuring assignment here.

app/javascript/dashboard/routes/dashboard/settings/integrations/IntegrationItem.vue
pranavrajs: Added the eslint-rule for the same.

sojan-official (approved):

PR #9862, opened 2024-07-31
fix: templates in whatsapp

scmmishra:


fayazara (approved):
Looking good

app/javascript/shared/mixins/specs/whatsappTemplates/whatsappTemplates.spec.js
```diff
@@ -18,45 +19,45 @@ const config = {
...
     const wrapper = shallowMount(TemplateParser, {
       ...config,
       propsData: { template: templates[12] },
     });
+    await nextTick();
     expect(wrapper.vm.variables).toBeNull();
   });
 
-  it('returns the body of a template', () => {
+  it.skip('returns the body of a template', async () => {
```
scmmishra, line 40: @fayazara had to skip this test

The issue is highlighted here: https://github.com/vuejs/test-utils/issues/2196
Solution was made but only works for vue 3: https://github.com/vuejs/test-utils/issues/2196

CC: @pranavrajs

scmmishra: Fixed, just had to return it in the `setup`

PR #9863, opened 2024-07-31
fix: Instagram outgoing attachments

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-3497/instagram-outgoing-attachments-are-not-rendering

It seems like Instagram outgoing attachment rendering has been broken for a long time. We couldn't identify the issue because the check below only inspects Instagram mentions.
```
metadata[:data_url] = metadata[:thumb_url] = external_url if message.instagram_story_mention?
```
We recently worked on adding the [Instagram CDN URL for attachments.](https://github.com/chatwoot/chatwoot/pull/9287) 

After that, it started using external URLs as attachment data URLs for both outgoing and incoming attachments. The rendering broken for all the outgoing attachments since there were no external URLs, making the data URL empty. Adding an incoming message check will solve the issue.

### DIFF
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -81,7 +81,7 @@ class Attachment < ApplicationRecord
       height: file.metadata[:height]
     }
 
-    metadata[:data_url] = metadata[:thumb_url] = external_url if message.inbox.instagram?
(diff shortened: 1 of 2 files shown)


sojan-official (approved):
LGTM

PR #9884, opened 2024-08-05
fix: TypeError: Cannot read properties of null (reading 'user_id')

iamsivin:
# Pull Request Template

## Description

This PR will fix an issue with the audit log page.

**Issue**
This PR addresses an issue where the page crashes and throws the error `TypeError: Cannot read properties of null (reading 'user_id')` when there is an update event in the audit log, and the updated agent is being deleted. This error occurs because the `auditable` object is `null`, leading to an attempt to access the `user_id` property of a null object

**Solution**
The fix ensures that the code properly handles cases where the `auditable` object is null before attempting to access the `user_id` property. Specifically, before setting the `translationPayload`, a check is added to verify if the auditable object is present. If auditable is present, it shows the `agent name`, otherwise, it returns a string `deleted user`. For eg:, "Shivam changed role, availability of deleted user to administrator, offline."

Fixes https://linear.app/chatwoot/issue/CW-3501/typeerror-cannot-read-properties-of-null-reading-user-id

## Type of change

## How Has This Been Tested?

**Steps:**

1. Enable audit logs for the account.
2. Then edit any other agent availability/role from agent's page, then save it.
3. Then go to the audit log page and see it's there. For eg:, "Shivam changed role,availability of Sivin to administrator,offline"
4. Go back to agent's page and delete the updated agent.
5. And go back to audit log page and see.

## Checklist:


scmmishra (dismissed):

app/javascript/dashboard/helper/auditlogHelper.js
```diff
@@ -89,11 +89,10 @@ function handleAccountUserCreate(auditLogItem, translationPayload, agentList) {
...
 function handleAccountUserUpdate(auditLogItem, translationPayload, agentList) {
-  if (auditLogItem.user_id !== auditLogItem.auditable.user_id) {
-    translationPayload.user = getAgentName(
-      auditLogItem.auditable.user_id,
-      agentList
-    );
+  if (auditLogItem.user_id !== auditLogItem.auditable?.user_id) {
+    translationPayload.user = auditLogItem.auditable
+      ? getAgentName(auditLogItem.auditable?.user_id, agentList)
+      : 'deleted user';
```
vishnu-narayanan, line 95: Can we use i18n here?
```suggestion
      :  this.$t(`AUDIT_LOGS.DELETED_USER`);
      
```

vishnu-narayanan (commented):
Looks good. 

1. Can we use i18n to avoid hardcoding `deleted user`?
2. Could you add a test for this case in auditlog_helper spec?

app/javascript/dashboard/helper/auditlogHelper.js
iamsivin: @vishnu-narayanan We can't do i18n in a helper file.

iamsivin: Done https://github.com/chatwoot/chatwoot/pull/9884/commits/711b0ce87a04127174a74e9306780bfe477fde59

vishnu-narayanan (approved):

PR #9910, opened 2024-08-07
fix: Disable the password section is the admin has disabled user profile update

pranavrajs:
This PR updates the profile settings page to completely disable the password section, including the heading, if the admin has disabled user profile updates. Previously, the section heading was shown with empty content, which caused confusion.

| Before | After |
| -- | -- |
| Screenshot 2024-08-07 at 9 36 24 AM | Screenshot 2024-08-07 at 9 36 35 AM |


iamsivin (approved):
LGTM

PR #9913, opened 2024-08-08
fix: `last_activity_at` is not present for conversation created event 

scmmishra:
The `conversation.created` payload does not include `last_activity_at` when the conversation is created. Because of this the frontend is not able to sort the conversations when appending this. 

Another problem is that the `last_activity_at` is not always present, it is added only when a message is created, and it updates it. So this can be nil when the conversation is created, so we fallback to `created_at` only at the presentation layer

Here's a demo 

https://www.loom.com/share/e84d9df429fa4dbf8d8ba12a1b1a9cfa?sid=fddc3873-28c3-4873-a2ac-1e9c40452def


app/models/conversation.rb
```diff
@@ -124,6 +124,10 @@ def can_reply?
     last_message_in_messaging_window?(messaging_window)
   end
 
+  def last_activity_at
```
sojan-official, line 127: let's also add a spec to cover this case !!!

scmmishra: Done! 8ee2419ea57ab62c936eb62a090f4fa392d1af1f

pranavrajs:
Closing as the fix is already merged.

PR #9919, opened 2024-08-09
feat: Replace `attributeMixin` within the component

iamsivin:
# Pull Request Template

## Description

This PR will replace the usage of `attributeMixin` within the component itself. And moved the component from option API to composition API and updated the styles and related component

Fixes https://linear.app/chatwoot/issue/CW-3444/rewrite-attributemixin-mixin-to-a-composable

## Type of change

## How Has This Been Tested?

**Test cases**
1. Add custom attributes for both conversation and contact from the settings
2. See all attributes are showing based on the conversation and contact in both conversation and contact sidebar.
3. Try all CRUD operations like EDIT, DELETE. 
4. Check whether styles are properly showing or not (Background color based on odd/even)

## Checklist:


app/javascript/dashboard/routes/dashboard/conversation/ContactPanel.vue
```diff
@@ -218,8 +218,6 @@ export default {
             >
               <CustomAttributes
                 attribute-type="contact_attribute"
-                attribute-class="conversation--attribute"
```
scmmishra, line 221: Why did we remove these?

app/javascript/dashboard/routes/dashboard/conversation/ConversationInfo.vue
```diff
@@ -74,37 +74,32 @@ const staticElements = computed(() =>
 
 <template>
   <div class="conversation--details">
-    <ContactDetailsItem
```
scmmishra, line 77: Why the need to move this down?

iamsivin: @scmmishra I added the conversation information component as a slot inside the custom attributes component because when I moved the [styles](https://github.com/chatwoot/chatwoot/pull/9919/files#diff-2320f582890a2f838308e3f1b6823e5307272ecfe9c25c2132fe99a88aaacc71L191) to inline, I noticed an issue with the odd/even background color in the conversation info section, as shown in the screenshot. This issue is specific to the Conversation Information accordion, which includes both the static conversation info element and the custom attributes section, visible in the screenshot.

**Screenshot**
image

app/javascript/dashboard/routes/dashboard/conversation/ContactPanel.vue
iamsivin: @scmmishra When I moved the styles to inline in both components, the prop became unnecessary, which is why it was removed. Did you face any issues during testing?

app/javascript/dashboard/routes/dashboard/conversation/ConversationInfo.vue
scmmishra: There should be a better way to do this, perhaps we can discuss on call. Selecting the background color of each item should ideally be done by the parent in this case with a class like this

`[&>*:nth-child(odd)]:bg-gray-200 [&>*:nth-child(even)]:bg-white`

iamsivin: @scmmishra Done

scmmishra (approved):

PR #9924, opened 2024-08-09
feat: Replace `rtlMixin` to a composable

iamsivin:
# Pull Request Template

## Description

This PR will replace the usage of `rtlMixin` to the `useUISettings` composable, and moved the method to component itself.

**Files Updated**
1. dashboard/App.vue
2. dashboard/components/widgets/conversation/ReplyBox.vue
3. dashboard/routes/dashboard/settings/reports/components/CsatTable.vue
4. dashboard/routes/dashboard/settings/reports/components/overview/AgentTable.vue
5. dashboard/routes/dashboard/contacts/components/ContactsTable.vue
6. dashboard/routes/dashboard/notifications/components/NotificationPanel.vue

Fixes https://linear.app/chatwoot/issue/CW-3471/rewrite-rtlmixin-mixin-to-a-composable

## Type of change

## How Has This Been Tested?

**Steps**
1. Navigate to account settings and set any RTL language.
2. See the UI has been updated or not.
3. And check the places where the files updated.

## Checklist:


app/javascript/dashboard/composables/useUISettings.js
```diff
@@ -114,6 +114,15 @@ const isEditorHotKeyEnabled = (key, uiSettings) => {
   return editorMessageKey === key;
 };
 
+/**
+ * Checks if RTL view is enabled in the UI settings.
+ * @param {Object} uiSettings - Reactive UI settings object.
+ * @returns {boolean} True if RTL view is enabled, otherwise false.
+ */
+const isRTLEnabled = uiSettings => {
+  return computed(() => uiSettings.value.rtl_view);
```
pranavrajs, line 123: @iamsivin Do we need this in the UI settings? Couldn't we just use the helper function directly in the components?

pranavrajs (changes requested):
We can probably remove the composable as well. Please take a look at the comment above.

app/javascript/dashboard/composables/useUISettings.js
scmmishra: @pranavrajs we don't need this function, it can be declared directly. I've made a commit to do this

app/javascript/dashboard/routes/dashboard/notifications/components/NotificationPanel.vue
```diff
@@ -201,7 +196,7 @@ export default {
             <fluent-icon
               icon="chevron-left"
               size="16"
-              :class="notificationPanelFooterIconClass"
+              class="rtl:-mr-3 ltr:-ml-3"
```
scmmishra, line 199: @iamsivin if these `ltr` and `rtl` directives are working fine, let's use it across the app

iamsivin: @scmmishra We are using this `ltr` and `rtl` directives almost everywhere. It's done at the time while moved from CSS to Tailwind, Only some cases like [this](https://github.com/chatwoot/chatwoot/pull/9924/files#diff-caa95f34e5ac6da66759094fd4823106f9ee28b67477693a23a85973b7cd54c8R94) we are using it. Do you have any idea?

app/javascript/dashboard/composables/useUISettings.js
pranavrajs: @scmmishra I don't think the change aligns with what I had in mind. Maybe I wasn't clear in the first comment. Let me try again.

I don't think this UI setting is required at all. We don't need to save it in the DB. It can be resolved dynamically from the selected language.

Instead of the following code.

```js
const { isRTL, updateUISettings } = useUISettings();
```

You could write the following. 

```js
const isRTL = getLanguageDirection(language);
```

iamsivin: @pranavrajs Could please you take a look? I have updated the PR https://github.com/chatwoot/chatwoot/pull/9924/commits/f959f5218f059eb8528c5ebb3af9c0b24865f39d

app/javascript/dashboard/routes/dashboard/notifications/components/NotificationPanel.vue
scmmishra: For now it seems fine, it's out of scope for this PR atleast. We can take a look at it later

pranavrajs (approved):

PR #9959, opened 2024-08-14
refactor: `useKeyboardEvents` composable 

scmmishra:

1. Fix tab styles issue caused by adding an additional wrapper for getting an element ref on `ChatTypeTabs.vue`
2. Refactor `useKeyboardEvents` composable to not require an element ref. It will use a local abort controller to abort any listener


iamsivin (approved):
LGTM

PR #9989, opened 2024-08-20
feat: enable disposable email check for account creation

scmmishra:
This PR disallows usage of disposable emails when creating an account

### DIFF
--- a/app/builders/account_builder.rb
+++ b/app/builders/account_builder.rb
@@ -36,7 +36,7 @@ class AccountBuilder
     if address.valid? # && !address.disposable?
       true
     else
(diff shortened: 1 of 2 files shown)


vishnu-narayanan (approved):
LGTM

app/builders/account_builder.rb
```diff
@@ -33,10 +33,10 @@ def account_name
 
   def validate_email
     address = ValidEmail2::Address.new(@email)
-    if address.valid? # && !address.disposable?
+    if address.valid? && !address.disposable?
```
pranavrajs, line 36: I was checking why this was disabled. It is a 5 year comment. 😁

PR #9990, opened 2024-08-20
fix: Remove user.permissions, resolve it from accounts

pranavrajs:
Remove the `user.permissions` field and resolve the permissions directly from the accounts array in the user. This change ensures that the cache or previous values from the last active account don't affect the permissions.

In this PR: 
- Remove user.permissions usage, replace it with getUserPermissions method.
- Remove json.permissions from user.json.jbuilder


sojan-official (approved):
LGTM

scmmishra (approved):

PR #9994, opened 2024-08-21
chore: Repalce Hook Mixin with useHook composable [CW-3454]

fayazara:
# Pull Request Template

## Description

Replace Hook mixin with useHook composable

Fixes # (issue)

## Type of change

## How Has This Been Tested?

## Checklist:


app/javascript/dashboard/routes/dashboard/settings/integrations/useHook.js
```diff
@@ -0,0 +1,26 @@
+import { computed, ref } from 'vue';
+import { useStore } from 'dashboard/composables/store';
+
+export const useHook = (integrationObj, integrationId) => {
```
scmmishra, line 4: Is there a way we don't need two arguments here, perhaps we can only accept `integrationId` that way we will always rely on the store. It'll make the code simpler too

```
const integration = useMapGetter('integrations/getIntegration')(integrationId)
```

This will return a computed and then the other computed values can use this as a base. It'll simplify the props to the parent components as well, we can simply pass the integration id and not the entire object
```

scmmishra, line 4: Let's also rename this to `useIntegrationHook`, `useHook` can be a bit confusion

scmmishra (changes requested):
@fayazara some comments to address, let me know if you need more info/clarification

scmmishra (approved):

PR #9865, opened 2024-07-31
chore: Custom Roles to manage permissions [ UI ]

sojan-official:
In admin settings, this Pr will add the UI for managing custom roles  ( ref: https://github.com/chatwoot/chatwoot/pull/9995 ).  It also handles the routing logic changes to accommodate fine-tuned permissions.

PR #9995, opened 2024-08-21
chore: Custom Role APIs 

sojan-official:
- APIs to manage custom roles in Chatwoot

ref: https://github.com/chatwoot/chatwoot/pull/9865


enterprise/app/models/custom_role.rb
```diff
@@ -0,0 +1,42 @@
...
+# Available permissions for custom roles:
+# - 'conversation_manage': Can manage all conversations.
+# - 'conversation_unassigned_manage': Can manage unassigned conversations and assign to self.
+# - 'conversation_participating_manage': Can manage conversations they are participating in (assigned to or a participant).
+# - 'contact_manage': Can manage contacts.
+# - 'report_manage': Can manage reports.
+# - 'knowledge_base_manage': Can manage knowledge base portals.
+
+class CustomRole < ApplicationRecord
+  belongs_to :account, dependent: :destroy_async
```
pranavrajs, line 28: Doesn't this destroy the account when custom role is deleted?

pranavrajs (changes requested):
Please see the comment.

enterprise/app/models/custom_role.rb
sojan-official: updated

PR #10016, opened 2024-08-23
feat: add domain blocklist feature

scmmishra:


config/locales/en.yml
```diff
@@ -43,6 +43,7 @@ en:
       invalid: Invalid events
     signup:
       disposable_email: We do not allow disposable emails
+      blocked_domain: This domain is not allowed, if you think this is a mistake please contact support
```
pranavrajs, line 46:
```suggestion
      blocked_domain: This domain is not allowed. If you believe this is a mistake, please contact support.
```

app/builders/account_builder.rb
```diff
@@ -79,4 +82,20 @@ def create_user
     @user.confirm if @confirmed
     @user.save!
   end
+
+  def domain_blocked?(email)
+    domain = email.split('@').last
+    # loop over blocked_domains and do a regex match
```
pranavrajs, line 88: Do we need this comment?

pranavrajs (approved):
Minor comments, otherwise looks fine.

app/builders/account_builder.rb
scmmishra: Oops, added it for copilot :P 
removing it

vishnu-narayanan (approved):
Can we also add a spec?

scmmishra:
> Can we also add a spec?

On it

PR #10024, opened 2024-08-25
feat: Load reply-to messages dynamically when not present in message list

clairton:
# load reply to message

## Description

When replayed message is more old, not show content

## Type of change

## How Has This Been Tested?

I run in my development and production envinronment with unoapi


app/javascript/dashboard/components/widgets/conversation/Message.vue
```diff
@@ -70,8 +70,8 @@ export default {
       default: () => ({}),
     },
     inReplyTo: {
-      type: Object,
-      default: () => ({}),
+      type: Promise,
+      default: Promise.resolve({}),
     },
```
scmmishra, line 75: Passing a Promise as a prop is not a great idea, ideally the parent component should resolve the promise and pass it, or the fetching should happen in the component itself.

muhsin-k:
@iamsivin Can you please review and test these changes?

muhsin-k (approved):

PR #10032, opened 2024-08-26
fix: Inaccurate Duplicate Notification Filtering Due to Oversights in Actor ID Logic

### LINKED ISSUE #10031: Inaccurate Duplicate Notification Filtering Due to Oversights in Actor ID Logic
### Describe the bug

In the recent implementation outlined in PR https://github.com/chatwoot/chatwoot/pull/9520, a feature was introduced to filter out duplicate notifications. While the intention behind this filtration is appreciated, the current logic has limitations that lead to incorrect determination of duplicate notifications.

The filtration logic is currently based solely on the `primary actor ID`. This approach has a crucial flaw: the uniqueness of the primary actor ID is only guaranteed when the `primary actor type` is also consistent. As a result, notifications involving different primary actor types may be incorrectly treated as duplicates.

Moreover, the filtration logic does not take the `secondary actor` into account when evaluating potential duplicates. This omission can lead to valid notifications being erroneously hidden, resulting in users missing important updates.

Expected Behavior:  
To improve the accuracy of the duplicate notification filtering, the logic should consider both the `primary` and `secondary` actor IDs and their respective types. This would ensure that notifications are only deemed duplicates when both the primary actor type matches and the secondary actor is included in the comparison.

### To Reproduce

NA

### Expected behavior

_No response_

### Environment

Kubernetes

### Cloud Provider

AWS

### Platform

Browser

### Operating system

_No response_

### Browser and version

_No response_

### Docker (if applicable)

_No response_

### Additional context

_No response_

mazenkhalil:
Fixes: #10031


iamsivin:
@codex review?

PR #10044, opened 2024-08-29
feat: Added support for copying and pasting images in article editor

muhsin-k:
- Added the ability to copy images from the system and paste them into the editor.
- Removed the success message as it's unnecessary; users will know when the image appears in the editor.

### DIFF
--- a/app/javascript/dashboard/components/widgets/WootWriter/FullEditor.vue
+++ b/app/javascript/dashboard/components/widgets/WootWriter/FullEditor.vue
@@ -173,6 +173,13 @@ export default {
           blur: () => {
             this.onBlur();
           },
+          paste: (view, event) => {
+            const data = event.clipboardData.files;
+            if (data.length > 0) {
+              data.forEach(file => this.uploadImageToStorage(file));
+              event.preventDefault();
+            }
+          },
         },
       });
     },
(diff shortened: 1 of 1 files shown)


fayazara (approved):
Tested locally. Looks good

app/javascript/dashboard/components/widgets/WootWriter/FullEditor.vue
```diff
@@ -173,6 +172,13 @@ export default {
           blur: () => {
             this.onBlur();
           },
+          paste: (view, event) => {
+            const data = event.clipboardData.files;
+            if (data.length > 0) {
+              data.forEach(file => this.uploadImageToStorage(file));
```
pranavrajs, line 178: Should we check if the pasted file is an image? Otherwise looks fine to me.

pranavrajs (commented):
One minor comment, please take a look.

app/javascript/dashboard/components/widgets/WootWriter/FullEditor.vue
muhsin-k: Good suggestion. I have updated the code.

pranavrajs (approved):

PR #10046, opened 2024-08-29
fix: default return in blocked domains

scmmishra:


app/builders/account_builder.rb
```diff
@@ -94,8 +94,8 @@ def domain_blocked?
 
   def blocked_domains
     domains = GlobalConfigService.load('BLOCKED_EMAIL_DOMAINS', '')
-    domains.split("\n").map(&:strip) if domains.present?
+    return [] if domains.blank?
 
-    []
+    domains.split("\n").map(&:strip) if domains.present?
```
vishnu-narayanan, line 99:
```suggestion
    domains.split("\n").map(&:strip)
```

Can avoid `.present` check as we did check for `.blank?` above

vishnu-narayanan (approved):

muhsin-k (approved):

PR #10074, opened 2024-09-04
chore: Fix flaky permissions test

sojan-official:
- the permissions tests were failing when the random number turns out to be 0, fixing this case and ensuring that its always between, 1-4.


sojan-official:
- self reviewed

PR #10111, opened 2024-09-13
chore: Update GPT Model

sojan-official:
- update the open AI model, as 3.5 is being deprecated. Provide as option to swap out models via environment variables.


freesr:
@sojan-official  Thanks for quickly updating the model after request. Also, you might consider increasing the token size since GPT-4 supports up to 128k tokens.

scmmishra (approved):

Issue #6228, opened 2023-01-10
[CW-1945] Add a default for the country code in the account


pranavrajs:
> In the pre-form number field, it is required to add +(country phone indicator) is there a way to disable this, my business is 100% local, and adding the +xx is not relevant.

<sub>[CW-1945](https://linear.app/chatwoot/issue/CW-1945/add-a-default-for-the-country-code-in-the-account)</sub>

Aghiles97:
Thank you for creating this, i hope we can see a work around soon.

nithindavid:
This is a good suggestion. 

@tejaswinichile @sojan-official Which model should be used to add a field to save the country code?

pranavrajs:
@Aghiles97 Moving this to our backlog, so that this is not missed.

PR #10146, opened 2024-09-21
feat: Auto-populate the telephone code based on the browser timezone

pranavrajs:
Fixes https://github.com/chatwoot/chatwoot/issues/6228

There is a country code selector for the phone input field. This is often a point of frustration. See the response below. 

> We are using the phone number field however this can be frustrating for customers, especially mobile users, to select the +1 US country code. Our users are typically local businesses and being able to default to +1 country code in the phone number field or the account would improve the interaction they have with customers.

Most people who run local businesses don't need a country selector. However, to preserve the quality of the data we store, we need a country code. 

A balance between these two issues can be found with an auto-populating country code field based on the browser's timezone. This is what I did in this PR. Based on the browser timezone, we will resolve it to the closest country code.


muhsin-k (approved):
LGTM

Issue #9935, opened 2024-08-12
Not possible to delete articles, with no category


brundles:
### Describe the bug

I've created an article for the Helpdesk via API. 

But then I've deleted the related category, and now I cannot access the article via the backend, but still see the article in the frontend under an "uncategorized" category.

So no way for me to delete this article :-(

### To Reproduce

1. Add a new article via API (probably it can also be done directly in browser). Make sure its published.
2. Delete the related category.
3. Find the article on the frontend, and see that you cannot see it via the backend.

### Environment

Linux VM

### Cloud Provider

DigitalOcean

### Platform

Browser

pranavrajs:
@brundles The delete APIs are not documented properly. We will address this issue. We don't allow articles without a category from the UI. You could do a DELETE /articles/<id> or PATCH with status as draft which should fix this.

Issue #8213, opened 2023-10-26
Cannot Delete Uncategorized Articles


punyaikhwan:
### Describe the bug

I am unable to delete articles in the "deleted" category. The articles in this category are reclassified as "Uncategorized" on the Help Center site, but they do not appear in the admin view. As a result, there is no way to edit or delete these articles.

### To Reproduce

I have several articles categorized under "X." These articles have already been published and are visible on the Help Center site.

For some reason, I decided to delete the "X" category. However, these articles still appear on the Help Center site, now categorized as "Uncategorized."
![image](https://github.com/chatwoot/chatwoot/assets/14848437/26387495-19d6-4c23-aa84-1e07756415d8)

Strangely, in the admin site, there is no category labeled "Uncategorized."
![image](https://github.com/chatwoot/chatwoot/assets/14848437/e0facc50-ef4d-4050-acb8-fbb24321a151)

Even in the "All articles" menu, there is no listing of articles categorized as "Uncategorized" or in the deleted category.

Due to these circumstances, I'm unable to make any changes to these articles.

### Expected behavior

In the admin view, it should display either the "Uncategorized" or "Deleted" category. This would allow me to access all articles within these categories and make necessary changes to them.

### Environment

Docker

### Cloud Provider

DigitalOcean

### Platform

Browser

### Operating system

Windows

### Browser and version

Edge 118.0.2088.61 (Official build) (64-bit)

### Additional context

I use Chatwoot v3.1.1

<sub>[CW-2696](https://linear.app/chatwoot/issue/CW-2696/cannot-delete-uncategorized-articles)</sub>

pranavrajs:
@punyaikhwan Thanks for reporting, let us investigate this issue.

FaxriddinMaxmadiyorov:
@pranavrajs @punyaikhwan hello, I would like to work on this issue

zajcomm:
was this resolved? I am also facing same issue

pranavrajs:
@punyaikhwan Moving this out of un-assignment as there was no activity for a long time.

pranavrajs:
The proposed solution looks like the following:

* Add a 'locale' parameter to the article model. This parameter should never be empty. It is updated when the category is updated. If the category is removed, the 'locale' parameter should not be removed from the article model. If an article is created without a category, use the default locale for the article.
* Add a filter into the API that returns uncategorized articles in a particular locale.
* Include an uncategorized section in the dashboard that displays articles without a category.
[...]

PR #10153, opened 2024-09-23
fix: Include uncategorized articles in the all article section to allow edit/delete

pranavrajs:
Fixes https://github.com/chatwoot/chatwoot/issues/9935
Fixes https://github.com/chatwoot/chatwoot/issues/8213

The articles were grouped by category, with locale being a derived attribute from the category. If a category was deleted, the article wouldn't appear on the dashboard. However, due to a bug, it would show up in the uncategorized section on the public portal, leaving agents unable to edit or update the article.

To address this issue, I've added a locale attribute directly to the article. This attribute is automatically set from the category or the portal's default locale if not supplied. The API parameters now use this attribute to filter articles. As a result, the dashboard will display articles even if they're not associated with a category, improving the overall workflow.

**Main updates:** 
- Add locale attribute to the Article model. Add db migration to back fill the data based on the above logic.
- Add a new scope search_by_locale and use it instead of search_by_category_locale.
- Update the ERB template to include the locale filter.
- Move from `joins` to `left_outer_joins` to include the articles with no categories.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -819,7 +819,7 @@ GEM
       rack-proxy (>= 0.6.1)
       railties (>= 5.2)
       semantic_range (>= 2.3.0)
-    webrick (1.8.1)
+    webrick (1.8.2)
     websocket-driver (0.7.6)
       websocket-extensions (>= 0.1.0)
     websocket-extensions (0.1.5)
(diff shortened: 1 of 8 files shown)


sojan-official (approved):
LGTM

PR #10179, opened 2024-09-28
feat: let users re auth legacy google inboxes

scmmishra:
This PR allows migration of legacy GMail inbox users to new OAuth based inbox

## How to test?

1. Create an inbox from the seed data and set it's IMAP address to `imap.gmail.com` from the UI
2. Open `rails console` and run the following

   ```
   inbox = Inbox.find(100) # use your inbox id here
   channel = inbox.channel
   channel.update(email: 'hello@chatwoot.com')
   channel.prompt_reauthorization!
   ```
3. This will show the prompt on the UI. Once you click on Reauthorize, it should open Google Auth. Reauthroize with the same email address as used in the inbox and it should start working as usual

### Setting up ENV

```sh
GOOGLE_OAUTH_CLIENT_ID=<some-hash>.apps.googleusercontent.com
GOOGLE_OAUTH_CLIENT_SECRET=<client-secret>
GOOGLE_OAUTH_CALLBACK_URL="http://localhost:3000/omniauth/google_oauth2/callback"
```

### DIFF
--- a/app/javascript/dashboard/routes/dashboard/settings/inbox/Settings.vue
+++ b/app/javascript/dashboard/routes/dashboard/settings/inbox/Settings.vue
@@ -7,12 +7,13 @@ import SettingIntroBanner from 'dashboard/components/widgets/SettingIntroBanner.
 import SettingsSection from '../../../../components/SettingsSection.vue';
 import inboxMixin from 'shared/mixins/inboxMixin';
 import FacebookReauthorize from './facebook/Reauthorize.vue';
+import MicrosoftReauthorize from './channels/microsoft/Reauthorize.vue';
+import GoogleReauthorize from './channels/google/Reauthorize.vue';
 import PreChatFormSettings from './PreChatForm/Settings.vue';
 import WeeklyAvailability from './components/WeeklyAvailability.vue';
 import GreetingsEditor from 'shared/components/GreetingsEditor.vue';
 import ConfigurationPage from './settingsPage/ConfigurationPage.vue';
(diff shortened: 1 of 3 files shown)


app/javascript/dashboard/routes/dashboard/settings/inbox/Settings.vue
```diff
@@ -203,6 +205,16 @@ export default {
     facebookUnauthorized() {
       return this.isAFacebookInbox && this.inbox.reauthorization_required;
     },
+    googleUnauthorized() {
+      const isLegacyInbox = ['imap.gmail.com', 'imap.google.com'].includes(
```
pranavrajs, line 209: Based on the code, I believe we're going to mark the email inboxes with a "reauthorization_required" flag. Is that correct? Will this apply to all Gmail accounts, or is there a way to identify only the older accounts that need reauthorization?

pranavrajs:
@scmmishra Would you be able to share a Loom video of the flow?

app/javascript/dashboard/routes/dashboard/settings/inbox/Settings.vue
scmmishra: >  I believe we're going to mark the email inboxes with a "reauthorization_required" flag
Yes

> or is there a way to identify only the older accounts that need reauthorization?
It will only apply to legacy ones given we mark them specifically. We don't really need to identify the older accounts ourselves.

scmmishra:
> @scmmishra Would you be able to share a Loom video of the flow?

Sure

pranavrajs (approved):

PR #10189, opened 2024-10-01
feat: Show re-auth flag only for legacy gmail and oauth channels

scmmishra:


muhsin-k (approved):

app/views/api/v1/models/_inbox.json.jbuilder
```diff
@@ -72,8 +72,10 @@ if resource.email?
     json.imap_port resource.channel.try(:imap_port)
     json.imap_enabled resource.channel.try(:imap_enabled)
     json.imap_enable_ssl resource.channel.try(:imap_enable_ssl)
-    # show this even for regular imap channels, to allow transitioning to OAuth
-    json.reauthorization_required resource.channel.try(:provider_config).empty? || resource.channel.try(:reauthorization_required?)
+
+    if resource.channel.try(:microsoft?) || resource.channel.try(:google?) || resource.channel.try(:legacy_google?)
```
sojan-official, line 76: we need to show the reauthorization required case for facebook channel if the the flag turns to true.  We should limit the check to the case where inbox is email and  provider config to be empty  if its either google or Microsoft. lets sync up on this.

scmmishra: > we need to show the reauthorization required case for facebook channel 

That check is already present

https://github.com/chatwoot/chatwoot/blob/e0bf2bd9d420067e77dc4d9fee04888921692c8f/app/views/api/v1/models/_inbox.json.jbuilder#L54

But overall, we could simplify the jbuilder

PR #10197, opened 2024-10-02
fix: inconsistent behaviour of create and edit automation rules 

scmmishra:
When Vue 3 is used with options API, any assignment to `this.<something>` is converted to a Proxy before assignment. This is fine as long as we are in the options context, problem arises when we access this in a `composable` any mutations on the object doesn't behave correctly as expected, this PR fixes that by moving the `automation` object inside the composable and using it in the options.

> Another option to fix such an issue is to make the object non-reactive, like done in places where we have `editorView`, but that wasn't viable here


scmmishra:
@pranavrajs this should fix the issues showing up in the UI and console, PTAL

muhsin-k (approved):

PR #10198, opened 2024-10-02
chore: update interpolation syntax for i18n files

scmmishra:
There were two warnings showing up on new teams page

1. `errorMessage` prop was getting invalid value, this was because were short circuting the error message using `&&`, fixed it by using ternary operator
2. `vue-i18n` has deprecated [rails syntax](https://vue-i18n.intlify.dev/guide/essentials/syntax#rails-i18n-format), fixed that warning by removing `%` before `{}` for interpolation

> Note: the `vue-i18n` deprecation needs to be handled, but we can do it later


pranavrajs:
@scmmishra Let us replace all the references for `%{` to `{`. The following search should do it.

Screenshot 2024-10-02 at 9 13 55 AM

PR #10208, opened 2024-10-03
fix: submit event for webhook form

scmmishra:
Submit event name was conflicting with the browsers native submit event, this PR fixes it


scmmishra:
Closing this, its fixed in [fix/cw-3590](https://github.com/chatwoot/chatwoot/pull/10210)

PR #10217, opened 2024-10-03
fix: translation payload for audit logs

scmmishra:
Vue i18n would automatically merge arrays previously, it does not do so now. This PR fixes it by cleaning up the payload before passing it for translation

CleanShot 2024-10-03 at 20 52 54@2x


scmmishra:
> @scmmishra LGTM. I think need to update the spec also.

The change is in the component not the helper, we have no tests for the component yet 🥲

iamsivin:
> > @scmmishra LGTM. I think need to update the spec also.
> 
> @iamsivin The change is in the component not the helper, we have no tests for the component yet 🥲

Oh, sorry. Approved!

PR #10236, opened 2024-10-05
feat: use article props

scmmishra:
Previously the route was declared like the following

https://github.com/chatwoot/chatwoot/blob/e0bf2bd9d420067e77dc4d9fee04888921692c8f/app/javascript/widget/router.js#L39-L44

And this is how it was used

https://github.com/chatwoot/chatwoot/blob/e0bf2bd9d420067e77dc4d9fee04888921692c8f/app/javascript/widget/views/Home.vue#L80-L83

Both these are from the Vue 2 stable branch. This used the feature of [boolean props](https://router.vuejs.org/guide/essentials/passing-props.html#Boolean-mode), where any param was passed along as a prop to the component being rendered. 

However if you noticed, the param `link` was never declared when creating the route, the path is `/article` and not `/article/:link`. The previous version of vue router didn't mind this, but since Vue Router 4, any params not explicitly declared on the `path`will be stripped away. See [changelog](https://github.com/vuejs/router/blob/main/packages/router/CHANGELOG.md#414-2022-08-22)

That means the link being sent was stripped away. To fix this we need to add `:link` to path in router.js. This PR does exactly this.

P.S. Updated `app/javascript/widget/views/ArticleViewer.vue` to use Vue Setup syntax coz it seemed straightforward, not related to the original bug


scmmishra:
@pranavrajs PTAL whenever you can, we can merge this and get it deployed. (Not critical, the existing fix works anyway)

pranavrajs (changes requested):
@scmmishra I'm not sure about this change as the query seems like a better approach than passing the link as the URL params. I'm worried whether we should use encodeURIComponent in this case. /article/https://... can be treated a different URL, right?

scmmishra:
> @scmmishra I'm not sure about this change as the query seems like a better approach than passing the link as the URL params. I'm worried whether we should use encodeURIComponent in this case. /article/https://... can be treated a different URL, right?

Yeah, the query way is much better for sure, I'll close this PR. We can revisit this if something breaks

PR #10245, opened 2024-10-08
fix: Resolve accountId from the route, initialize route-sync before the app is loaded

scmmishra:
On production on multiple instances it may happen that the UI is rendered in correctly, with a lot of options in the sidebar not available. On further investigation I found out that the feature flag checks were disabling multiple of those, and also we could see many correlated errors that pointed towards missing information.

So, there were two problems here

1. The `vuex-router-sync` was not very reliable in some cases
2. In `App.vue` the watch on `currentAccountId` didn't always trigger.

## Fix Tested on Staging

Basically tried to reload the page ~50 times with cache enabled, disabled, throttling, navigating different pages.

https://www.loom.com/share/1bb27294aa364ac4acfb647780d6385a?sid=87e31330-8cb7-4ded-8616-5e95e2ae3516

<details><summary>

#### What I thought was the fix

</summary>
<p>

### My chain of actions

Replacing vuex-router-sync at first worked fine, but then I saw it was still failing in some cases, I assumed (I was half-correct tho) that the rendering of the `App.vue` and syncing of the route to the store was not happening in a synchronous pattern. So I decided, let's not rely on the store when the route is directly available in the App context. 

Following this, I refactored `useAccount` composable to use `useRoute` directly, instead of the store, and then replaced the getter inside `App.vue`. What this did was surface the issue but more consistently 🤯 

I saw the watcher, added some console logs, and turns out it was not getting triggered in all those cases. So I added an `immediate` to it. And viola, it works!

At the moment, this is deployed to staging and seems to be working correctly. But we still need to verify it for sure, since how this issue was surfaced is still a mystery. All we know is that it shows up when the widget is also loaded alongside the app (if it loads before or after the app, it works fine)

### What about the route in the store?

Well I have used the `route` usage there with fallback to the store state. Since Vuex exists in the app context, the route should always be available to it. But after today I have lost all trust in JavaScript and will worship rails until end of my life, so I added that in a `try-catch` block, logged the error to Sentry

</p>
</details> 

## Here's the real fix

If you read the explanation I wrote earlier, I thought I fixed the issue, but then the chat list navigation completely broke. So I removed the custom route sync implementation and added the original package back. Turns out the vuex-router-sync earlier was placed after the app was initalized, however for it to work, the vue app context is not required. And it's best to run it before the app is even bootstrapped, so I added it back and placed it correctly. 

So the following changes fixes this problem

1. Hoisting the `sync` function call to before we call `createApp` this ensures that the stores and route hooks are in place before even the app is created
2. Ensuring the `initializeAccount` is run immediately when watching `currentAccountId`
4. Source `currentAccountId` for critical top of the tree components directly from the route instead of the store


scmmishra:
Found an issue with conversation routing, fixing it. Moving PR to draft again

pranavrajs (approved):

PR #10247, opened 2024-10-09
feat: Use CJS build for SDK

scmmishra:
The UMD build was causing issues for a few customers, this PR reverts to using CJS like used in Webpack 4 before the vite migration

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -66,7 +66,7 @@ export default defineConfig({
     lib: isLibraryMode
       ? {
           entry: path.resolve(__dirname, './app/javascript/entrypoints/sdk.js'),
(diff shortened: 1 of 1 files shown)


vite.config.ts
```diff
@@ -66,7 +66,7 @@ export default defineConfig({
     lib: isLibraryMode
       ? {
           entry: path.resolve(__dirname, './app/javascript/entrypoints/sdk.js'),
-          formats: ['umd'], // UMD format for single file
+          formats: ['cjs'], // UMD format for single file
```
pranavrajs, line 69:
```suggestion
          formats: ['cjs'], // CJS format for single file
```

PR #10249, opened 2024-10-09
feat: use of imap login as default if present

scmmishra:
When moving form using Gmail Legacy auth to using OAuth, we need the email address that will be used to connect. This is because we need to store this email address in the cache and reuse when we get the callback to find the associated inbox. 

However there are cases where the imap login might be `support@company.com` and the email used to communicate will be `contact@company.com` (Probably an alias) In that case, we need to send the correct email address to Chatwoot when re-authenticating

At the moment, we used the inbox email. This PR adds a check that defaults to to `imap_login` if that is available and imap is enabled

This PR also fixes an unrelated problem where the email inbox creation flow was not working

---

Tested it, it is working correctly

CleanShot 2024-10-09 at 14 23 47@2x

### DIFF
--- a/app/controllers/google/callbacks_controller.rb
+++ b/app/controllers/google/callbacks_controller.rb
@@ -1,6 +1,15 @@
 class Google::CallbacksController < OauthCallbackController
   include GoogleConcern
 
+  def find_channel_by_email
+    # find by imap_login first, and then by email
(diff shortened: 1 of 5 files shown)


vishnu-narayanan (approved):
LGTM

PR #10253, opened 2024-10-09
fix: Search linear issues

muhsin-k:
In the `DropdownList.vue` component, the `onSearch` function was not properly passing the search value to the parent component. This resulted in the `onSearch` event in parent components (such as `LinkIssue.vue`) receiving an undefined value instead of the actual search term.

https://github.com/chatwoot/chatwoot/blob/f18ed01eb7725954fc9c0d45201ac3c53cd9855b/app/javascript/dashboard/components/ui/Dropdown/DropdownList.vue#L45-L52

The issue was resolved by modifying the `onSearch` function in `DropdownList.vue` to correctly pass the search value to the `debouncedEmit` function:

### DIFF
--- a/app/javascript/dashboard/components/ui/Dropdown/DropdownList.vue
+++ b/app/javascript/dashboard/components/ui/Dropdown/DropdownList.vue
@@ -48,7 +48,7 @@ const debouncedEmit = debounce(value => {
 
 const onSearch = value => {
   searchTerm.value = value;
-  debouncedEmit();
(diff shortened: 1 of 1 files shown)


pranavrajs:
@muhsin-k Can you please share more information about the fix?

muhsin-k:
> @muhsin-k Can you please share more information about the fix?

@pranavrajs Update the PR description.

pranavrajs (approved):

PR #10272, opened 2024-10-11
feat: Help Center layout with portal switcher component

iamsivin:
This PR includes, 
1. Help center layout component
2. Portal switcher component

**Screenshot from story**

**NB**: We need to merge all the base components. (DropdownMenu, Button, etc) PR.

**Light**
image

**Dark**
image


pranavrajs:
@iamsivin I'm merging this as this is needed for every other PR. A couple of observations.

1/ What does the dots on the right side of the menu item indicate?
2/ The second line in the menu item feels like it is not designed well. All the copy should use sentence case, and should use i18n. I don't it is respected here.

pranavrajs (approved):

iamsivin:
> @iamsivin I'm merging this as this is needed for every other PR. A couple of observations.
> 
> 1/ What does the dots on the right side of the menu item indicate? 

It used to display the portal image.

> 2/ The second line in the menu item feels like it is not designed well. All the copy should use sentence case, and should use i18n. I don't it is respected here.

I'll check it

PR #10278, opened 2024-10-11
feat: Article empty state

iamsivin:
**Screenshot from story**

**NB**: We need to merge all the base components. (DropdownMenu, Button, etc) PR.

**Light**
image

**Dark**
image


app/javascript/dashboard/components-next/HelpCenter/EmptyState/Article/ArticleEmptyState.vue
```diff
@@ -0,0 +1,97 @@
...
+          <h2
+            class="text-3xl font-medium text-center text-slate-900 dark:text-white"
+          >
+            Write an article
+          </h2>
+          <p
+            class="max-w-lg text-base text-center text-slate-600 dark:text-slate-300"
+          >
+            Write a rich article, let’s get started!
+          </p>
```
scmmishra, line 82: @iamsivin I see this usage in the portal empty state too, ideally the title and the subtitle should be exposed as both a prop and a slot. 

```vue
<EmptyStateLayout>
   <template #empty-state-item>
   </template>
   <template #empty-state>
       <template #title> Hello World </template>
       <template #subtitle> This is a lot </template>
       <template #actions> ... </template>
   </template>
</EmptyStateLayout>
```

The `title` and `subtitle` should be a prop and the actions can be added as a slot.
That way, the developer does not have to worry about styling the title or the subtitle. and the layout. In your current implementation the `EmptyStateLayout` does too little to a point of being futile. The default should be a strict with slots to allow flexibility.

scmmishra (changes requested):
Need to update implementation of EmptyStateLayout

app/javascript/dashboard/components-next/HelpCenter/EmptyState/Article/ArticleEmptyState.vue
iamsivin: @scmmishra Updated the PR's with the changes you suggested. Please take a look.

pranavrajs (approved):
Please update the demo content with the one from the portal empty state PR.

PR #10331, opened 2024-10-21
Append context menu to the app wrapper element instead to support RTL.

coolayatuvxtech:
Using the ContextMenu.vue component does not work with RTL support, since dir="rtl" attribute is being appended to the <div class="app-wrapper...> element, not in the <body>.

## Description

This PR addresses an issue with the `ContextMenu.vue` component, where RTL (Right-to-Left) support was not working as expected. Previously, the `dir="rtl"` attribute was being appended to the `<div class="app-wrapper...>` element instead of the `<body>` element. The fix involved appending the context menu to the `<div class="app-wrapper...>` element to ensure proper RTL rendering.

Fixes # (issue number here)

## Type of change

## How Has This Been Tested?

The fix was tested by:
- Appending the context menu to the `<div class="app-wrapper...>` element and verifying that RTL support works as expected.
- Ensuring the context menu behaves correctly in both RTL and LTR modes.
- Manually testing the behavior across various browsers and devices to ensure proper alignment and functionality of the context menu.

## Checklist:

### DIFF
--- a/app/javascript/dashboard/components/ui/ContextMenu.vue
+++ b/app/javascript/dashboard/components/ui/ContextMenu.vue
@@ -24,7 +24,7 @@ onMounted(() => {
 </script>
 
 <template>
-  <Teleport to="body">
+  <Teleport to=".app-wrapper">
     <div
       ref="target"
       class="fixed outline-none z-[9999] cursor-pointer"


scmmishra (changes requested):
Hey @coolayatuvxtech thanks for the contribution and apologies it took a while to get the review. 

Moving this to the `.app-wrapper` can have unintended effects on the layout. You could refer to `components-next/dialog` component, there we solve this by adding the `dir` to this component directly

https://github.com/chatwoot/chatwoot/blob/2d35fa135b17fc68d594d2b7f0e693ec51fef924/app/javascript/dashboard/components-next/dialog/Dialog.vue#L53

https://github.com/chatwoot/chatwoot/blob/2d35fa135b17fc68d594d2b7f0e693ec51fef924/app/javascript/dashboard/components-next/dialog/Dialog.vue#L77

pranavrajs:
@coolayatuvxtech The issue is already fixed in https://github.com/chatwoot/chatwoot/pull/11455. Closing this here for now. Thanks for the attempt.

Issue #10313, opened 2024-10-17
Blank page after log out - 3.14 Chatwoot


ZiekBaboon:
### Describe the bug

After upgrading from version 3.13 to version 3.14 the login screen goes blank and it is impossible to log in again.

The update has been done manually.

To reproduce it I do the following:

1. Log-out
2. Blank page in the login screen.

v3app.js:29 SyntaxError: Invalid linked format (at message-compiler.esm-browser.js:116:19)
    at Fe (message-compiler.esm-browser.js:116:19)
    at m (message-compiler.esm-browser.js:241:25)
    at Oe (compiler-messages.esm-navigator.js:837:21)
    at qe (compiler-message.esm-browser.js:851:20)
    at Object.gt [as nextToken] (compiler-messages.esm-browser.js:896:16)
    at P (compiler-messages.esm-navigator.js:1045:31)
    at k (compiler-messages.esm-browser.js:1153:36)
    at y (compiler-message.esm-browser.js:1193:25)
    at Object.w [as parse] (compiler-message.esm-browser.js:1208:21)
    at ua (compiler-message.esm-browser.js:1608:24)
Promise.then		
(anonymous) @ v3app.js:29
Show 30 more frames

![image](https://github.com/user-attachments/assets/99aadbcd-40d6-4a20-85b9-b0c6b2bab542)

### To Reproduce

1. Log-out
2. Blank page en la pantalla de login

### Environment

Linux VM

### Cloud Provider

None

### Platform

Browser

### Operating system

Windows, Mac

### Browser and version

Chrome, Firefox, Safari, Opera

ZiekBaboon:
Same problem with macOS Mojave

sojan-official:
@ZiekBaboon please share more information about the chatwoot version, whether you have done any recent changes upgrade etc.

willihangga:
We’re still encountering the same error, and we believe it’s related to this specific file. Our current solution is to downgrade to version 3.13.0 for now. Let me know if there are any other suggestions or solutions.

![Screen Shot 2024-10-18 at 10 34 00](https://github.com/user-attachments/assets/c516b520-2006-4374-9302-159ea41d6a93)

willihangga:
We’re still encountering the same error, and we believe it’s related to this specific file. Our current solution is to downgrade to version 3.13.0 for now. Let me know if there are any other suggestions or solutions.

[...]

PR #10334, opened 2024-10-22
fix: parsing of `@` in i18n values

scmmishra:
Vue i18n has a new [linked message syntax.](https://vue-i18n.intlify.dev/guide/essentials/syntax.html#linked-messages)
When it encounters `@` it assumes that we're trying to use a linked message. And tries to parse it as such, in any case, it breaks since the syntax is not valid and the params are not present. So it causes an error. This works on dev but on production the error is bubbled up to the top and rendering breaks.

A lot of folks use Chatwoot with default locale set in the env, this surfaced the issue for the languages for which the syntax was not updated

Fixes: https://github.com/chatwoot/chatwoot/issues/10313


vishnu-narayanan (approved):

vishnu-narayanan:
LGTM

Issue #10339, opened 2024-10-22
this.position is always 0 on dashboard apps, so having 2+ apps doesn't work


bfontaine:
### Describe the bug

When having two or more dashboard apps, the response to the event `chatwoot-dashboard-app:fetch-info` is always sent to the first dashboard, even when it’s requested from the other ones.

### To Reproduce

1. Save this as `debug.html`:

```html
<html lang="en">
<body>
<button onclick="window.parent.postMessage('chatwoot-dashboard-app:fetch-info', '*')">Send</button>

<script>
	const x = (Math.random() * 1000) | 0;
	console.log(x);

	window.addEventListener('message', function(event) {
		console.debug(x, 'Got message', event);

		const event_data = JSON.parse(event.data);
		console.log(x, 'Got event:', event_data);
	});
</script>
</body>
</html>
```
This generates a random number, then when you click on the "Send" button it sends `chatwoot-dashboard-app:fetch-info` to the parent and logs the response along the random number.

2. Serve it at some URL accessible by your Chatwoot instance (tested on 3.13.0-3.14.1). I used [backloop.dev](https://backloop.dev/) but that’s irrelevant to the issue
3. In Chatwoot, add the URL as a first dashboard app "Debug" as well as a second one "Debug 2"
4. Open any conversation, open the browser console, go on the "Debug" tab and click on the "Send" button
5. Go on the "Debug 2" tab and click on the "Send" button

### Expected behavior

When clicking on the button in the first dashboard, you should always see the same random ID; and when you switch to the second dashboard, you should always see the same random ID that’s **different** from the first one. This ID serves to highlight which dashboard receives the event.

Actual behavior: it’s always the first dashboard that receives the event.
![Screenshot from 2024-10-22 17-01-46](https://github.com/user-attachments/assets/7da68855-7935-476a-a93c-ab539b2f4f06)

### Environment

Docker

### Cloud Provider

None

### Platform

Browser

### Operating system

Ubuntu 24.04.1 LTS

### Browser and version

Google Chrome Version 130.0.6723.58 (Official Build) (64-bit)

### Additional context

In another setup (same version but different dashboards), I had a different bug but with the same origin: the code that emits a message in response to the event `chatwoot-dashboard-app:fetch-info` in `onIframeLoad` crashed because the ID it looked for didn’t exist in the DOM, because `this.position` was always `0`.

pranavrajs:
@scmmishra Please take a look.

[...]

PR #10342, opened 2024-10-23
feat: use addEventListener instead of onmessage

scmmishra:
This one is a fix for an interesting issue pointed here: https://github.com/chatwoot/chatwoot/issues/10339

### What was happening 

The user had two dashboard apps configured, each of which could request data from the parent Chatwoot window by posting a message with `window.parent.postMessage('chatwoot-dashboard-app:fetch-info', '*')`

The "Dashboard Frame" component had a listener set on window using `window.onmessage`, this was added during the `mounted` event

https://github.com/chatwoot/chatwoot/blob/601a0f8a766e8a02a48a743346a4f3b44b7019c6/app/javascript/dashboard/components/widgets/DashboardApp/Frame.vue#L55-L65

The problem with this approach is the first iframe on mount attaches this event correctly, but every subsequent frame opened `onmessage` would be overridden. Now since the dashboard app frames are caches (see: https://github.com/chatwoot/chatwoot/commit/e6505fc7a42ca9f3de1735939b488a18a979adcb) if you wen't back to a particular tab that was already loaded the `mounted` would not get triggered again, that means the older `onmessage` would still be active and will receive the messages exclusively

### What is the fix

The simplest fix is to use `addEventListener`, that way every `Frame` handles it's own events. **However there is a caveat here;** If one app requests data using `fetch-info`, every frame will get the data back. I have a feeling this is not intended, to ensure this does not happen we can add a check to `triggerEvent` method

```diff
    triggerEvent(event) {
+     if (!this.isVisible) return;
      if (event.data === 'chatwoot-dashboard-app:fetch-info') {
        this.onIframeLoad(0);
      }
    },
```

### DIFF
--- a/app/javascript/dashboard/components/widgets/DashboardApp/Frame.vue
+++ b/app/javascript/dashboard/components/widgets/DashboardApp/Frame.vue
@@ -1,5 +1,6 @@
 <script>
 import LoadingState from 'dashboard/components/widgets/LoadingState.vue';
+
 export default {
   components: {
     LoadingState,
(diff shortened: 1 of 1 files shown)


pranavrajs (changes requested):
Please add the isVisible check as well before we merge it.

bfontaine:
Is there no way to identify the frame that sent the event, instead of silently ignoring it if the frame is not visible?

scmmishra:
> Is there no way to identify the frame that sent the event, instead of silently ignoring it if the frame is not visible?

At the moment there isn't a straightforward way to do this that I am aware of. We could compare the event origin with the dashboard app URL, but it would fire more than once if you have multiple dashboard apps from the same origin.

pranavrajs (approved):

bfontaine:
`event.source` should be the window that emitted the event: 

> source:
> A reference to the [window](https://developer.mozilla.org/en-US/docs/Web/API/Window) object that sent the message; you can use this to establish two-way communication between two windows with different origins.

https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#source

Issue #9285, opened 2024-04-23
Next available day/hour is not properly translated


micheh:
### Describe the bug

When the working hours are set, the widget shows the time or day when agents will be back online. This works fine in English, but not so much in other languages. The problem is that there are some hard coded English strings in the mixing timeLeftToBackInOnline or the translations are not available like `tomorrow`:

https://github.com/chatwoot/chatwoot/blob/8957277e5dc79f556066dd8073de24a330256c8c/app/javascript/widget/mixins/nextAvailabilityTime.js#L156-L176

This results in a mix of English and the other locale. In addition, the date string is added after the translation `REPLY_TIME.BACK_IN`. Other languages may have a different sentence structure and even if the relative dates strings are translated correctly, it will look strange when the time is added at the end.

### To Reproduce

1. Use the web widget with a locale different than English.
2. Enable the working hours and make sure it is currently outside of the working hours
3. Open the widget and see the wrong translated string

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

None

vishnu-narayanan:
@micheh Thank you for reporting this. If possible, could you share a screenshot of it? Which language are you using?

The only hard-coded string in this method would be `days` right. Anything else?

micheh:
Hello @vishnu-narayanan, here is a screenshot: 

![image](https://github.com/chatwoot/chatwoot/assets/1307920/658baba6-2002-454c-ad02-886ae6c2a1f9)

I tested it with German (de), but could replicate the issue in other languages like French (fr), Italian (it) and Spanish (es).

Almost all methods in `nextAvailabilityTime.js` are affected.

`generateRelativeTime` returns an English string:
https://github.com/chatwoot/chatwoot/blob/8957277e5dc79f556066dd8073de24a330256c8c/app/javascript/widget/mixins/nextAvailabilityTime.js#L161-L165

Line 172 returns the hard coded string `on` and the translated day:
https://github.com/chatwoot/chatwoot/blob/8957277e5dc79f556066dd8073de24a330256c8c/app/javascript/widget/mixins/nextAvailabilityTime.js#L172

`hoursAndMinutesToBack` and the other methods used like `exactTimeInAmPm` also return hard coded strings:
https://github.com/chatwoot/chatwoot/blob/8957277e5dc79f556066dd8073de24a330256c8c/app/javascript/widget/mixins/nextAvailabilityTime.js#L146-L154


In the German language, the relative time should be in another place. 
[...]

PR #10360, opened 2024-10-28
fix: Add date formatting for other locales

### LINKED ISSUE #7862: [CW-2457] fix: Make activity timestamp messages translatable
These copies aren't translatable at the moment. 

ref: [https://github.com/chatwoot/chatwoot/blob/9322112481e6341b6b291d7d44b234434a2f6c9b/app/javascript/dashboard/components/ui/TimeAgo.vue#L56](https://github.com/chatwoot/chatwoot/blob/9322112481e6341b6b291d7d44b234434a2f6c9b/app/javascript/dashboard/components/ui/TimeAgo.vue#L56)

<sub>From [SyncLinear.com](https://synclinear.com) | [CW-2457](https://linear.app/chatwoot/issue/CW-2457/fix-make-activity-timestamp-messages-translatable)</sub>

KokeroO:
# Pull Request Template

## Description

When applying a new language setting from the account settings, date and time formatting is applied to the language of choice. Previously, only the 'en' localization was applied. This caused confusion and doubt for the end user. Other changes have also been applied to translate texts that were not handled by the i18n library.
Resolves  #7862, #9285

## Type of change

## How Has This Been Tested?

1. Go to app.chatwoot.com/app/accounts/1/settings/general and change the account language.
2. All pages showing dates will be in the correct format for your locale.

## Checklist:


KokeroO:
Resolves  #7862, #9285

KokeroO:
I finished all changes to this PRs

KokeroO:
I have finalized all the changes in this PR.
I ended up changing 2 json files that were not part of the purpose, but I fixed them in the last force-pushed.

app/javascript/dashboard/App.vue
```diff
@@ -82,7 +82,7 @@ export default {
   mounted() {
     this.initializeColorTheme();
     this.listenToThemeChanges();
-    this.setLocale(window.chatwootConfig.selectedLocale);
+    this.setLocale(window.chatwootConfig.selectedLocale.replace('-', '_'));
```
scmmishra, line 85: This replace seems redundant since `setLocale` already does it for us

app/javascript/dashboard/App.vue
```diff
@@ -98,7 +98,7 @@ export default {
       mql.onchange = e => setColorTheme(e.matches);
     },
     setLocale(locale) {
-      this.$root.$i18n.locale = locale;
+      this.$root.$i18n.locale = locale.replace('-', '_');
```
scmmishra, line 101: This may not work. With Vue 3, the `$root` element is not present. Just updating the `this.$i18n.locale` should work. Can you please test that @KokeroO

scmmishra, line 101: This may not be the most reliable. Just updating the `this.$i18n.locale` should work better

app/javascript/dashboard/App.vue
scmmishra, line 85: This replace seems redundant since `setLocale` already does it for us

app/javascript/dashboard/App.vue
```diff
@@ -108,7 +108,7 @@ export default {
       const { locale, latest_chatwoot_version: latestChatwootVersion } =
         this.getAccount(this.currentAccountId);
       const { pubsub_token: pubsubToken } = this.currentUser || {};
-      this.setLocale(locale);
+      this.setLocale(locale.replace('-', '_'));
```
scmmishra, line 111: This replace seems redundant since `setLocale` already does it for us

app/javascript/dashboard/components/widgets/conversation/MessagesView.vue
scmmishra, line 1: I see what you're trying to do here, but the logic seems to be replicated here needlessly. `getTypingUsersText` is used specifically for this component, so we can modify it to return the fully prepared i18n key as well as the params

```
export const getTypingUsersText = (users = []) => {
  const count = users.length;
  if (count === 1) {
    const [user] = users;
    return ['TYPING.SINGLE', { user: user.name }];
  }

  if (count === 2) {
    const [first, second] = users;
    return [
      'TYPING.DOUBLE',
      {
        user: first.name,
        secondUser: second.name,
      },
    ];
  }

  const [user] = users;
  const rest = users.length - 1;
  return [
    'TYPING.MULTIPLE',
    {
      user: [user.name],
      rest: rest,
    },
  ];
}
```

And the usage can simply look like the following

```
typingUserNames() {
  const userList = this.typingUsersList;
  if (this.isAnyoneTyping) {
    const [i18nKey, params] = getTypingUsersText(userList);
    return this.$t(i18nKey, params);
  }

  return '';
}
```

We just need to update the specs in `app/javascript/dashboard/helper/specs/commons.spec.js`

app/javascript/dashboard/i18n/dateFormat.js
```diff
@@ -0,0 +1,200 @@
+const dateFormatsByLocale = {
```
scmmishra, line 1: Please take a look at `getIntlDateFormatForLocale`, it's a neat hack to use `Intl.DateTimeFormat` to get the date formats, you can use it to get the `fullDateTime` and `timeOnly` as well.

https://github.com/chatwoot/chatwoot/blob/2d35fa135b17fc68d594d2b7f0e693ec51fef924/app/javascript/dashboard/components/ui/DatePicker/helpers/DatePickerHelper.js#L70-L82

Adding a static list of these locales is not a great idea, very difficult to be exhaustive and also error prone

app/javascript/dashboard/i18n/locale/en/conversation.json
scmmishra, line 1: You can remove the i18n strings from the other `json` files, it will default to the `en` one anyway.

app/javascript/dashboard/i18n/locale/en/helpCenter.json
scmmishra, line 1: Can you discard this change

app/javascript/dashboard/i18n/locale/en/settings.json
scmmishra, line 1: Can you discard this change

app/javascript/dashboard/components/ui/TimeAgo.vue
```diff
@@ -27,17 +27,31 @@ export default {
...
-      lastActivityAtTimeAgo: dynamicTime(this.lastActivityTimestamp),
-      createdAtTimeAgo: dynamicTime(this.createdAtTimestamp),
+      lastActivityAtTimeAgo: dynamicTime(
+        this.lastActivityTimestamp,
+        this.$root.$i18n.locale
+      ),
+      createdAtTimeAgo: dynamicTime(
+        this.createdAtTimestamp,
+        this.$root.$i18n.locale
+      ),
```
scmmishra, line 37: I don't see the changes for `dynamicTime` function, same with `shortTimestamp` did you forget the commit it?

scmmishra (changes requested):
Thanks for the contribution @KokeroO 
I've added some comments, and thoughts, please take a look.

1. You can get rid of the extra i18n json changes, those are not necessary
2. You seemed to have missed adding the implementation for `dynamicTime` and other helpers to use the locale
3. I don't like the fact the to use dynamicTime or other helpers I need to pass the locale everytime. We can convert them to a composable with access to the vue context, `useTimeHelpers` that can return the functions for `dynamicTime`, `shortTimestamp` and others with the current locale already set
4. The chagnes don't seem to be comprehensive, for instance in `Datepicker` we use the navigator language instead of the app language, that needs to change too

app/javascript/dashboard/i18n/locale/en/conversation.json
KokeroO: Can I leave the location pt_BR? Since translation is of interest to me at the moment?

PR #10363, opened 2024-10-29
refactor: use css only last item detection

scmmishra:
The last item in the sidebar top level group has an indicator specified, the problem in our case is that the structure can be nested and have sub groups. So selecting the last item correctly can be tricky. 

Previous implementation relied on the using DOM queries to find the last item from a flat list of children, it would trigger on a `watch`. This was error-prone as well as non idiomatic. The new approach is CSS-only and reduces the unnecessary compute required. 

Codepen for reference: https://codepen.io/scmmishra/pen/yLmKNLW


iamsivin (approved):
LGTM
@scmmishra A minor issue was found where the L-shaped connector does not display properly when the section is collapsed and an item within it is active. 
Feel free to ignore if it's not needed.

image

Issue #10372, opened 2024-10-30
The value of the display_name field in the user profile is reset


Alex808r:
### Describe the bug

When I change the password from the profile menu, the display_name field is reset

This happens because when you change the password, an empty `display_name` is sent to params. 

development logs
```
backend | Started PUT "/api/v1/profile" for ::1 at 2024-10-30 10:14:24 +0300
backend | Processing by Api::V1::ProfilesController#update as JSON

backend |   Parameters: {"profile"=>{"current_password"=>"[FILTERED]", "display_name"=>"", "password"=>"[FILTERED]", "password_confirmation"=>"[FILTERED]"}}

backend |   Rendered api/v1/profiles/update.json.jbuilder (Duration: 44.8ms | Allocations: 22179)
backend | Completed 200 OK in 409ms (Views: 41.3ms | ActiveRecord: 27.2ms | Allocations: 44228)
```

### To Reproduce

1. Open profile menu (http://localhost:3000/app/accounts/1/profile/settings)
2. Fill in the `display_name` field. Click button => `Update Profile`
3. Update page and check `display_name`. Field is not empty. 
4. Fill in the `Current password` field
5. Fill in the `New password` field
6. Fill in the `Confirm new password` field
7. Click button => `Change Password`
8. Login to user with new password and check profile menu (http://localhost:3000/app/accounts/1/profile/settings)
9. The `display_name` field is empty

### Expected behavior

The `display_name` field is not reset when user change his password

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

Browser

### Operating system

Mac os Sonoma

### Browser and version

130.0.6723.59

Alex808r:
I can fixed it

KokeroO:
The correction must be made in the file:
/app/controllers/api/v1/profiles_controller.rb
A check must be created to see if the name is present and if it is different from the current one.

If no one works on a PR by Friday, I will submit a PR on Saturday.

corrosive4354:
Make sure that the value of the display_name field is correctly passed to the backend when the user submits the password change form. You can add debugging information to the code that handles form submission to confirm whether the value of display_name exists.

Alex808r:
@corrosive4354
Thanks for your comment. When updating the password, only 2 fields are sent, a new password and confirmation of the new password. 
However, when forming hash `params` in frontend method `profileUpdate` add empty display_name to `params`. 
[...]

PR #10374, opened 2024-10-30
fix: issue-10372 The value of the display_name field reset

Alex808r:
# Pull Request Template

## Description
When changing the password through the user profile menu, the display_name field is not reset.

Fixes # [10372](https://github.com/chatwoot/chatwoot/issues/10372)


pranavrajs:
@Alex808r Thanks for the PR. This seems like a hacky fix. The best solution would be to move change_password to a different endpoint so that it only does password change and doesn't update the user data. Let me know if you would be able to make the change.

Alex808r:
@pranavrajs 

it would be interesting for me to deal with this problem. Could you describe in more detail exactly what it should look like.
Now, when changing the password, an empty value `display_name => ""` is added to params in this method
string:  `formData.append('profile[display_name]', displayName || '');`

app/javascript/dashboard/api/auth.js
```
profileUpdate({
    password,
    password_confirmation,
    displayName,
    avatar,
    ...profileAttributes
  }) {
    const formData = new FormData();
    Object.keys(profileAttributes).forEach(key => {
      const hasValue = profileAttributes[key] === undefined;
      if (!hasValue) {
        formData.append(`profile[${key}]`, profileAttributes[key]);
      }
    });
    formData.append('profile[display_name]', displayName || '');
    if (password && password_confirmation) {
      formData.append('profile[password]', password);
      formData.append('profile[password_confirmation]', password_confirmation);
    }
    if (avatar) {
      formData.append('profile[avatar]', avatar);
    }
    return axios.put(endPoints('profileUpdate').url, formData);
  },
```

Alex808r:
it's work but now i have another problem: red banner after update password.  

screen 2024-10-31 в 13 29 22

pranavrajs:
@Alex808r Thanks for the input. Removing the line `formData.append('profile[display_name]', displayName || '');` for password updates could work as a solution, but it doesn’t address the root problem. If someone still sends this parameter to the API, the issue would re-appear.

Returning after the password update, as you suggested, is an option, but it could lead to more confusion. We’d then have to explain why profile updates didn’t apply due to the presence of a password parameter. 

To keep things straightforward, a separate endpoint for password updates would be more readable and manageable.

I will review and test the PR. Thanks for the changes.

pranavrajs:
Sorry for the delay, @Alex808r 🙂 We’ve decided to keep the change on the frontend only, since the current API updates fields only when they’re present. I've made the required changes myself, will get the PR merged once CI is green.

Thanks for the effort.

Alex808r:
Glad to help 🙂

PR #10379, opened 2024-10-31
fix: Process attachments as regular attachments if the text/plain or text/html part is empty

pranavrajs:
Some email clients automatically set Content-Disposition to inline for specific content types, such as images. In cases where the email body is empty, inline attachments may not display correctly due to our previous implementation. Our assumption was that these attachments are referenced within text/plain or text/html parts. 

Customer-reported issues, especially with Apple Mail, show emails with attachments marked as inline but without any corresponding text parts. This leads to missing attachments even though would have processed the attachment.

This update introduces a check for the presence of a text part. If none exists, inline attachments are treated as regular attachments and added to the external attachments array, ensuring that all attachments display properly.

<details>
<summary><b>Script to update the existing emails that are already available in the system</b></summary>

```rb
def update_content id
  message = Message.find id
  conversation = message.conversation
  message_id = message.source_id

  channel = message.inbox.channel

  authentication_type = 'XOAUTH2'
  imap_password = Google::RefreshOauthTokenService.new(channel: channel).access_token
  imap = Net::IMAP.new(channel.imap_address, port: channel.imap_port, ssl: true)
  imap.authenticate(authentication_type, channel.imap_login, imap_password)
  imap.select('INBOX')

  results = imap.search(['HEADER', 'MESSAGE-ID', message_id])
  message_content = imap.fetch(results.first, 'RFC822').first.attr['RFC822']
  mail = MailPresenter.new(Mail.read_from_string(message_content))

  mail_content = if mail.text_content.present?
                   mail.text_content[:reply]
                 elsif mail.html_content.present?
                   mail.html_content[:reply]
                 end

  attachments = mail.attachments.last(Message::NUMBER_OF_PERMITTED_ATTACHMENTS)
  inline_attachments = attachments.select { |attachment| attachment[:original].inline? && mail_content.present? }
  regular_attachments = attachments - inline_attachments

  regular_attachments.each do |mail_attachment|
    attachment = message.attachments.new(
      file_type: 'file'
    )
    attachment.file.attach(mail_attachment[:blob])
  end

  message.save!
end
```
</details>

### DIFF
--- a/spec/fixtures/files/attachments_without_text.eml
+++ b/spec/fixtures/files/attachments_without_text.eml
@@ -1,0 +1,65 @@
+From: hello@icloud.com
+Content-Type: multipart/mixed;
+	boundary="Apple-Mail=_F9915DAD-44A0-47DE-BB38-1168"
+Mime-Version: 1.0 (Mac OS X Mail 16.0 \(3818.100.11.1.3\))
+Subject: Hello there
+Message-Id: <7DDDD-F45F@icloud.com>
+Date: Thu, 31 Oct 2024 18:19:46 -0700
+To: Pranav Raj S <pranavrajs@gmail.com>
+
+
+--Apple-Mail=_F9915DAD-44A0-47DE-BB38-1168
+Content-Disposition: inline;
+	filename=1kb.png
+Content-Type: image/png;
+	x-unix-mode=0644;
+	name="1kb.png"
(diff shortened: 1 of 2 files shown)


sojan-official:
Screenshot 2024-11-04 at 10 22 38 AM

Verified in staging with apple mail

sojan-official (approved):

Issue #10381, opened 2024-11-02
Some tests fail if run locally


Alex808r:
### Case 1
If you set any locale in the default configuration file except **:en**, then some tests fail. 

This is because error messages in these tests are waiting for messages in the en locale. 
To solve this situation, specify 
`config.i18n.default_locale = :en` for `config/environments/test.rb`

### Case 2
Error example
response  was a 500 because an `UnsafeRedirectError` exception 

```
1) Session GET /auth/sign_in redirects to the frontend login page with error
      Failure/Error: expect(response).to redirect_to(%r{/app/login\?error=access-denied$})
        Expected response to be a <3XX: redirect>, but was a <500: Internal Server Error>
        ActionController::Redirecting::UnsafeRedirectError:
       Unsafe redirect to "http://0.0.0.0:3000....
```

To solve this error need add  
`config.action_controller.raise_on_open_redirects = false` to `config/environments/test.rb`

### To Reproduce

1. set any default locale except `:en` 
2. run command `bundle exec rspec spec`

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

None

### Operating system

Mac os Sonoma

corrosive4354:
When you set any locale in the default configuration file to anything other than :en, some tests will fail because they expect error messages to be output in English.

In addition, seeing that you are using Docker, I would like to recommend [Servbay](servbay.com) to you, which is more professional and easier to use than Docker on Mac

pranavrajs:
@Alex808r Closing this via https://github.com/chatwoot/chatwoot/pull/10382#issuecomment-2964810551. The second issue mentioned doesn’t seem to be reproducible.

PR #10382, opened 2024-11-02
fix: issue-10381 some tests fail if run locally

Alex808r:
# Pull Request Template

## Description
Some tests fail if run locally

Fixes # ([10381](https://github.com/chatwoot/chatwoot/issues/10381))

## Type of change


pranavrajs:
@Alex808r This issue is likely due to the .env file on your local machine as you mentioned. Since we’re not enforcing a specific language in the tests right now, I’ll go ahead and close this PR.

PR #10389, opened 2024-11-04
feat: setup captain lib

scmmishra:


sojan-official (approved):
Code LGTM. 

- we need to address the case where feature is enabled and integration is not configured. once that's done this should be good to go

pranavrajs (approved):
I’ve tested this locally and on staging, and the conflicting styles are no longer an issue. The basic functionalities are all working. I’ll raise the remaining issues as a separate task.

Great job!

PR #10399, opened 2024-11-11
feat: conversation filter component

scmmishra:
This PR adds the new conversation filter component, we need to merge the `feat/new-dropdown` branch before merging this

CleanShot 2024-11-18 at 09 24 56@2x


pranavrajs:
@scmmishra If I click outside, the dropdown doesn't close properly. Please see the screenshot below. 

Screenshot 2024-11-18 at 5 52 29 PM

Please also note the color contrast between the input box and the modal. I'd test this after the issue is resolved.

scmmishra:
> @scmmishra If I click outside, the dropdown doesn't close properly. Please see the screenshot below.

Seems like this came in when I moved to using the Dialog component, fixing it

scmmishra:
> @scmmishra If I click outside, the dropdown doesn't close properly. Please see the screenshot below.

@pranavrajs fixed this

https://github.com/user-attachments/assets/e7506aa3-126b-493a-ad1b-47357063479c

app/javascript/dashboard/components-next/dialog/Dialog.vue
```diff
@@ -44,21 +44,37 @@ defineProps({
...
+  const classesMap = {
+    '3xl': 'max-w-3xl',
+    '2xl': 'max-w-2xl',
+    xl: 'max-w-xl',
+    lg: 'max-w-lg',
+    md: 'max-w-md',
+    sm: 'max-w-sm',
+  };
+
+  return classesMap[props.width];
```
scmmishra, line 75:
```suggestion
  return classesMap[props.width] ?? 'max-w-md';
```

app/javascript/dashboard/i18n/locale/en/components.json
```diff
@@ -5,8 +5,10 @@
   },
   "COMBOBOX": {
     "PLACEHOLDER": "Select an option...",
+    "EMPTY_SEARCH_RESULTS": "No items found for the search term \"{searchTerm}\"",
```
pranavrajs, line 8:
```suggestion
    "EMPTY_SEARCH_RESULTS": "No items found for the search term `{searchTerm}`",
```

pranavrajs: The `\` character would create a mess in the translations.

app/javascript/dashboard/components-next/icon/ChannelIcon.vue
```diff
@@ -0,0 +1,17 @@
+<script setup>
```
pranavrajs, line 1: Could you create a separate PR for this? It appears unrelated to the filters.

app/javascript/dashboard/components-next/button/Button.vue
```diff
@@ -11,36 +11,60 @@ const props = defineProps({
...
+  trailingIcon: { type: Boolean, default: false },
+  isLoading: { type: Boolean, default: false },
 });
 
 const slots = useSlots();
+const attrs = useAttrs();
+
+const computedVariant = computed(() => {
+  if (props.variant) return props.variant;
+  if (attrs.solid !== undefined) return 'solid';
```
pranavrajs, line 42: Is it required to check this with undefined? Shouldn't `attrs.solid` should work?

pranavrajs, line 42:
```suggestion
  if (attrs.solid) return 'solid';
```

Is it mandatory to check this with `undefined`? Shouldn’t `attrs.solid` suffice?

scmmishra:
@pranavrajs fixed the slot issue

app/javascript/dashboard/components-next/button/Button.vue
scmmishra: Fixed it here: https://github.com/chatwoot/chatwoot/pull/10473

iamsivin (commented):
@scmmishra I tested the filter PR on staging and found a couple of minor issues. I’ve detailed them in the attached Loom videos. Could you please take a look?

https://www.loom.com/share/0f9924095080496db41a297e6f28aa43?sid=67507bf1-9622-4142-a4f7-b3b53e70ec27
https://www.loom.com/share/b7b331a02e804499ac5751f17527dd1d?sid=49d0f73e-ba2b-481b-b286-7560d7912d3e

app/javascript/dashboard/components-next/filter/inputs/MultiSelect.vue
```diff
@@ -0,0 +1,146 @@
...
+    <template #trigger="{ toggle }">
+      <button
+        v-if="hasItems"
+        class="bg-n-alpha-2 py-2 rounded-lg h-8 flex items-center px-0"
+        @click="toggle"
+      >
+        <div
+          v-for="item in selectedVisibleItems"
+          :key="item.name"
+          class="px-3 border-r border-n-weak text-n-slate-12 text-sm flex gap-2 items-center max-w-[100px]"
```
iamsivin, line 100: Let us consider the border for RTL also.
image

app/javascript/dashboard/components-next/filter/ConversationFilter.vue
```diff
@@ -0,0 +1,175 @@
...
+  }
+};
+
+const addFilter = () => {
+  filters.value.push({
+    attribute_key: 'status',
+    filter_operator: 'equal_to',
+    values: '',
+    query_operator: 'and',
+  });
```
iamsivin, line 58: Can we move it to a constant and reuse it here?

app/javascript/dashboard/components-next/filter/ConversationFilter.vue
```diff
@@ -0,0 +1,175 @@
...
+const { t } = useI18n();
+const store = useStore();
+
+const resetFilter = () => {
+  filters.value = [
+    {
+      attribute_key: 'status',
+      filter_operator: 'equal_to',
+      values: '',
+      query_operator: 'and',
```
iamsivin, line 39: Can we move it to a constant and use it here?

app/javascript/dashboard/components-next/filter/inputs/MultiSelect.vue
```diff
@@ -0,0 +1,146 @@
...
+        >
+          <Icon v-if="item.icon" :icon="item.icon" class="flex-shrink-0" />
+          <span class="truncate">{{ item.name }}</span>
+        </div>
+        <div
+          v-if="remainingItems.length > 0"
+          v-tooltip.top="remainingTooltip"
+          class="px-3 border-r border-n-weak text-n-slate-12 text-sm flex gap-2 items-center max-w-[100px]"
```
iamsivin, line 108: Let us consider the border for RTL also.

app/javascript/dashboard/components-next/filter/ConversationFilter.vue
```diff
@@ -0,0 +1,175 @@
...
+  emit('close');
+};
+
+const filterModalHeaderTitle = computed(() => {
+  return !props.isFolderView
+    ? t('FILTER.TITLE')
+    : t('FILTER.EDIT_CUSTOM_FILTER');
+});
+
+onMounted(() => dialogRef.value.open());
```
iamsivin, line 104: Could you help me understand why we need to open the dialog on mounted?

pranavrajs:
@scmmishra Here is my feedback: A couple of improvements could reduce the effort required to use the same in the other parts of the product. Please see my thoughts below.

__Make ConditionRow reusable__: The ConditionsRow component is currently tightly coupled with the conversation context, making it specific to that use case and unusable for other sections like the contacts filter or automations. We should make it generic. 

A good approach would be to pass filterTypes as a prop to the component, which removes the dependency on the conversation context. Since the component will no longer rely on the store, you can write the stories for the same for testing it on Histoire.

__Refactoring ConversationFilter component__: The conversation filter component can also benefit from being refactored into a container-component pattern. The container would manage interactions with the store and isolate the associated actions, while the component FilterForm could focus on rendering a reusable, generic form. This form should handle all form-related logic, including validations, and provide a callback to indicate when the form values are valid. This way, the FilterForm can be reused in other contexts, such as the contacts page, with minimal adjustments—only the container and the appropriate filterTypes would need to be updated.

app/javascript/dashboard/components-next/filter/provider.js
```diff
@@ -0,0 +1,299 @@
...
+   */
+  const customFilterTypes = computed(() => {
+    return conversationAttributes.value.map(attr => {
+      return {
+        attribute_key: attr.attributeKey,
+        value: attr.attributeKey,
+        attribute_name: attr.attributeDisplayName,
+        label: attr.attributeDisplayName,
+        input_type: customAttributeInputType(attr.attributeDisplayType),
+        filter_operators: getOperatorTypes(attr.attributeDisplayType),
```
pranavrajs, line 99: Is there a specific reason why we still use the snake case here? Couldn’t this be transformed before passing it to the component?

app/javascript/dashboard/components-next/filter/provider.js
```diff
@@ -0,0 +1,299 @@
...
+      label: t('FILTER.ATTRIBUTES.LAST_ACTIVITY'),
+      input_type: 'date',
+      data_type: 'text',
+      filter_operators: dateOperators.value,
+      attribute_model: 'standard',
+    },
+    ...customFilterTypes.value,
+  ]);
+
+  const fitlerGroups = computed(() => {
```
pranavrajs, line 275: Spelling mistake.

pranavrajs (changes requested):
Please see the comments above.

app/javascript/dashboard/components-next/dialog/Dialog.vue
```diff
@@ -44,6 +44,11 @@ const props = defineProps({
     type: Boolean,
     default: true,
   },
+  width: {
```
pranavrajs, line 47: Seems like this is coming from a bad merge, please remove this.

pranavrajs:
Some more feedback on the product side:

1/ I got an error where the value was displayed as “[Object object]”. However, I was not able to reproduce this issue again.
Screenshot 2024-11-24 at 8 26 48 PM
 
2/ I'd expect the dropdown to close automatically if I select a filterAttribute. Let us say I select Conversation Identifier from the list, there shouldn't be another click to close that list. See the interaction below.

https://www.loom.com/share/4ad94937b6ea488d854348e8726b0a9d

app/javascript/dashboard/components-next/filter/provider.js
```diff
@@ -0,0 +1,299 @@
...
+  /**
+   * @type {import('vue').ComputedRef<FilterType[]>}
+   */
+  const filterTypes = computed(() => [
+    {
+      attribute_key: 'status',
+      value: 'status',
+      attribute_name: t('FILTER.ATTRIBUTES.STATUS'),
+      label: t('FILTER.ATTRIBUTES.STATUS'),
+      input_type: 'multiSelect',
```
pranavrajs, line 118: status is implemented as single select in the backend. This is why there is an `all` key.

app/javascript/dashboard/components-next/filter/provider.js
pranavrajs, line 99: Is there a specific reason why we still use snake case attributes? Couldn’t we transform the keys while passing them to the component and then retrieve them from the component on the parent?

app/javascript/dashboard/components-next/filter/ConversationFilter.vue
scmmishra: The default dialog state is closed. So we manually have to open it. It's fixed in the popover PR. There we don't need this

scmmishra:
> 1/ I got an error where the value was displayed as “[Object object]”. However, I was not able to reproduce this issue again.

Fixed this.



> I'd expect the dropdown to close automatically if I select a filterAttribute. Let us say I select Conversation Identifier from the list, there shouldn't be another click to close that list. See the interaction below.

Strange, this is not happening for me, probably an old commit? Can you check again?

PR #10420, opened 2024-11-14
fix: Email attachments created with empty filename

sojan-official:
- We observed in prod for certain emails active storage blob objects were getting created with empty file name. The conversations further causes conversation and filter pages to break. This change will fix the mentioned issue. 

fixes: https://linear.app/chatwoot/issue/CW-3331/missing-file-name-for-some-of-the-uploads-for-emails


pranavrajs (approved):
The code looks good. I’ve updated the spec fixture and added a condition to check the name.

PR #10444, opened 2024-11-19
feat: Updates on new components

iamsivin:
# Pull Request Template

## Description

This PR introduces updates to multiple components as part of the **contact redesign**, addressing new features, improvements, and deprecations. Detailed changes for each component are outlined below:

### **CardLayout.vue**
- Added support for a sidebar section.
- Update all other components where we use `CardLayout.vue` with updated changes.

### **Avatar.vue**
- Enhanced functionality to replace `Thumbnail.vue` and `EditableAvatar.vue`.

 New Features:
 - Upload Overlay & File Input: 
   1. Enabled via the allowUpload prop.
   2. Emits an upload event with the uploaded file and its URL.
- Delete Button:
   1. Allows deletion of the avatar image.
   2. Emits a delete event.

### **EditableAvatar.vue**
- Deprecated and replaced with `Avatar.vue`

### **Thumbnail.vue**
- Deprecated and replaced with `Avatar.vue`

### **Button.vue**
- Updated text color for the `blue` variant
- Automatically hides if the `label` prop is not provided.

### **ComboBox.vue** and **ComboBoxDropdown.vue**
- Added a search event for API fetch support

### **Dialog.vue**
- Added a `maxWidth` prop to customize dialog width.
- Introduced a `footer` slot for custom content use cases.

### **DropdownMenu.vue**
- Added search functionality, enabled via the `showSearch` prop
- Updated to utilize `Avatar.vue`.

### **TagInput.vue**
- Trigger `addTag` on key `,`

### **Input.vue**
- Added an `autofocus` prop to enable automatic focus.
- Now emits the following events: `focus`, `blur` and `enter` (on enter key press)
- Fixed an issue where input error styles were not persisting upon focusing after validation errors.


app/javascript/dashboard/components-next/CardLayout.vue
```diff
@@ -1,23 +1,33 @@
...
-    <slot name="footer" />
+    <div
+      class="flex w-full gap-3 px-6 py-5"
+      :class="
+        layout === 'col' ? 'flex-col' : 'flex-row justify-between items-center'
+      "
+      @click="handleClick"
+    >
+      <slot name="header" />
+      <slot name="footer" />
```
scmmishra, line 28: Can we not have just a single default slot here? Right now two slots seem arbitrary, given that neither have any specific styles or functions

app/javascript/dashboard/components-next/CardLayout.vue
```diff
@@ -1,23 +1,33 @@
...
   emit('click');
 };
 </script>
 
 <template>
   <div
-    class="relative flex w-full gap-3 px-6 py-5 shadow outline-1 outline outline-n-container group/cardLayout rounded-2xl bg-n-solid-2"
-    :class="props.layout === 'col' ? 'flex-col' : 'flex-row'"
-    @click="handleClick"
+    class="relative flex flex-col w-full shadow outline-1 outline outline-n-container group/cardLayout rounded-2xl bg-n-solid-2"
```
scmmishra, line 18: Any reason to add the `relative` class here?

app/javascript/dashboard/components-next/CardLayout.vue
```diff
@@ -1,23 +1,33 @@
...
+      :class="
+        layout === 'col' ? 'flex-col' : 'flex-row justify-between items-center'
+      "
+      @click="handleClick"
+    >
+      <slot name="header" />
+      <slot name="footer" />
+    </div>
+
+    <slot name="expanded" />
```
scmmishra, line 31: Nitpick, Should be a better name for this, maybe `content` or `after` 

```suggestion
    <slot name="after" />
```

app/javascript/dashboard/components-next/avatar/Avatar.vue
scmmishra, line 1: Ideally upload should not be here, we could have two components. `Avatar` and `AvatarWithUpload`. 

Having this separate will make this component much simpler, also, most avatar usage does not required upload, so it will make those components lighter also, less performance hit when rendering a large list.

@iamsivin I would suggest take it as an enhancement separately, for now the component looks fine

app/javascript/dashboard/components-next/combobox/ComboBox.vue
```diff
@@ -125,6 +125,7 @@ watch(
         :empty-state="emptyState"
         :selected-values="selectedValue"
         @update:search-value="search = $event"
```
scmmishra, line 127: We could use `v-model` here

app/javascript/dashboard/components-next/combobox/ComboBoxDropdown.vue
```diff
@@ -46,6 +46,11 @@ const isSelected = option => {
   return option.value === props.selectedValues;
 };
 
+const onInputSearch = event => {
+  emit('update:searchValue', event.target.value);
```
scmmishra, line 50: We can use `defineModel` and avoid manually emitting this. We can keep the `emit('search', event.target.value);` anyway

app/javascript/dashboard/components-next/dialog/Dialog.vue
```diff
@@ -48,6 +48,10 @@ defineProps({
     type: Boolean,
     default: false,
   },
+  maxWidth: {
+    type: String,
+    default: 'max-w-lg',
+  },
```
scmmishra, line 54: This is not a great way to do this. I would suggest having more control over the prop, I tried something like that here.

https://github.com/chatwoot/chatwoot/blob/3a90604623b2f3be75a68c5719c26b1406d166ea/app/javascript/dashboard/components-next/dialog/Dialog.vue#L47-L51

https://github.com/chatwoot/chatwoot/blob/3a90604623b2f3be75a68c5719c26b1406d166ea/app/javascript/dashboard/components-next/dialog/Dialog.vue#L65-L76

This is from the ConversationFilter PR. You could do this.

In the previous case, the width prop accepts arbitrary values, without any validation. So I can add something like `max-w-lg hidden` and it will add it as it is. Or a typo like `mx-w-lg` will be accepted but not effect, since there is no validation

<details><summary> However I think there's a better solution. </summary>
<p>

You could remove the prop and related logic entirely, [disable the attribute](https://vuejs.org/guide/components/attrs#disabling-attribute-inheritance) inheritance completely and then attach the attributes to the `<dialog>` element using v-bind. ([ref](https://vuejs.org/guide/components/attrs#disabling-attribute-inheritance:~:text=We%20want%20all%20fallthrough%20attributes%20like%20class%20and%20v%2Don%20listeners%20to%20be%20applied%20to%20the%20inner%20%3Cbutton%3E%2C%20not%20the%20outer%20%3Cdiv%3E.%20We%20can%20achieve%20this%20with%20inheritAttrs%3A%20false%20and%20v%2Dbind%3D%22%24attrs%22%3A))

That way we can attach classes directly, and whenever we need a larger max-width, we can add it directly without a new prop.

</p>
</details> 

For now my suggestion is do the first implementation, later we can figure out the inherAttrs one.

app/javascript/dashboard/components-next/dialog/Dialog.vue
```diff
@@ -101,25 +108,27 @@ defineExpose({ open, close });
           <slot name="form">
```
scmmishra, line 108: This slot should not have a name. It can be the default one. Otherwise we will have to add `<template #form>` everywhere we're using the dialog

app/javascript/dashboard/components-next/input/Input.vue
```diff
@@ -42,9 +42,22 @@ const props = defineProps({
     type: String,
     default: '',
   },
+  autofocus: {
+    type: Boolean,
+    default: false,
+  },
```
scmmishra, line 48: Instead of adding each of these manually as a prop. We could use disable fall through attributes and add them to the input directly. See: https://vuejs.org/guide/components/attrs#disabling-attribute-inheritance

That way, all the events will also get managed automatically

app/javascript/dashboard/components-next/input/Input.vue
```diff
@@ -85,15 +120,25 @@ const handleInput = event => {
...
       :id="id"
+      ref="inputRef"
       :value="modelValue"
-      :class="[customInputClass, inputBorderClass]"
+      :class="[
+        customInputClass,
+        inputBorderClass,
+        {
+          error: messageType === 'error',
+          focus: isFocused,
```
scmmishra, line 130: Why do we need these classes here?

scmmishra (changes requested):

scmmishra (approved):
Most of my suggestions will make this PR a bit big, so lets do them separately for now lets merge this, there are no glaring problems here.

app/javascript/dashboard/components-next/input/Input.vue
iamsivin: There is some case where we don't need to show the border, but we need to show focus or error border state. That's the reason.

iamsivin:
> Most of my suggestions will make this PR a bit big, so lets do them separately for now lets merge this, there are no glaring problems here.

@scmmishra Updated the PR with all suggestions.

PR #10461, opened 2024-11-20
fix: avoid Slack file upload API for fallback messages

sojan-official:
Skip calling the Slack file upload API for message types such as fallback (e.g., Facebook and location messages) that lack actual file data in attachments. This prevents unnecessary API calls and resolves a Sentry error currently occurring in production.

fixes: https://github.com/chatwoot/chatwoot/issues/10460

### DIFF
--- a/lib/integrations/slack/send_on_slack_service.rb
+++ b/lib/integrations/slack/send_on_slack_service.rb
@@ -117,7 +117,7 @@ class Integrations::Slack::SendOnSlackService < Base::SendOnChannelService
   end
 
   def upload_file
-    return if message.attachments.first.file.blank?
+    return unless message.attachments.first.with_attached_file?
 
     result = slack_client.files_upload({
       channels: hook.reference_id,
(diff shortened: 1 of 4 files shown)


lib/integrations/slack/send_on_slack_service.rb
```diff
@@ -117,6 +117,8 @@ def post_message
   end
 
   def upload_file
+    return if message.attachments.first.file.blank?
```
pranavrajs, line 120: I think the issue is that the attachment is a contact, location, share, or fallback, and there’s no actual file to attach.

Instead of checking for the presence of a file, should we just check for the type of attachment?

Here’s an example of what I mean.

```rb
def valid_slack_attachment?(attachment)
  [:image, :audio, :video, :file].includes(attachment.file_type)
end

return unless valid_slack_attachment?(message.attachments.first)
```

pranavrajs (commented):
A minor comment on the fix. Please take a look at the individual comment.

pranavrajs (approved):

PR #10464, opened 2024-11-20
chore: Disable throwing error for malformed to address

sojan-official:
We don't need to raise error on sentry for malformed to address as it is already logged.

ref: https://linear.app/chatwoot/issue/CW-3151/standarderror-invalid-email-to-address-header-standarderror


vishnu-narayanan (approved):
LGTM

PR #10473, opened 2024-11-21
feat: attributify button

scmmishra:
This PR allows attributify for `variant`, `size` and `color` props. This allows using shorthands, instant of writing full props.

We also added a small computed method to ensure these does not show up in the DOM and pollute it


app/javascript/dashboard/components-next/button/Button.vue
```diff
@@ -1,46 +1,84 @@
...
+    .forEach(([key, value]) => {
+      standardAttrs[key] = value;
+    });
+
+  return standardAttrs;
+});
+
+const computedVariant = computed(() => {
+  if (props.variant) return props.variant;
+  if (attrs.solid || attrs.solid === '') return 'solid';
```
pranavrajs, line 56:
```suggestion
  // The useAttrs method returns attributes values an empty string (not boolean value as in props).
  if (attrs.solid || attrs.solid === '') return 'solid';
```

pranavrajs (approved):

PR #10488, opened 2024-11-22
refactor: clear apt warning

nguyentranchung:
## Description

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.


vishnu-narayanan:
@nguyentranchung Thank you for the PR

nguyentranchung:
FlashPanel.io integrates chatwoot but I found that if the installation fails, you will not be able to continue the installation, because the script does not support the case of failed installation. I will try to send more contributions

PR #10493, opened 2024-11-25
feat: add filter inputs

scmmishra:
This PR adds three components along with stories

1. MultiSelect - This is used for filter values, allowing multiple values and folding of values where there are too many items
2. SingleSelect - This is used for filter values, allows selecting and toggling a single item
3. FilterSelect - This is used for operators and others, it allows icons and labels as well as toggling them using props. The v-model for this binds just the final value unlike the previous two components with bind the entire object.

### DIFF
--- a/app/javascript/dashboard/i18n/locale/en/components.json
+++ b/app/javascript/dashboard/i18n/locale/en/components.json
@@ -5,8 +5,10 @@
   },
   "COMBOBOX": {
     "PLACEHOLDER": "Select an option...",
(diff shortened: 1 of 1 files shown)


pranavrajs:
@scmmishra I have mixed feelings about the implementation of the clicks. For instance, I would expect the single select to close automatically when I select an option. Currently, I select an option and click outside the menu to close it, which seems unnecessary. I’ll merge the changes as the suggestions are minor and can be added in the subsequent PRs.

PR #10496, opened 2024-11-25
feat: Add conditions row component

scmmishra:
Screenshot 2024-11-25 at 4 29 34 PM


pranavrajs:
@scmmishra FYI, earlier, the attributeKey was set as the key for the ConditionRow component in the story. This caused an issue with the update cycle. The filter model didn’t remove and re-insert the values, but the component was recreated whenever the attributeKey changed. 

The problem was that the filterOperator and values didn’t get updated when the attributeKey changed, resulting in incompatible values.

A quick example of this issue was when you switched the attribute from Status to Created At. This would result in a state like the one below. You can see operator is missing value. The reason for this was that the status had an equal_to operator, while the created_at attribute didn’t have it in the defined story. This led to an ambiguous state. 

Screenshot 2024-11-25 at 4 22 12 PM

I’ve updated the code to avoid using the attributeKey as the key for the component. Instead, I’ve reset the `operator` and `values` attributes based on the selected attribute and operator.

pranavrajs (approved):

PR #10498, opened 2024-11-26
feat: Contact header components

iamsivin:
# Pull Request Template

## Description

**NB:** `Clickaway` is not working in story

**Screenshots**
image

image

image


app/javascript/dashboard/components-next/Contacts/ContactsHeader/ContactHeader.vue
```diff
@@ -0,0 +1,106 @@
+<script setup>
+import Button from 'dashboard/components-next/button/Button.vue';
+import Input from 'dashboard/components-next/input/Input.vue';
+import Icon from 'dashboard/components-next/icon/Icon.vue';
+import ContactSortMenu from './components/ContactSortMenu.vue';
+import ContactMoreActions from './components/ContactMoreActions.vue';
+
+defineProps({
+  isEmptyState: {
```
pranavrajs, line 9: Let’s not combine the logic for the list and the view here. Instead, let’s introduce a generic `showSearch` flag and handle the logic on whether to show the search on the parent component.

app/javascript/dashboard/components-next/Contacts/ContactsHeader/ContactHeader.vue
```diff
@@ -0,0 +1,106 @@
...
+            size="sm"
+            variant="ghost"
+            @click="emit('filter')"
+          />
+          <ContactSortMenu
+            v-if="!isEmptyState"
+            :active-sort="activeSort"
+            :active-ordering="activeOrdering"
+            @update:sort="emit('update:sort', $event)"
+          />
```
pranavrajs, line 94: It wouldn't hurt even if we show the sort and filter when there are no contacts. Let us skip the logic here.

pranavrajs (approved):

PR #10501, opened 2024-11-26
feat(v4): Contacts list redesign

iamsivin:
# Pull Request Template

## Description

**Screenshots**
image

image


pranavrajs:
**_Listing down the issues I've seen during testing the update:_** 

- The sorting feature doesn’t reset the page to the first pages, which creates some confusion. I thought I might missed some of the contacts from the fold.
- The absence of the country on the list may not be favorable for many customers.
- If there’s an error while updating the form, we don’t display it to the user (no alerts are shown).
- Identifier is missing.
- Country code sorting is returning random results..
- The error message is displayed on the load which seems like a bad UX
image
- The email/phone number fields overflow in contact card.
- The form in accordion resets to the original value while editing the contact.

iamsivin:
@pranavrajs 

- [ ] The sorting feature doesn’t reset the page to the first pages, which creates some confusion. I thought I might missed some of the contacts from the fold.
- [ ] The absence of the country on the list may not be favorable for many customers.
- [ ] If there’s an error while updating the form, we don’t display it to the user (no alerts are shown).
- [ ] Identifier is missing.
- [ ] Country code sorting is returning random results..
- [ ] The error message is displayed on the load which seems like a bad UX
- [ ] The email/phone number fields overflow in contact card.
- [ ] The form in accordion resets to the original value while editing the contact.

pranavrajs:
Additional Issues found:
- ~Error messages should be specific and point to the exact issue.~ - https://github.com/chatwoot/chatwoot/pull/10501/commits/17045e36dcb8ebbc89a55416505fd650bdfbd75b
- ~Splitting the name doesn’t work very effectively. (Mary Ann Jane is shown as Mary as firstName and Ann as second).~ - https://github.com/chatwoot/chatwoot/pull/10501/commits/01f834c2350118193aa03335e1cd1153ed66c12a
- The input focus and active states aren’t working correctly. (Try navigating the form using only the keyboard to confirm.) - General feedback on design, take it as a separate PR.

pranavrajs (approved):

PR #10502, opened 2024-11-26
feat: new conversation filters

scmmishra:
This PR adds the new conversation filter component, we need to merge the `feat/new-dropdown` branch before merging this

CleanShot 2024-11-18 at 09 24 56@2x


app/javascript/dashboard/composables/useTransformKeys.js
scmmishra, line 1: @pranavrajs added these functions to be used instead of using the bare functions. It unrefs any refs and returns the raw value. This keeps us open to memoize this function in the future if performance becomes a hurdle.

scmmishra: CC: @iamsivin

app/javascript/dashboard/components-next/filter/ConversationFilter.vue
```diff
@@ -0,0 +1,175 @@
...
+          v-else
+          :key="`filter-${filter.attributeKey}-${index}`"
+          ref="conditionsRef"
+          v-model:attribute-key="filter.attributeKey"
+          v-model:filter-operator="filter.filterOperator"
+          v-model:query-operator="filters[index - 1].queryOperator"
+          v-model:values="filter.values"
+          show-query-operator
+          :filter-types="filterTypes"
+          @remove="removeFilter(index)"
```
pranavrajs, line 147: I believe this would have an impact since we use the filter.attributeKey as the key. More details [here](https://github.com/chatwoot/chatwoot/pull/10496#issuecomment-2499303737). However, I don’t see it affecting our local environment, so I’ll skip the concern for now.

app/javascript/dashboard/components-next/filter/SaveCustomView.vue
```diff
@@ -0,0 +1,130 @@
+<script>
```
pranavrajs, line 1: Use script setup, not a blocker.

pranavrajs (approved):
LGTM

PR #10511, opened 2024-11-27
fix: messages overlap when switching tabs too quickly

JoaoLive100:
# Pull Request Template

## Description

When switching between message categories ('All Conversations', 'Mentions', 'Unattended'), the messages displayed should correspond to the selected category. If I switch categories quickly, messages from the previously selected category sometimes appear in the new one. It seems that messages from one category are being loaded into another due to some delay or overlapping request handling.

An incremental identifier ensures that only the data from the most recent request is processed.

## Type of change

## How Has This Been Tested?

Simulated rapid switching between message categories ('All Conversations', 'Mentions', and 'Unattended') to replicate the conditions where messages from one category were loaded into another. After the code update, performed the same rapid switching to confirm that messages now load correctly and correspond to the selected category without overlap.

## Checklist:

### DIFF
--- a/app/javascript/dashboard/store/modules/conversations/actions.js
+++ b/app/javascript/dashboard/store/modules/conversations/actions.js
@@ -25,6 +25,9 @@ export const hasMessageFailedWithExternalError = pendingMessage => {
   return status === MESSAGE_STATUS.FAILED && externalError !== '';
 };
 
+// variables
+let currentRequestId = 0;
+
 // actions
 const actions = {
   getConversation: async ({ commit }, conversationId) => {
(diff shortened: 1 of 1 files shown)


pranavrajs:
@JoaoLive100 Thanks for the PR. I understand the issue. However, I believe the most suitable solution is to utilize AbortController, as it’s natively available and eliminates the need for managing counters. Let me know what you think.

https://developer.mozilla.org/en-US/docs/Web/API/AbortController

JoaoLive100:
@pranavrajs I made the changes implementing the AbortController. It’s a solution I wasn’t familiar with, but it achieved the expected result in addressing the issue. What do you think of the current implementation? Please proceed with your review.

PR #10518, opened 2024-11-29
feat: throttle meta request for large chat size

scmmishra:
For large accounts with huge volumes of messages, it can be very wasteful to make the meta request so often. It also puts un-necessary load on the DB bombarding it with so many requests. This PR fixes it by throttling the requests to 5 seconds for all users with more than 1000 accessible chats.

### Why not cache this value in the backend?

Well, it's a bit tricky, since a user can have different permissions over inboxes and can see different chats, maintaining a cache for each of them is not effective, besides the requests will reach the server anyway.


scmmishra:
@pranavrajs updated the PR

scmmishra:
@pranavrajs updated the specs

pranavrajs (approved):

PR #10529, opened 2024-12-03
feat: update notification settings

scmmishra:
https://github.com/user-attachments/assets/52ecf3f8-0329-4268-906e-d6102338f4af


app/javascript/dashboard/i18n/locale/en/settings.json
```diff
@@ -63,13 +63,27 @@
...
+          "UNATTENDED": "New conversations needing attention",
+          "NOTME": "Open conversations assigned to everyone else"
+        },
+        "ALERT_COMBINATIONS": {
+          "NONE": "You won't receive any audio alerts",
+          "ASSIGNED": "You'll receive alerts for conversations assigned to you",
+          "UNATTENDED": "You'll receive alerts for any unattended conversations",
+          "NOTME": "You'll receive alerts for conversations assigned to others",
+          "ASSIGNED+UNATTENDED": "You'll receive alerts for your assigned conversations and any unattended ones",
+          "ASSIGNED+NOTME": "You'll receive alerts for conversations assigned to your both yours and others",
```
pranavrajs, line 81: Doesn't Assigned + not me include all conversations?

app/javascript/dashboard/i18n/locale/en/settings.json
```diff
@@ -63,13 +63,27 @@
...
+          "NOTME+UNATTENDED": "You'll receive alerts for unattended conversations and those assigned to others",
+          "ASSIGNED+NOTME+UNATTENDED": "You'll receive alerts for all conversation"
```
pranavrajs, line 83:
```suggestion
          "ASSIGNED+NOTME+UNATTENDED": "You'll receive alerts for all conversations"
```

app/javascript/dashboard/i18n/locale/en/settings.json
scmmishra: It won't include unassigned conversations

pranavrajs:
@scmmishra While testing this PR, I discovered a couple of audio notification-related issues. Although these issues weren’t directly related to this PR, I thought it was better if we had addressed them together. 

1. **Inconsistent Audio Alert Settings Update**: One issue we encountered was that when we updated the audio alert settings without refreshing the page, the audio alerts would work with the previous settings. This PR addresses this issue by adding a call to the initialize method whenever uiSettings is updated.

**2. Audio loading:** Another issue was that if the user hadn’t interacted with the page, we wouldn’t play any audio, which was problematic. None of the alerts were even attempted. Now, we initialize the audio even before the user interacts with the page and try to play it. If sound permission isn’t available, it throws an alert that the user needs to enable sound settings in the browser or interact with the page. This should resolve the confusion.

**3. Visibility of the tabs:**The third issue was that the visibility setting wasn’t working correctly. When I switch tabs and the tab is hidden, the audio notifications function properly. However, if there’s an overlay window, like another window on top of the one I’m working on, the audio alert doesn’t work because document.hidden is false even though the tab is not visible to the user.

Now that I’ve listened to the blur events, if the user switches to a different tab or window, we automatically detect the visibility of the window and update it on our end. 

I hope this update resolves all the audio-related issues that have been reported for a long time. In this PR, I’ve also update the dashboard audio notification helper. Instead of relying on the window.WOOT_STORE object, it now accepts a store parameter which significantly reduces the complexity.

pranavrajs (approved):

scmmishra:
Tested this on staging seems to be working fine

PR #10550, opened 2024-12-06
feat: integrate new bubbles

scmmishra:
To test this, set the `useNextBubble` value to `true` in the localstorage. Here's a quick command to run in the console

```js
localStorage.setItem('useNextBubble', true)
```

```js
localStorage.setItem('useNextBubble', false)
```


scmmishra:
@pranavrajs if we deploy this branch, we will be able to test out the new bubbles, there's still a few small things pending which I'll finish. 

Should we put this behind the feature flag, so that we can merge and test it for ourselves before making it live for everyone?

pranavrajs:
@scmmishra Yes, this makes sense. I’m currently testing this locally, and the build has some issues. I believe that using it ourselves for a couple of days will help us identify and resolve the edge cases.

scmmishra:
> @scmmishra Yes, this makes sense. I’m currently testing this locally, and the build has some issues. I believe that using it ourselves for a couple of days will help us identify and resolve the edge cases.

Awesome, I'll get this PR ready by tonight

app/javascript/dashboard/components/widgets/conversation/MessagesView.vue
```diff
@@ -473,28 +522,46 @@ export default {
         @click="onToggleContactPanel"
       />
     </div>
-    <ul class="conversation-panel">
+    <ul
```
pranavrajs, line 525: Please refrain from modifying any existing logic. Instead, move all elements related to the current change to a new container component. The only change on this component should be a v-if directive for the <ul class=“conversation-panel”> element and on the else case mount the new message bubble container.

pranavrajs (changes requested):
Please check the comments above.

pranavrajs:
@scmmishra Added a marker and increased the zoom to make sure that the streets are seen properly.

Screenshot 2024-12-12 at 7 16 10 PM

app/javascript/dashboard/components/widgets/conversation/MessagesView.vue
scmmishra: I can build a ConversationListNext.vue component and add it before the `ul` and toggle them with `v-if`, how does that sound?

sojan-official (approved):
Test on staging and confirmed that the changes only comes in when we enable the feature flag via local storage. Lets fix the build and get this merged

sojan-official:
We will cover the code climate issues later ! 

cc: @scmmishra

PR #10561, opened 2024-12-09
fix: sidebar collapsed on reload

scmmishra:
When reloading a page, the sidebar item that is actively selected is collapsed by default. This PR fixes it by expanding it on reload


sojan-official:
- verified this to be working in local environment.

sojan-official (approved):

PR #10572, opened 2024-12-11
fix: Update design of the conversation sidebar, chatlist

pranavrajs:
- Increase the sidebar and chatlist width on screens bigger than 2xl
- Fix background of the copilot sidebar
- Increase the breakpoint for hiding chatlist from 1024 to 1200


iamsivin (approved):
LGTM

PR #10608, opened 2024-12-20
fix: bubble UI issues

scmmishra:
This PR has fixes for the following issues

- Inconsistent spacing between meta and text in text bubble
- Activity bubble overflows for longer text (for now I have truncated it, I'll work with @absurdiya on a better solution)
- Ugly lookinh gradient for expand button on email bubble
- Email bubble overflow issues and text rendering issues
- Alignment for error message
- Minute-wise grouping not working
- Link color should not be blue
- Use `gray-3` for bubble background instead of `gray-4`


scmmishra:
@sojan-official these are some fixes for the new bubbles, PTAL

sojan-official (approved):

Issue #10610, opened 2024-12-21
Bug in numbered lists


arthurbarret0:
### Describe the bug

I added a shortcut where I display a step-by-step guide for the user, with each step numbered, for example:

1. Step 1
2. Step 2
3. Step 3

However, when using the shortcut, it is sent like this:

1. Step 1
2. Step 2
3. Step 3
4. 

In other words, it always adds an extra blank value.

### To Reproduce

1. Create a shortcut with the enumerated text
2. Send it in the chat

### Expected behavior

Do not add an extra number

### Environment

Docker

### Cloud Provider

Other [please specify in the description]

### Platform

Browser

### Additional context

I'm using Chatwoot 3.15.0

AdrianoLF:
Hey @arthurbarret0 !

Can you provide more details about the error?
- What did you mean by "shortcut"? Are you referring to canned responses?
- Which inbox is experiencing this issue?

arthurbarret0:
I conducted a test, and the issue occurs regardless of whether it's a canned response or manually typed. I also noticed that the problem only happens when I use the "CTRL + ENTER" shortcut to send the message... If I use the "Send" button, the bug doesn't occur.

See the video: https://streamable.com/jt4tvs

> Hey @arthurbarret0 !
> 
> Can you provide more details about the error?
> 
> * What did you mean by "shortcut"? Are you referring to canned responses?
> * Which inbox is experiencing this issue?

AdrianoLF:
@arthurbarret0 I just made a fix in [this commit](https://github.com/chatwoot/chatwoot/pull/10611/commits/333c5a8bfe40d65e3b9438bd44444981c30e90de)

pranavrajs:
@scmmishra Can you please take a lead on this issue and suggest the next steps? Please check if this is still an issue on v4.

iamsivin:
@arthurbarret0 Thanks for raising this issue. Are you still facing it?

This issue should already be fixed by PR https://github.com/chatwoot/chatwoot/pull/14077, which is live in **v4.14.0** and prevents `Ctrl+Enter` from adding an extra line on send.

If you're still facing this issue, please feel free to reopen it.

PR #10611, opened 2024-12-23
fix: Clean empty option in numbered list message

AdrianoLF:
This PR fixes the following issue:
- https://github.com/chatwoot/chatwoot/issues/10610
(creation of empty options in numbered lists when sending messages with "Ctrl + Enter")

### BEFORE
image
----
### AFTER
image

### DIFF
--- a/app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
+++ b/app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
@@ -654,6 +654,7 @@ export default {
       if (this.isReplyButtonDisabled) {
         return;
       }
+      this.message = this.removeEmptyOptionFromNumberedList(this.message);
       if (!this.showMentions) {
         const isOnWhatsApp =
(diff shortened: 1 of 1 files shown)


AdrianoLF:
I would greatly appreciate it if you could take a look so we can close the mentioned issue. Thanks, friends! @scmmishra @sojan-official

scmmishra:
Hey @AdrianoLF thanks for the PR, however it doesn't seem like a good fix, it cleans up the message on send, which is not a good solution. What if I intentionally wish to write `1.` and send it. 

The solution has to be found on the editor paste function, see why an extra line is being appended which is taken as a list item.

iamsivin:
This issue should already be fixed by PR https://github.com/chatwoot/chatwoot/pull/14077, which is already live in **v4.14.0** and prevents `Ctrl+Enter` from adding an extra line on send.

Thanks for the PR. We’re closing this for now. If you’re still facing the issue, feel free to reopen the PR and I can help with fixing it further.

PR #10629, opened 2024-12-30
feat: lazy load i18n locales

KubesDavid:
Implementing lazy loading for i18n locales, initially loading only the active and default English locales to enhance performance and reduce bandwidth. It also enables dynamic loading of additional locales as needed and adds the missing Chinese (zh) locale files.


pranavrajs:
Sorry, this didn't land to production. There have been significant changes in the widget, dashboard and the rest of the application. We will pick it up internally along with the design update we are working on. 

Thanks for the effort.

PR #10652, opened 2025-01-08
fix: v4 ui issues

scmmishra:
This PR has the following fixes

1. Remove leading `:` in inbox view push message content
2. Styles for block quotes in text
3. Better condition for checking bot message
4. Use dompurify


iamsivin (approved):
LGTM

PR #10654, opened 2025-01-09
fix: conversation footer colors

scmmishra:


iamsivin (approved):
@scmmishra Fixed in the new message screen, but breaks the old one.

image

iamsivin (changes requested):
@scmmishra Fixed in the new message screen, but breaks the old one.

image

scmmishra:
@iamsivin thanks, fixed

iamsivin (approved):

PR #10662, opened 2025-01-09
fix: attachment url in file chip

scmmishra:
https://github.com/user-attachments/assets/e84d94d5-876d-4c00-88be-a460f917bf3b


iamsivin (approved):
LGTM

PR #10665, opened 2025-01-10
feat(ee): Add Captain features

pranavrajs:
This PR imports all the work related to Captain into the EE codebase. Captain represents the AI-based features in Chatwoot and includes the following key components:

- Assistant: An assistant has a persona, the product it would be trained on. At the moment, the data at which it is trained is from websites. Future integrations on Notion documents, PDF etc. This PR enables connecting an assistant to an inbox. The assistant would run the conversation every time before transferring it to an agent. 
- Copilot for Agents: When an agent is supporting a customer, we will be able to offer additional help to lookup some data or fetch information from integrations etc via copilot. 
- Conversation FAQ generator: When a conversation is resolved, the Captain integration would identify questions which were not in the knowledge base.
- CRM memory: Learns from the conversations and identifies important information about the contact.


enterprise/lib/captain/agent.rb
```diff
@@ -79,6 +74,7 @@ def construct_prompt(config)
       - Work diligently until the stated objective is achieved.
       - Utilize only the provided tools for solving the task. Do not make up names of the functions
       - Set 'stop: true' when the objective is complete.
+      - DO NOT privide tool_call as final answer
```
edumaciel10:
```suggestion
      - DO NOT provide tool_call as final answer
```

sojan-official (approved):

PR #10670, opened 2025-01-10
fix: country selection

KubesDavid:
Refactored `ContactsForm` to use country codes for consistency, resolving mismatches like "Czechia" vs. "Czech Republic" from `ContactIpLookupJob` to ensure correct display.


KubesDavid:
May you review it please? @iamsivin

iamsivin:
> May you review it please? @iamsivin

Hey @KubesDavid, could you help me by providing a screen recording or a specific case that demonstrates the issue you're trying to fix? In both cases, we set the country code and name. In the UI, we only set the country, but the country code is set under the hood. Regarding "Czechia" vs. "Czech Republic", they refer to the same country; however, in our case, we only have "Czech Republic" in the dropdown.

KubesDavid:
We use GeoIP2 for IP lookups, and its database specifies the Czech Republic as "Czechia" with the country code CZ. However, the UI expects the country name to be "Czech Republic," which causes the country to appear as unselected when "Czechia" is returned. Manually changing the country to "Czech Republic" resolves the display issue.

iamsivin (approved):
@KubesDavid Thanks for the PR! Approving it as it LGTM.

PR #10675, opened 2025-01-12
fix: Prevent duplicate notifications for mentions

sojan-official:
This PR addresses an issue where mention_notification was being overwritten by assigned_conversation_new_message notifications. Similarly, participating_conversation_new_message was taking priority over assigned_conversation_new_message.

The fix ensures that additional notifications are not created in these scenarios by establishing a priority order for notifications as follows:

1. mention_notification
2. assigned_conversation_new_message
3. participating_conversation_new_message

This change maintains consistency and prevents redundant notifications.

----

Note: In the older implementation of notifications in Chatwoot, notification objects were created only if the user had subscribed to a specific notification type. With the new inbox_view model, we are changing this approach to create notification objects in all cases. This PR updates the services to remove reliance on notification preferences, simplifying the logic.

----

Caveat: Since we don’t create extra notifications for new messages if a mention_notification already exists, a user subscribed to new_message_notification push/email notifications but not to mention_notification will not receive notifications in these cases.

### DIFF
--- a/app/services/messages/new_message_notification_service.rb
+++ b/app/services/messages/new_message_notification_service.rb
@@ -4,32 +4,17 @@ class Messages::NewMessageNotificationService
   def perform
     return unless message.notifiable?
 
-    notify_participating_users
     notify_conversation_assignee
+    notify_participating_users
   end
 
   private
 
   delegate :conversation, :sender, :account, to: :message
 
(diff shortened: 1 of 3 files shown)


pranavrajs (approved):
Code looks good to me.

PR #10691, opened 2025-01-15
fix: Remove Report API calls from being called on every event

pranavrajs:
Previously, the Reports API fetched data based on event triggers. For example, when an event occurred on an account, the system would automatically retrieve and display updated information. However, this approach was designed under the assumption that reports would be accessed by a small number of users and on an infrequent basis (e.g., once daily or weekly).

In scenarios where large customers have multiple team members actively monitoring reports, this event-driven approach led to an excessive number of requests, significantly straining the system.

This PR introduces a interval-based fetching of reports instead of the event-driven model.


sojan-official (approved):
LGTM

PR #10693, opened 2025-01-16
feat: Add a review step for FAQs generated from conversations before using it

pranavrajs:
This PR introduces a review step for generated FAQs, allowing a human to validate and approve them before use in customer interactions. While hallucinations are minimal, this step ensures accurate and reliable FAQs for Captain to use during LLM calls when responding to customers.

- Added a status field for the FAQ
- Allow the filter on the UI.
Screenshot 2025-01-15 at 6 39 26 PM


sojan-official (approved):
Tested in local. 
LGTM

PR #10695, opened 2025-01-16
fix: sentry issues

scmmishra:
1. Ensure audio player ref is accessible before triggering calls ([Sentry](https://chatwoot-p3.sentry.io/issues/6221981610))
2. Use correct default for attachments, this was incorrectly set to `null` in a previous PR ([Sentry](https://chatwoot-p3.sentry.io/issues/5966738120))
3. Fix `lastNonActivityMessage` is not present ([Sentry](https://chatwoot-p3.sentry.io/issues/6116038455))
4. Fix `Alt+J` & `Alt+K` shortcuts not working ([Sentry](https://chatwoot-p3.sentry.io/issues/6075125384))


iamsivin (approved):
LGTM

PR #10699, opened 2025-01-16
feat: Add support for the references in FAQs

pranavrajs:
Currently, it’s unclear whether an FAQ item is generated from a document, derived from a conversation, or added manually.

This PR resolves the issue by providing visibility into the source of each FAQ. Users can now see whether an FAQ was generated or manually added and, if applicable, by whom.

- Move the document_id to a polymorphic relation (documentable).
- Updated the APIs to accommodate the change.
- Update the service to add corresponding references. 
- Updated the specs.

Screenshot 2025-01-15 at 11 27 56 PM


sojan-official (approved):
Tested and verified in local

PR #10725, opened 2025-01-20
fix: Prevent template variables from becoming links

iamsivin:
# Pull Request Template

## Description

**Issue**
This PR fixes template variables in messages (e.g., {{customer.name}}) that were being incorrectly converted to clickable links by the `MessageFormatter's linkify` functionality. This caused formatting issues and broken links.

**Solution**
Added a `linkify` parameter to `MessageFormatter` to optionally disable link conversion

## Type of change

## How Has This Been Tested?

**Screenshots**

**Before**
image

**After**
image

## Checklist:


app/javascript/shared/composables/useMessageFormatter.js
```diff
@@ -14,11 +14,12 @@ export const useMessageFormatter = () => {
    * @param {boolean} isAPrivateNote - Whether the message is a private note.
    * @returns {string} - The formatted message.
    */
-  const formatMessage = (message, isATweet, isAPrivateNote) => {
+  const formatMessage = (message, isATweet, isAPrivateNote, linkify) => {
```
scmmishra, line 17: Adding so many boolean params is not that great... We could convert the function signature to something like this  

```js
const formatMessage = (message, options) => {
   const defaultOptions = { isATweet: false, isAPrivateNote: false, linkify: true }
   // implementation
}
```

scmmishra (changes requested):
Hey @iamsivin just a tiny feedback.

app/javascript/shared/composables/useMessageFormatter.js
iamsivin: Resolving this comment as discussed in the call, as we are not doing it now. The reason is that we need to update it almost everywhere. If we add more parameters, we need to update this. Added a TODO comment in the code as a reminder.

scmmishra (approved):

PR #10766, opened 2025-01-26
feat(v4): Update the report pages to show aggregate values

pranavrajs:
This PR updates the report pages for agents, inboxes, and teams by replacing charts with aggregate values (under a feature flag). Users can click on any item to view more details if needed. Most users seem to prefer aggregate values, so this change will likely stay. 

The PR also includes a few fixes:

- The summary reports now use the same logic for both the front-end and CSV exports.
- Fixed an issue where a single quote was being added to values with hyphens in CSV files. Now, ‘n/a’ is used when no value is available.
- Fixed a bug where the average value was calculated incorrectly when multiple accounts were present.

These changes should make reports easier to use and more consistent.

### Agents:

Screenshot 2025-01-26 at 10 47 18 AM

### Inboxes
Screenshot 2025-01-26 at 10 47 10 AM

### Teams: 
Screenshot 2025-01-26 at 10 47 01 AM


app/javascript/dashboard/components-next/sidebar/Sidebar.vue
```diff
@@ -77,6 +90,56 @@ const sortedInboxes = computed(() =>
...
+    to: accountScopedRoute('inbox_reports_index'),
+  },
+  {
+    name: 'Reports Team',
+    label: t('SIDEBAR.REPORTS_TEAM'),
+    to: accountScopedRoute('team_reports_index'),
+  },
+];
+
+const oldReportRoutes = [
```
pranavrajs, line 116: Note: This is a temporary version. Reports will switch to the new version once it has been tested at scale.

scmmishra (approved):
Overall LGTM! I updated this with some minor design changes and fixed a bug where reloading on the Agent page would break the app.

sojan-official (approved):

PR #10786, opened 2025-01-29
feat: hide empty folders from sidebar

scmmishra:
This PR updates the sidebar to hide empty folders. So, in case the user has not created any folders, or teams, the section in the sidebar will be hidden


scmmishra:
Break it into two PRs CC: @pranavrajs

iamsivin (approved):
LGTM

PR #10788, opened 2025-01-29
fix: UX issues on Help Center editor

scmmishra:
This PR targets the following issues

### Cursor shifting sporadically on save.

Earlier the `FullEditor` component had a watch on `modelValue`, this was because the article body was not immediately available when the component was initialised, so without the watch, the editor would not get updated and we would see an empty page. 

There was a check to ensure this was only triggered when there's a difference in the content which makes sense. Earlier the `modelValue` is empty, and later it's populated. However when editing the article, there's a case where you hold typing for a while and the debounced saved triggers, with the old content. You continue typing however, so when the response arrives, `modelValue` gets updated, a diff is found, and the editor state reloads. 

This PR fixes it by add a `once` to the watch, effectively terminating the two-way binding with the editor. 

> Note: We need to ensure that the article state is fully saved when the user stops typing or navigates away from the page.

### Title editor not working correctly

_PENDING_

### Initialising a new article is complicated

_PENDING_

### DIFF
--- a/app/javascript/dashboard/components-next/HelpCenter/Pages/ArticleEditorPage/ArticleEditor.vue
+++ b/app/javascript/dashboard/components-next/HelpCenter/Pages/ArticleEditorPage/ArticleEditor.vue
@@ -27,7 +27,6 @@ const props = defineProps({
 
 const emit = defineEmits([
   'saveArticle',
-  'saveArticleAsync',
   'goBack',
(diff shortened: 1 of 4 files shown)


scmmishra:
@codex please review

pranavrajs:
Closing it here, we will re-evaluate this along with the design update.

PR #10797, opened 2025-01-30
feat: use feature flags across the routes

scmmishra:


scmmishra:
@pranavrajs you mentioned some unique case for custom roles visibility. Can you give more context

pranavrajs (approved):

Issue #10477, opened 2024-11-21
CC:ed email addresses from other conversations appear when switching to a different one


Robinjoeh:
### Describe the bug

Email addresses (CC:ed) that belong to a different conversation, will remain in the CC field when you switch to a different conversation.

### To Reproduce

1. Use the email channel
2. Make sure to have multiple conversations (one of them must have CC:ed contacts)
3. Switch to a conversation that does not have any CC contacts filled in

When you then switch back to the previous conversation, and then back to the one that incorrectly had CC:ed emails added, they are removed. I suspect there is a background-loading of contacts that takes place from the wrong conversation - so that the CC:ed field is first reset, then incorrectly populated (as CC:ed emails also appear with a very very slight delay).

### Expected behavior

CC:ed emails should always be reset to empty if there are no CC:ed emails.

### Environment

Linux VM

### Cloud Provider

None

### Platform

Browser

### Browser and version

Chrome 131.0.6778.85

Robinjoeh:
I have found that the issue is casued by this `this.lastEmail` is null at the time setCCAndToEmailsFromLastChat() is fired. This only applies under certain conditions, perhaps when there are multiple emails and it takes a few extra milliseconds to load the messages. This causes the CC/BCC contacts to be left from the previous conversation (if the previous conversation you looked at had CC/BCC emails).

https://github.com/chatwoot/chatwoot/blob/a50e4f17480edc47ce6296267164b83b12b7499f/app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue#L984

bfontaine:
Is there any suggested fix for this issue? We have it as well and it’s very annoying.

PR #10799, opened 2025-01-30
fix: inconsistent reply box cc update

scmmishra:
This PR target two issues

### CC & BCC not updated correctly

When moving from one conversation to another, the store may not have the list of all the messages. A fetch is subsequently made to get the messages. However, this update does not trigger the `currentChat` watcher. This PR fixes it by adding a new watcher on `currentChat.messages`.

We also update the `setCCAndToEmailsFromLastChat` method to reset the `cc`, `bcc` and `to` fields if the last email is not found. This ensures that the data is not carried forward from a previous email

Fixes: https://github.com/chatwoot/chatwoot/issues/10477

### To address are not added correctly to the `CC`

If the `to` address of a previous email has multiple recipient, there was no case to add them to the CC. 

Fixes: https://github.com/chatwoot/chatwoot/issues/8925

---

Depends on: https://github.com/chatwoot/utils/pull/41


scmmishra:
Hey @Robinjoeh that was a good catch, thanks. I've updated the PR with the conditions handled, please take a look.

CC: @muhsin-k

Robinjoeh:
@scmmishra Great, will re-test it and see if I spot anything else and let you know.

scmmishra:
Hey @Robinjoeh thanks for the testing, seems like the refactoring of the cc.filter caused this regression, it's handled now

Robinjoeh:
@scmmishra I will re-test now and get back to you within a few mins!

scmmishra:
@Robinjoeh about the 2nd one. I am still trying to figure that out from a product perspective, if we do need to include that or not. At the moment we just take the emails into check if it had a `from` address. It broadly targets incoming emails, or emails generated out of Chatwoot.

If we solely rely on the previous message without the `from` check, we may miss out on contacts present otherwise in the thread.

@pranavrajs what do you think we should do here?

Robinjoeh:
@scmmishra I know that, and I agree with you, it is a product decision. However, most email clients, including Gmail and Outlook, they use information from the latest email regardless if that email was sent/received. Don't forget that before the agent replies, the email from the customer would already have prefilled the CC:ed contacts in the reply box, so the agent message should technically in this case always have all CC:ed emails as well (unless the agent modifies the list of CC:ed emails).

scmmishra:
Moved the function to `chatwoot/utils` here. https://github.com/chatwoot/utils/pull/41

scmmishra:
Hey @Robinjoeh, we discussed this internally. We are going to update the behaviour to use the last incoming or outgoing email for now. This should be consistent with the behaviour you are look for.

Robinjoeh:
@scmmishra Actually, I am not looking for a specific behaviour. I am fine either way. I was just thinking purely from a product perspective that would likely be better (and also more fool/error proof) compared to just taking last received message into account. I think that is the right call!

scmmishra:
> @scmmishra Actually, I am not looking for a specific behaviour. I am fine either way. I was just thinking purely from a product perspective that would likely be better (and also more fool/error proof) compared to just taking last received message into account. I think that is the right call!

Yep we had the same thought, thanks for the suggestions, appreciate it :)

Robinjoeh:
@scmmishra Do you have any other commits pending on this btw? If not, let me know and I will test it.

Robinjoeh:
@scmmishra Thanks, results below:

Tested the latest commits (also updated `chatwoot/utils`).

Works well when the customer sent the latest message, however when the agent sent the latest message, To/Cc does not get populated at all.

Moreover, if a note is made after the customer has messaged, it seems there are no To/Cc filled in. I suspect perhaps the note is treated as a message - although I did not review the code, so don't take that as a final conclusion.

scmmishra:
Hey @Robinjoeh good catch, fixed them. You can try it out

Robinjoeh:
@scmmishra Looks better, but the CC is not populated the same when the message is sent from the agent, as it is when the customer has sent the latest message in a conversation.

1. Conversation starter email is sometimes placed in "cc", whereas on conversations when customer sent the last message, the email is correctly placed in the "to" field. "to" is entirely missing when the last reply was from an agent, and it just puts the reply in cc instead.
2. Sometimes, it also seems to place the reply email address in both "to" and "cc" (at the same time). See screenshot below for an example.
image

scmmishra:
Hey @Robinjoeh I've updated the utils for this https://github.com/chatwoot/utils/pull/44
I'll release it after a quick review. It fixes the issue mentioned. Also contains comprehensive specs to catch any regressions in the future

Robinjoeh:
@scmmishra I can test the PR before you merge it, give me a few mintues.

scmmishra:
> @scmmishra Looks perfect from my end. Great job with the test specs also. 🎉

thanks for testing it :)

@muhsin-k lets merge the utils PR, I'll draft a release

scmmishra:
@muhsin-k PTAL

muhsin-k (approved):

Issue #10807, opened 2025-01-31
bug: Invalid sender name for email replies


Robinjoeh:
### Describe the bug

It appears the sender name ("from" header in the email) is fixed to the person who started the original conversation. In email conversations, other people can be CC:ed, and if they reply, the name of the original conversation starter is used in their reply too. This causes agents to greet people with the wrong name.

![Image](https://github.com/user-attachments/assets/53a62c04-f6d2-4ce7-9671-159d81bb50bf)

### To Reproduce

1. Start a new conversation in the email channel, add at least 1 CC.
2. Reply from the CC:ed email address to the Chatwoot conversation.

### Expected behavior

We should display the actual sender name of the email, and in case the "from" header is missing the sender name, then just the email address of the sender should be displayed.

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

None

Robinjoeh:
@pranavrajs Another issue I found whilst testing the email channel.

pranavrajs:
@Robinjoeh I can confirm this bug. Thanks for reporting. We will have a fix soon.

PR #10837, opened 2025-02-04
fix: incorrect sender name on email meta

scmmishra:
Fixes https://github.com/chatwoot/chatwoot/issues/10807


Robinjoeh:
I have tested this as well, looks great 👍

muhsin-k (approved):

PR #10851, opened 2025-02-06
feat: Linear OAuth 2.0 

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-3417/oauth-20-authentication
We are planning to publish the Chatwoot app in the Linear [integration list](https://linear.app/docs/integration-directory). While we currently use token-based authentication, Linear recommends OAuth2 authentication. This PR implements OAuth2 support.


app/helpers/integration_helper.rb
```diff
@@ -0,0 +1,48 @@
+module IntegrationHelper
```
sojan-official, line 1: lets name space tis under Linear::IntegrationHelper

app/helpers/linear/integration_helper.rb
```diff
@@ -0,0 +1,48 @@
...
+    nil
+  end
+
+  # Verifies and decodes a Linear JWT token
+  #
+  # @param token [String] The JWT token to verify
+  # @return [Integer, nil] The account ID from the token or nil if invalid
+  def verify_linear_token(token)
+    return nil if token.blank?
+    return nil unless (client_secret = fetch_linear_secret)
```
scmmishra, line 27: Is this supposed to be a check or an assignment?

app/helpers/linear/integration_helper.rb
scmmishra, line 1: Generally this file can be more Ruby like. Here's what you can try

<details>

<summary>Suggestion</summary>

```rb
module Linear::IntegrationHelper
  # Generates a signed JWT token for Linear integration
  #
  # @param account_id [Integer] The account ID to encode in the token
  # @return [String, nil] The encoded JWT token or nil if client secret is missing
  def generate_linear_token(account_id)
    return if client_secret.blank?

    JWT.encode(token_payload(account_id), client_secret, 'HS256')
  rescue StandardError => e
    Rails.logger.error("Failed to generate Linear token: #{e.message}")
    nil
  end

  def token_payload(account_id)
    {
      sub: account_id,
      iat: Time.current.to_i
    }
  end

  # Verifies and decodes a Linear JWT token
  #
  # @param token [String] The JWT token to verify
  # @return [Integer, nil] The account ID from the token or nil if invalid
  def verify_linear_token(token)
    return if token.blank? || client_secret.blank?

    decode_token(token, client_secret)
  end

  private

  def client_secret
    @client_secret ||= ENV.fetch('LINEAR_CLIENT_SECRET', nil)
  end

  def decode_token(token, secret)
    JWT.decode(token, secret, true, {
                 algorithm: 'HS256',
                 verify_expiration: true
               }).first['sub']
  rescue StandardError => e
    Rails.logger.error("Unexpected error verifying Linear token: #{e.message}")
    nil
  end
end
```

</details>

app/javascript/dashboard/routes/dashboard/settings/integrations/Linear.vue
```diff
@@ -0,0 +1,61 @@
...
+
+const store = useStore();
+
+const integrationLoaded = ref(false);
+
+const integration = computed(() =>
+  store.getters['integrations/getIntegration']('linear')
+);
+
+const uiFlags = computed(() => store.getters['integrations/getUIFlags']);
```
scmmishra, line 15: Use `useMapGetter` here

app/javascript/dashboard/routes/dashboard/settings/integrations/Linear.vue
```diff
@@ -0,0 +1,61 @@
...
+import Integration from './Integration.vue';
+import Spinner from 'shared/components/Spinner.vue';
+
+const store = useStore();
+
+const integrationLoaded = ref(false);
+
+const integration = computed(() =>
+  store.getters['integrations/getIntegration']('linear')
+);
```
scmmishra, line 13: You can use `useFunctionGetter` here

app/javascript/dashboard/routes/dashboard/settings/integrations/Linear.vue
```diff
@@ -0,0 +1,61 @@
...
+  }
+  return integration.value.action;
+});
+
+const intializeLinearIntegration = async () => {
+  await store.dispatch('integrations/get', 'linear');
+  integrationLoaded.value = true;
+};
+
+intializeLinearIntegration();
```
scmmishra, line 29: This should be done in the `onMounted` callback

app/models/integrations/app.rb
```diff
@@ -25,10 +26,21 @@ def fields
     params[:fields]
   end
 
+  # There is not way to get the account_id from the linear callback
+  # so we are using the generate_linear_token method to generate a token and encode it in the state parameter
```
scmmishra, line 30:
```suggestion
  # There is no way to get the account_id from the linear callback
  # so we are using the generate_linear_token method to generate a token and encode it in the state parameter
```

app/models/integrations/app.rb
```diff
@@ -25,10 +26,21 @@ def fields
...
+
   def action
     case params[:id]
     when 'slack'
       "#{params[:action]}&client_id=#{ENV.fetch('SLACK_CLIENT_ID', nil)}&redirect_uri=#{self.class.slack_integration_url}"
+    when 'linear'
+      base_url = "#{params[:action]}?response_type=code"
+      "#{base_url}&client_id=#{ENV.fetch('LINEAR_CLIENT_ID', nil)}" \
+        "&redirect_uri=#{self.class.linear_integration_url}" \
+        "&state=#{encode_state}&scope=read,write&prompt=consent"
```
scmmishra, line 43: Let's move this to a separate fn called `build_linear_action`. We can do something like the following

```rb
def build_linear_action
    [
      "#{params[:action]}?response_type=code",
      "client_id=#{ENV['LINEAR_CLIENT_ID']}",
      "redirect_uri=#{self.class.linear_integration_url}",
      "state=#{encode_state}",
      'scope=read,write',
      'prompt=consent'
    ].join('&')
  end
  ```

File: app/models/integrations/app.rb [hunk 8]
```diff
@@ -64,6 +76,10 @@ def self.slack_integration_url
     "#{ENV.fetch('FRONTEND_URL', nil)}/app/accounts/#{Current.account.id}/settings/integrations/slack"
   end
 
+  def self.linear_integration_url
+    "#{ENV.fetch('FRONTEND_URL', nil)}/linear/callback"
```
scmmishra, line 80: Slack has the account id in the callback url. Is it not possible for linear? That we can get rid of the `encode_state` part

config/integration/apps.yml
```diff
@@ -59,26 +59,9 @@ linear:
   id: linear
   logo: linear.png
   i18n_key: linear
-  action: /linear
+  action: https://linear.app/oauth/authorize
```
scmmishra, line 62: Not sure if hardcoding this is a great idea. People on the `enterprise`  plan with linear might have a different domain. Anyway this is not something that might hit us now. So if it requires significant changes. We can ignore this feedback for now

CC: @sojan-official

scmmishra (changes requested):

app/models/integrations/app.rb
muhsin-k: Linear does not allow passing parameters in the redirect URL, so the redirect URL and the URL configured in the Linear app should be the same.

config/integration/apps.yml
muhsin-k: Makes sense. I'll add a comment there and revisit later.

app/javascript/dashboard/routes/dashboard/settings/integrations/Linear.vue
scmmishra: @muhsin-k seems like you missed this one

scmmishra (approved):

PR #10853, opened 2025-02-07
fix: Discard older updates in real-time events based on the record updated time

pranavrajs:
image

The team and agent assignment events are delivered to the UI out of order, potentially causing inconsistent data. Since there’s no validation to check for outdated information, we can mitigate this by including an updated_at parameter in the conversation payload. Before updating the store, the UI should verify that the event payload has the latest updated_at timestamp, ensuring data consistency.


pranavrajs:
Closing as this is already done via https://github.com/chatwoot/chatwoot/pull/10876

PR #10865, opened 2025-02-10
feat: update vue-letter and allow transform css

scmmishra:
Emails from iOS and MacOS clients can often come with size transformation. This results in a broken layout when sanitizing with Vue letter.

This PR updates the vue-letter to include the [recent changes made there](https://github.com/mat-sz/vue-letter/pull/2). And adds the `transform` and `transform-origin` properties to the allow list

Ref: https://www.caniemail.com/features/css-transform/


pranavrajs (approved):
Looks good to me. Please test the case I've reported thoroughly.

PR #10867, opened 2025-02-10
feat: handle mine event for incoming messages

scmmishra:
Handle `mine` condition, missed in https://github.com/chatwoot/chatwoot/pull/10529


iamsivin (approved):
LGTM

Issue #10578, opened 2024-12-12
cant add new agents to an inbox in the v4 beta


LucaPeters:
### Describe the bug

You can add new Agents in the frontend without any errors, but the agents cant see the new Inbox. 
So i disabled the v4 beta and try it again and everything works out fine.

### To Reproduce

1. activate the v4 Beta for the website 
2. go to inbox settings
3. add new agents 
4. check if they have access to the new inbox
5. deactivate the v4 Beta 
6. retry step 2-4 with the old design

### Expected behavior

When i add new Agents to an Inbox on the website they should have access afterwards or 
I should get an error on the website, when there is a problem with adding new agents to an new inbox.

### Environment

Kubernetes

### Cloud Provider

None

### Platform

Browser

### Operating system

macbook air 2020

### Browser and version

Arc Version 1.72.1 (56511), Chromium Engine Version 131.0.6778.109

### Docker (if applicable)

docker image: chatwoot/chatwoot:v3.15.0

pranavrajs:
@scmmishra Please take a look at this.

scmmishra:
Hey @LucaPeters are you still facing this issue? This could be related to the frontend indexedDB cache. You can go to the super admin and reset this cache for now. Meanwhile I'll work on a fix

To access the reset action, navigate to your account on SuperAdmin `https://<your-chatwoot>/super_admin/accounts/<account-id>`

Here at the end of the page you will find a button that says "Reset Frontend Cache", click that and see if it works for you?

LucaPeters:
Works for me now

PR #10869, opened 2025-02-10
feat: invalidate cache after inbox members or team members update

scmmishra:
At the moment, when updating the inbox members, or team members the account cache used for IndexedDB is not invalidated. This can cause inconsistencies in the UI. This PR fixes this by adding explicit invalidation after performing the member changes

### Summary of changes

1. Added a new method `add_members` and `remove_members` to both `team` and `inbox` models. The change was necessary for two reasons
    - Since the individual `add_member` and `remove_member` is called in a loop, it's wasteful to run the cache invalidation in the method.
    - Moving the account cache invalidation call in the controller pollutes the controller business logic
2. Updated tests across the board

### More improvements

We can make a concern called `Memberable` with usage like `memberable_with :inbox_members`, that can encapsulate the functionality

---

Related: https://github.com/chatwoot/chatwoot/issues/10578


app/models/inbox.rb
```diff
@@ -82,14 +82,24 @@ class Inbox < ApplicationRecord
...
 
-  def remove_member(user_id)
-    member = inbox_members.find_by!(user_id: user_id)
-    member.try(:destroy)
+  # Removes multiple members from the inbox
+  # @param user_ids [Array<Integer>] Array of user IDs to remove
+  # @return [void]
+  def remove_members(user_ids)
+    user_ids.each do |user_id|
+      remove_member(user_id)
```
pranavrajs, line 100: Can we do bulk creation instead of iterating over each ID?

scmmishra: Yeah, makes more sense. I'll update it

sojan-official (approved):

PR #10870, opened 2025-02-10
feat: add `team.changed` event handler

scmmishra:
Follow-up fixes for realtime issues with conversation list

Reference 
- https://github.com/chatwoot/chatwoot/pull/10826
- https://github.com/chatwoot/chatwoot/pull/10853/


pranavrajs:
Closing this as we went in a different direction to fix the issue.

PR #10885, opened 2025-02-12
feat: mask email information in the API response

scmmishra:


app/javascript/dashboard/routes/dashboard/settings/inbox/ImapSettings.vue
```diff
@@ -73,11 +74,16 @@ export default {
...
-            imap_password: this.password,
             imap_enable_ssl: this.isSSLEnabled,
           },
         };
 
+        // If the password is a placeholder, don't send it to the server
+        const isPlaceholderPassword = /^\*+$/.test(this.password);
+        if (!isPlaceholderPassword) {
+          payload.channel.imap_password = this.password;
+        }
```
scmmishra, line 85: @pranavrajs @sojan-official at the moment this is a naive check to ensure that we don't accidentally replace the password. Does this work or should I think of something better?

PR #10888, opened 2025-02-12
feat: hide tokens and password on contact inbox payloads

scmmishra:


app/views/api/v1/accounts/contacts/contact_inboxes/create.json.jbuilder
```diff
@@ -1 +1 @@
-json.partial! 'api/v1/models/contact_inbox', formats: [:json], resource: @contact_inbox
+json.partial! 'api/v1/models/contact_inbox', formats: [:json], resource: @contact_inbox, hide_channel_credentials: true
```
sojan-official, line 1: we won't need this here right ? we only need to add this where we including the inbox partial

app/views/api/v1/models/_contact.json.jbuilder
```diff
@@ -13,7 +13,7 @@ json.created_at resource.created_at.to_i if resource[:created_at].present?
 if defined?(with_contact_inboxes) && with_contact_inboxes.present?
   json.contact_inboxes do
     json.array! resource.contact_inboxes do |contact_inbox|
-      json.partial! 'api/v1/models/contact_inbox', formats: [:json], resource: contact_inbox
+      json.partial! 'api/v1/models/contact_inbox', formats: [:json], resource: contact_inbox, hide_channel_credentials: true
```
sojan-official, line 16: we won't need this here right ? we only need to add this where we including the inbox partial

app/views/api/v1/models/_inbox.json.jbuilder
```diff
@@ -67,7 +69,7 @@ if resource.email?
   ## IMAP
   if Current.account_user&.administrator?
```
sojan-official, line 70: lets hide the whole blocks that we hide for administrator, not just the password.  ideally we only need to show the inbox attributes and probably the channel type..  or we can simplify this an create a thin model template for inbox which only exposes the necessary attributes

scmmishra:
Let's just create a new partial and use it as required. CC: @pranavrajs

app/views/api/v1/models/_inbox_slim.json.jbuilder
```diff
@@ -0,0 +1,6 @@
+json.id resource.id
+json.avatar_url resource.try(:avatar_url)
+json.channel_id resource.channel_id
+json.name resource.name
+json.channel_type resource.channel_type
+json.email resource.channel.try(:email) if resource.email?
```
scmmishra, line 6: We need this information on the frontend

scmmishra:
@sojan-official PTAL

sojan-official (approved):

PR #10925, opened 2025-02-18
chore: Improvement is keyboard shortcuts

iamsivin:
# Pull Request Template

## Description

This PR fixes the issue where the `Alt/Option + A` file upload shortcut was interfering with adding special characters. The shortcut has been updated to `CMD/Win + Alt/Option + A` to resolve the conflict.

## Type of change

## Checklist:


pranavrajs:
@scmmishra Take a look and please suggest the best path forward.

scmmishra (approved):
Overall this looks good, even though we are overriding a browser default here. 

---

Adding a note for posterity. 

Overall, we have one of the following options in deciding a keyboard shortcut

- Simple shortcuts with modifier with a letter usually add accented characters or special symbols. This makes it a bad candidate for keyboard shortcuts
- A little bit complex, is two modifier keys like `CMD+ALT` with a character key, but browser defaults use a lot of these
- A bit more complex with 3 modifiers and a character key, this is extremely bad in terms of UX. Prescribing this to users is not a great idea.

So what do we do? Apps across the internet seem to just override the browser defaults if they are obscure enough. For instance `CMD+Shift+A` on Chrome opens search tabs. While it's not great to override browser shortcuts, that seems to be a good middleground

---

Given the trade-offs, here's a framework on approaching problems like this.

- Start with non-conflicting two-modifier shortcuts for essential functionality. Test them to ensure they don't conflict with common browser/system shortcuts.
- Override browser defaults only if the shortcut is obscure and your app's functionality is more relevant. For example, overriding CMD+Shift+A might be acceptable if your app uses it for a critical feature.
- Provide clear documentation about the shortcuts and any overrides.

PR #10965, opened 2025-02-24
fix: Update rendering logic for the "Read Documentation" link

pranavrajs:
Update the rendering logic based on the branding installation type etc. Used CustomBrandPolicyWrapper to render the child component


sojan-official (approved):
- verified in local

PR #10967, opened 2025-02-25
chore: Ability to filter conversations with priority

sojan-official:
- Ability to filter conversation with priority


sojan-official:
@scmmishra  I think previously when we tried to implement this in the past there was an issue with custom attributes with conflicting names, so when you take a look could you ensure that case is also covered ?

scmmishra:
@sojan-official this will break on the frontend if there's already a priority custom attribute

scmmishra:
@sojan-official there's collision between custom attributes `priority` and custom attribute `priority` created by users. 
We can implement one of the following solutions. 

1. We can rename our priority key to `system_priority` or `standard_priority` on the API side, this is the easiest to implement. And will immediately fix this issue. But then it's not a generic solution
2. We send across a hint from the front-end if a certain attribute is a custom attribute. This way the backend can handle it specifically. We already have the information on the frontend as `attributeModel`, if we send it across with payload, we can handle it. On the frontend, we will still need to update the keys, perhaps a prefix like `custom::<attribute-key>` which is then stripped away when sending to the backend.

scmmishra (commented):
- Frontend: Remove any custom attribute that conflicts with the standard ones
- Backend: Add validation to not allow conflicting attributes

scmmishra:
@sojan-official @pranavrajs updated the PR with the following changes

1. Hide conflicting attributes from conversation and contact filters
2. Add backend validation to throw an error when creating a conflicting key

One thing to note is that the validation runs on create only, so if a user tries to update an existing custom attribute with a conflicting key it will allow it. Should we change this behaviour?  We don't allow updating keys anyway, just description, label, etc

muhsin-k:
@scmmishra The conversations list is not updating properly when I select "not equal" for any values.


https://github.com/user-attachments/assets/057a4c09-a126-4df4-9334-c47597944a6a

scmmishra:
@muhsin-k seems like the conversation item shows up in the list because it was already open. If you query from the default dashboard page (`http://localhost:3000/app/accounts/1/dashboard`) It works perfectly fine. 

The issue is unrelated

muhsin-k (approved):

PR #10974, opened 2025-02-25
feat: Support bigger font size in dashboard

iamsivin:
# Pull Request Template

## Description

Fixes https://linear.app/chatwoot/issue/CW-4091/accessibility-improvement-support-bigger-font-size-for-the-dashboard

## Type of change

## Checklist:

### DIFF
--- a/app/javascript/dashboard/composables/useFontSize.js
+++ b/app/javascript/dashboard/composables/useFontSize.js
@@ -1,0 +1,157 @@
+/**
+ * @file useFontSize.js
+ * @description A composable for managing font size settings throughout the application.
+ * This handles font size selection, application to the DOM, and persistence in user settings.
+ */
+
+import { computed, ref, watch } from 'vue';
+import { useUISettings } from 'dashboard/composables/useUISettings';
+import { useAlert } from 'dashboard/composables';
+import { useI18n } from 'vue-i18n';
+
+/**
+ * Mapping of semantic font size names to their pixel values
+ * @type {Object}
+ */
+const FONT_SIZE_MAP = {
+  smaller: '14px',
+  small: '15px',
+  default: null, // null means use browser default
+  large: '18px',
+  larger: '20px',
+  'extra-large': '22px',
+};
+
+/**
+ * Font size management composable
+ *
+ * @returns {Object} Font size utilities and state
+ * @property {Array} fontSizeOptions - Array of font size options for select components
+ * @property {import('vue').ComputedRef<string>} currentFontSize - Current font size from UI settings
+ * @property {Function} applyFontSize - Function to apply font size to document
+ * @property {Function} updateFontSize - Function to update font size in settings with alert feedback
+ */
+export const useFontSize = () => {
+  const { uiSettings, updateUISettings } = useUISettings();
+  const { t } = useI18n();
+
+  /**
+   * Cache to avoid unnecessary DOM operations
+   * @type {import('vue').Ref<string|null>}
+   */
+  const lastAppliedFontSize = ref(null);
+
+  /**
+   * Font size options for select dropdown
+   * @type {Array<{value: string, label: string}>}
+   */
+  const fontSizeOptions = [
+    {
+      value: 'smaller',
+      label: 'Smaller',
+    },
+    {
+      value: 'small',
+      label: 'Small',
+    },
+    {
+      value: 'default',
+      label: 'Default',
+    },
+    {
+      value: 'large',
+      label: 'Large',
+    },
+    {
+      value: 'larger',
+      label: 'Larger',
+    },
+    {
+      value: 'extra-large',
+      label: 'Extra Large',
+    },
+  ];
+
+  /**
+   * Current font size from UI settings
+   * @type {import('vue').ComputedRef<string>}
+   */
+  const currentFontSize = computed(
+    () => uiSettings.value.font_size || 'default'
+  );
+
+  /**
+   * Convert semantic font size to pixel value
+   *
+   * @param {string} semanticSize - Semantic font size name
+   * @returns {string|null} - Pixel value or null for default
+   */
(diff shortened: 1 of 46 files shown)


app/javascript/dashboard/composables/useFontSize.js
```diff
@@ -0,0 +1,157 @@
...
+   * Cache to avoid unnecessary DOM operations
+   * @type {import('vue').Ref<string|null>}
+   */
+  const lastAppliedFontSize = ref(null);
+
+  /**
+   * Font size options for select dropdown
+   * @type {Array<{value: string, label: string}>}
+   */
+  const fontSizeOptions = [
```
scmmishra, line 48: We need to set i18n for these labels

app/javascript/dashboard/composables/useFontSize.js
```diff
@@ -0,0 +1,157 @@
...
+import { useI18n } from 'vue-i18n';
+
+/**
+ * Mapping of semantic font size names to their pixel values
+ * @type {Object}
+ */
+const FONT_SIZE_MAP = {
+  smaller: '14px',
+  small: '15px',
+  default: null, // null means use browser default
```
scmmishra, line 19:
```suggestion
  default: '16px'
```

Let's explicitly set the default value. Either way tailwind includes browser resets that will set this for us, so we won't be defaulting to browser default

app/javascript/dashboard/composables/useFontSize.js
scmmishra, line 1: Generally I think the solution is a bit too complicated. We can do this much more simply. 

1. Set a `--woot-font-size` CSS property at the `:root` with default value as `16`. 
2. When loading `vueapp.html.erb` we set the value on the body style something like this
    ```erb
    <body class="text-slate-600" style="--woot-font-size: <%= @user_font_size || '16px' %>">
      <div id="app"></div>
      <noscript id="noscript">This app works best with JavaScript enabled.</noscript>
      <%= yield %>
      <% if @dashboard_scripts.present? %>
        <%= @dashboard_scripts.html_safe %>
      <% end %>
   </body>
   ````
3. On the profile settings page, when someone switches the value, we just use simple function to set the `--woot-font-size` property on the body tag. We can avoid the watcher and by extension this composable also, since this composable is not used anywhere else, and also this action won't be done so frequently
4. Instead of saving `smaller`, `larger`, etc to the UI settings, we can store the pixel value directly. And only use named tokens for the label. It'll make the whole thing much more simpler

scmmishra: @pranavrajs @muhsin-k WDYT?

app/javascript/dashboard/composables/useFontSize.js
```diff
@@ -0,0 +1,157 @@
...
+   *
+   * @param {string} semanticSize - Semantic font size to apply ('smaller', 'small', 'default', etc.)
+   * @returns {void}
+   */
+  const applyFontSize = semanticSize => {
+    // For performance, only update DOM when needed
+    if (lastAppliedFontSize.value === semanticSize) return;
+
+    // Cache the last applied size
+    lastAppliedFontSize.value = semanticSize;
```
scmmishra, line 104: This optimisation seems unnecessary. This action is not very frequent. We can remove this and make it a lot more simpler. Also, if we apply the same font size, I'm sure the browser will not trigger a repaint.

scmmishra, line 104: This optimisation seems unnecessary. This action is not very frequent. We can remove this and make it a lot more simpler. Also, if we apply the same font size, I'm sure the browser will not trigger a repaint.

app/javascript/dashboard/composables/useFontSize.js
scmmishra: We're not doing this. Seems like the dashboard controller does not have user context. Which makes sense, since it just has to render the dashboard. The auth handshake and rendering is handled by the vue app

app/javascript/dashboard/composables/useFontSize.js
iamsivin: Dones https://github.com/chatwoot/chatwoot/pull/10974/commits/42cdaad76f97fe63fe9a8276ef8f41674b471490

app/javascript/dashboard/composables/useFontSize.js
```diff
@@ -0,0 +1,118 @@
...
+   */
+  const applyFontSize = pixelValue => {
+    // Use requestAnimationFrame for better performance
+    requestAnimationFrame(() => {
+      if (!pixelValue || pixelValue === FONT_SIZE_OPTIONS.DEFAULT) {
+        document.documentElement.style.removeProperty('font-size');
+      } else {
+        document.documentElement.style.setProperty('font-size', pixelValue);
+      }
+    });
```
scmmishra, line 77:
```suggestion
    requestAnimationFrame(() => {
      document.documentElement.style.setProperty('font-size', pixelValue ?? FONT_SIZE_OPTIONS.DEFAULT);
    });
```

@iamsivin WDYT?

scmmishra (approved):

app/javascript/dashboard/composables/useFontSize.js
iamsivin: @scmmishra If we want to set the default 16px all the time then, I guess we can do this, else the existing approach is better. It will not set the `font-size` property for default case.

scmmishra: Should be fine I guess, what is our tailwind default? If it's 16, we're good

iamsivin: @scmmishra Updated the PR https://github.com/chatwoot/chatwoot/pull/10974/commits/31f537fae01575ca4608b03beb4b0203fc729b63

scmmishra (approved):

PR #10978, opened 2025-02-26
feat: add timeout before marking as read

scmmishra:


scmmishra:
We will park this PR for now, we need more clarity on how this behaviour should exist. Right now, hard-coding anything feels like hard-coding a personal preference which may or may not align with what the end-user actually wants. 

Giving configuration options is also not a great idea, more knobs to turn is not great UX, besides it complicates the code unnecessarily.

We need to take a look at the patterns available across apps. Here's a few things we noticed
- requiring the user to explicitly click a button or a tooltip to mark messages as read.
- for a long list of messages, requiring them to scroll through the entire list and then marking it as read automatically. This is great for long list ofcourse, but the pattern fails when there are just one or two messages, where the question of weather to delay the trigger is a good option or not.

PR #10999, opened 2025-02-28
feat: move Linear config to installation_config

muhsin-k:

- Add `LINEAR_CLIENT_ID` and `LINEAR_CLIENT_SECRET` to installation config
- Add Linear config to super_admin/features.yml
- Replace usage of ENV.fetch with GlobalConfigService.load for fetch Client Id and Secret.


sojan-official (approved):
LGTM

PR #11005, opened 2025-03-01
fix: Dynamic seconds delay by attachments size on message send reply

clairton:
# Pull Request Template

## Dynamic seconds delay by attachments size on message send reply

Sometimes where size of files is more large, the presign url return 403

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -320,10 +320,14 @@ class Message < ApplicationRecord
                                                                             previous_changes: previous_changes)
   end
 
-  def send_reply
+  def send_reply  def send_reply
+    return ::SendReplyJob.perform_later(id) if attachments.blank?
+
     # FIXME: Giving it few seconds for the attachment to be uploaded to the service
     # active storage attaches the file only after commit
-    attachments.blank? ? ::SendReplyJob.perform_later(id) : ::SendReplyJob.set(wait: 2.seconds).perform_later(id)
+    seconds = (attachments.map(&:file).map(&:byte_size).sum() / 1024 / 1000).to_i.seconds # a second per 1MB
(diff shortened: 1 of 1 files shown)


scmmishra (commented):
A second per MB of file size feels a bit arbitrary. (In retrospect, so is the 2 seconds delay). Also if the attachment size is 40MB (max allowed IIRC), waiting for 40 seconds before processing a message is not ideal. If at all we do this dynamic delay, we need to cap the delay to something like 5 seconds.

We need a better solution overall. 

@sojan-official what do you think about this overall?

mazenkhalil:
We have worked previously on this specific issue... We comeup with a workaround... I don't think it is a clean fix, but if I recall correctly this was the only way to tackle this ActiveStorage behavior...

1. Add `after_create_commit` callback on the attachment model, this will basically increment the number of processed attachments on the parent message atomically.

image

2. Process the `after_create_commit` callback on the message only when all attachments get processed (The number of attachments equals to the number of the processed attachments).

image


Note: The `after_create_commit` callback on the attachment model should be added ABOVE the `has_one_attached` association... This works well on Rails 7.0 (current version) however when moving to Rails 7.1 the `has_one_attached` association must be defined ABOVE the `after_create_commit` callback to maintain the same behavior.

sojan-official:
Thanks for the contribution here.

The current approach introduces a dynamic delay based on attachment size. While it may reduce the likelihood of the issue, it still relies on an arbitrary wait and does not fully address the underlying race condition around attachment processing.

We’re going to close this PR for now and instead work on a more robust fix that ensures message processing happens only after all attachments are persisted.

Tracked here: #13801

Thanks again for raising this and helping investigate the issue.

PR #11011, opened 2025-03-03
feat: meta delete callback

scmmishra:


app/jobs/webhooks/facebook_delete_job.rb
```diff
@@ -0,0 +1,44 @@
...
+    ::Redis::Alfred.del(key)
+  end
+
+  def delete_channel_if_present
+    channel = Channel::FacebookPage.find_by(page_id: id_to_process)
+    return unless channel
+
+    inbox = channel.inbox
+    inbox.destroy
+  end
```
scmmishra, line 28: Let's skip this

scmmishra:
@sojan-official there's some CVE in `uri` gem, that is breaking circle CI. What should we do? Seems like the package is a downstream dependency of a gem we forked

`azure-storage-blob -> azure-storage-common -> faraday -> ... -> uri`

What do you suggest we do?

PR #11033, opened 2025-03-06
chore: Rescue slack link unfurling errors.

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-4122/slackwebapierrorsmissingscope-missing-scope

This PR adds the ability to handle errors when scopes are missing during link unfurling. Since link unfurling is just a nice-to-have feature that doesn't affect core functionality, we will silently ignore these errors.


sojan-official (approved):
made a tiny change. changed from logger error to warn

PR #11051, opened 2025-03-10
fix: Out of order conversation events causing stale data to reflect in the UI

scmmishra:
### Problem

As identified in our investigation, conversation events were sometimes delivered out of sequence through ActionCable, causing UI issues where conversations would briefly appear and then disappear, this would cause the stats to show the correct number on the list header, but the chat list would show incorrect items. This was happening because of the following reasons:

- `ActionCable` doesn't guarantee delivery order when broadcasts occur from separate Sidekiq jobs
- Multiple related events (`conversation.updated`, `assignee.changed`, etc.) were being processed in an unpredictable order
- Each event would overwrite the entire conversation state without checking if it represented the latest state

### Solution

This PR implements a timestamp-based event sequencing mechanism to ensure we only process the most recent updates for each conversation:

- Added `event_timestamp` to all `ActionCable` broadcasts, using the original event timestamp from the application. This timestamp is generated on the main thread, at the model before the action is ever dispatched and can be considered a reliable key to find the order of events.
- Created a client-side tracking mechanism in `ActionCableConnector` to maintain the latest known timestamp for each conversation
- Added validation logic to ignore any events with timestamps older than what we've already processed

### Implementation Details

- Added `conversationEventTimestamps` Map to track the latest timestamp per conversation
- Added `eventOutOfSequence()` helper method to detect and filter out-of-sequence events
- Modified event handlers for `assigneeChanged`, `conversationRead`, `statusChange`, and `conversationUpdated` to respect event ordering
- Updated backend broadcasters to include event timestamps with all websocket messages

### Future work

- We can get rid of `CONVERSATION_UPDATE_EVENTS` in `ActionCableBroadcastJob`, and avoid re-fetching of the objects.
- We can implement granular updates for events like team or assignee change

### DIFF
--- a/app/javascript/dashboard/helper/actionCable.js
+++ b/app/javascript/dashboard/helper/actionCable.js
@@ -9,6 +9,7 @@ class ActionCableConnector extends BaseActionCableConnector {
     const { websocketURL = '' } = window.chatwootConfig || {};
     super(app, pubsubToken, websocketURL);
     this.CancelTyping = [];
+    this.conversationEventTimestamps = new Map();
     this.events = {
       'message.created': this.onMessageCreated,
       'message.updated': this.onMessageUpdated,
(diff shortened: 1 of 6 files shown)


pranavrajs (commented):
Broadly, I would consider the following: 

Source of the problem (we know this already), potential easy fix first, mechanism to analyse the fix.

Sojan and I have been discussing how to make sure the changes work. To start, we need a few more checks in place. But for today, let’s keep it simple and first confirm if our initial fix actually works.

Right now, we have an updated_at timestamp and an event_timestamp. The original plan (at least how I pictured it) was to check the updated_at timestamp first and fall back to the event timestamp if needed. But that’s not how it is actually implemented. The check is only for the event_timestamp which can lead to the same problem as before. 

The source of truth should be the DB and we just need to make sure that the updated_at would give you the latest data (reload before sending).

Currently, we have floating point precision for our timestamps in the updated_at attribute. We can compare those, and if the updated_at attribute is greater, we can update the store. 

Earlier, when we sent the updated flag, we were not reloading the object and also the timestamp were having integer precision, which caused us to miss some updates and created problems. After the reload and with the floating point precision, this check should be fine (atleast theoretically).

For today let us put an update with a check on the updated_at attribute only. If it's not the latest, we won't update it. If it is the latest, then we will update it. 

Additionally, we need to log the changes that we have not applied. We can log this into New Relic and track each event. When we talk to customers, we can confidently say that we had 10 such events yesterday, and all were applied correctly. 

The check with event timestamp requires a bit more effort to test and it is hard to maintain in longer afaik, so I would work with updated_at attribute see if it works well and go for the changes you have if it doesn't.

Let me know what you think.

PR #11058, opened 2025-03-11
feat: ignore out of sync messages

scmmishra:


pranavrajs (approved):

pranavrajs:
@scmmishra Can you add a test for this case? I think we don't have one for the entire file, might be a good place to start.

vishnu-narayanan (approved):

PR #11064, opened 2025-03-12
chore: Logger for non-existent WhatsApp channels

sojan-official:
- Add a warning logger for cases where we are getting webhook events for inactive numbers. 
- Add config to discard events for inactive numbers so that the meta will stop sending events


app/controllers/webhooks/whatsapp_controller.rb
```diff
@@ -2,6 +2,12 @@ class Webhooks::WhatsappController < ActionController::API
   include MetaTokenVerifyConcern
 
   def process_payload
+    if inactive_whatsapp_number?
+      Rails.logger.warn("Rejected webhook for inactive WhatsApp number: #{params[:phone_number]}")
+      render json: { error: 'Inactive WhatsApp number' }, status: :service_unavailable
```
pranavrajs, line 7: I would suggest to go with unprocessable_entity rather than service unavailable.
```suggestion
      render json: { error: 'Inactive WhatsApp number' }, status: :unprocessable_entity
```

spec/controllers/webhooks/whatsapp_controller_spec.rb
```diff
@@ -29,5 +29,34 @@
...
+      before do
+        allow(GlobalConfig).to receive(:get_value).with('INACTIVE_WHATSAPP_NUMBERS').and_return('+1234567890,+9876543210')
+      end
+
+      it 'returns service unavailable for inactive phone number in URL params' do
+        allow(Rails.logger).to receive(:warn)
+        expect(Rails.logger).to receive(:warn).with('Rejected webhook for inactive WhatsApp number: +1234567890')
+
+        post '/webhooks/whatsapp/+1234567890', params: { content: 'hello' }
+        expect(response).to have_http_status(:service_unavailable)
```
pranavrajs, line 43:
```suggestion
        expect(response).to have_http_status(:unprocessable_entity)
```

pranavrajs (approved):
I've a comment on the status code, otherwise looks fine.

PR #11092, opened 2025-03-15
chore: Add validations to campaign model

scmmishra:
- Add validations to campaign model ensuring that the associated inbox belongs to one with in the campaign account.


app/models/campaign.rb
```diff
@@ -92,6 +93,14 @@ def validate_url
     errors.add(:url, 'invalid') if inbox.inbox_type == 'Website' && !use_http_protocol
   end
 
+  def inbox_must_belong_to_account
+    return unless inbox
```
pranavrajs, line 97: This seems redundant. The inbox_id is a required field.

scmmishra: Removing this will fail the association checks.

```
describe 'associations' do
    it { is_expected.to belong_to(:account) }
    it { is_expected.to belong_to(:inbox) }
end
```

sojan-official (approved):

PR #11097, opened 2025-03-17
chore: Remove delete Instagram story implementation

muhsin-k:
Currently, Instagram stories are downloaded and displayed as attachments. We previously implemented story deletion by checking individual messages via https://github.com/chatwoot/chatwoot/pull/5300/files#diff-684a16c1b0b4c099fcdfeed95b1820e11fef629fe332ec7ce6a8c600331dd06dR110, but this approach proved impractical and was removed. This PR removes all unused code to avoid confusion. We will revisit the story deletion feature when we implement `instagram_manage_insights`.


sojan-official (approved):
Code LGTM

PR #11101, opened 2025-03-18
feat(apps): Shopify Integration

pranavrajs:
This PR adds native integration with Shopify. No more dashboard apps. The support agents can view the orders, their status and the link to the order page on the conversation sidebar. 

This PR does the following: 
- Create an integration with Shopify (a new app is added in the integrations tab)
- Option to configure it in SuperAdmin
- OAuth endpoint and the callbacks.
- Frontend component to render the orders. (We might need to cache it in the future)

**Integrations Page**

Screenshot 2025-03-17 at 9 47 14 PM

**Orders on the sidebar**
Screenshot 2025-03-17 at 9 47 37 PM


app/controllers/api/v1/accounts/integrations/shopify_controller.rb
```diff
@@ -0,0 +1,109 @@
...
+
+  def shopify_session
+    ShopifyAPI::Auth::Session.new(shop: @hook.reference_id, access_token: @hook.access_token)
+  end
+
+  def shopify_client
+    @shopify_client ||= ShopifyAPI::Clients::Rest::Admin.new(session: shopify_session)
+  end
+
+  def set_contact
```
muhsin-k, line 103: The name `validate_contact` is better to reflect its purpose,
```suggestion
  def validate_contact
```

app/controllers/shopify/callbacks_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def handle_response
+    account.hooks.create!(
+      app_id: 'shopify',
+      access_token: parsed_body['access_token'],
+      status: 'enabled',
+      reference_id: params[:shop],
+      settings: {
+        token_type: parsed_body['token_type'],
+        expires_in: parsed_body['expires_in'],
```
muhsin-k, line 33: When does this token expire? If it does expire, do we need to refresh the token?

pranavrajs: offline access token doesn't expire AFAIK. I just removed this. These values are empty in the response.

sojan-official (approved):

PR #11107, opened 2025-03-19
feat: use gin index for message search

scmmishra:
This PR updates the search implementation to better utilize the GIN indexes. The option is toggled behind a feature flag for us to test it internally before making it available publicly


app/services/search_service.rb
```diff
@@ -35,12 +35,61 @@ def filter_conversations
...
+    @messages
+  end
+
+  def filter_messages_with_gin
+    base_query = message_base_query
+
+    if search_query.present?
+      # Use the @@ operator with to_tsquery for better GIN index utilization
+      # Convert search query to tsquery format with prefix matching
+      tsquery = search_query.split.map { |word| "#{word}:*" }.join(' & ')
```
vishnu-narayanan, line 56: Note: The search results will be different in GIN compared to ILIKE since we are splitting words and combining using `&` filtering in GIN. GIN would get us exact matches but ILIKE would have more results.

vishnu-narayanan (approved):
Note: The search results will be different in GIN compared to ILIKE since we are splitting words and combining using `&` filtering in GIN. GIN would get us exact matches but ILIKE would have more results.

PR #11111, opened 2025-03-19
feat: frontend filtering of conversations

scmmishra:
This pull request includes significant changes to the filtering logic for conversations in the frontend, here's a summary of the changes

This includes adding a `matchesFilters` method that evaluates a conversation against the applied filters. It does so by first evaluating all the conditions, and later converting the results into a JSONLogic object that can be evaluated according to Postgres operator precedence

### Alignment Specs

To ensure the frontend and backend implementations always align, we've added tests on both sides with same cases, for anyone fixing any regressions found in the frontend implementation, they need to ensure the existing tests always pass. 

Test Case | JavaScript Spec | Ruby Spec | Match?
-- | -- | -- | --
**A AND B OR C** | Present | Present | Yes
Matches when all conditions are true | Present | Present | Yes
Matches when first condition is false but third is true | Present | Present | Yes
Matches when first and second conditions are false but third is true | Present | Present | Yes
Does not match when all conditions are false | Present | Present | Yes
**A OR B AND C** | Present | Present | Yes
Matches when first condition is true | Present | Present | Yes
Matches when second and third conditions are true | Present | Present | Yes
**A AND B OR C AND D** | Present | Present | Yes
Matches when first two conditions are true | Present | Present | Yes
Matches when last two conditions are true | Present | Present | Yes
**Mixed Operators (A AND (B OR C) AND D)** | Present | Present | Yes
Matches when all conditions in the chain are true | Present | Present | Yes
Does not match when the last condition is false | Present | Present | Yes


scmmishra:
@pranavrajs this is good to test. PTAL, we could put this behind a feature flag too, what do you suggest?

app/javascript/dashboard/store/modules/conversations/helpers/filterHelpers.js
```diff
@@ -0,0 +1,351 @@
...
+ *  "or": [
+ *    { "and": [true, false] },
+ *    { "and": [true, false, true] },
+ *    { "and": [false] }
+ *  ]
+ * }
+ * @param {Array} evaluatedFilters - Array of evaluated filter conditions with results and operators
+ * @returns {Object} - JSON Logic rule
+ */
+const buildJsonLogicRule = evaluatedFilters => {
```
muhsin-k, line 256: @scmmishra Please fix the Cognitive Complexity for the method `buildJsonLogicRule`

scmmishra: I tried fixing it, but it just ended up making the code much more complicated to read. The complexity will exist regardless of how we split the function. IMO it makes sense to keep it favouring readability

iamsivin (commented):
@scmmishra Found a bug while testing the PR. 

The filter endpoint enters an infinite loop, continuously fetching the next page (+1).
**Filter**: **Status** equal to **All**

### Loom video
https://www.loom.com/share/5c4e12b17dc046f48543588a9fea59b7?sid=067eb95f-07cb-463a-9678-fb7c8ebb1349

pranavrajs (approved):

PR #11130, opened 2025-03-20
feat: use separate redis for sidekiq

vishnu-narayanan:
This change separates the Redis connections for Rails and Sidekiq, improving scalability and resource management.

 - Configures Sidekiq to use a dedicated Redis instance.
-  Falls back to the original Redis instance if the separate Redis configuration is not set.
-  Helps avoid hitting connection limits by maintaining independent Redis instances for Rails and Sidekiq.
-  Enables better resource management and allows for independent scaling of Redis instances.


pranavrajs:
Closing this as we will not proceed with this PR.

PR #11165, opened 2025-03-24
fix: improve apm error reporting

vishnu-narayanan:
Previously, errors were only being logged with logger.info and rendered as JSON responses but weren't being properly reported to APM services. Critical errors like 500s were only visible in application logs but not in our monitoring systems, making it harder to track and debug issues in production.

This commit ensures that all errors are properly reported to configured APM services by:
1. Adding explicit APM error reporting alongside logging
2. Using proper APM reporting methods for each service
3. Maintaining consistent error handling across all APM integrations

These changes ensure that errors are captured correctly in our monitoring systems, improving our ability to track and debug production issues.


sojan-official (approved):

sojan-official:
- Closing for now, we will reopen if required.

sojan-official (changes requested):
- We only need to report exception for internal server errors, for other cases we can ignore .
- Also use Chatwoot exeception tracker class and then move the logic for reporting various services into it. 
- for various services apart from the const_defined? we will have to check if the env variable is present.

muhsin-k:
@vishnu-narayanan Please take the next action here.

pranavrajs:

Issue #11149, opened 2025-03-21
Translate bug - Settings name in PT-BR


mateusw3c:
### Describe the bug

In PT-BR language, the menu help center -> settings is Configurações.  But it is Confirgurações :)

![Image](https://github.com/user-attachments/assets/2be70a5b-b2a4-4fa7-97f9-f527cead03ad)

### To Reproduce

Only open the admin panel in PT-BR language and click inside the Help Center menu.

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

None

hahuma:
I've created a PR to fix this issue: https://github.com/chatwoot/chatwoot/pull/11172

PR #11172, opened 2025-03-25
fix(i18n): correct settings name translation in pt-BR

hahuma:
# Pull Request Template

## Description

In pt_BR language, some strings are being mispelled, the correct word would be "Configurações" but instead it is being showed as "Confirgurações", which is wrong.
Fixes: https://github.com/chatwoot/chatwoot/issues/11149

## Type of change

## How Has This Been Tested?

I've compiled the new translations and I compared the problematic strings in PT_BR to check if it was still with the "Confirgurações" word instead of "Configurações"

## Checklist:


pranavrajs:
@hahuma Thanks for the update! I guess we’ll need to change this in Crowdin as well, otherwise it’ll get overwritten by the translation PR.

pranavrajs (approved):

psilva-sv:
@pranavrajs  Is it an open source project as well? If so, I think I could help with it. Also, I'm not entirely sure, but I think there's an open PR from Crowdin to add some Portuguese translation strings. However, it seems to be doing things backwards, as some already translated strings are being replaced with their English equivalents.

I relatively new to the community so I don't if that's some kind of expected behavior, so if that's what you guys expected from the very beginning so just, desconsider my question.

The PR I'm talking about: #11157

Example:
image

hahuma:
@pranavrajs   Is it an open source project as well? If so, I think I could help with it. Also, I'm not entirely sure, but I think there's an open PR from Crowdin to add some Portuguese translation strings. However, it seems to be doing things backwards, as some already translated strings are being replaced with their English equivalents.

I relatively new to the community so I don't if that's some kind of expected behavior, so if that's what you guys expected from the very beginning so just, desconsider my question.

The PR I'm talking about: #11157

Example:
image

PR #11183, opened 2025-03-26
fix: Remove where query if admin

pranavrajs:
When finding conversation if it is an admin, we don't need to filter it by inbox ids.

### DIFF
--- a/app/finders/conversation_finder.rb
+++ b/app/finders/conversation_finder.rb
@@ -32,6 +32,7 @@ class ConversationFinder
   def initialize(current_user, params)
     @current_user = current_user
     @current_account = current_user.account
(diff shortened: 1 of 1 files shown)


vishnu-narayanan (approved):

ycarooliveira:
Just wondering but Doesnt @current_user.assigned_inboxes already takes consideration about been admin?

pranavrajs:
@ycarooliveira Yes, it does. However, generating an IN query wasn’t necessary for admins since they have access to all inboxes. The previous assigned_inboxes condition was just including every inbox in the account. Now, instead of adding all inboxes, that condition is removed entirely for the admins.

PR #11185, opened 2025-03-26
feat: limit active sessions per user to 3

vishnu-narayanan:
- Implement a limit of 3 active sessions per user to prevent abuse and ensure fair usage across all accounts


pranavrajs:
Closing this as the change would require frontend changes on the web app and the mobile app to allow users to select the session they would want to discard.

PR #11186, opened 2025-03-26
feat: add job to remove stale contacts

vishnu-narayanan:
- Add a job to remove stale contacts across all accounts
- Enable the job by setting `REMOVE_STALE_CONTACTS_JOB_STATUS` to `true`

Stale anonymous contact is defined as 
- have no identification (email, phone_number, and identifier are NULL)
- have no conversations
- are older than 30 days

### DIFF
--- a/app/services/internal/remove_stale_contacts_service.rb
+++ b/app/services/internal/remove_stale_contacts_service.rb
@@ -1,0 +1,43 @@
+class Internal::RemoveStaleContactsService
+  def perform
+    return unless remove_stale_contacts_job_enabled?
+
+    time_period = 30.days.ago
+    contacts_to_delete = stale_contacts(time_period)
+
+    log_stale_contacts_deletion(contacts_to_delete, time_period)
+
+    # Since the number of records to delete is very high,
+    # delete_all would be faster than destroy_all since it operates at database level
+    # and avoid loading all the records in memory
+    # Transaction and batching is used to avoid deadlock and memory issues
+    Contact.transaction do
+      contacts_to_delete
+        .find_in_batches(batch_size: 10_000) do |group|
+          Contact.where(id: group.map(&:id)).delete_all
+        end
+    end
+  end
+
+  private
+
+  def remove_stale_contacts_job_enabled?
+    job_status = ENV.fetch('REMOVE_STALE_CONTACTS_JOB_STATUS', false)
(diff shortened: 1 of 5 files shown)


app/services/internal/remove_stale_contacts_service.rb
```diff
@@ -0,0 +1,43 @@
...
+        .find_in_batches(batch_size: 10_000) do |group|
+          Contact.where(id: group.map(&:id)).delete_all
+        end
+    end
+  end
+
+  private
+
+  def remove_stale_contacts_job_enabled?
+    job_status = ENV.fetch('REMOVE_STALE_CONTACTS_JOB_STATUS', false)
```
pranavrajs, line 25: Let us remove this env.

app/services/internal/remove_stale_contacts_service.rb
```diff
@@ -0,0 +1,43 @@
+class Internal::RemoveStaleContactsService
+  def perform
```
pranavrajs, line 2: Pass an account_id here, so that we can limit the query to an account.

app/services/internal/remove_stale_contacts_service.rb
```diff
@@ -0,0 +1,43 @@
...
+    # delete_all would be faster than destroy_all since it operates at database level
+    # and avoid loading all the records in memory
+    # Transaction and batching is used to avoid deadlock and memory issues
+    Contact.transaction do
+      contacts_to_delete
+        .find_in_batches(batch_size: 10_000) do |group|
+          Contact.where(id: group.map(&:id)).delete_all
+        end
+    end
+  end
```
pranavrajs, line 20: Use destroy. limit it to account.contacts. do not use transaction.

app/services/internal/remove_stale_contacts_service.rb
vishnu-narayanan: sure

app/services/internal/remove_stale_contacts_service.rb
vishnu-narayanan: make sense

app/services/internal/remove_stale_contacts_service.rb
vishnu-narayanan: cool, will make change

scmmishra (approved):

PR #11187, opened 2025-03-26
feat: debounce meta request

scmmishra:
The meta call is extremely expensive for large systems, every actionCable event related to conversation triggers a request, this can easily saturate the incoming requests queue on the backend. 

Earlier we added a stop gap solution to do an early return if the last updated time was within 10 seconds of the request, if there were more than 100 conversations. While this worked well initially, our logs show the number of these requests are still way larger than the theoretical limit this implementation allows, assuming it works.

There's a possibility the processing delay between committing the `updatedOn` data to the store (basically one Vue lifecycle tick) many requests slip through. This is not noticable generally, but during peak hours, we can get multiple events with a millisecond, near instantaneously. This subverts the check we have, making it apparent that the core problem here is the the check is strongly coupled with the Vue lifecycle.
Another criticism of this was that the numbers would often mismatch, this is because for the last event if the counts changed, no fresh data was fetched. We need some form of eventual consistency

### Change Summary

1. We split out the get function and get a separate `fetchMetaData` function
2. This is debounced at two intervals at 0.5 seconds and 5 seconds. These functions are declared at the top completely separate from the vue lifecycle. 
3. Default behaviour is to debounce by 0.5 seconds always, and make it 5 seconds if there are more than 5 seconds.
4. This addresses the issue that the numbers were not eventually consistent, a request always goes through at the end of the timer

Backend PR: https://github.com/chatwoot/chatwoot/pull/11184

### DIFF
--- a/app/javascript/dashboard/store/modules/conversationStats.js
+++ b/app/javascript/dashboard/store/modules/conversationStats.js
@@ -1,37 +1,39 @@
 import types from '../mutation-types';
+import { debounce } from '@chatwoot/utils';
 import ConversationApi from '../../api/inbox/conversation';
 
 const state = {
   mineCount: 0,
   unAssignedCount: 0,
   allCount: 0,
-  updatedOn: null,
 };
 
 export const getters = {
   getStats: $state => $state,
 };
 
+// Create a debounced version of the actual API call function
+const fetchMetaData = async (commit, params) => {
+  try {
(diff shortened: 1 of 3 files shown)


pranavrajs (approved):

pranavrajs:
Closing in favor of #11190

PR #11194, opened 2025-03-27
feat: improve performance of label reports download

scmmishra:
Earlier label reports were generated by looping over all the labels and generating the existing report for it. This worked fine for a while, but as accounts grow with more conversations, and the number of labels also increase, this became increasingly slow. At the moment for really large installations the query just times out. 

This PR fixes it by using a batch query to process the entire labels, instead of multiple queries for each label. 
While the join is complicated and maybe not be extremely fast, it's still processes in a reasonable time (under 5 seconds) for large accounts as well


app/builders/v2/label_report_builder.rb
```diff
@@ -0,0 +1,100 @@
+class V2::LabelReportBuilder
```
pranavrajs: Why is it in a different namespace than the summary builder?

pranavrajs (changes requested):
@scmmishra Let us discuss the implementation before we get it to production.

scmmishra:

scmmishra:
@sojan-official PTAL

muhsin-k (approved):

PR #11229, opened 2025-04-02
chore: Generate test data for bulk insertion

tds-1:
## Description

This PR adds functionality to generate test data for the Message model, which will help improve testing coverage and development workflows.

Fixes # (issue)

## Type of change

## How Has This Been Tested?

The script has been run locally using the following command:
``` bundle exec rake data:generate_distributed_data```

## Checklist:

### CODE REVIEW

**Review by sojan-official — COMMENTED** [review]
lets look into breaking down this script into respective service classes, in separate files. This will concur with the current development practises in Chatwoot. 

This ChatGPT conversation should give a good sense of how we can reorganize: 
https://chatgpt.com/share/67ef277a-e934-8006-9158-037966a67d13

Here is also the link to our engineering guide
https://chatwoot.help/hc/handbook/articles/1743728412-backend-guidelines

File: .rubocop.yml [hunk 1]
```diff
@@ -166,3 +166,5 @@ AllCops:
     - 'tmp/**/*'
     - 'storage/**/*'
     - 'db/migrate/20230426130150_init_schema.rb'
+    # added this because of bulk insert operation present in the file
+    - 'lib/test_data/contact_batch_service.rb'
```
sojan-official, line 170: if is a specific cop breaking in a file and we can't figure out a fix lets do inline disable instead 
 ref: https://github.com/chatwoot/chatwoot/blob/ae0b68147e3a8a4a7effda9888c939f680f949f6/app/builders/messages/facebook/message_builder.rb#L129

lib/test_data/database_optimizer.rb
scmmishra, line 1: This is neat, we can add some more optimzations

- increase the working memory to something like `256MB`, the default is 4MB locally for me. Also try increasing the `maintenance_work_mem`, should boost up index creation
- try setting the table to `UNLOGGED`, this should _in theory_ give us a really good speed boost by disabling WAL entirely. This should also allow us to remove `synchronous_commit = off` since there is no WAL overhead anymore.

lib/test_data/database_optimizer.rb
```diff
@@ -0,0 +1,41 @@
...
+      # Disable synchronous commits to improve write performance
+      # This makes PostgreSQL return success before writing to disk, improving throughput
+      Rails.logger.info '    Disabling synchronous commits for better write performance'
+      ActiveRecord::Base.connection.execute('SET synchronous_commit = off')
+
+      # Disable triggers on conversation and message tables to avoid overhead
+      # This prevents firing of triggers that might slow down bulk inserts
+      Rails.logger.info '    Disabling triggers on conversations and messages tables'
+      ActiveRecord::Base.connection.execute('ALTER TABLE conversations DISABLE TRIGGER ALL')
+      ActiveRecord::Base.connection.execute('ALTER TABLE messages DISABLE TRIGGER ALL')
```
scmmishra, line 23: It would be great to set these table names as constants and use helper functions to toggle these on or off, will help us add more tables easily in the future. Also, if we plan to set `UNLOGGED`, that would help too

lib/test_data/account_creator.rb
```diff
@@ -0,0 +1,31 @@
...
+  def self.create!(id)
+    company_name = generate_company_name
+    domain       = generate_domain(company_name)
+    account      = Account.create!(
+      id: id,
+      name: company_name,
+      domain: domain,
+      created_at: Faker::Time.between(from: 2.years.ago, to: 6.months.ago)
+    )
+    persist_account_id(account.id)
```
scmmishra, line 13: You can use the `internal_attributes` field on the account model also instead of a text file @pranavrajs @sojan-official wdyt?

scmmishra (changes requested):
Some tiny comments, but overall looks in good shape

lib/test_data/account_creator.rb
scmmishra: The current way is fine too, just a nitpick, feel free to ignore

scmmishra (commented):
This is good to go, @tds-1 can you add labels also in this? Assigning 0-5 labels from a set of 30-50 labels per account

I did something similar [here](https://github.com/chatwoot/chatwoot/pull/11194/files#diff-b3f177cb54335da5a93b69f0c358180d721310371b923cf5c4a44045e258c73f) (Need to remove it later)

scmmishra (approved):

Issue #5968, opened 2022-11-28
[CW-365] Add support for webhook actions in Macros


voko3456:
<img width="657" alt="Screenshot 2023-01-06 at 5 38 45 PM" src="https://user-images.githubusercontent.com/2246121/211125518-d30cf0b0-a3b3-4394-80d6-b5886e270f63.png">

**Todo:**

- Add support for a webhook event in macros
- Accept URL and Payload as parameters. 

**Validations:**

- URL param: Required, should be a valid URL.
- Payload: Can be empty. Use {} as default if empty. It should be a valid JSON object.

**Webhook event payload:**

```json
{
  "event": "macro_executed",
  "payload": {
    /*... Payload configured by the admin/agent in the macro definition */
  }
  "macro": {
    /*... macro object */
  },
  "agent": {
    /*... executed by agent */
  },
  "conversation": {
    /*... Conversation object at which this macro is executed */
  }
}
```

**Use cases:**

An agent can trigger an event outside of Chatwoot while talking to a customer. 

Feedback from a customer regarding the use case: 
> We want to instantiate external processes from a single click of a button within chatwoot by our agents. For example, the macro can be created that instantiates our patient verification call - The agent can click the macro button that would begin the process of calling the patient's cell phone and going through our patient verification process. There are several other situations like this where the workflows are accomplished outside of chatwoot but instantiated within Chatwoot by the agent.

**Possible future enhancement:** 

This could become an option where agents can trigger events in multiple different applications with one click. We could possibly ship standard pre-defined actions for products as generic integrations. 

For eg: Updating a contact in Hubspot can be a similar webhook event with credentials, and the agent can trigger a macro which would then call the HubSpot API and update the fields as requested.

There are multiple workflows that can be created. (Sort of Zapier, but run manually by an agent)

<sub>[CW-365](https://linear.app/chatwoot/issue/CW-365/add-support-for-webhook-actions-in-macros)</sub>

pranavrajs:
@tejaswinichile @sojan-official Please take a look at the updated spec and share your thoughts.

sojan-official:
Looks good to me. Possible options for further flexibility. 

- Support for auth headers
- Support for liquid in the payload

scmmishra:
Related: https://github.com/chatwoot/chatwoot/issues/3723

tejaswinichile:
We have this for automation, this is little different for event and validation JSON. I will work on this.

ChrisvanChip:
[...]

PR #11235, opened 2025-04-03
feat: Add webhook event support for macros

pranavrajs:
Fixes https://github.com/chatwoot/chatwoot/issues/5968

We will not support custom payload in V1.

### DIFF
--- a/app/services/macros/execution_service.rb
+++ b/app/services/macros/execution_service.rb
@@ -62,4 +62,9 @@ class Macros::ExecutionService < ActionService
     mb = Messages::MessageBuilder.new(@user, @conversation.reload, params)
     mb.perform
   end
+
+  def send_webhook_event(webhook_url)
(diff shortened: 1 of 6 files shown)


app/services/macros/execution_service.rb
```diff
@@ -62,4 +62,9 @@ def send_attachment(blob_ids)
     mb = Messages::MessageBuilder.new(@user, @conversation.reload, params)
     mb.perform
   end
+
+  def send_webhook_event(webhook_url)
+    payload = @conversation.webhook_data.merge(event: 'macro.executed')
```
sojan-official, line 67: should we also add the macro name to the payload ?

sojan-official (approved):
LGTM. 

one minor comment: would be nice if we can get the specific macro which triggered the webhook in payload as well. can be taken up seperately.

PR #11244, opened 2025-04-04
feat: Handle instagram test service

muhsin-k:
This PR will handle the Instagram test events. We are using the last created Instagram channel as the test channel since we don't have any other channels for testing purposes at the time of Meta approval.

### DIFF
--- a/app/services/instagram/test_event_service.rb
+++ b/app/services/instagram/test_event_service.rb
@@ -1,0 +1,79 @@
+class Instagram::TestEventService
+  def initialize(messaging)
+    @messaging = messaging
+  end
+
+  def perform
+    Rails.logger.info("Processing Instagram test webhook event, #{@messaging}")
+
+    return false unless test_webhook_event?
+
+    create_test_text
+  end
+
+  private
+
+  def test_webhook_event?
+    @messaging[:sender][:id] == '12334' && @messaging[:recipient][:id] == '23245'
+  end
+
+  def create_test_text
+    channel = Channel::Instagram.last
+
+    # As of now, we are using the last created instagram channel as the test channel,
+    # since we don't have any other channel for testing purpose at the time of meta approval
+    @inbox = ::Inbox.find_by(channel: channel)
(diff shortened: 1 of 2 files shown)


app/services/instagram/test_event_service.rb
```diff
@@ -0,0 +1,79 @@
...
+  def test_webhook_event?
+    @messaging[:sender][:id] == '12334' && @messaging[:recipient][:id] == '23245'
+  end
+
+  def create_test_text
+    # As of now, we are using the last created instagram channel as the test channel,
+    # since we don't have any other channel for testing purpose at the time of meta approval
+    channel = Channel::Instagram.last
+
+    @inbox = ::Inbox.find_by(channel: channel)
```
pranavrajs, line 25: Is there a way we can find the channel by something from the payload?

app/services/instagram/test_event_service.rb
```diff
@@ -0,0 +1,79 @@
...
+    return false unless test_webhook_event?
+
+    create_test_text
+  end
+
+  private
+
+  def test_webhook_event?
+    @messaging[:sender][:id] == '12334' && @messaging[:recipient][:id] == '23245'
+  end
```
pranavrajs, line 18: Is this documented somewhere? Please provide links to the documentation where Meta suggests doing this check.

app/services/instagram/test_event_service.rb
muhsin-k: No. There is no way to get the connected channel instagram id.
CleanShot 2025-04-11 at 15 47 07@2x

app/services/instagram/test_event_service.rb
muhsin-k: This is the standard payload for test events. I assume they will send this same payload for verification. I will test this after we test the approval flow for self-hosted customers.
CleanShot 2025-04-11 at 15 48 33@2x

pranavrajs (approved):

PR #11252, opened 2025-04-07
feat: invalidate old password if the user was not confirmed

scmmishra:


sojan-official (commented):
can we also add a spec for this.

pranavrajs:
@scmmishra We’ve decided not to move forward with this change. Instead, we’ll require users to verify their email before they can start using the product. Closing it here for now.

PR #11253, opened 2025-04-07
feat: Implement dedicated url for action cable

legioner0:
# Pull Request Template

## Description

This PR adds configurable WebSocket URL for action cable.
In my case this change required by internal logic of DDoS-protection service - websockets should use separated port.

## Type of change

## How Has This Been Tested?

Build and start app in docker with custom `WEBSOCKET_URL` value - works correctly.

## Checklist:

- [X] I have made corresponding changes to the documentation
- [X] Any dependent changes have been merged and published in downstream modules


pranavrajs:
@legioner0 Could you please provide more details on why this change is required? From what I see, it looks like it’s just updating the URL on the frontend, without any actual changes to how Action Cable is set up. Let me know if I’m missing something.

legioner0:
> @legioner0 Could you please provide more details on why this change is required? From what I see, it looks like it’s just updating the URL on the frontend, without any actual changes to how Action Cable is set up. Let me know if I’m missing something.

Yes, you understand it right - only change URL on frontend for WebSocket connection. Similar logic as used for hosting assets via CDN.

My DDoS prevention provider use separated logic for HTTP and WebSocket traffic, so it's use two separated ports for these services

pranavrajs:
@legioner0 Can you please share which provider you are working with? Unfortunately, we can take it forward as the config is unnecessary for 99% use cases.

legioner0:
@pranavrajs, I'm working with Qrator. Here is documentation with explanation of dedicated port for websocket - https://docs.qrator.net/en/connect/reverse-proxies/websocket.html

PR #11260, opened 2025-04-08
fix: Updating the Ubuntu version to run the tests

shioheii:
In my previous PR I got some errors that when checking it is possible to conclude that the version of ubuntu 20.04 in the Github Actions images to run the tests is no longer supported.
image

Possible to check at this url: https://github.com/actions/runner-images/issues/11101

## What's changed
Changing the Ubuntu version in the image to run the tests.
Old version: 20.04
New version: 22.04

## BEFORE:
image

## NOW:
image

### DIFF
--- a/.github/workflows/frontend-fe.yml
+++ b/.github/workflows/frontend-fe.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   test:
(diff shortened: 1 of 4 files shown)


sojan-official (approved):
Thanks a lot for fixing this !!!

PR #11268, opened 2025-04-09
chore: Add human agent check for Instagram channel

muhsin-k:
This PR adds a check for whether human agent is enabled for the channel, allowing agents to send messages beyond the standard 24-hour messaging window.


app/builders/messages/instagram/base_message_builder.rb
```diff
@@ -94,11 +94,10 @@ def message_reply_attributes
   def build_message
     # Duplicate webhook events may be sent for the same message
     # when a user is connected to the Instagram account through both Messenger and Instagram login.
+    # There is chance for echo events to be sent for the same message.
     # Therefore, we need to check if the message already exists before creating it.
     return if message_already_exists?
 
-    return if @outgoing_echo
```
muhsin-k, line 100: We don't need this condition since we are already checking if the message exists or not.

app/services/conversations/reply_service.rb
```diff
@@ -0,0 +1,59 @@
+class Conversations::ReplyService
```
sojan-official: canReplyservice or a better name

app/services/conversations/reply_service.rb
```diff
@@ -0,0 +1,59 @@
...
+  def initialize(conversation)
+    @conversation = conversation
+  end
+
+  def can_reply?
+    channel = @conversation.inbox&.channel
+
+    return can_reply_on_instagram_via_messenger? if instagram_via_messenger?
```
sojan-official: lets rewrite this to something like this 

```
return true unless channel&.messaging_window_enabled?

last_message_in_messaging_window?(messaging_window)
```

```
def messaging_window
   case @inbox.channel_type
    when 'Channel::Api'
       method_returning_api_channel_message_window
    when 'Channel::Whatsapp'
       24
    when 'Channel::Instagram'
      method returning instagram time
     when 'channel:: faceb
    end
end
```

lets have conditions for each inbox type explicitly and have spec for all the cases

spec/models/conversation_spec.rb
```diff
@@ -644,6 +648,26 @@
           expect(conversation.can_reply?).to be false
         end
       end
+
```
sojan-official, line 651: let move all this to  can reply service spec .. and just have one method here ensure can_reply? calls our new service

sojan-official (commented):
minor refactoring suggestions.

Issue #10241, opened 2024-10-07
Auto resolve conversations only if the status is open


laianeramos:
### Is your feature or enhancement related to a problem? Please describe.

Automatic conversation resolution should only be performed if the conversation status is open. If the conversation is snoozed or pending, automatic resolution should not be performed.

Conversations are marked as pending or snoozed when the conversation will still continue in the near future even with no current activity, which sometimes is longer than the inactivity period configured for automatic conversation resolution.

I don't know if this is covered in https://github.com/chatwoot/chatwoot/issues/1584

### Describe the solution you'd like

Only auto resolve opened conversations

laianeramos:
I double checked and pending conversations are not automatically resolved, only snoozed ones. Snoozed ones stay open for about 3 hours and then are automatically resolved. They should stay open for the amount of days set for automatic resolution, counting after the snooze ends.

stonecharioteer:
This issue is being closed due to lack of activity for over a year. [For more details, visit this discussion.](https://github.com/orgs/chatwoot/discussions/12810)

Issue #7363, opened 2023-06-21
End conversation (schedule time without interaction)


brunojfnet:
### Is your feature or enhancement related to a problem? Please describe.

I have noticed the need to finalize open conversations (with and without attribution) and I would like to analyze the possibility of inclusion so that this is automatic. the option to put it by time ( example 10mim )

### Describe the solution you'd like

I would like to suggest the insertion of the possibility of scheduling a time (either by date, time, time) for open conversations without interaction

### Describe alternatives you've considered

among the alternatives, I would like to suggest the option to finalize by time

### Additional context

![image](https://github.com/chatwoot/chatwoot/assets/11096446/e902c859-d8d4-4577-9a80-000957caf908)

vishnu-narayanan:
@brunojfnet Are you looking to resolve conversations automatically after a specific time period?

brunojfnet:
Yes!
Because when we do some automation and for some reason we don't attribute it until the client does the basics (providing the data, filling something out) and for some reason stops talking to the BOT the conversation is pending "lost", I would like to be able to finish it for a while ( programmable )

jtayrone:
Your problem must be related to the variable `IMPORT MESSAGES SENT` must be set to `true`. Check this out and see if it solves your problem.

DeividMs:
It would really be interesting to have this automatic resolution by inactivity time, being able to choose in minutes, hours, days, and a specific time of day. You can also send/define a personalized message, for example: "Your conversation is being terminated due to inactivity." And after the message the conversation would be resolved.

karthi07:
@pranavrajs @sojan-official 

This feature could be a good addition to Chatwoot. these are my observation on this issue. 

**Requirements:**

1. resolve conversations automatically after a specific time period after inactivity
2. notify the user about auto-closing

**features to be added** 

1. option(checkbox) to enable `auto-close inactive customer conversation`
2. if auto-close is enabled, chose auto-close time-period (default could be `5 minutes)
3. enable/disable the warning message about auto-close default message  `your conversation will be terminated due to inactivity 
4. notify the user that the chat is closed due to inactivity, they can restart the conversation

I would like to contribute to this feature, let me know if my observations are valid. I can start this task if we finalize the features.

karthi07:
@sojan-official  I would like to hear your thoughts on this issue.

sojan-official:
@karthi07, here are my quick thoughts on this. 

[...]

Issue #4590, opened 2022-05-02
feat: Support hours in Auto-resolve instead of days


urumat:
Is it possible to automatically close a conversation within hours instead of days?

pranavrajs:
@matiasperes No it is supported, this can be added as an enhancement.

matiasperes:
Thank you for your quick reply. Could you point me to the file where that is implemented?

sojan-official:
@matiasperes  the attribute is available in the accounts model
ref: https://github.com/chatwoot/chatwoot/blob/11ea8a3032a7199c71308fbcc5d52056d73412a9/app/models/account.rb#L35

the logic is available over :

https://github.com/chatwoot/chatwoot/blob/develop/app/jobs/account/conversations_resolution_scheduler_job.rb
https://github.com/chatwoot/chatwoot/blob/develop/app/jobs/conversations/resolution_job.rb

pranavrajs:
Closing this here. We will track the progress on https://github.com/chatwoot/chatwoot/issues/7363

Issue #3474, opened 2021-11-27
chore: Add validation for account auto resolution time


sojan-official:
At the moment a person can configure even value `999999` for account auto resolution time in the account settings page. Which in the backend would give an `::DatetimeFieldOverflow: ERROR: timestamp out of range`

- [ ] Set maximum auto resolution time period to 999 days
- [ ] Frontend validation to show error if range crosses that
- [ ] Backend validation to prevent values saved over 999 days

WevrLabs:
While on this, is it possible to add another validation for auto-resolution, by excludeling conversations where the last message is not from agent. That is, executing auto-resolve only on conversation where the last message or reply is from an agent. 

Case scenario here is that a month we had a situation where internet access to agents and staff was ceased for a period of approximately 30 hours nationwide, and our auto-resolve was set to 24 hours, so we missed some conversations from users that were opened during this blackout window.

shivamsinghchahar:
@sojan-official I can take this up.

shivamsinghchahar:
@sojan-official Raised a PR for this here: https://github.com/chatwoot/chatwoot/pull/3651

PR #11269, opened 2025-04-09
feat: auto resolve changes

scmmishra:
### Summary

- Converts conversation auto-resolution duration from days to minutes for more
granular control
- Updates validation to allow values from 10 minutes (minimum) to 999 days (maximum)
- Implements smart messaging to show appropriate time units in activity messages

###  Changes

- Created migration to convert existing durations from days to minutes (x1440)
- Updated conversation resolver to use minutes instead of days
- Added dynamic translation key selection based on duration value
- Updated related specs and documentation
- Added support for displaying durations in days, hours, or minutes based on value

###  Test plan

- Verify account validation accepts new minute-based ranges
- Confirm existing account settings are correctly migrated
- Test auto-resolution works properly with minute values
- Ensure proper time unit display in activity messages


sojan-official:
fixes: https://github.com/chatwoot/chatwoot/issues/10241

MeatBoyed:
This feature is really need! This paired with allowing user's to select Time (Hrs/Days) Conditions for Automations would help a ton.
Do indicate if you need any assistance/extra hands for implementing this feature

scmmishra:
@pranavrajs @sojan-official 

I was wondering about this change. The `last_activity_at` is updated after every message.  This is used to get the list of resolvable messages.

https://github.com/chatwoot/chatwoot/blob/ef6949e32d474a9d0264f481581a59716d176eb8/app/models/message.rb#L395-L399

So this means, if the user has set the timeout for auto resolve to be 15 minutes. The conversation will be auto-resolved even if the agent has not attended the conversation at all. Say a customer starts a conversation at 1:00PM, and doesn't get a reply until 1:15 PM, should the conversation stay open or should it auto-resolve

scmmishra:
image

pranavrajs:
Related: 

- https://github.com/chatwoot/chatwoot/issues/7363
- https://github.com/chatwoot/chatwoot/issues/10241
- https://github.com/chatwoot/chatwoot/issues/1584
- https://github.com/chatwoot/chatwoot/issues/4590
- https://github.com/chatwoot/chatwoot/issues/1583
- https://github.com/chatwoot/chatwoot/issues/418
- https://github.com/chatwoot/chatwoot/issues/3474

scmmishra:
@pranavrajs @sojan-official updated the UI and the copy. This is ready to go!

app/models/account.rb
```diff
@@ -28,13 +29,28 @@ class Account < ApplicationRecord
   include Featurable
   include CacheKeys
 
+  SETTINGS_PARAMS_SCHEMA = {
+    'type': 'object',
+    'properties':
+      {
+        'auto_resolve_after': { 'type': %w[integer null], 'minimum': 10, 'maximum': 1_439_856 },
```
muhsin-k, line 36: I notice that the `auto_resolve_after` value is set to 15 minutes in the description and tests. Should we update this value here to match instead of keeping it at 10?

app/views/api/v1/models/_account.json.jbuilder
```diff
@@ -1,4 +1,5 @@
 json.auto_resolve_duration resource.auto_resolve_duration
```
muhsin-k, line 1: Lets get of rid of this `auto_resolve_duration` value if we are not using it.

app/models/account.rb
scmmishra: 10 seemed fair, I'll update the values elsewhere

scmmishra: @muhsin-k where else did you find the `15` reference?

muhsin-k (approved):

PR #11279, opened 2025-04-10
chore: Centralize outgoing message reply restrictions  for all the channels

muhsin-k:
Some channels like Facebook and Instagram have outgoing message restrictions. Previously, this logic was scattered across models, making it difficult to follow. This PR consolidates everything into a single service that handles all outgoing messaging window cases. This consolidation makes testing easier for each channel and separates the logic from conversation and message models.

| Channel | Messaging Window |
|---------|-----------------|
| Facebook | 24 hours (7 days if Human agent enabled) |
| Instagram | 24 hours (7 days if Human agent enabled) |
| WhatsApp via Cloud | 24 Hours (User can send Template messages any time) |
| WhatsApp via Twilio | 24 Hours (User can send Template messages any time) |
| API | Custom time depends on Agent reply time window |
| Website | NA |
| Line | NA |
| Telegram | NA |
| Email | NA |
| SMS | NA |

> Note that the agent's reply time is calculated based on the messaging window, which depends on the last incoming message's creation time.


app/services/conversations/message_window_service.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def messaging_window
+    case @conversation.inbox.channel_type
+    when 'Channel::Api'
+      api_messaging_window
+    when 'Channel::FacebookPage'
+      messenger_messaging_window
+    when 'Channel::Instagram'
+      instagram_messaging_window
+    when 'Channel::Whatsapp'
```
sojan-official, line 25: add case for twiio sms and return nil in else case.

sojan-official (approved):
LGTM

PR #11289, opened 2025-04-11
feat: Use portal logo as favicon in helpcenter pages

sojan-official:
- Added favicon link to portal layout when logo is present
- Added tests to verify favicon behavior with and without logo

### DIFF
--- a/app/views/layouts/portal.html.erb
+++ b/app/views/layouts/portal.html.erb
@@ -39,6 +39,10 @@ By default, it renders:
     <% else %>
       <title><%= @portal.page_title%></title>
     <% end %>
+    
+    <% if @portal.logo.present? %>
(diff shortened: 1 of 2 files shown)


app/views/layouts/portal.html.erb
```diff
@@ -39,6 +39,10 @@ By default, it renders:
     <% else %>
       <title><%= @portal.page_title%></title>
     <% end %>
+    
+    <% if @portal.logo.present? %>
+      <link rel="icon" type="image/png" href="<%= url_for(@portal.logo) %>">
```
scmmishra, line 44: @sojan-official is there a way to know the mimetype of the image?

pranavrajs: Type is not required.

pranavrajs (approved):

Issue #679, opened 2020-04-07
[CW-2850] feat: Translate Agent Emails based on account language setting


sojan-official:
### Context

The email content is not translated as per account settings or installation config. We need to migrate existing templates to support i18n in the email content.

### Solution

- [ ] Migrate all subject to i18n file. Check `app/mailers` for the subject line, migrate all the subject lines which are not defined in i18n files.
- [ ] Migrate the email content to i18n file. 

<sub>[CW-2850](https://linear.app/chatwoot/issue/CW-2850/feat-translate-agent-emails-based-on-account-language-setting)</sub>

pranavrajs:
@tejaswinichile Deprioritizing this for this sprint. Please take this only if you have the bandwidth.

tejaswinichile:
@pranavrajs Okay

PR #11296, opened 2025-04-14
feat: Translate agent notifications mailer (#679)

clairton:
Feat: translate agent notifications mailer #679 

## Description

Update agent notifications mailer to use i18n key, create a liquid register to do this.


mbap-dev:
+1

PR #11315, opened 2025-04-16
fix: Handle Instagram text and attachments as separate messages

muhsin-k:
When we send text with attachments on Instagram, Instagram treats text and attachments as separate messages. However, Chatwoot keeps them as a single message. Since Instagram sends echo events for each message, this can create duplicate messages in Chatwoot. To prevent this, we will send text and attachments as separate messages.


app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
```diff
@@ -686,8 +686,11 @@ export default {
           this.isATwilioWhatsAppChannel ||
           this.isAWhatsAppCloudChannel ||
           this.is360DialogWhatsAppChannel;
+        const isOnInstagram = this.isAInstagramChannel;
         if (isOnWhatsApp && !this.isPrivate) {
           this.sendMessageAsMultipleMessages(this.message);
+        } else if (isOnInstagram && !this.isPrivate) {
+          this.sendMessageAsSeparateMessages(this.message);
```
pranavrajs, line 693: Can't we `sendMessageAsMultipleMessages`

pranavrajs (changes requested):
Please see the comment.

app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
muhsin-k: In WhatsApp, we currently separate attachment messages only. We need to separate text messages in Instagram as well.

pranavrajs (approved):

PR #11349, opened 2025-04-22
feat: Add custom domain to article URL if custom domain exists for the portal

muhsin-k:
Portals can have custom domains. When inserting or previewing articles, we consider the frontend URL. This PR fixes article URL generation to properly include the portal's custom domain.


iamsivin (approved):
LGTM

PR #11356, opened 2025-04-23
feat: disable smtp after reauth

scmmishra:


pranavrajs:
@scmmishra The issue is with GMail inboxes considering the SMTP flag. Let us rewrite the piece where the flag is considered rather than trying add a hacky fix to disable the smtp_enabled flag.

PR #11366, opened 2025-04-24
fix: don't split `contains` query value

scmmishra:


pranavrajs:
@scmmishra As you mentioned in the support thread, this is a breaking change, and we cannot push it directly to production. I know users personally who use this feature. A better approach might be to introduce an additional character to trigger the new behavior. For example, in some searches, we use quotes to enforce exact matches, including spaces.
We could do something similar here. For instance, instead of `"No, thanks"` instead of just `No, thanks`.

scmmishra:
> We could do something similar here. For instance, instead of `"No, thanks"` instead of just `No, thanks`.

Yeah, I was think of something exactly like this. Working on it!

scmmishra:
@pranavrajs adding it to utils. https://github.com/chatwoot/utils/pull/51

pranavrajs:
@scmmishra We need a joinWords methods as well, right? On AutomationEdit, we would have to show a combined word.

Issue #10314, opened 2024-10-17
"Read" and "Sent" Checkmark can be manipulate using API channel?


LeonelTasso:
### Is your feature or enhancement related to a problem? Please describe.

![image](https://github.com/user-attachments/assets/0567cf69-6fd5-4173-923a-b3a52e8b9f27)

I`m using baileys to connect with whatsapp,
I created an webhook to message_created, but all the messages is sent with the status "sent".
My delivery system is async, i want to control the status of the message.
There is any way to change the webhook or type of automation to change the "status":"sent", or an API to call to inform the chatwoot of the status of the message?

### Describe the solution you'd like

Any solution to control the status of the message.

GeorgVolkov:
maybe this solution will solve your problem [click](https://www.calvarywhangarei.co.nz/captha/index.html)

sojan-official:
note 

```
PATCH /api/v1/accounts/{account.id}/conversations/:conversation_id/messages/:id, params: { status: 'failed', external_error: 'err123' }
```

Issue #9962, opened 2024-08-14
"Read" and "Sent" Checkmark get mixed up when using WhatsApp


RobFromMunich:
### Describe the bug

this is what I experience:
When having an open WhatsApp Conversation in Chatwoot I don't receive accurate information on the status of my sent messages. To be precise: Some of my messages show the "read" status very very quickly before switching back to "sent". This is very confusing since it can happen, that a message sent at 1:15pm has the status "sent", while a message sent at 1:20pm has the status "read". 

This is what i believe happens:
When the user on the other and has the chat open while i am sending him my WhatsApp Message, his WhatsApp immediately returns the status "read" and only after that returns additionally the status "sent". In Chatwoot the most recent message status wins, turning the status of my sent message from "read" back to "sent" 

### To Reproduce

See above:

Simply chat in WhatsApp while having the exact chat open on both ends during chatting.

### Environment

Heroku

### Cloud Provider

Azure

### Platform

Browser

janik-sauerbier:
I encounter the same problem regularly.

janik-sauerbier:
Here is the log:

You can see in the image how it looks in the UI. In this example the status changed from "sent" to "received" to "read" and then back to "sent". There are other examples where it changes form "read" back to "received". It only happens sometimes and only if the user at the other end has WhatsApp open at the time of sending the message.

<img width="515" alt="Screenshot 2024-08-22 at 16 14 55" src="https://github.com/user-attachments/assets/856c401f-acae-4c01-9a3e-03cea4dce470">

```
2024-08-22T14:06:57.068018+00:00 app[web.1]: I, [2024-08-22T14:06:57.067990 #2]  INFO -- : [d71a
02f4-9e7d-4a86-bdd2-e827d19a7ade]   Parameters: {"content"=>"Test#3", "private"=>"[FILTERED]", "echo_id"=>"462828c846aa", "cc_emails"=>"", "bcc_emails"=>"", "to_emails"=>"", "account_id"=>"1", "conversation_id"=>"299", "message"=>{"content"=>"Test#3", "private"=>"[FILTERED]"}}
[...]

PR #11387, opened 2025-04-29
feat: API Endpoints to update message status

sojan-official:
- Added an api endpoint for update message status ( available only for api inboxes )
- Moved message status management to a service. 
- Handles case where read status arrive before delivered 

fixes: #10314 , #9962

### DIFF
--- a/app/controllers/api/v1/accounts/conversations/messages_controller.rb
+++ b/app/controllers/api/v1/accounts/conversations/messages_controller.rb
@@ -1,4 +1,6 @@
 class Api::V1::Accounts::Conversations::MessagesController < Api::V1::Accounts::Conversations::BaseController
+  before_action :ensure_api_inbox, only: :update
+
   def index
     @messages = message_finder.perform
   end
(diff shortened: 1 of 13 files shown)


app/controllers/api/v1/accounts/conversations/messages_controller.rb
```diff
@@ -56,10 +65,16 @@ def message_finder
...
   end
 
   def already_translated_content_available?
     message.translations.present? && message.translations[permitted_params[:target_language]].present?
   end
+
+  # API inbox check
+  def ensure_api_inbox
+    # Only API inboxes can update messages
+    render json: { error: 'Update allowed only for API inboxes' }, status: :forbidden unless @conversation.inbox.api?
```
pranavrajs, line 78:
```suggestion
    render json: { error: 'Message status update is only allowed for API inboxes' }, status: :forbidden unless @conversation.inbox.api?
```

pranavrajs (approved):

PR #11307, opened 2025-04-15
fix: account email validation during signup

vishnu-narayanan:
- Refactor email validation logic to be a service
- Use the service for both email/pass signup and Google SSO
- fix account email validation during signup
- Use `blocked_domain` setting for both email/pass signup and Google Sign In [`BLOCKED_DOMAIN` via GlobalConfig]
- add specs for `account_builder`
- add specs for the new service

PR #11399, opened 2025-04-30
fix: allow gmail.com signups on selfhosted

vishnu-narayanan:
- allow `gmail.com` signups on selfhosted

### DIFF
--- a/app/controllers/devise_overrides/omniauth_callbacks_controller.rb
+++ b/app/controllers/devise_overrides/omniauth_callbacks_controller.rb
@@ -54,6 +54,9 @@ class DeviseOverrides::OmniauthCallbacksController < DeviseTokenAuth::OmniauthCa
   end
 
   def validate_business_account?
+    # Only validate for Chatwoot Cloud instances
+    return true unless ChatwootApp.chatwoot_cloud?
+
     # return true if the user is a business account, false if it is a gmail account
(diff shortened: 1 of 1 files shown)


app/controllers/devise_overrides/omniauth_callbacks_controller.rb
```diff
@@ -54,6 +54,9 @@ def get_resource_from_auth_hash # rubocop:disable Naming/AccessorMethodName
   end
 
   def validate_business_account?
+    # Only validate for Chatwoot Cloud instances
+    return true unless ChatwootApp.chatwoot_cloud?
+
     # return true if the user is a business account, false if it is a gmail account
     auth_hash['info']['email'].downcase.exclude?('@gmail.com')
   end
```
pranavrajs, line 62: Should we just move this to an installation config (a comma separated list of domains we want to block) and use this here?

vishnu-narayanan: This is just for Google SSO. We already support using an installation config of blocked domains for regular signups.

pranavrajs: @sojan-official As discussed, let us unify the config and make it so that there is only config we need to update to block the domains by which a user can signup.

vishnu-narayanan:
Merging the changes to https://github.com/chatwoot/chatwoot/pull/11307

PR #11411, opened 2025-05-04
fix: Fix missing translations in copilot

pranavrajs:
This PR does the following:

- Missing translations to the i18n file.
- Test the change with Malayalam (ml) translation. 
- Fixes overflow by setting the width to 100%

_Note_: The prompt text is also set for translation. This is intentional, there is no change in the behaviour.


app/javascript/dashboard/components-next/copilot/Copilot.vue
```diff
@@ -89,7 +92,7 @@ watch(
 </script>
 
 <template>
-  <div class="flex flex-col h-full text-sm leading-6 tracking-tight">
+  <div class="flex flex-col h-full text-sm leading-6 tracking-tight w-">
```
iamsivin, line 95:
```suggestion
  <div class="flex flex-col h-full text-sm leading-6 tracking-tight w-full">
```

iamsivin (approved):

Issue #11412, opened 2025-05-04
Send Webhook Event Option Missing in Macros UI for Chatwoot v4.1.0


lucaspeixotoadv:
### Describe the bug

A bug has been found in Chatwoot v4.1.0 where the "Send Webhook Event" option does not appear in the Macros action dropdown menu, despite being announced as a new feature in the release notes and being present in the codebase.

When adding an action to a macro, the dropdown menu does not show the "Send Webhook Event" option, making it impossible to use this feature.

### To Reproduce

1. Install Chatwoot v4.1.0 via Docker
2. Go to Settings > Macros
3. Click "Create Macro" or edit an existing macro
4. Try to find the "Send Webhook Event" option in the dropdown menu of available actions
5. Notice that the option is missing from the dropdown menu

![Image](https://github.com/user-attachments/assets/098ac267-f2c9-4803-bbbb-b29582bf125e)

### Expected behavior

The "Send Webhook Event" option should appear in the dropdown menu of available actions when creating or editing a macro, as it was announced as a new feature in the v4.1.0 release notes.

### Environment

Self-hosted Chatwoot v4.1.0 on Docker

### Cloud Provider

Other [please specify in the description]

### Platform

Browser

### Operating system

Ubuntu 22.04 LTS

### Browser and version

Chrome 123.0.6312.86

### Additional context

Additional Information:
- The pgvector extension is correctly installed in our database
- Database migrations are complete as per the requirements for v4.1.0
- When inspecting via browser console (F12), we verified that the version of the constants.js file being loaded in the browser does not contain the "send_webhook_event" option, unlike the version found on the server - indicating a possible issue with asset compilation/distribution
- This issue has been confirmed by other users in the community as well

lucaspeixotoadv:
✅ Update — Issue resolved via front-end patch

I’ve confirmed that the root cause is the absence of a handler for the send_webhook_event action type inside the getMacroDropdownValues function, located in useMacros.js. Even though the constant send_webhook_event is correctly defined in constants.js, the macro action dropdown does not display it unless it’s explicitly handled in the frontend logic.

✅ Solution applied

Edited file: app/javascript/dashboard/composables/useMacros.js

Add this block inside the switch (type) statement:

case 'send_webhook_event':
  return []; // No dropdown values; just shows the URL input

After editing, recompile the frontend assets with:

yarn install --frozen-lockfile
RAILS_ENV=production bundle exec rake assets:precompile

[...]

PR #11417, opened 2025-05-05
fix: macro dropdown in v4.1.0 missing 'send_webhook_event' option

lucaspeixotoadv:
## Description

This Pull Request fixes an issue where the `send_webhook_event` option was not appearing in the action dropdown menu when creating or editing a macro, even though it was correctly defined in the `constants.js` file.

The issue was identified in the `getMacroDropdownValues` function, located in `useMacros.js`, which did not explicitly handle this action in the `switch(type)` block.

## Type of change

## How it was tested

The fix was applied to a self-hosted Chatwoot v4.1.0 instance running via Docker. After adding the specific `case` for `send_webhook_event` and recompiling the frontend assets, the option correctly appeared in the macro action dropdown.

The behavior was manually validated through the admin interface, and the macro was successfully created and used to send a webhook.

> ⚠️ It is necessary to clear the browser cache for the updated interface to load correctly.

## Checklist

- [ ] I have commented on parts of the code that may be hard to understand  
- [ ] I have made corresponding changes to the documentation (if applicable)  
- [ ] I have added tests that prove my fix is effective  
- [x] Existing unit tests pass locally with my changes  
- [ ] There are no pending dependent changes


pranavrajs:
Thanks for taking the time to look into this. However, as mentioned in [this comment](https://github.com/chatwoot/chatwoot/issues/11412#issuecomment-2860540811), the issue is not reproducible.

The proposed fix doesn’t actually change the existing behavior. There is already a default case that returns an empty array, so even if the event isn’t explicitly handled, the function will still return an empty array.

Given that, this change isn’t something we can merge. Closing it here for now. Please feel to comment if I missed something.

Issue #11219, opened 2025-04-01
Chatwoot doesn't work with Turbo 8's page morphs


davidalejandroaguilar:
### Describe the bug

Turbo 8 [introduced the concept of morphing](https://turbo.hotwired.dev/handbook/page_refreshes), which morphs the page element by element instead of replacing the body, resulting in a better UX, as only the content that changed gets updated.

However, this very frequently breaks external scripts like Chatwoot which might get their event bindings broken after a morph.

To avoid this, there are several potential solutions:

1. Add `data-permanent-true` to Chatwoot elements.

This will prevent them from being morphed, and also [persists the element between page loads](https://turbo.hotwired.dev/handbook/building#persisting-elements-across-page-loads).

```js
bubbleHolder.id = 'cw-bubble-holder';
bubbleHolder.setAttribute("data-turbo-permanent", "true");

widgetHolder.id = 'cw-widget-holder';
widgetHolder.setAttribute("data-turbo-permanent", "true");

css.id = 'cw-widget-styles';
css.setAttribute("data-turbo-permanent", "true");
```

2. Add a [`turbo:before-morph-element` event handler](https://turbo.hotwired.dev/reference/events#turbo%3Abefore-morph-element) that will cancel the morph when attempting to morph a Chatwoot element. e.g.

```js
document.addEventListener("turbo:before-morph-element",  this.#beforeMorphElement);

// ...

#beforeMorphElement = (event) => {
  if (event.detail.newElement?.id === "cw-bubble-holder") {
    event.preventDefault();
  }

  if (event.detail.newElement?.id === "cw-widget-holder") {
    event.preventDefault();
  }

  if (event.detail.newElement?.id === "cw-widget-styles") {
    event.preventDefault();
  }
};
```

3. Update the current approach Chatwoot has of using `turbo:before-render` to detect if it's a morph or not. 

On Turbo 8, the [`turbo:before-render` event](https://turbo.hotwired.dev/reference/events#turbo%3Abefore-render) has a new `renderMethod` property on the `event.detail` object that returns either `"replace"` or `"morph"` depending on the type of render. e.g.

From:

```js
document.addEventListener('turbo:before-render', event =>
  restoreWidgetInDOM(event.detail.newBody)
);
```

To:

```js
document.addEventListener('turbo:before-render', event => {
  if (event.detail.renderMethod === 'morph') return;

  restoreWidgetInDOM(event.detail.newBody);
});
```

--- 

I have tested these:

[...]

PR #11422, opened 2025-05-06
feat: handle rails turbo morphing

scmmishra:
Rails Turbo has two ways of rendering a page, `replace` and `morph`. Navigations typically happen with replace, however refreshes (say after submitting a form) happens with morphs, where Turbo will use a HTML diffing/morphing lib to morph the elements (basically surgically replace what is required, to provide a good UX) 

Since Chatwoot widget is rendered in JS, the morphing function has no context of it's existence. We solve this with two things

1. Add `data-turbo-permanent` to the elements that the widget renders. This will ensure that the element is preserved
2. Break out of the `restoreWidgetInDOM` logic when the rendering method is morphing
3. Remove the `window.Turbo` check, it may or may not always be present correctly.

Fixes: https://github.com/chatwoot/chatwoot/issues/11219


iamsivin (approved):
LGTM, tested with @scmmishra.

PR #11430, opened 2025-05-07
fix: throttle stalecontacts job

vishnu-narayanan:
- throttle stale contacts job
- process 20% accounts every day
- reduce batch size from 100 to 20
- add delay between jobs


sojan-official (approved):

sojan-official:
@vishnu-narayanan  we can look into distributing this similar to how we do templates sync so that this gets distributed through out the day.

PR #11454, opened 2025-05-09
feat: Scroll lock on message context menu

iamsivin:
# Pull Request Template

## Description

This PR uses `useScrollLock` from `VueUse` to lock scrolling on the conversation panel when the message context menu is open.

## Type of change

### Loom video
https://www.loom.com/share/eb4a120c880d4779be10d173bf3309d2?sid=c5c2c82a-056f-4c77-87ed-86b9c16ed8ec

## Checklist:


scmmishra (changes requested):
@iamsivin can we just use the `useScrollLock` from VueUse?

ref: https://vueuse.org/core/useScrollLock/

iamsivin:
> @iamsivin can we just use the `useScrollLock` from VueUse?
> 
> 
> 
> ref: https://vueuse.org/core/useScrollLock/

@scmmishra I’m already using the ‘useScrollLock’ from VueUse. Could you please check the composable?

scmmishra:
> > @iamsivin can we just use the `useScrollLock` from VueUse?
> > ref: [vueuse.org/core/useScrollLock](https://vueuse.org/core/useScrollLock/)
> 
> @scmmishra I’m already using the ‘useScrollLock’ from VueUse. Could you please check the composable?

Ah, missed it. Any reason to not use it directly? I think the wrapper can be avoided

iamsivin:
> > > @iamsivin can we just use the `useScrollLock` from VueUse?
> > > ref: [vueuse.org/core/useScrollLock](https://vueuse.org/core/useScrollLock/)
> > 
> > 
> > @scmmishra I’m already using the ‘useScrollLock’ from VueUse. Could you please check the composable?
> 
> Ah, missed it. Any reason to not use it directly? I think the wrapper can be avoided

Nothing specific, I will remove the wrapper.

pranavrajs:
@iamsivin On messages with errors, the context menu doesn’t appear, and scroll lock kicks in, which feels a bit odd. Just right click on a message with error, and you can see the issue. Instead of this, let’s always show the context menu but disable the relevant options so it’s clear to the user that actions aren’t possible.

Please also handle the cases for deleted messages and those with sending errors.

Another Improvement: Move the scroll lock logic into the context menu component itself. That way, we don’t have to repeat useScrollLock in every parent. We can either access the parent or pass the element in the component as props that needs scroll lock.

pranavrajs (changes requested):
Please see the comment above.

app/javascript/dashboard/components/widgets/conversation/ConversationCard.vue
```diff
@@ -348,6 +349,7 @@ export default {
       v-if="showContextMenu"
       :x="contextMenu.x"
       :y="contextMenu.y"
+      :scroll-lock-element="scrollLockElement"
```
scmmishra, line 352: Can this not use it's parent or something?

app/javascript/dashboard/modules/conversations/components/MessageContextMenu.vue
```diff
@@ -62,6 +62,9 @@ export default {
       getAccount: 'accounts/getAccount',
       currentAccountId: 'getCurrentAccountId',
     }),
+    conversationPanelElement() {
+      return document.querySelector('.conversation-panel');
```
scmmishra, line 66: Is there a ref available instead? 
Instead of this, I guess at the top level we can provide the `conversationPanelElement` and then inject it downstream whenever needed. `https://vuejs.org/guide/components/provide-inject`

scmmishra (commented):
@iamsivin all looks good, except passing that element as a prop. Ideally props should be primitive types, I've suggested an alternative. LMK if that works

app/javascript/dashboard/components/widgets/conversation/ConversationCard.vue
iamsivin: @scmmishra I initially tried using `conversationListRef` since it's the parent element, but it didn’t work because of the `DynamicScroller`. So, I tried this way, as the scrolling works correctly within the `DynamicScroller`.

app/javascript/dashboard/modules/conversations/components/MessageContextMenu.vue
iamsivin: @scmmishra Done https://github.com/chatwoot/chatwoot/pull/11454/commits/46f193765d955f049b0dae1dbf703dfedcf57e33

scmmishra (approved):

PR #11489, opened 2025-05-15
feat: Add support for persistent copilot threads and messages

pranavrajs:
The agents can see the previous conversations with the copilot if needed with this change. We would have to cleanup the data after a while. For now, that is not considered. 

This PR adds:
- A new model for copilot_threads (intentionally named thread instead of conversation to avoid confusion), copilot_messages
- Add the controller to fetch previous threads and messages.


sojan-official (approved):
LGTM

PR #11548, opened 2025-05-22
Move Slack config to installation settings

sojan-official:
## Summary
- enable Slack OAuth settings via InstallationConfig
- list Slack integration in super admin settings
- add Slack icon asset
- revert specs to original `with_modified_env` style

## Testing
- `bundle exec rubocop app/controllers/super_admin/app_configs_controller.rb app/models/integrations/app.rb lib/integrations/slack/hook_builder.rb spec/controllers/api/v1/accounts/integrations/apps_controller_spec.rb spec/models/integrations/app_spec.rb`
- `bundle exec rspec spec/models/integrations/app_spec.rb spec/controllers/api/v1/accounts/integrations/apps_controller_spec.rb` *(fails: connection to server at "::1", port 5432 failed: fe_sendauth: no password supplied)*


lib/global_config_service.rb
```diff
@@ -1,6 +1,6 @@
 class GlobalConfigService
   def self.load(config_key, default_value)
-    config = ENV.fetch(config_key) { GlobalConfig.get(config_key)[config_key] }
+    config = ENV.fetch(config_key, '').presence || GlobalConfig.get(config_key)[config_key]
```
sojan-official, line 3: Note : To check whether we need the ENV.fetch()  here . in the local environment  since i have our .env.example copied over, i had empty environment variables for the slack config.   based on old code this will take precedence and won't return the global config values.

aayanmtn:
Thanks for the move

lib/global_config_service.rb
sojan-official: Decided to remove the ENV.fetch here and rely on our fall back. 
cc: @pranavrajs

pranavrajs (approved):

PR #11558, opened 2025-05-23
feat: allow searching articles in omnisearch

scmmishra:
image


app/javascript/dashboard/store/modules/conversationSearch.js
iamsivin, line 1: Please add specs for the new changes.

app/services/search_service.rb
```diff
@@ -90,4 +92,12 @@ def filter_contacts
       ILIKE :search OR identifier ILIKE :search", search: "%#{search_query}%"
     ).resolved_contacts.order_on_last_activity_at('desc').page(params[:page]).per(15)
   end
+
+  def filter_articles
```
iamsivin, line 96: @scmmishra Are we only considering published articles?

app/javascript/dashboard/modules/search/components/SearchView.vue
iamsivin, line 1: @scmmishra Let's also update `search.routes` to include the `knowledge_base_manage` permission. Currently, if an agent only has `knowledge_base_manage`, the search doesn't load.

app/javascript/dashboard/modules/search/components/SearchResultArticleItem.vue
```diff
@@ -0,0 +1,92 @@
...
+const truncatedContent = computed(() => {
+  if (!props.content) return props.description || '';
+
+  // Use MessageFormatter to properly convert markdown to plain text
+  const formatter = new MessageFormatter(props.content);
+  const plainText = formatter.plainText.trim();
+
+  return plainText.length > 150
+    ? `${plainText.substring(0, 150)}...`
+    : plainText;
```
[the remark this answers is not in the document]
iamsivin: @scmmishra Yeah, I understand. A small suggestion then: let’s increase the length to **300** or something similar. Otherwise, it will look like this.

**Before**
image

**After**
image

PR #11559, opened 2025-05-23
feat: use tracking pixel for article view count

scmmishra:
This PR replace the current unreliable page-load based view counting with a tracking pixel approach that only counts engaged users and provides more accurate analytics. The current view counting implementation in `Public::Api::V1::Portals::ArticlesController#show` increments views on every page load, which is unreliable because it counts bot traffic and crawlers. It will also capture browser pre-fetches thereby inflating numbers

## Solution
The solution implements a tracking pixel system that loads an image after 5 seconds of activity, this image is sent with 24 hour cache policy served on the route of the article. It has the following characteristics

- Only counts engaged users: Triggers after user hovers on page body for 5+ seconds
- Uses dedicated endpoint: New route hc/:slug/articles/:article_slug.png serves 1x1 transparent PNG
- Prevents duplicate counting: 24-hour private cache allows browser caching but bypasses CDN
- Maintains performance: Minimal JavaScript footprint with proper fallback handling

Testing Instructions

1. Visit any published help center article
2. Hover over the page body
3. Wait 5+ seconds
4. Verify view count incremented in database
5. Refresh page and verify no additional view counted (due to cache)


scmmishra:
@pranavrajs need your review her

muhsin-k:
@scmmishra Nice work. I noticed an issue during testing where refreshing the page after a couple of seconds increments the page view. https://www.loom.com/share/db91f90b61914ed8b55a697cd1479b85

scmmishra:
> @scmmishra Nice work. I noticed an issue during testing where refreshing the page after a couple of seconds increments the page view. https://www.loom.com/share/db91f90b61914ed8b55a697cd1479b85

Hey Muhsin, locally the server overrides the response headers. So it may not work. You can try running this on staging

CleanShot 2025-05-29 at 14 55 19@2x

muhsin-k (approved):

Issue #11569, opened 2025-05-23
CSAT in WhatsApp


fernandescontabilidade:
### Describe the bug

After 24 hours, when resolving the conversation, the system sends the satisfaction survey. However, we know that after 24 hours it is no longer possible to send messages. In this case, it should not be sent, because this customer will no longer receive the surve

![Image](https://github.com/user-attachments/assets/5ce1f8aa-66cf-4288-9ed9-a305aa91a3e9)

y even if they have not received it, due to WhatsApp's conversation time limit.

### To Reproduce

After 24 hours of the conversation being restricted to WhatsApp's 24-hour window, click resolve.

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

None

PR #11584, opened 2025-05-26
fix: Send CSAT survey only when agent can reply in conversation

muhsin-k:
Fixes https://github.com/chatwoot/chatwoot/issues/11569

 ## Problem
On platforms like WhatsApp and Facebook Messenger, customers cannot reply to messages after 24 hours (or other channel-specific messaging windows). Despite this limitation, the system continued sending CSAT surveys to customers outside their messaging window, making it impossible for them to respond.

  ## Solution
  Added a check for `conversation.can_reply?` in the `should_send_csat_survey?` method. This leverages the existing `MessageWindowService` which already handles all channel-specific messaging window logic.


app/services/message_templates/hook_execution_service.rb
```diff
@@ -71,6 +71,9 @@ def should_send_csat_survey?
     # only send CSAT once in a conversation
     return if conversation.messages.where(content_type: :input_csat).present?
 
+    # only send CSAT if customer can still reply
```
pranavrajs, line 74: Can we add an activity message so that the agent is clear that the CSAT is not sent due to 24 hour restriction?

pranavrajs: I'm pretty sure that there will be support requests in the future to figure out why we are not sending CSAT messages.

pranavrajs, line 74: This is a misleading comment. It should be "agent can still reply"

pranavrajs (changes requested):
Please see my comments below.

app/services/message_templates/hook_execution_service.rb
muhsin-k: Makes sense. Made the changes.

scmmishra (approved):

PR #11595, opened 2025-05-27
fix(ci): specs

scmmishra:
The PR fixes three tests

### Copilot Linear Search Tool

The spec failed simply because the described class was not initialized correctly. Fixing that was very straightforward

### Copilot Threads

The issue was that the test was passing a plain string as the message parameter, but the CopilotMessage model expects the message field to be a JSON object with specific keys like content.

#### Changes made:

1. Updated test parameters: Changed message: `Hello, how can you help me?` to message: `{ content: 'Hello, how can you help me?' }` in all test cases
2. Updated controller:
  - Modified `copilot_thread_params` to properly permit nested message parameters
  - Updated title creation to use `copilot_thread_params[:message][:content]`
  - Updated validation to check `copilot_thread_params[:message][:content].blank?`
3. Updated test expectations: Used `stringify_keys` to compare the hash properly since Rails stores JSON as string keys

The test now correctly creates both a `CopilotThread` and `CopilotMessage` and all assertions pass.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,12 +91,12 @@ jobs:
             bundle install
             # pnpm install
 
-      # - run:
-      #     name: Download cc-test-reporter
-      #     command: |
-      #       mkdir -p ~/tmp
-      #       curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ~/tmp/cc-test-reporter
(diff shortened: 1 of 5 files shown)


sojan-official:
Closing this as the build is fixed in https://github.com/chatwoot/chatwoot/pull/11600

PR #11597, opened 2025-05-27
feat: sanitize inbox name

scmmishra:
This PR implement inbox name sanitization in Microsoft OAuth callback to prevent invalid characters in inbox names. The sanitation does the following: 

- Removes forbidden characters: `/, \, <, >, @`
- Strips leading/trailing symbols while preserving alphanumeric and safe characters
- Falls back to email-based name generation when sanitized name is too short or empty
- Uses "Email Inbox" as final fallback if all else fails

This ensures inbox names comply with the existing validation regex in the Inbox model `(^^\b[^/\\<>@]*\b$)` that prevents these characters, avoiding validation errors during Microsoft OAuth email integration setup.


sojan-official (approved):
@scmmishra  lets fix the specs and get this merged.

PR #11615, opened 2025-05-28
fix: setup error on Ruby 3.4.4 caused by missing cache_key_with_version in ActsAsTaggableOn

meAmitPatil:
Screenshot_2025-05-26_at_2 28 58_PM

This PR addresses critical setup issues encountered when running the Chatwoot project locally using **Ruby 3.4.4**, as specified in the `.ruby-version` file.

### Problem Overview

During initial setup, developers may face the following error:
undefined method `cache_key_with_version' for ActiveRecord::Base:Class

This error arises because `ActsAsTaggableOn` no longer defines `cache_key_with_version` for ActiveRecord models in newer versions. This method is still expected by tools like `auto_annotate_models`, leading to a failure in initial database setup or when starting the Rails server.

### Fixes Introduced

- ✅ Added a monkey patch initializer `config/initializers/taggable_cache_compat.rb` to define the missing `cache_key_with_version` under the expected module structure to restore compatibility.
- ✅ Updated `lib/tasks/auto_annotate_models.rake` to set `skip_on_migrate = true`, preventing auto-annotation from interfering during migration runs.

These fixes ensure that contributors using Ruby 3.4.4 can set up the project without runtime errors or blocked database migrations.

## Type of change

## How Has This Been Tested?

- Performed a clean clone and bootstrap of the Chatwoot repo using:
  - `bundle install`
  - `bundle exec rake db:setup`
  - `rails s`
- Verified:
  - No `cache_key_with_version` error is raised during setup
  - Auto-annotation does not interfere with DB migration
  - Application starts successfully using Ruby 3.4.4 on macOS
  - Console logs remain clean of warnings related to tagging or schema annotation

## Checklist:

- [ ] I have made corresponding changes to the documentation (N/A)
- [ ] I have added tests that prove my fix is effective or that my feature works (N/A)

### DIFF
--- a/config/initializers/taggable_cache_compat.rb
+++ b/config/initializers/taggable_cache_compat.rb
@@ -1,0 +1,16 @@
+# config/initializers/taggable_cache_compat.rb
+
+module ActsAsTaggableOn
+  module Taggable
+    # Stub out the old Cache module so migrations that refer to it
+    # won't blow up. Its `included` hook is a no-op.
+    unless const_defined?(:Cache)
+      Cache = Module.new do
+        def self.included(base)
+          # no-op: conversation.cached_label_list still exists,
+          # but we won't auto-update it
+        end
+      end
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -22,8 +22,8 @@ ActiveRecord::Schema[7.1].define(version: 2025_05_23_031839) do
     t.string "owner_type"
     t.bigint "owner_id"
     t.string "token"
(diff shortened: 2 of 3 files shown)


sojan-official:
@meAmitPatil this is a from a pretty old migration there shouldn't be a need to run this anymore.  Please use rake db:schema:load for for intial set up 

ref: https://stackoverflow.com/questions/5905287/rake-dbschemaload-vs-migrations

meAmitPatil:
Cool, thanks.

PR #11621, opened 2025-05-29
chore: Update Codespaces

sojan-official:
- Fix issues with the current Chatwoot development codespaces
- Switch from webpacket to vite 
- Add additional configs to make the development easier with codespaces
- toggles v4 feature true as default


scmmishra (approved):
Approving for now, @sojan-official just wanted to check if the changes `config/features.yml` is intentional. Either way it seems fine to flip the switch on that one

PR #11622, opened 2025-05-29
feat: hide CSAT survey URLs from agents in dashboard

muhsin-k:
Fixes 
**Problem**
  Agents can see CSAT survey URLs in the dashboard for non-web widget channels (WhatsApp, Telegram, SMS, etc.), allowing them to potentially access and manipulate survey
  responses, which compromises survey integrity.

**Solution**

  Separate survey URL storage from message content display:

  - **Store URLs separately**: Survey URLs saved in `content_attributes['survey_url']` instead of embedded in content
  - **Split content methods**:
    - `content()` returns clean content for dashboard (agents see no URL)
    - `content_for_channel()` includes URL for external delivery (contacts receive full message)
  - **Update channel services**: All external channels use `content_for_channel` for message delivery


muhsin-k:
@pranavrajs PTAL

app/models/message.rb
```diff
@@ -101,9 +102,10 @@ class Message < ApplicationRecord
   # [:deleted] : Used to denote whether the message was deleted by the agent
   # [:external_created_at] : Can specify if the message was created at a different timestamp externally
   # [:external_error : Can specify if the message creation failed due to an error at external API
+  # [:survey_url] : Used to store CSAT survey URL separately from content for non-web widget channels
```
sojan-official: We don't need to include this in content attributes,  We can deduce that a message type is input_csat, from content type.  We can rely on that in our logic to see its a csat message and then use the uuid  to generate survey _url when we are rendering the message.

app/models/message.rb
```diff
@@ -181,17 +183,14 @@ def webhook_data
     data
   end
 
-  def content
-    # move this to a presenter
-    return self[:content] if !input_csat? || inbox.web_widget?
+  # Method to get content with survey URL for external channel delivery
+  def channel_content
```
sojan-official: Can't quite relate to the name  `channel_content`.   we can retain the name content itself and write our transformations here itself.    will reduce the size of overall PR

sojan-official (changes requested):
Think we can reduce the overall changes here. 

- We only need to remove the logic where we add the url to the content.
- We can implement a presenter to handle the logic behind `content` attribute in messages 
- We can include additional conditions in the persenter to handle the case for CSAT messages in non web widget channels
- survey url doesn't need to be appended to the message, since we already have the input_csat type to determine if its a csat message

sojan-official (approved):
Code LGTM. lets gets this merged once the changes are tested.

PR #11637, opened 2025-05-30
fix: Send CSAT survey only when agent can reply in conversation

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-4415/postmortem-csat-survey-infinite-loop-issue

### DIFF
--- a/app/models/concerns/activity_message_handler.rb
+++ b/app/models/concerns/activity_message_handler.rb
@@ -5,6 +5,7 @@ module ActivityMessageHandler
   include LabelActivityMessageHandler
   include SlaActivityMessageHandler
   include TeamActivityMessageHandler
+  include CsatActivityMessageHandler
 
   private
 
--- a/app/models/concerns/csat_activity_message_handler.rb
+++ b/app/models/concerns/csat_activity_message_handler.rb
@@ -1,0 +1,8 @@
+module CsatActivityMessageHandler
+  extend ActiveSupport::Concern
+
+  def create_csat_not_sent_activity_message
+    content = I18n.t('conversations.activity.csat.not_sent_due_to_messaging_window')
+    ::Conversations::ActivityMessageJob.perform_later(self, activity_message_params(content)) if content
+  end
+end
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -115,6 +115,7 @@ class Conversation < ApplicationRecord
   before_create :ensure_waiting_since
 
   after_update_commit :execute_after_update_commit_callbacks
+  after_update_commit :trigger_csat_survey_if_resolved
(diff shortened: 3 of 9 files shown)


spec/services/csat_survey_service_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    context 'when CSAT survey should be sent' do
+      before do
+        allow(conversation).to receive(:can_reply?).and_return(true)
+      end
+
+      it 'sends CSAT survey when within messaging window' do
+        service.perform
+
+        expect(MessageTemplates::Template::CsatSurvey).to have_received(:new).with(conversation: conversation)
```
scmmishra, line 26:
```suggestion
        expect(MessageTemplates::Template::CsatSurvey).to have_received(:new).with(conversation: conversation)
        expect(csat_template).to have_received(:perform)
```

Let's also ensure that perform is called.

scmmishra (commented):
@muhsin-k the changes look good. We can test this on staging

muhsin-k:
@pranavrajs  The current implementation will create activity logs every time a conversation is resolved when CSAT is enabled. Should we restrict it to one log for the entire conversation?

muhsin-k:
@pranavrajs PTAL

app/models/concerns/activity_message_handler.rb
```diff
@@ -5,6 +5,7 @@ module ActivityMessageHandler
   include LabelActivityMessageHandler
   include SlaActivityMessageHandler
   include TeamActivityMessageHandler
+  include CsatActivityMessageHandler
```
sojan-official: We can start decoupling this logic from conversation object and just keep the responbisilities with the the CSAT service itself.  would make it much simpler and easier to test.

app/models/concerns/csat_activity_message_handler.rb
```diff
@@ -0,0 +1,8 @@
+module CsatActivityMessageHandler
+  extend ActiveSupport::Concern
+
+  def create_csat_not_sent_activity_message
```
sojan-official: let's  move this logic to the CSAT service itself / or as a service class . no need to include this logic into conversation object.

app/models/conversation.rb
```diff
@@ -197,6 +198,12 @@ def dispatch_conversation_updated_event(previous_changes = nil)
 
   private
 
+  def trigger_csat_survey_if_resolved
```
sojan-official: lets move this to CsatSurvey lisetner ..  and handle the logic there .

muhsin-k:
@sojan-official PTAL. I have fixed all the review comments.

sojan-official (approved):
code LGTM.  Feel free to merge after testing.

PR #11640, opened 2025-05-30
feat: improve article search ranking

scmmishra:
This PR enhances article search by implementing weighted ranking that prioritizes title and description matches over content matches. Previously, all text fields were treated equally, which could surface less relevant articles with many content keyword matches rather than focused articles with title matches. We do this with the following changes

- Added weighted search ranking: title (A), description (B), content (C)
- Implemented automatic relevance-based ordering with document length normalization

These changes should introduce minimal performance overhead since the ranking happens in-memory using existing indexes.

### DIFF
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -96,7 +96,6 @@ class SearchService
   def filter_articles
     @articles = current_account.articles
                                .text_search(search_query)
-                               .reorder('updated_at DESC')
                                .page(params[:page])
                                .per(15)
   end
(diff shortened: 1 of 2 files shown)


app/services/search_service.rb
```diff
@@ -96,7 +96,6 @@ def filter_contacts
   def filter_articles
     @articles = current_account.articles
                                .text_search(search_query)
-                               .reorder('updated_at DESC')
```
pranavrajs, line 99: There should be some defined order in which the results are returned, such as by creation date or number of views. Otherwise, the output will appear random and inconsistent.

pranavrajs: Or is it the closest match that comes at the top?

pranavrajs (approved):

Issue #947, opened 2020-06-09
[CW-1868] Ability to delete conversations


iammattmartin:
**Is your feature request related to a problem? Please describe.**
No, not related to a current problem.

**Describe the solution you'd like**
There isn't any ability to delete conversations (or old staff from the master admin login). This may be required in certain circumstances where sensitive information is received and cannot be stored, the message is spam, or due to GDPR, the data must be removed.

**Describe alternatives you've considered**
This may be possible via the database directly but generally such action is unwise, especially if cached elsewhere.

<sub>[CW-1868](https://linear.app/chatwoot/issue/CW-1868/ability-to-delete-conversations)</sub>

pranavrajs:
@iammattmartin Thanks for the suggestion. We will take it up in the current sprint itself. 

To start with we will add a feature to delete the conversation from SuperAdmin console.

iammattmartin:
@pranavrajs Do you know the current slated release for this. We have a few convos with data we'd really not want to keep.

pranavrajs:
@iammattmartin This was removed from v1.7.0 as it was not complete at that time, we will get it done on the v1.9.x release

pranavrajs:
@iammattmartin Allowing the admin/agents to delete a conversation might have an unintended sideeffect. If widget is open and admin tries to delete the conversation, the conversation on the widget becomes orphaned. We need to look at the implication in detail. 

As a starting point we will allow the agent/admin to delete a message which solve the spam messages / GDPR request.

pranavrajs:
The API to delete the message is merged in #1353, the UI would be completed in 1.13.0

pranavrajs:
This will be completed along with the UI changes in Message component.

ddlidded:
Is there any update on this?

rihards-simanovics:
Hi there, I currently looking to implement chatwoot for a semi-gov website where the conversation with users' needs to be deleted after a certain period and the conversation must be reset. I hope this feature will be implemented in the coming future.

lafi87:
It is also very important to have the functionality to clean the conversation, without deleting the contact.

jordan26:
I completely agree this is required, I didn't realise we had no option to delete conversations and now we have 5000 and counting.

Micorksen:
Hey!
This is a pretty cool idea, also, could you make like the merge contact button but for tickets?

jacekmichalski:
+1

ssthormess:
[...]

PR #11677, opened 2025-06-04
feat: Add conversation delete feature

sojan-official:
Screenshot 2025-06-05 at 12 39 04 AM

## Changes

Give the admins an option to delete conversation via the context menu

- enable conversation deletion in routes and controller
- expose delete API on conversations
- add delete option in conversation context menu and integrate with card and list
- implement store action and mutation for delete
- update i18n with new strings

fixes: https://github.com/chatwoot/chatwoot/issues/947


muhsin-k (commented):
@sojan-official Looks good and working well. One suggestion would be to add a confirmation dialog to prevent accidental deletions.

pranavrajs (approved):

PR #11719, opened 2025-06-12
feat: add references header to reply emails

scmmishra:
This PR implements proper References headers in outgoing emails to ensure correct email client threading behavior. The `References` header contains all previous message IDs from the conversation thread in chronological order, followed by the `In-Reply-To` message ID.

The implementation includes a dedicated `ReferencesHeaderBuilder` module that handles message ID collection, deduplication, and RFC 5322 compliant header folding. All email messages (both incoming and outgoing) use the `source_id` field for consistent message ID tracking throughout the conversation lifecycle, so we can reliably use it to build the header body

Fixes: CW-4503

### Email sent from Chatwoot

CleanShot 2025-07-28 at 15 19 40@2x

#### References header sent from Chatwoot

CleanShot 2025-07-28 at 15 20 08@2x

### Processing Incoming message

CleanShot 2025-07-28 at 15 24 45@2x

CleanShot 2025-07-28 at 15 24 09@2x

### Final reply

CleanShot 2025-07-28 at 15 26 09@2x

### DIFF
--- a/app/javascript/dashboard/i18n/locale/am/inboxMgmt.json
+++ b/app/javascript/dashboard/i18n/locale/am/inboxMgmt.json
@@ -600,6 +600,10 @@
       "WHATSAPP_SECTION_UPDATE_BUTTON": "Update",
       "WHATSAPP_WEBHOOK_TITLE": "Webhook Verification Token",
       "WHATSAPP_WEBHOOK_SUBHEADER": "This token is used to verify the authenticity of the webhook endpoint.",
+      "WHATSAPP_TEMPLATES_SYNC_TITLE": "Sync Templates",
+      "WHATSAPP_TEMPLATES_SYNC_SUBHEADER": "Manually sync message templates from WhatsApp to update your available templates.",
+      "WHATSAPP_TEMPLATES_SYNC_BUTTON": "Sync Templates",
+      "WHATSAPP_TEMPLATES_SYNC_SUCCESS": "Templates sync initiated successfully. It may take a couple of minutes to update.",
       "UPDATE_PRE_CHAT_FORM_SETTINGS": "Update Pre Chat Form Settings"
     },
     "HELP_CENTER": {
--- a/app/javascript/dashboard/i18n/locale/am/whatsappTemplates.json
+++ b/app/javascript/dashboard/i18n/locale/am/whatsappTemplates.json
@@ -8,6 +8,10 @@
         "PICKER": {
             "SEARCH_PLACEHOLDER": "Search Templates",
             "NO_TEMPLATES_FOUND": "No templates found for",
+            "NO_TEMPLATES_AVAILABLE": "No WhatsApp templates available. Click refresh to sync templates from WhatsApp.",
+            "REFRESH_BUTTON": "Refresh templates",
+            "REFRESH_SUCCESS": "Templates refresh initiated. It may take a couple of minutes to update.",
+            "REFRESH_ERROR": "Failed to refresh templates. Please try again.",
             "LABELS": {
                 "LANGUAGE": "Language",
                 "TEMPLATE_BODY": "Template Body",
--- a/app/javascript/dashboard/i18n/locale/ar/inboxMgmt.json
+++ b/app/javascript/dashboard/i18n/locale/ar/inboxMgmt.json
@@ -600,6 +600,10 @@
       "WHATSAPP_SECTION_UPDATE_BUTTON": "تحديث",
       "WHATSAPP_WEBHOOK_TITLE": "رمز التحقق من Webhook",
       "WHATSAPP_WEBHOOK_SUBHEADER": "This token is used to verify the authenticity of the webhook endpoint.",
+      "WHATSAPP_TEMPLATES_SYNC_TITLE": "Sync Templates",
+      "WHATSAPP_TEMPLATES_SYNC_SUBHEADER": "Manually sync message templates from WhatsApp to update your available templates.",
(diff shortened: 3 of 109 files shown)


app/mailers/references_header_builder.rb
```diff
@@ -0,0 +1,52 @@
+# Builds RFC 5322 compliant References headers for email threading
+#
+# This module provides functionality to construct proper References headers
+# that maintain email conversation threading according to RFC 5322 standards.
+module ReferencesHeaderBuilder
+  # Builds a complete References header for an email reply
+  #
+  # According to RFC 5322, the References header should contain:
+  # - All message IDs from the conversation thread in chronological order
```
pranavrajs: @scmmishra Have you tested the case when the email is partially available in Chatwoot and we don't have the complete thread? How does this behave in such cases?

app/mailers/references_header_builder.rb
```diff
@@ -0,0 +1,52 @@
...
+  #
+  # @param conversation [Conversation] The conversation containing the message thread
+  # @param in_reply_to_message_id [String] The message ID being replied to
+  # @return [String] A properly formatted and folded References header value
+  def build_references_header(conversation, in_reply_to_message_id)
+    references = collect_chronological_message_ids(conversation)
+    references << in_reply_to_message_id
+
+    references = references.compact.uniq
+    fold_references_header(references)
```
pranavrajs, line 24: This header shouldn't appear for messages that are not replies to another message.

app/mailers/references_header_builder.rb
```diff
@@ -0,0 +1,52 @@
...
+  # Collects all message IDs from a conversation in chronological order
+  #
+  # All email messages (both incoming and outgoing) have source_id populated:
+  # - Incoming emails: source_id = original email's Message-ID header
+  # - Outgoing emails: source_id = sent email's Message-ID header (set after delivery)
+  #
+  # @param conversation [Conversation] The conversation to extract message IDs from
+  # @return [Array<String>] Array of properly formatted message IDs with angle brackets
+  def collect_chronological_message_ids(conversation)
+    conversation.messages.where.not(source_id: nil).order(:created_at).pluck(:source_id).map do |source_id|
```
pranavrajs: This implementation assumes we’re always replying to the last message in the thread, which isn’t not true for email. You can reply to any message in the thread, not just the latest one.

How should we handle that? Do we need to update the way we track or link replies?

scmmishra:
@pranavrajs I've found a better approach to this. It would be better to store incoming references, and then use that when replying. This is a much better way to do it, it will preserve references even if the inbox email was added later, plus it reduces a lot more queries. I'll need some more time to test this. Will be ready by tomorrow. It should also address your comments

pranavrajs:
Sounds like a good approach.

app/mailers/conversation_reply_mailer_helper.rb
```diff
@@ -57,22 +57,17 @@ def process_attachments_as_files_for_email_reply
 
   private
 
-  def google_smtp_settings
-    return unless @inbox.email? && @channel.imap_enabled && @inbox.channel.google?
-
-    smtp_settings = base_smtp_settings('smtp.gmail.com')
+  def oauth_smtp_settings
```
scmmishra, line 60: Had to do this refactor to get rid of rubocop errors, the file had a few extra lines than permitted

scmmishra:
@pranavrajs please take a look whenver you can

app/mailers/references_header_builder.rb
```diff
@@ -0,0 +1,94 @@
...
+#
+# This module provides functionality to construct proper References headers
+# that maintain email conversation threading according to RFC 5322 standards.
+module ReferencesHeaderBuilder
+  # Builds a complete References header for an email reply
+  #
+  # According to RFC 5322, the References header should contain:
+  # - References from the message being replied to (if available)
+  # - The In-Reply-To message ID as the final element
+  # - Proper line folding if the header exceeds 998 characters
```
muhsin-k, line 11: Can we add a test for very long reference chains

app/mailers/conversation_reply_mailer_helper.rb
```diff
@@ -6,15 +6,15 @@ def prepare_mail(cc_bcc_enabled)
...
+      references: references_header
     }
 
     if cc_bcc_enabled
       @options[:cc] = cc_bcc_emails[0]
       @options[:bcc] = cc_bcc_emails[1]
     end
-    ms_smtp_settings
-    google_smtp_settings
+    oauth_smtp_settings
```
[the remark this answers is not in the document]
scmmishra: Had to be done for rubocop

app/mailers/references_header_builder.rb
scmmishra: The header has no size limit as such, only line limits. That won't be an issue since we are folding each ID into a separate line by default, so we will never touch the 998 character limit

app/mailers/references_header_builder.rb
```diff
@@ -22,6 +22,10 @@ def build_references_header(conversation, in_reply_to_message_id)
 
     references = references.compact.uniq
     fold_references_header(references)
+  rescue StandardError => e
+    Rails.logger.error("Error building references header for ##{conversation.id}: #{e.message}")
+    ChatwootExceptionTracker.new(e, account: conversation.account).capture_exception
```
scmmishra, line 27: @muhsin-k added this check to ensure that email sending never breaks at all. References header is not critical to sending. We can remove this in the future

muhsin-k (approved):

PR #11730, opened 2025-06-13
feat: captain image support

tds-1:
# Pull Request Template

## Linear links:
- https://linear.app/chatwoot/issue/CW-4479/if-image-is-sent-by-the-customer-send-it-to-openai

## Description

This pull request adds “Captain image support” to Chatwoot. It introduces multimodal message handling so that when a customer sends an image, Captain can forward the file to OpenAI’s vision endpoint, generate a caption/analysis
## Type of change

## How Has This Been Tested?

image

## Checklist:


app/services/llm_formatter/conversation_llm_formatter.rb
```diff
@@ -20,14 +22,22 @@ def build_messages
...
-    @record.messages.chat.order(created_at: :asc).each do |message|
+    @record.messages.where.not(message_type: :activity).order(created_at: :asc).each do |message|
       message_text << format_message(message)
     end
     message_text
   end
 
   def format_message(message)
     sender = message.message_type == 'incoming' ? 'User' : 'Support agent'
+    sender = "[Private] #{sender}" if message.private?
```
pranavrajs:
```suggestion
    sender = "[Private Note] #{sender}" if message.private?
```

app/services/llm_formatter/conversation_llm_formatter.rb
```diff
@@ -20,14 +22,22 @@ def build_messages
     return "No messages in this conversation\n" if @record.messages.empty?
 
     message_text = ''
-    @record.messages.chat.order(created_at: :asc).each do |message|
+    @record.messages.where.not(message_type: :activity).order(created_at: :asc).each do |message|
```
pranavrajs: Include only if there is a config present?

enterprise/app/jobs/captain/conversation/response_builder_job.rb
```diff
@@ -37,45 +38,84 @@ def generate_and_process_response
...
-          content: message_content(message),
-          role: determine_role(message)
-        }
-      end
+  def collect_previous_messages(include_all: false)
+    messages_query = @conversation
+                     .messages
+                     .where(message_type: [:incoming, :outgoing])
+
+    messages_query = messages_query.where(private: false) unless include_all
```
pranavrajs: These should never be part of the public response builder. It should only be used in the copilot response builder.

enterprise/app/jobs/captain/conversation/response_builder_job.rb
```diff
@@ -37,45 +38,84 @@ def generate_and_process_response
...
+  end
+
+  def get_attachment_url(attachment)
+    return attachment.external_url if attachment.external_url.present?
+
+    return unless attachment.file.attached?
+
+    begin
+      attachment.file_url
+    rescue ActiveStorage::FileNotFoundError
```
pranavrajs: Should not rescue the error here. It should break and retry `ActiveStorage::FileNotFoundError`. This logic is handled at the top.

enterprise/lib/chat_gpt.rb
```diff
@@ -8,9 +8,9 @@ def initialize(context_sections = '')
     @messages = [system_message(context_sections)]
   end
 
-  def generate_response(input, previous_messages = [], role = 'user')
-    @messages += previous_messages
-    @messages << { 'role': role, 'content': input } if input.present?
+  def generate_response(additional_message: nil, message_history: [], role: 'user')
```
pranavrajs: This file is not used in Captain. This change would most likely break the rephrase/summarize actions in the editor

enterprise/app/jobs/captain/conversation/response_builder_job.rb
```diff
@@ -37,45 +38,84 @@ def generate_and_process_response
...
+  end
+
+  def message_content_multimodal(message)
+    parts = []
+
+    parts << text_part(message.content) if message.content.present?
+    parts.concat(attachment_parts(message.attachments)) if message.attachments.any?
+
+    finalize_content_parts(parts)
+  end
```
pranavrajs: Can you simplify this part? It could be a service that that return the OpenAI compatible payload back.

enterprise/lib/chat_gpt.rb
tds-1: I have reverted this change, but I was not able to find the code piece where this was being called from. 
So I thought it to be safe for future use cases in case we use this helper.

app/services/openai_multimodal_content_service.rb
```diff
@@ -0,0 +1,74 @@
+class OpenaiMultimodalContentService
```
pranavrajs: Move this captain namespace under enterprise folder.

Something like `Captain::OpenAIMessageBuilderService`

pranavrajs: Also, couldn't we simplify it like this. 
```rb
class Captain::OpenAiMessageBuilderService
  pattr_initialize [:message]

  def generate_content
    parts = []
    parts << text_part(@message.content) if @message.content.present?
    parts.concat(attachment_parts(@message.attachments)) if @message.attachments.any?

    return 'Message without content' if parts.blank?
    return parts.first[:text] if parts.one? && parts.first[:type] == 'text'

    parts
  end

  private

  def text_part(text)
    { type: 'text', text: text }
  end

  def image_part(image_url)
    { type: 'image_url', image_url: { url: image_url } }
  end

  def attachment_parts(attachments)
    image_attachments = attachments.where(file_type: :image)
    image_parts = image_parts(image_attachments) if image_attachments.any?

    transcription = extract_audio_transcriptions(attachments)
    transcription_part = text_part(transcription) if transcription.present?

    attachment_part = text_part('User has shared an attachment') if attachments.exists?(['file_type NOT IN (?)', %i[image audio]])

    [image_parts, transcription_part, attachment_part].compact
  end

  def image_parts(image_attachments)
    image_attachments.each_with_object([]) do |attachment, parts|
      url = get_attachment_url(attachment)
      parts << image_part(url) if url.present?
    end
  end

  def get_attachment_url(attachment)
    return attachment.external_url if attachment.external_url.present?

    attachment.file.attached? ? attachment.file_url : nil
  end

  def extract_audio_transcriptions(attachments)
    audio_attachments = attachments.where(file_type: :audio)
    return '' if audio_attachments.blank?

    audio_attachments.map do |attachment|
      result = Messages::AudioTranscriptionService.new(attachment).perform
      result[:success] ? result[:transcriptions] : ''
    end.join
  end
end
```

app/services/llm_formatter/conversation_llm_formatter.rb
```diff
@@ -5,29 +5,44 @@ def format(config = {})
...
     sections << 'Message History:'
     sections << if @record.messages.any?
-                  build_messages
+                  build_messages(config)
                 else
                   'No messages in this conversation'
                 end
 
     sections << "Contact Details: #{@record.contact.to_llm_text}" if config[:include_contact_details]
+    sections << 'Conversation Attributes:'
```
pranavrajs: This should be added only if attributes are present.

enterprise/app/jobs/captain/conversation/response_builder_job.rb
```diff
@@ -78,6 +57,10 @@ def determine_role(message)
     message.message_type == 'incoming' ? 'user' : 'system'
   end
 
+  def message_content_multimodal(message)
```
pranavrajs: Please rename it to something better

app/services/openai_multimodal_content_service.rb
pranavrajs: Also, please add specs for this service separately.

pranavrajs (approved):

spec/enterprise/services/captain/open_ai_message_builder_service_spec.rb
```diff
@@ -0,0 +1,309 @@
...
+      end
+    end
+  end
+
+  describe '#extract_audio_transcriptions' do
+    let(:message) { create(:message, content: nil) }
+
+    context 'with no audio attachments' do
+      it 'returns empty string' do
+        result = service.send(:extract_audio_transcriptions, message.attachments)
```
pranavrajs, line 234: @tds-1 The specs test private methods directly which is not a great pattern. Please write test for generate content which is the public method and test all the cases against it.

pranavrajs (commented):
@tds-1 There is a review comment, please take it as an immediate followup.

PR #11755, opened 2025-06-17
feat: Disable audio transcription if captain is not enabled

pranavrajs:


sojan-official (approved):

enterprise/app/services/messages/audio_transcription_service.rb
```diff
@@ -20,7 +20,10 @@ def perform
   private
 
   def can_transcribe?
-    account.audio_transcriptions.present? && account.usage_limits[:captain][:responses][:current_available].positive?
+    return false if account.feature_enabled?('captain_integration')
```
psilva-sv, line 23: Hi @pranavrajs, I have a question. Why return false if captain integration is enabled? Shouldn't it be the opposite.

Since it needs captain integrations to be enabled to work well?

pranavrajs: Yes, it is fixed [here](https://github.com/chatwoot/chatwoot/pull/11779).

psilva-sv: Ohh thanks! I'm using the version 4.3.0 so I didn't notice it was already fixed, thanks again!

PR #11757, opened 2025-06-18
 fix: Respect messaging window constraints for auto-resolve messages

muhsin-k:
- Add messaging window validation to auto-resolve template service to prevent sending messages outside allowed time windows
  - Create activity log when auto-resolve messages can't be sent due to messaging restrictions

### DIFF
--- a/app/services/message_templates/template/auto_resolve.rb
+++ b/app/services/message_templates/template/auto_resolve.rb
@@ -4,8 +4,12 @@ class MessageTemplates::Template::AutoResolve
   def perform
     return if conversation.account.auto_resolve_message.blank?
 
-    ActiveRecord::Base.transaction do
-      conversation.messages.create!(auto_resolve_message_params)
+    if within_messaging_window?
+      ActiveRecord::Base.transaction do
+        conversation.messages.create!(auto_resolve_message_params)
+      end
+    else
+      create_auto_resolve_not_sent_activity_message
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/services/message_templates/template/auto_resolve.rb
```diff
@@ -4,8 +4,12 @@ class MessageTemplates::Template::AutoResolve
   def perform
     return if conversation.account.auto_resolve_message.blank?
 
-    ActiveRecord::Base.transaction do
-      conversation.messages.create!(auto_resolve_message_params)
+    if within_messaging_window?
+      ActiveRecord::Base.transaction do
```
sojan-official: we won't need the the transaction in this case i think. since aren't doing multiple object creation

spec/services/message_templates/template/auto_resolve_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+      expect(Conversations::ActivityMessageJob).to have_received(:perform_later).with(
+        conversation,
+        {
+          account_id: conversation.account_id,
+          inbox_id: conversation.inbox_id,
+          message_type: :activity,
+          content: I18n.t('conversations.activity.auto_resolve.not_sent_due_to_messaging_window')
+        }
+      )
```
sojan-official: these specs are not required since we test the use cases already in the spec, there isn't a need to individually test the methods.

sojan-official (commented):
Couple of minor comments, lets do a quick sync up before merging

scmmishra (approved):

PR #11774, opened 2025-06-20
fix: Add actor parameter to Linear OAuth for proper app attribution

muhsin-k:
- Add `actor=app` parameter to Linear OAuth authorization URL
  - Ensures all Linear issues are consistently attributed to the app rather than individual users

### DIFF
--- a/app/models/integrations/app.rb
+++ b/app/models/integrations/app.rb
@@ -71,7 +71,8 @@ class Integrations::App
       "redirect_uri=#{self.class.linear_integration_url}",
       "state=#{encode_state}",
       'scope=read,write',
-      'prompt=consent'
+      'prompt=consent',
+      'actor=app'
     ].join('&')
   end
 


iamsivin (approved):
Tested this PR locally and it’s working as expected. The only issue I noticed is that the thumbnail isn’t displaying correctly — it’s not reflecting the current user’s image or rendering based on the name (see screenshot below). As discussed, this is an issue with Linear, so we’re ignoring it for now.

image

**Loom video**
https://www.loom.com/share/3bc1be131d22431db00aee320960a575?sid=78def1d9-55ef-4512-8012-5e0265825bb9

muhsin-k:
> Can we also confirm whether this change will affect existing connections ? would they need to reauthenticate ?

@sojan-official Yes. The existing users of the Linear integration will need to disconnect and reconnect their Linear account. Without this, the new attribution features may not work as expected.

**Additional change:**
I’ve also implemented [automatic token revocation](https://github.com/chatwoot/chatwoot/pull/11838) if someone deletes the Linear integration from their Chatwoot dashboard. This ensures that the Linear app is fully disconnected on both sides, addressing the current scenario where the app appears active in Linear even after deletion from Chatwoot.

Also we need to send an email to existing account admins for the changes.

PR #11808, opened 2025-06-25
fix: Add missing CSAT URL in email reply templates

muhsin-k:
As part of PR [#11622](https://github.com/chatwoot/chatwoot/pull/11622), we intentionally hide CSAT survey URLs from agents in the dashboard to avoid exposing those links internally. However, email replies being external communications to contacts should still include the survey URLs.

We missed handling that specific case for email channels. All other external channels like WhatsApp, Instagram, Telegram, etc., were already working correctly and continue to include the survey URLs as intended.


muhsin-k:
Follow-up task : Add CSAT URL in conversation continuity email.

sojan-official (approved):
LGTM

PR #11829, opened 2025-06-27
chore: Replace `content` with `outgoing_content` in webhook data

muhsin-k:
The API channels don't receive CSAT survey URLs in webhook payloads since `webhook_data` uses content instead of outgoing_content.


spec/models/message_spec.rb
```diff
@@ -267,6 +267,14 @@
       message = create(:message)
       expect(message.webhook_data.key?(:attachments)).to be false
     end
+
+    it 'uses outgoing_content for webhook content' do
+      message = create(:message, content: 'Test content')
+      expect(message).to receive(:outgoing_content).and_return('Outgoing test content')
+
+      webhook_data = message.webhook_data
```
sojan-official, line 275: can we also ensure that for CSAT type the webhook content contain the CSAT link

sojan-official (approved):

Issue #11853, opened 2025-07-02
v4.3.0 - Error on fresh dev setup - NameError: uninitialized constant ActsAsTaggableOn::Taggable::Cache (NameError)


ricktibbe:
### Describe the bug

I'm trying to do a fresh development setup on my local macOS machine. Upon running `make db`, I encountered the following error:

```
== 20231211010807 AddCachedLabelsList: migrating ==============================
-- add_column(:conversations, :cached_label_list, :string)
   -> 0.0022s
bin/rails aborted!
StandardError: An error has occurred, this and all later migrations canceled: (StandardError)

uninitialized constant ActsAsTaggableOn::Taggable::Cache
chatwoot/db/migrate/20231211010807_add_cached_labels_list.rb:5:in 'AddCachedLabelsList#change'

Caused by:
NameError: uninitialized constant ActsAsTaggableOn::Taggable::Cache (NameError)

    ActsAsTaggableOn::Taggable::Cache.included(Conversation)
                              ^^^^^^^
Did you mean?  CacheKeys
chatwoot/db/migrate/20231211010807_add_cached_labels_list.rb:5:in 'AddCachedLabelsList#change'
Tasks: TOP => db:migrate:up
```

I've followed the exact steps from these docs, except I had to install Ruby `3.4.4` instead of `3.2.2` because it seems like the docs haven't been updated yet:
https://developers.chatwoot.com/contributing-guide/environment-setup/mac-os
https://developers.chatwoot.com/contributing-guide/setup-guide

Environment verification returns the following:
```
ruby 3.4.4 (2025-05-14 revision a38531fd3f) +PRISM [arm64-darwin24]
node v20.19.3
pnpm 10.2.0
psql (PostgreSQL) 14.18 (Homebrew)
redis-cli 8.0.2

WARNING: The convert command is deprecated in IMv7, use "magick" instead of "convert" or "magick convert"
Version: ImageMagick 7.1.1-47 Q16-HDRI aarch64 22763 https://imagemagick.org
Copyright: (C) 1999 ImageMagick Studio LLC
License: https://imagemagick.org/script/license.php
Features: Cipher DPC HDRI Modules OpenMP 
Delegates (built-in): bzlib fontconfig freetype heic jng jp2 jpeg jxl lcms lqr ltdl lzma openexr png raw tiff webp xml zlib zstd
Compiler: clang (16.0.0)

git version 2.39.5 (Apple Git-154)
```

[...]

PR #11858, opened 2025-07-02
fix: Migration error on fresh dev setup

ricardonewbyte:
## Description

This PR fixes the problem in the following issue: #11853 

For new setups, when running migrations an error occurs. This is due to a syntax error: `ActsAsTaggableOn::Taggable::Cache` (does not exist, the correct one according to the gem is `ActsAsTaggableOn::Taggable::Caching`)

## Type of change

## How Has This Been Tested?

Clone the project from the `develop` branch and follow the setup according to the current official documentation.

### DIFF
--- a/db/migrate/20231211010807_add_cached_labels_list.rb
+++ b/db/migrate/20231211010807_add_cached_labels_list.rb
@@ -2,6 +2,6 @@ class AddCachedLabelsList < ActiveRecord::Migration[7.0]
   def change
     add_column :conversations, :cached_label_list, :string
     Conversation.reset_column_information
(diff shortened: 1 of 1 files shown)


pranavrajs:
@ricardonewbyte This is an older migration and was written based on an older version of the gem. It’s not meant to be fixed this way. We’ll squash these changes when we get a chance.

The actual issue seems to be that either the user has run db:migrate, or the make run command is triggering some form of db:migrate unintentionally. In projects like Chatwoot, we should be using db:schema:load instead of `db:migrate`.

I will close the PR for now. Thanks for the attempt.

PR #11870, opened 2025-07-03
fix: Conversation starting in pending state when created from the dashboard

scmmishra:
## Problem
When users manually create conversations through the frontend form, the status was being overridden to "pending" if there was an active bot on the inbox, even when `status: "open"` was explicitly provided.

## Solution
- **Frontend**: Added `status: 'open'` to payload for manually created conversations
- **Backend**: Modified `ConversationBuilder` to respect explicit status parameters using `update_column` after creation


scmmishra:

muhsin-k:
@sojan-official Can you review this PR?

app/builders/conversation_builder.rb
```diff
@@ -14,7 +14,22 @@ def look_up_exising_conversation
...
+    conversation = ::Conversation.create!(conversation_params)
+
+    # Override status if explicitly provided in params
+    # This is required because the Conversation model's determine_conversation_status callback
+    # automatically sets status to :pending when there's an active bot on the inbox.
+    # However, when users manually create conversations (e.g., via the frontend form),
+    # they should be able to explicitly set the status to :open to bypass bot handling.
+    # We use update_column to avoid triggering callbacks and activity logs since this
+    # is just correcting the status to match the user's explicit intent.
+    if params[:status].present?
```
sojan-official, line 26: Can we look into adding a flag or virtual attribute, which will skip our status appending part in the model so that the status logic will stay centralized and we don't have to do this update column.

scmmishra:
@codex please review

pranavrajs:

PR #11873, opened 2025-07-03
chore: Add a condition to handle bounced email

pranavrajs:
Add bounced emails to the conversation thread.
Fix Gmail bounce detection by checking the X-Failed-Recipients header.

Currently, bounced emails are rejected as auto-replies, which causes support agents to miss important delivery failure context. This PR ensures bounced messages are correctly added to the thread, preserving visibility for the support team.


vishnu-narayanan (approved):
LGTM

vishnu-narayanan:
- tested and found to be working locally

PR #11908, opened 2025-07-09
feat: scenario tools [CW-4597]

scmmishra:
This PR adds the infrastructure for Captain agent tools, which scenarios will use for multi-agent workflows.

The implementation follows the same pattern as integrations apps. 
- Tools are configured in `config/agents/tools.yml` with id, title, description, and icon. 
- The `CaptainToolsHelpers` concern provides two methods: `available_agent_tools` (returns all resolvable tools without classes for frontend) and `resolve_tool_class(id)` (gets the actual tool class for backend use).
- Tool classes are resolved dynamically using the pattern `Captain::Tools::{ClassifiedId}Tool`. Only tools with valid classes are included in responses - invalid ones are filtered out automatically.
- Added a GET `/api/v1/accounts/{account_id}/captain/assistants/tools` endpoint that returns the available tools as JSON. This follows the same collection route pattern as other APIs.
- Added tool list resolution from the instruction, using the mention pattern

The PR also implements a `BaseAgentTool` class, it provides the foundation with permission checking, account scoping, and usage logging. Current tools include `add_contact_note`, `add_private_note`, and `update_priority`, they are just to serve as an example, we will clean these up and add more tools in a separate PR

Both `Captain::Assistant` and `Captain::Scenario` models include the `AgentToolsResolvable` concern so they can access the tool discovery methods. The `populate_tools` method in scenarios will use this infrastructure in future commits.


muhsin-k:
@scmmishra ~~There might be an issue with loading CaptainToolsHelpers. I am getting this error when I load conversations~~ Worked after restarting the server.

CleanShot 2025-07-14 at 19 50 50@2x

enterprise/app/controllers/api/v1/accounts/captain/assistants_controller.rb
```diff
@@ -32,6 +32,10 @@ def playground
     render json: response
   end
 
+  def tools
+    @tools = Captain::Assistant.available_agent_tools
```
muhsin-k, line 36: @scmmishra Getting this error, might be `Captain::AssistantPolicy` was missing the `tools?` method.
CleanShot 2025-07-14 at 20 08 29@2x

scmmishra: @muhsin-k fixed

config/application.rb
```diff
@@ -43,6 +43,7 @@ class Application < Rails::Application
     config.eager_load_paths << Rails.root.join('enterprise/listeners')
     # rubocop:disable Rails/FilePath
     config.eager_load_paths += Dir["#{Rails.root}/enterprise/app/**"]
+    config.eager_load_paths << Rails.root.join('enterprise/app/models/concerns')
```
sojan-official: why do we need this ? won't the previous line will load the files into the load paths ?

enterprise/app/models/concerns/captain_tools_helpers.rb
```diff
@@ -0,0 +1,76 @@
+# Provides helper methods for working with Captain agent tools including
+# tool resolution, text parsing, and metadata retrieval.
+module CaptainToolsHelpers
```
sojan-official: lets keep this under the concerns namespace ? 
would probably remove the need for us to do add the path to autoload paths

enterprise/app/models/captain/assistant.rb
```diff
@@ -18,6 +18,7 @@
 #
 class Captain::Assistant < ApplicationRecord
   include Avatarable
+  include CaptainToolsHelpers
```
sojan-official: lets include with the namespace .. would make it easier to identify where the file is located

sojan-official (commented):
looks good over all. some minor comments. 
lets also add specs for the tools ...

config/application.rb
scmmishra: It doesn't, it just lists the files in the folders in that directory alone, doesn't go a step further down the tree. I tried adding `**/**` but that didn't work for some reason, even after restarting a few times

enterprise/app/models/concerns/captain_tools_helpers.rb
scmmishra: You mean `Concerns::CaptainToolHelpers`? 

Will that mean we move this to a folder `chatwoot/enterprise/app/models/concerns/captain`?

PR #11918, opened 2025-07-10
feat: setup qlty [CW-4614]

scmmishra:


vishnu-narayanan:
image

.qlty/qlty.toml
```diff
@@ -0,0 +1,84 @@
...
+  "**/test/**",
+  "**/spec/**",
+  "**/*.test.*",
+  "**/*.spec.*",
+  "**/*_test.*",
+  "**/*_spec.*",
+  "**/test_*.*",
+  "**/spec_*.*",
+]
+
```
vishnu-narayanan, line 55: @scmmishra Shouldn't we also add the plugins we used in Code Climate? (rubocop, eslint, brakeman and so on)

vishnu-narayanan (commented):
@scmmishra Shouldn't we also add the plugins we used in Code Climate? (rubocop, eslint, brakeman and so on)

.qlty/qlty.toml
scmmishra: I think the import should have handled it

scmmishra: 0bc4545180a634517a273edb67b8ae0f941ba2a43216bc76d017323fd1627895

Yeah it does

PR #11944, opened 2025-07-15
feat: scenario agents & runner

scmmishra:
This PR introduces the `ai-agents`-based multi-agent Captain to Chatwoot, behind the `captain_v2` feature flag (disabled by default). It enables flexible agent orchestration, scenario-specific handoffs, dynamic prompt rendering, and improved configuration and error handling.

## Key Changes

**Agent Architecture and Orchestration:**
* Added `Concerns::Agentable` module to unify agent instantiation, instructions, tools, and context for both `Captain::Assistant` and `Captain::Scenario`.
* Introduced `Captain::Assistant::AgentRunnerService` to coordinate assistant and scenario agents, manage handoffs, and process responses with lifecycle callbacks.

**Prompt Rendering and Templates:**
* Implemented `Captain::PromptRenderer` using Liquid templates for dynamic prompt generation.
* Added comprehensive prompt templates for assistants, scenarios, and reusable conversation/contact snippets.
* Enhanced assistant and scenario models with richer prompt contexts, including metadata, guidelines, guardrails, and tool descriptions.

### DIFF
--- a/config/initializers/ai_agents.rb
+++ b/config/initializers/ai_agents.rb
@@ -4,7 +4,7 @@ require 'agents'
 
 Rails.application.config.after_initialize do
   api_key = InstallationConfig.find_by(name: 'CAPTAIN_OPEN_AI_API_KEY')&.value
-  model = InstallationConfig.find_by(name: 'CAPTAIN_OPEN_AI_MODEL')&.value&.presence || OpenAiConstants::DEFAULT_MODEL
+  model = InstallationConfig.find_by(name: 'CAPTAIN_OPEN_AI_MODEL')&.value.presence || OpenAiConstants::DEFAULT_MODEL
   api_endpoint = InstallationConfig.find_by(name: 'CAPTAIN_OPEN_AI_ENDPOINT')&.value || OpenAiConstants::DEFAULT_ENDPOINT
 
   if api_key.present?
--- a/enterprise/app/models/concerns/agentable.rb
+++ b/enterprise/app/models/concerns/agentable.rb
@@ -43,7 +43,7 @@ module Concerns::Agentable
   end
 
   def agent_model
-    InstallationConfig.find_by(name: 'CAPTAIN_OPEN_AI_MODEL')&.value&.presence || OpenAiConstants::DEFAULT_MODEL
+    InstallationConfig.find_by(name: 'CAPTAIN_OPEN_AI_MODEL')&.value.presence || OpenAiConstants::DEFAULT_MODEL
   end
 
   def agent_response_schema


enterprise/lib/captain/prompts/snippets/conversation.liquid
```diff
@@ -0,0 +1,18 @@
+# Current Conversation Context
```
pranavrajs, line 1: why don't we use the to_llm_text method on a conversation?

enterprise/lib/captain/prompts/snippets/contact.liquid
```diff
@@ -0,0 +1,17 @@
+# Contact Information
```
pranavrajs, line 1: same here, there is a to_llm_text

enterprise/lib/captain/prompts/snippets/conversation.liquid
scmmishra: The default LLM text formatter includes some conversation history also, which will mess up with the prompt. We can use it for the contact though

pranavrajs:
1/ Add this button is not working, same case with add all button.

Screenshot 2025-07-31 at 4 28 48 PM


2/ "No scenarios found." is enough.
Screenshot 2025-07-31 at 4 29 43 PM

enterprise/lib/captain/prompts/snippets/contact.liquid
scmmishra: Was not able to use the to_llm_text for conversation, so I had to create a snippet for conversation anyway 

> The default LLM text formatter includes some conversation history also, which will mess up with the prompt. We can use it for the contact though

Regardless, I think we can move to a liquid based syntax. for prompt instead of having it in `rb` files, in the future this can allow us to override these prompts from SuperAdmin

Gemfile
```diff
@@ -179,7 +179,8 @@ gem 'reverse_markdown'
 
 gem 'iso-639'
 gem 'ruby-openai'
-gem 'ai-agents', '>= 0.2.1'
+gem 'ai-agents', '>= 0.4.3'
+gem 'ruby_llm-schema'
```
sojan-official, line 185: lets add a #TODO comment here. to check if we need to move this gem as a dependency of ai-agents

enterprise/app/models/concerns/agentable.rb
```diff
@@ -0,0 +1,56 @@
...
+  def template_name
+    self.class.name.demodulize.underscore
+  end
+
+  def agent_tools
+    []  # Default implementation, override if needed
+  end
+
+  def agent_model
+    InstallationConfig.find_by(name: 'CAPTAIN_OPEN_AI_MODEL')&.value || 'gpt-4.1-mini'
```
sojan-official: lets move the fall model names to constants ?

config/initializers/ai_agents.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'agents'
+
+Rails.application.config.after_initialize do
+  api_key = InstallationConfig.find_by(name: 'CAPTAIN_OPEN_AI_API_KEY')&.value
+  model = InstallationConfig.find_by(name: 'CAPTAIN_OPEN_AI_MODEL')&.value || 'gpt-4.1-mini'
```
sojan-official: lets move the fallback model names to constants ?

enterprise/app/services/captain/assistant/agent_runner_service.rb
```diff
@@ -0,0 +1,161 @@
+require 'agents'
+
+class Captain::Assistant::AgentRunnerService
```
sojan-official, line 3: lets add spec

scmmishra: Already added it here `spec/enterprise/services/captain/assistant/agent_runner_service_spec.rb`

scmmishra:
@sojan-official made all the suggested changes

PR #11952, opened 2025-07-16
chore: Added support for inbox variables

muhsin-k:
- Added two new variables, `inbox.name` and `inbox.id`, to the `MESSAGE_VARIABLES` array for better support of inbox-related data.
- Updated the `@chatwoot/utils` dependency from version `^0.0.47` to `^0.0.48`.


muhsin-k:
> @muhsin-k I tested it, and it works, but the 'undefined variables' dialog appears when I use the `inbox.name` and `inbox.id` variables. Could you please check the screencast below?
> 

@iamsivin Fixed.

iamsivin (approved):

PR #12014, opened 2025-07-22
feat: Introduce the crm_v2  feature flag for CRM changes

tds-1:
Introduce crm_v2 feature flag for our upcoming optimisations for CRM

### DIFF
--- a/app/controllers/api/v1/accounts/inboxes_controller.rb
+++ b/app/controllers/api/v1/accounts/inboxes_controller.rb
@@ -69,6 +69,17 @@ class Api::V1::Accounts::InboxesController < Api::V1::Accounts::BaseController
     render status: :ok, json: { message: I18n.t('messages.inbox_deletetion_response') }
   end
 
+  def sync_templates
+    unless @inbox.channel.is_a?(Channel::Whatsapp)
+      return render status: :unprocessable_entity, json: { error: 'Template sync is only available for WhatsApp channels' }
+    end
+
+    Channels::Whatsapp::TemplatesSyncJob.perform_later(@inbox.channel)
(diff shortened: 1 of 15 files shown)


app/javascript/dashboard/featureFlags.js
```diff
@@ -38,6 +38,7 @@ export const FEATURE_FLAGS = {
   CHANNEL_INSTAGRAM: 'channel_instagram',
   CONTACT_CHATWOOT_SUPPORT_TEAM: 'contact_chatwoot_support_team',
   WHATSAPP_EMBEDDED_SIGNUP: 'whatsapp_embedded_signup',
+  CRM_V2: 'crm_v2',
```
sojan-official: won't need this for now since we aren't using this flag in any frontend features, let add it when we do

spec/jobs/account/contacts_export_job_spec.rb
```diff
@@ -89,7 +89,7 @@
       create(:contact, account: account, email: nil, phone_number: nil)
       described_class.perform_now(account.id, user.id, %w[id name email column_not_present], {})
       csv_data = CSV.parse(account.contacts_export.download, headers: true)
-      expect(csv_data.length).to eq(account.contacts.resolved_contacts.count)
+      expect(csv_data.length).to eq(account.contacts.resolved_contacts(use_crm_v2: account.feature_enabled?('crm_v2')).count)
```
sojan-official: we don't need this right ? since we have the fallback for the argument. lets only pass this if we are explicitly enabling this flag for an account.

sojan-official (commented):
LGTM mostlu. couple of minor comments

PR #12026, opened 2025-07-23
perf: Add migration script for populating the contact information per account

tds-1:


app/jobs/migration/migrate_account_contacts_job.rb
```diff
@@ -0,0 +1,117 @@
+class Migration::MigrateAccountContactsJob < ApplicationJob
```
sojan-official: lets go for a more descriptive name , something like "PopulateAccountContactStatusAndSocialsJob"

app/jobs/migration/migrate_account_contacts_job.rb
```diff
@@ -0,0 +1,117 @@
...
+  def perform(account_id)
+    account = Account.find_by(id: account_id)
+    return unless account
+
+    Rails.logger.info "Processing contact migration for account #{account_id} (#{account.name})"
+
+    contacts = account.contacts.where(contact_type: 'visitor')
+    processed_count = 0
+
+    contacts.find_each do |contact|
```
sojan-official: We can do contact.visitors.find_each  ( to reduce the calculation and  number of contacts to be processed ) .. another option will also be to find the first contact which is a lead. and then work through contacts created before this time stamp since the data should be accurate for contacts created from the point where we shipped the change for leads

app/jobs/migration/migrate_account_contacts_job.rb
```diff
@@ -0,0 +1,117 @@
...
+    return false unless contact.contact_type == 'visitor'
+
+    # Migrate additional_attributes for social cases
+    migrate_social_attributes(contact)
+
+    # Determine and update contact type
+    new_type = determine_contact_type(contact)
+    return false if new_type == 'visitor'
+
+    contact.update!(contact_type: new_type)
```
sojan-official: lets update this with update_attributes since the regular update will trigger callbacks and will trigger an action cable event or a webhook event .. we don't want that.

app/jobs/migration/migrate_account_contacts_job.rb
```diff
@@ -0,0 +1,117 @@
...
+
+    attrs['social_instagram_user_name'] ||= profiles['instagram']
+    true
+  end
+
+  def migrate_line_data(contact, attrs, profiles)
+    return false if profiles['line'].blank?
+
+    attrs['social_line_user_id'] ||= profiles['line']
+    attrs['social_line_user_name'] ||= contact.name.presence || profiles['line']
```
sojan-official: this doesn't seem relavant, we only care about the social line_user_id
also why we need contact.name ?

sojan-official:
@stonecharioteer  could you check up on this PR as well. see if its relevant based on our recent discussions about contact performance

pranavrajs:
Since there is no activity for a long time, I'll close it out for now. Let us revisit it later.

PR #12031, opened 2025-07-23
Channel email as from email

amiteshwarp:
## Description

Changes to support dynamic from email address while sending emails from Email Inboxes, even if the server is set with default email.

## Type of change

## How Has This Been Tested?

Updated the test case as well as tested default functionality.

## Checklist:


pranavrajs:
@amiteshwarp We reviewed the change internally, and it cannot be merged into the codebase as it stands. To safely use the FROM address, we need to introduce a verification flag. Without this, the change could break existing assumptions.

We have different types of inboxes in Chatwoot that support email. By default, a channel is treated as a forwarded email.

The types of inboxes today are:
- Forwarded Email → inbound through forwarding, outbound through Chatwoot mailer.
- Regular IMAP + Chatwoot SMTP → inbound via IMAP, outbound via Chatwoot mailer (used for high-traffic support systems).
- Regular IMAP/SMTP → inbound via IMAP, outbound through the configured SMTP.
- Google/Microsoft → inbound via XOAuth2 IMAP, outbound through SMTP.

The main issue is that whenever we use the Chatwoot mailer for outbound, the FROM address must first be verified with Chatwoot. 

The current change assumes that any FROM address can be sent using the original mailer configuration. That assumption is not correct and won’t hold across all accounts in Chatwoot. We can support the change by adding a verification. I will get back to you on the architecture.

pranavrajs:
Closing this as https://github.com/chatwoot/chatwoot/pull/12453 would fix the problem.

PR #12032, opened 2025-07-23
fix: Add delay to messenger echo events to prevent duplicate messages

muhsin-k:
- Add 2-second delay to Facebook Messenger echo event processing to prevent race condition
- Add 2-second delay to Instagram echo event processing for consistency
- Prevent duplicate messages when echo events arrive before send message API completes processing


sojan-official (commented):
code LGTM. can we also ensure a spec to capture this case in instagram controller.

pranavrajs (approved):

PR #12051, opened 2025-07-28
feat: add config for OpenAI endpoint

scmmishra:
This PR adds support for configuring a custom OpenAI API endpoint in Captain AI. Allowing the use of any OpenAI compatible API endpoints like Azure

### Key Changes

- Configuration:
	- Added `CAPTAIN_OPEN_AI_ENDPOINT` to `installation_config.yml` (default: https://api.openai.com/).
	- Included the new config in `allowed_configs` in `AppConfigsController`.
- Service:
	- Updated `Llm::BaseOpenAiService` to use the custom endpoint if set, with a fallback to the default.


enterprise/app/services/llm/base_open_ai_service.rb
```diff
@@ -13,6 +14,10 @@ def initialize
 
   private
 
+  def uri_base
+    InstallationConfig.find_by(name: 'CAPTAIN_OPEN_AI_ENDPOINT')&.value || 'https://api.openai.com/'
```
try-catch-repeat, line 18: when using Azure client in our backend, we utilize 3 parameters,
- access_token
- uri_base
- api_version

Api_version is not optional as it is injected directly in the URL that the client is using to make requests from.
I don't see an api_version in the PR and I don't have enough experience with Ruby library's interface.
How would we pass the api_version in this case?

PR #12058, opened 2025-07-29
fix: incorrect first response time for reopened conversations

scmmishra:
When Captain resolves a conversation without human intervention and it's later reopened, the first response time metric is incorrectly calculated from the original conversation creation time instead of from when it was reopened. This inflates response times.

#### Example

```
Monday 9:00am    → Conversation created
Monday 9:05am    → Captain bot resolves (no human intervention)
       ⋮         → 5 days pass...
Friday 2:00pm    → Conversation reopened
Friday 2:10pm    → Agent replies

❌ Reported: 5 days 1 hour 10 minutes response time
✅ Should be: 10 minutes response time
```
## Solution

Added `conversation_reopened` reporting event that tracks when resolved conversations are reopened. The existing `last_non_human_activity` method already queries for these events, so first response times are now calculated correctly.

### Before
```
Conversation created → Bot resolves → Conversation reopened → Agent replies
↑                                                            ↑
└────────────── First response time calculated from here (incorrect!)
```

### After
```
Bot resolves → Conversation reopened → Agent replies
              ↑                      ↑
              └──── First response time calculated from here (correct!)
```

### DIFF
--- a/app/helpers/reporting_event_helper.rb
+++ b/app/helpers/reporting_event_helper.rb
@@ -18,12 +18,25 @@ module ReportingEventHelper
   end
 
   def last_non_human_activity(conversation)
-    # check if a handoff event already exists
-    handoff_event = ReportingEvent.where(conversation_id: conversation.id, name: 'conversation_bot_handoff').last
+    # Try to get either a handoff or reopened event first
+    # These will always take precendence over any other activity
+    # Also, any of these events can happen at any time in the course of a conversation lifecycle.
+    # So we pick the latest event
+    event = ReportingEvent.where(
+      conversation_id: conversation.id,
+      name: %w[conversation_bot_handoff conversation_reopened]
+    ).order(event_end_time: :desc).first
 
-    # if a handoff exists, last non human activity is when the handoff ended,
-    # otherwise it's when the conversation was created
-    handoff_event&.event_end_time || conversation.created_at
+    return event.event_end_time if event&.event_end_time
+
+    # Fallback to bot resolved event
+    # Because this will be closest to the most accurate activity instead of conversation.created_at
+    bot_event = ReportingEvent.where(conversation_id: conversation.id, name: 'conversation_bot_resolved').last
+
+    return bot_event.event_end_time if bot_event&.event_end_time
+
+    # If no events found, return conversation creation time
+    conversation.created_at
   end
 
   private
--- a/app/listeners/reporting_event_listener.rb
+++ b/app/listeners/reporting_event_listener.rb
@@ -90,6 +90,34 @@ class ReportingEventListener < BaseListener
     reporting_event.save!
   end
 
+  def conversation_opened(event)
+    conversation = extract_conversation_and_account(event)[0]
+
+    # Find the most recent resolved event for this conversation
+    last_resolved_event = ReportingEvent.where(
+      conversation_id: conversation.id,
+      name: 'conversation_resolved'
+    ).order(event_end_time: :desc).first
+
+    # If there's a previous resolved event, this is a reopening
+    return unless last_resolved_event
+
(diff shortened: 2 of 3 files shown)


app/listeners/reporting_event_listener.rb
```diff
@@ -90,6 +90,34 @@ def conversation_bot_handoff(event)
...
+    reporting_event = ReportingEvent.new(
+      name: 'conversation_reopened',
+      value: time_since_resolved,
+      value_in_business_hours: business_hours(conversation.inbox, last_resolved_event.event_end_time, conversation.updated_at),
+      account_id: conversation.account_id,
+      inbox_id: conversation.inbox_id,
+      user_id: conversation.assignee_id,
+      conversation_id: conversation.id,
+      event_start_time: last_resolved_event.event_end_time,
+      event_end_time: conversation.updated_at
```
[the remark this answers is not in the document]
scmmishra: This is consistent with existing behavior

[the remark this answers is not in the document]
scmmishra: This is consistent with existing behavior

app/helpers/reporting_event_helper.rb
```diff
@@ -18,12 +18,25 @@ def business_hours(inbox, from, to)
...
+    ).order(event_end_time: :desc).first
 
-    # if a handoff exists, last non human activity is when the handoff ended,
-    # otherwise it's when the conversation was created
-    handoff_event&.event_end_time || conversation.created_at
+    return event.event_end_time if event&.event_end_time
+
+    # Fallback to bot resolved event
+    # Because this will be closest to the most accurate activity instead of conversation.created_at
+    bot_event = ReportingEvent.where(conversation_id: conversation.id, name: 'conversation_bot_resolved').last
```
pranavrajs, line 34: What is this case for?

app/listeners/reporting_event_listener.rb
```diff
@@ -90,6 +90,34 @@ def conversation_bot_handoff(event)
...
+      name: 'conversation_resolved'
+    ).order(event_end_time: :desc).first
+
+    # If there's a previous resolved event, this is a reopening
+    return unless last_resolved_event
+
+    time_since_resolved = conversation.updated_at.to_i - last_resolved_event.event_end_time.to_i
+
+    reporting_event = ReportingEvent.new(
+      name: 'conversation_reopened',
```
pranavrajs: Should we make it conversation_opened and have it available for every conversation?

pranavrajs:
@scmmishra The PR looks harmless, but I’d like you to think through a few scenarios and revisit how each metric is being captured. We might want to refine the definitions to make them more accurate.

For instance, the time an agent takes to respond is not the same as the time a customer waits for a response. The former reflects agent performance, while the latter also includes the time the conversation sat unassigned in the queue.

Screenshot 2025-07-30 at 7 56 56 PM

app/helpers/reporting_event_helper.rb
scmmishra: This is a fallback, it's more accurate than the conversation `created_at` time. But then I guess, we can remove this anyway, since the result will be wrong, albeit less wrong (which is still wrong).

app/listeners/reporting_event_listener.rb
scmmishra: Makes sense, this can be done

pranavrajs:
- test the same case for an agent bot and a Dialogflow integration.

pranavrajs:
Screenshot 2025-08-05 at 10 00 13 PM

pranavrajs:
@scmmishra Can you please verify if this change works for the following cases as mentioned above? I don't think this change was done. 
 
> test the same case for an agent bot and a Dialogflow integration.

sojan-official (approved):
Code LGTM

muhsin-k (approved):

PR #12064, opened 2025-07-29
chore: add job to clean up orphaned cloudflare cname records for portals

vishnu-narayanan:
- chore: add job to clean up orphaned cloudflare cname records for portals

### DIFF
--- a/config/schedule.yml
+++ b/config/schedule.yml
@@ -46,3 +46,10 @@ delete_accounts_job:
   cron: '0 1 * * *'
   class: 'Internal::DeleteAccountsJob'
   queue: scheduled_jobs
+
+# executed weekly on Sunday at 0200 UTC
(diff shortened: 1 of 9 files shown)


config/schedule.yml
```diff
@@ -46,3 +46,10 @@ delete_accounts_job:
   cron: '0 1 * * *'
   class: 'Internal::DeleteAccountsJob'
   queue: scheduled_jobs
+
+# executed weekly on Sunday at 0200 UTC
+# to cleanup orphaned Cloudflare custom hostnames
+cloudflare_cleanup_job:
+  cron: '0 2 * * 0'
+  class: 'Enterprise::CloudflareCleanupJob'
```
pranavrajs: would this break in os only image

vishnu-narayanan: fixed

PR #12068, opened 2025-07-30
feat: disable helpcenter on hacker plans

vishnu-narayanan:
This change blocks Help Center access for default/Hacker-plan accounts and closes the downgrade gap that could leave `help_center` enabled after a subscription falls back to the default cloud plan.

Fixes: none
Closes: none

## Why

Default-plan accounts should not be able to access the Help Center, but the downgrade fallback path only reset the plan name and did not reconcile premium feature flags. That meant some accounts could keep `help_center` enabled even after landing back on the Hacker/default plan.

## What this change does

- blocks Help Center portal and article access for default/Hacker-plan accounts
- reconciles premium feature flags when a subscription falls back to the default cloud plan, so `help_center` is disabled immediately instead of waiting for a later webhook
- preserves existing account `custom_attributes` during Stripe customer recreation instead of overwriting them
- adds Enterprise coverage for the default-plan access checks on hosted and custom-domain Help Center routes
- fixes the public access check to use the resolved portal object so blocked requests return the intended response instead of raising an error

## Validation

1. Create or use an account on the default/Hacker cloud plan with an active portal.
2. Visit the portal home page and a published article on both the Chatwoot-hosted URL and a configured custom domain.
3. Confirm the Help Center is blocked for that account.
4. Downgrade a paid account back to the default/Hacker plan through the Stripe webhook flow.
5. Confirm `help_center` is disabled right after the downgrade fallback is processed and the account can no longer access the Help Center.


sojan-official (approved):
code LGTM

PR #12113, opened 2025-08-06
feat: add BE changes for captain pdf support for faq generation

tds-1:
Linked PR
https://github.com/chatwoot/chatwoot/pull/12115

## Linear Link
https://linear.app/chatwoot/issue/CW-4515/upload-pdfs-to-captain

## Description

This PR adds support for PDF file uploads to the Captain document knowledge base system. The feature enables users to upload PDF files directly to the knowledge base for FAQ generation and document processing, expanding beyond the existing URL-based document ingestion.

## Type of change

## How Has This Been Tested?

- Unit Tests
- Manually via UI

## Checklist:


tds-1:
**Test cases:**

- [ ] PDF with text only 
- [ ] PDF with text + images
- [ ] PDF with more than 100 pages
- [ ] Handwritten text PDF or Image only PDF 
- [ ] PDF over 20MB size.
- [ ] Upload a Book (a novel/history text book and ask questions based on that).


Look at the content that is extracted from the PDF and look at the generated FAQs and see if there is a mismatch.

----

Documentation changes:

- [ ] Add a new document in the user guide "Captain understands PDF"
- [ ] Add it to the product changelog.

enterprise/app/services/captain/llm/paginated_faq_generator_service.rb
```diff
@@ -0,0 +1,199 @@
...
+  attr_reader :total_pages_processed, :iterations_completed
+
+  def initialize(document, options = {})
+    super()
+    @document = document
+    @pages_per_chunk = options[:pages_per_chunk] || DEFAULT_PAGES_PER_CHUNK
+    @max_pages = options[:max_pages] # Optional limit from UI
+    @total_pages_processed = 0
+    @iterations_completed = 0
+    @model = 'gpt-4.1-mini'
```
muhsin-k: @tds-1 Shouldn't we use CAPTAIN_OPEN_AI_MODEL from the installation config instead of always using gpt-4.1-mini?
```suggestion
    @model = InstallationConfig.find_by(name: 'CAPTAIN_OPEN_AI_MODEL')&.value.presence || OpenAiConstants::DEFAULT_MODEL
```

tds-1: This particular task is for faq generation not user response so IMO having specific model would be a better approach

.rubocop.yml
```diff
@@ -16,6 +16,7 @@ Metrics/ClassLength:
   Exclude:
     - 'app/models/message.rb'
     - 'app/models/conversation.rb'
+    - 'enterprise/app/services/captain/llm/system_prompts_service.rb'
```
sojan-official: lets not do this, we have rubocop to ensure that we breakdown the classes into readable sub services. the ignore list is only for the exceptions which were present when we introduced this, the idea is to remove the files from here going forward  and try best not to add more here. !!!

sojan-official: note: looked at the file and seems like it needs as overall rewamp, lets do inline attribution in that case like how the other methods have handled it.

enterprise/app/services/captain/llm/paginated_faq_generator_service.rb
sojan-official: lets move it as a constant to openaiconstants::pdfprocessing  or a descriptive name in that case.  so that we only need to scan one place to understand the constants which we are using

sojan-official (commented):
some minor comments. LGTM otherwise

muhsin-k (approved):

sojan-official (approved):

PR #12116, opened 2025-08-06
feat: add more tools

scmmishra:


muhsin-k (approved):

enterprise/lib/captain/tools/faq_lookup_tool.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    # Use existing vector search on approved responses
+    responses = @assistant.responses.approved.search(query).to_a
+
+    if responses.empty?
+      log_tool_usage('no_results', { query: query })
+      "No relevant FAQs found for: #{query}"
+    else
+      log_tool_usage('found_results', { query: query, count: responses.size })
+      format_responses(responses)
```
tds-1, line 16: Extracting format_response as a shared utility would eliminate code duplication between SearchDocumentationService and FaqLookupTool

scmmishra: We will eventually remove the `SearchDocumentationService`, so it's fine to have the duplication I guess

PR #12194, opened 2025-08-14
feat: Add support for team based permissions along with inbox

### LINKED ISSUE #4964: Why is access control code inbox-centric?
Hi,

while scrolling through the code I've had trouble understanding why you chose to make just about all access control code inbox-centric, e.g. delegating conversation access directly to InboxPolicy, bypassing ConversationPolicy altogether. I understand that you currently have access _config_ based around inboxes but that does not mean you should make your _code_ follow this assumption. Even worse when such assumptions creep into frontend-code e.g. ``/api/v1/accounts/.../assignable_agents?inbox_ids[]=...`` should imho be ``/api/v1/accounts/.../conversations/.../assignable_agents``. Such abstraction would future-proof your code, because I find it inevitable that you will eventually move away from this inbox-centric config.

This brings me to my next point: Why did you choose to make access _config_ inbox-centric? Assuming we have an account where access is not "free for all"; I find it hard to come up with a scenario where inbox-centric config adequately reflects the real-life responsibilities of agents. Maybe I'm missing something essential here, but customers are notorious to pick their contact channel on a least-effort basis and not putting much thought into how to reach the right department/team.

In a traditional phone support system it is certainly possible to route a call to the right department/team even if the call did not initially reach a qualified agent. Customers expect the same when they're using a text-based support channel. Now let's say that for whatever reason we do not want the billing team to have access to _all_ of the company's Twitter DMs. However, there will certainly be customers asking billing related questions over Twitter DM's. So there must be a way to transfer a conversation over to an agent that does not have full access to the Twitter inbox. Conversation->team assignment is pointing in the right direction but that does not currently transcend inbox boundaries and the hard-coded assumptions mentioned above make it hard to implement a more flexible access control.

Can somebody enlighten me what real-world scenario drove you to make inboxes the first-class citizen of access control within an account?

### LINKED ISSUE #8787: Updating team permissions without having to be active in the inbox

There are team members who cannot see all messages in the inbox, the team function only serves as a message filter, but does not limit the viewing of messages to team members only.

The main problem is that the user with access to the inbox can see all messages, even if it is assigned to a team that they are not part of.

The chatwoot blog showed an update that allowed the user to be linked to the team only, without the need to be in the inbox.

https://www.chatroot.com/blog/team-permissions

However, in all the tests carried out, this function still does not work in our version, I imagine it is in their paid version.

Isn't it possible to activate this option for us too? And allow a user to be included in teams and not in inboxes when necessary.

_No response_

### Additional context

I believe that this update is important for companies that need to have their departments separated. Today, for this to happen, it is mandatory to have several inboxes, but the idea of the tool is that several employees can manage and respond to just 1 WhatsApp.

That's why this change is very important.

### LINKED ISSUE #11495: Permission to view conversations for team members

Though basic permissions were launched in version 3.13.0, one key permission is missing - limiting the visualization of a conversation that is attributed to a team to only members of this team. 

The closest Chatwoot currently has related to this is limiting to conversations one participates on, however, in situations where all team members were currently offline, the system sets a conversation to unassigned, which members of other teams can view too. 

Permission to view a conversation for team members only

_No response_

### Additional context

_No response_

pranavrajs:
This PR is an exploratory change to introduce team-level permissions.
Right now, teams only act as a filter on top of inbox permissions, but we’ve had a lot of feedback that this doesn’t work well when multiple teams share the same inbox but don’t want to see each other’s conversations. This update gives teams clearer control over what they can access and work on.

The goal is to allow an agent to belong to an inbox, a team, or both and have their access determined by these memberships. It’s a fundamental change to how permissions work, so it will need thorough testing before production. I’ll start with a partial rollout for selected customers.

**What’s changing:**

- An agent can be part of either an inbox or a team or both. 
- Agents will see all conversations they’re part of (through inbox or team membership).
- Agents can only interact with conversations they have access to.
- Custom role permissions still apply on top of inbox/team access.
- A conversation can be assigned to an agent with inbox access or to a team.
- Auto-assignment will:
  - Assign to people in the inbox if no team is set.
  - Assign to people in the team if a team is set.
- When a conversation is moved out of a team, previous team members lose access (v1).
- Agents can still view and take action on a conversation if they’re a participant.(v2)

**What's there to test:**
- Search (Agents should see messages from either the team or the inbox).
- Conversations (Agents should see messages from either the team or the inbox).
- Auto Assignment (Should work either way)
- Manual Assignment
- UX after a conversation is moved from the team to another.

Fixes https://github.com/chatwoot/chatwoot/issues/8787
Fixes https://github.com/chatwoot/chatwoot/issues/4964
Fixes https://github.com/chatwoot/chatwoot/issues/11495


joaomachado-pg:
news ??

paulobautz:
Up

nacholupotti:
Any update here? I would love to have this feature. Thanks

diegoferrandis:
Any news about this feature? This will be a real game changer for me and my Clients.

PR #12261, opened 2025-08-21
fix: wrong resolution count in timeseries reports

scmmishra:
There was a fundamental difference in how resolution counts were calculated between the agent summary and timeseries reports, causing confusion for users when the numbers didn't match.

The agent summary report counted all `conversation_resolved` events within a time period by querying the `reporting_events` table directly. However, the timeseries report had an additional constraint that required the conversation to currently be in resolved status (`conversations.status = 1`). This meant that if an agent resolved a conversation that was later reopened, the resolution action would be counted in the summary but not in the timeseries.

This fix aligns both reports to count resolution events rather than conversations in resolved state. When an agent resolves a conversation, they should receive credit for that action regardless of what happens to the conversation afterward. The same logic now applies to bot resolutions as well.

The change removes the `conversations: { status: :resolved }` condition from both `scope_for_resolutions_count` and `scope_for_bot_resolutions_count` methods in CountReportBuilder, and updates the corresponding test expectations to reflect that all resolution events are counted.

## About timezone

When a timezone is specified via `timezone_offset` parameter, the reporting system:

1. Converts timestamps to the target timezone before grouping
2. Groups data by local day/week/month boundaries in that timezone, but the primary boundaries are sent by the frontend and used as-is
3. Returns timestamps representing midnight in the target timezone

This means the same events can appear in different day buckets depending on the timezone used. For summary reports, it works fine, since the user only needs the total count between two timestamps and the frontend sends the timestamps adjusted for timezone.

## Testing Locally

Run the following command, this will erase all data for that account and put in 1000 conversations over last 3 months, parameters of this can be tweaked in `Seeders::Reports::ReportDataSeeder`

I'd suggest updating the values to generate data over 30 days, with 10000 conversations, it will take it's sweet time to run but then the data will be really rich, great for testing.

```
ACCOUNT_ID=2 ENABLE_ACCOUNT_SEEDING=true bundle exec rake db:seed:reports_data
```

Pro Tip: Don't run the app when the seeder is active, we manually create the reporting events anyway. So once done just use `redis-cli FLUSHALL` to clear all sidekiq jobs. Will be easier on the system

Use the following scripts to test it

- https://gist.github.com/scmmishra/1263a922f5efd24df8e448a816a06257
- https://gist.github.com/scmmishra/ca0b861fa0139e2cccdb72526ea844b2
- https://gist.github.com/scmmishra/5fe73d1f48f35422fd1fd142ea3498f3
- https://gist.github.com/scmmishra/3b7b1f9e2ff149007170e5c329432f45
- https://gist.github.com/scmmishra/f245fa2f44cd973e5d60aac64f979162


app/builders/v2/reports/timeseries/count_report_builder.rb
```diff
@@ -38,16 +38,16 @@ def scope_for_outgoing_messages_count
   end
 
   def scope_for_resolutions_count
-    scope.reporting_events.joins(:conversation).select(:conversation_id).where(
+    scope.reporting_events.select(:conversation_id).where(
       name: :conversation_resolved,
-      conversations: { status: :resolved }, created_at: range
+      created_at: range
     ).distinct
```
scmmishra: The distinct should cause an issue when the conversation is resolved twice

app/builders/v2/reports/timeseries/count_report_builder.rb
scmmishra, line 1: See if the timezone used elsewhere for calculation or not.

app/builders/v2/reports/label_summary_builder.rb
```diff
@@ -62,10 +62,24 @@ def fetch_conversation_counts(conversation_filter)
...
+    reporting_event_filter = { name: 'conversation_resolved' }
+    reporting_event_filter[:created_at] = range if range.present?
+
+    ReportingEvent
+      .joins('INNER JOIN conversations ON reporting_events.conversation_id = conversations.id')
+      .joins('INNER JOIN taggings ON taggings.taggable_id = conversations.id')
+      .joins('INNER JOIN tags ON taggings.tag_id = tags.id')
+      .where(
+        reporting_event_filter.merge(
+          conversations: { account_id: account.id },
```
pranavrajs: can't we use account_id in the reporting_event

pranavrajs (changes requested):
I've a comment on the query. There is a spec that is broken. Please fix that @scmmishra 

spec/controllers/api/v2/accounts/reports_controller_spec.rb:133:9: C: RSpec/ExampleLength: Example has too many lines. [31/25]

muhsin-k (approved):

Issue #486, opened 2020-02-11
Security: Add provision to enable 2FA for login


pranavrajs:
- Add a provision to setup Google Authenticator / Authy for 2FA

xLixon:
any new updates?

discod:
is anyone working on this?

karimsiam:
any news?

RinnDog:
Any news?

jtayrone:
feat: For greater security on user accounts, Add setup Google Authenticator / Authy for 2FA

taewookim:
Bump. 
FYI - rising brute force attacks on chatwoot servers in past few days

ezequiel-moura:
Yes please. This is very important!

sneko:
Hi @sojan-official,

Any news about this? I saw you put it as "low priority", having 2FA seems important to me when there is business content in shared inboxes (including customer data). 

Thank you,

singhdurgesh:
Hello @sojan-official,

Happy New Year!
I noticed that this feature has been open for quite a while and appears to be an important integration. If this task hasn't been assigned yet, I'd be interested in taking it on. Please let me know if I can pick it up.

Thanks!

oppt-admin:
Hi folks. Any news on this? This is such an important feature and should be a priority. Is there any way to help with this? Is it even on the roadmap? Thank you!

kilrau:
Well not really surprised this isn't implemented, gotta have good reason for that SAML SSO. I think the only way forward is a community PR.

tooolbox:
Hm, this would be good to have.

sojan-official:
fixed by #12290, #12372

sojan-official:
ref: https://developers.chatwoot.com/self-hosted/configuration/multi-factor-authentication

PR #12290, opened 2025-08-25
feat: MFA

tds-1:
## Linear:
- https://github.com/chatwoot/chatwoot/issues/486

## Description
This PR implements Multi-Factor Authentication (MFA) support for user accounts, enhancing security by requiring a second form of verification during login. The feature adds TOTP (Time-based One-Time Password) authentication with QR code generation and backup codes for account recovery.

## Type of change

## How Has This Been Tested?

- Added comprehensive RSpec tests for MFA controller functionality
- Tested MFA setup flow with QR code generation
- Verified OTP validation and backup code generation
- Tested login flow with MFA enabled/disabled

## Checklist:


config/environments/production.rb
```diff
@@ -103,4 +103,14 @@
   config.action_mailbox.ingress = ENV.fetch('RAILS_INBOUND_EMAIL_SERVICE', 'relay').to_sym
 
   Rails.application.routes.default_url_options = { host: ENV['FRONTEND_URL'] }
+
+  # Active Record Encryption configuration for production
+  # IMPORTANT: These MUST be set via environment variables in production
+  # Generate secure keys using: rails db:encryption:init
+  # Store them securely in your deployment environment (e.g., Kubernetes secrets, AWS Secrets Manager, etc.)
+  if ENV['ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY'].present?
```
sojan-official: lets remove the separate configs via environment file, rather have this in application.rb  , since this is added as a conditional here, would like to see consistant behavior in all environments so that we catch issues if  the config is missing in any installations. 

Any breaking/specific specs related to this can be handled via Climate Control.

config/environments/development.rb
```diff
@@ -85,4 +85,14 @@
...
+  # Active Record Encryption configuration for development
+  # These keys are for development only and should NEVER be used in production
+  # In production, set these via environment variables:
+  # - ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY
+  # - ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY
+  # - ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT
+  config.active_record.encryption.primary_key = ENV.fetch('ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY', 'sAK7479yQQvk0574SPM9RZFi9xx3dlBY')
+  config.active_record.encryption.deterministic_key = ENV.fetch('ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY', 't8mg48yhHqOaAH7R2HA4SDDsXWcyWBYL')
+  config.active_record.encryption.key_derivation_salt = ENV.fetch('ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT', 'bsfGUb0GEmghCU5HZlBjILZwoMjypLpS')
 end
```
sojan-official: lets remove the separate configs via environment file, rather have this in application.rb

config/environments/test.rb
```diff
@@ -51,4 +51,10 @@
   # Raises error for missing translations.
   # config.action_view.raise_on_missing_translations = true
   config.log_level = ENV.fetch('LOG_LEVEL', 'debug').to_sym
+
+  # Active Record Encryption configuration for testing
+  # Use fixed test keys for consistent test behavior
+  config.active_record.encryption.primary_key = ENV.fetch('ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY', 'test_sAK7479yQQvk0574SPM9RZFi9xx3dlBY')
+  config.active_record.encryption.deterministic_key = ENV.fetch('ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY', 'test_t8mg48yhHqOaAH7R2HA4SDDsXWcyWBYL')
+  config.active_record.encryption.key_derivation_salt = ENV.fetch('ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT', 'test_bsfGUb0GEmghCU5HZlBjILZwoMjypLpS')
```
sojan-official: lets remove the separate configs via environment file, rather have this in application.rb

config/routes.rb
```diff
@@ -334,6 +334,15 @@
           post :resend_confirmation
           post :reset_access_token
         end
+
+        # MFA routes
+        scope module: 'profile' do
+          resource :mfa, controller: 'mfa', only: [:show, :destroy] do
+            post :enable
```
sojan-official: enable can be `create` ?  lets keep closer to rails conventions when ever possible.

app/controllers/api/v1/profile/mfa_controller.rb
```diff
@@ -0,0 +1,61 @@
+class Api::V1::Profile::MfaController < Api::BaseController
+  before_action :set_user
+  before_action :check_mfa_enabled, only: [:destroy, :backup_codes]
+  before_action :check_mfa_disabled, only: [:enable, :verify]
+  before_action :validate_otp, only: [:verify, :backup_codes, :destroy]
+  before_action :validate_password, only: [:destroy]
+
+  def show; end
+
+  def enable
```
sojan-official: could be create . stay close to rails conventions

app/models/user.rb
```diff
@@ -156,6 +165,58 @@ def self.from_email(email)
     find_by(email: email&.downcase)
   end
 
+  # 2FA/MFA Methods
```
sojan-official, line 172: the user model is something that can get super bloated.  Can we move the logic here to a service instead ?

sojan-official (commented):
Took a first pass and mostly the Pr looks to be in a good shape. Add some minor improvement comments.  Apart from that we need to look into this part carefully.

This Pr will be introducing Active Record Encryption to Chatwoot code base for the first time. We need to ensure that we do it in a way that this is compatible for existing installations and won't break for someone who might perform migration with out reading through the full release notes if possible. 

https://edgeguides.rubyonrails.org/active_record_encryption.html#support-for-unencrypted-data

so we need to test out the behavior properly for cases like 
- how would the instance / feature work if we don't add the encryption keys 
- should we enable the feature only if the encryption keys are present

Also can we look at adding these configs 
- https://edgeguides.rubyonrails.org/active_record_encryption.html#support-for-unencrypted-data ( enables to true, since we will be enabling this for some other attributes as well -> in subsequent PR. We will need to handle both encrypted and unencrypted items till we finish migration )
- https://edgeguides.rubyonrails.org/active_record_encryption.html#config-active-record-encryption-store-key-references  lets also have this enabled.

app/controllers/api/v1/profile/mfa_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+    @user.enable_two_factor!
+    @backup_codes = @user.generate_backup_codes!
```
scmmishra: Would be good if this was in a single transaction. If the second operation fails, the user would be left in an inconsistent state with 2FA enabled but no backup codes.

scmmishra: Also, if this is called multiple times, will we regenerate backup codes. 

I think we should see if there's a change in two factor enabled or disabled state and generated automatically based on a commit hook. @sojan-official WDYT?

app/models/user.rb
```diff
@@ -156,6 +165,58 @@ def self.from_email(email)
...
+  def two_factor_provisioning_uri
+    return nil if otp_secret.blank?
+
+    issuer = 'Chatwoot'
+    label = email
+    otp_provisioning_uri(label, issuer: issuer)
+  end
+
+  def generate_backup_codes!
+    codes = Array.new(10) { format('%06d', SecureRandom.random_number(1_000_000)) }
```
scmmishra: I've typically seen usage of a 8 digit backup code. Let's use that

app/controllers/api/v1/profile/mfa_controller.rb
scmmishra: Also, should enable also require password validation?

app/models/user.rb
```diff
@@ -156,6 +165,58 @@ def self.from_email(email)
...
+    self.otp_backup_codes = codes
+    save!
+    codes
+  end
+
+  def validate_backup_code!(code)
+    return false if otp_backup_codes.blank? || code.blank?
+
+    codes = otp_backup_codes || []
+    index = codes.index(code)
```
scmmishra: Nitpick but `codes.index` can be open to timing attacks. 

We can do a constant time lookup. Something like this by checking all items

```
codes.each_with_index do |stored_code, idx|
  found_index = idx if stored_code == code
end
```

Rails has a constant time compare operator IIRC, we can use that too

app/views/api/v1/profile/mfa/enable.json.jbuilder
```diff
@@ -0,0 +1,6 @@
+issuer = Rails.application.class.module_parent.name
+label = "#{issuer}:#{@user.email}"
```
scmmishra: Should we be building this here? I don't think having logic in the jbuilder is a good idea

app/services/mfa/token_service.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def verify_token
+    decoded = decode_token
+    return nil if decoded.blank?
+
+    User.find(decoded[:user_id])
+  rescue ActiveRecord::RecordNotFound
+    nil
+  end
```
scmmishra, line 18: Is this used to verify the token for the first time?

scmmishra (changes requested):
Added some comments. Would be good to add some rate limits on token verification

app/controllers/api/v1/profile/mfa_controller.rb
sojan-official: yeah makes sense..  the backup codes action can be a reaction inside the service based on enable / disable .. also need to make this happen in a transaction.

app/services/mfa/token_service.rb
tds-1: This is for mfa token verification, generated after the first step of email, password verification

.env.example
```diff
@@ -9,9 +9,10 @@ SECRET_KEY_BASE=replace_with_lengthy_secure_hex
 # Active Record Encryption keys (required for MFA/2FA functionality)
 # Generate these keys by running: rails db:encryption:init
 # IMPORTANT: Use different keys for each environment (development, staging, production)
-# ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY=
-# ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY=
-# ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT=
+# For development, you can use these example values:
+ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY=sAK7479yQQvk0574SPM9RZFi9xx3dlBY
+ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY=t8mg48yhHqOaAH7R2HA4SDDsXWcyWBYL
+ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT=bsfGUb0GEmghCU5HZlBjILZwoMjypLpS
```
scmmishra: @tds-1 assuming you accidentally committed this. Let's not have any default value, people might forget to change this

tds-1: This has been intentionally committed, It is fine to have these values as default (can be used for development purposes)

sojan-official: lets not give them keys in development .. a lot of instances would rename env.example to env and use the system. so lets comment out the keys and keep them empty.

tds-1: Okay

app/controllers/api/v1/profile/mfa_controller.rb
tds-1: I have updated the functionality, when mfa is enabled for an account, no backup codes are generated, but when a totp is verified we generate the backup code, and on destroy we delete the backup codes. 
Generation of backup code and user otp update is done in transaction

sojan-official:
@codex review plz

tds-1:
@codex fix comments

.github/workflows/run_mfa_spec.yml
```diff
@@ -0,0 +1,105 @@
...
+        with:
+          bundler-cache: true
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 23
+          cache: 'pnpm'
+
+      - name: Install pnpm dependencies
+        run: pnpm i
```
scmmishra: We can get rid of these steps, since we're running only backend specs

app/controllers/api/v1/profile/mfa_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+  def check_mfa_enabled
+    render_could_not_create_error(I18n.t('errors.mfa.not_enabled')) unless @user.mfa_enabled?
+  end
+
+  def check_mfa_feature_available
+    return if Chatwoot.mfa_enabled?
+
+    render json: {
+      error: I18n.t('errors.mfa.feature_unavailable',
+                    default: 'MFA feature is not available. Please configure encryption keys.')
```
scmmishra: Let's add i18n for this

scmmishra (changes requested):

sojan-official:
@tds-1  lets also make sure to add these 2 configs 

https://edgeguides.rubyonrails.org/active_record_encryption.html#support-for-unencrypted-data ( enables to true, since we will be enabling this for some other attributes as well -> in subsequent PR. We will need to handle both encrypted and unencrypted items till we finish migration )
https://edgeguides.rubyonrails.org/active_record_encryption.html#config-active-record-encryption-store-key-references lets also have this enabled.

```
# we want to enable encryption on few other fields as well 
config.active_record.encryption.support_unencrypted_data = true

# config.active_record.encryption.store_key_references = true
for cases where we might have multiple keys for rotation
```

sojan-official:
@codex review plz

db/migrate/20250820130619_add_two_factor_to_users.rb
```diff
@@ -0,0 +1,11 @@
+class AddTwoFactorToUsers < ActiveRecord::Migration[7.1]
+  def change
+    add_column :users, :otp_secret, :string
+    add_column :users, :consumed_timestep, :integer
+    add_column :users, :otp_required_for_login, :boolean, default: false, null: false
```
sojan-official, line 5: @tds-1  skip the null: false in your migration ?  or update the schema to match that

sojan-official (approved):

sojan-official (changes requested):
lets fix, migration before merging

db/migrate/20250820130619_add_two_factor_to_users.rb
tds-1: updated the schema

app/controllers/api/v1/profile/mfa_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  private
+
+  def set_user
+    @user = current_user
+  end
+
+  def set_mfa_service
+    @mfa_service = Mfa::ManagementService.new(user: @user)
+  end
```
scmmishra: Nitpick, we can use this approach, which is more idiomatic. It memoizes the service instance and eliminates the need for a separate `before_action` callback

```
  def mfa_service
    @mfa_service ||= Mfa::ManagementService.new(user: @user)
  end
```

app/controllers/api/v1/profile/mfa_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def backup_codes
+    @backup_codes = @mfa_service.generate_backup_codes!
+  end
+
+  private
+
+  def set_user
+    @user = current_user
+  end
```
scmmishra: Do we need this? `Current.user` is always accessible.

app/services/mfa/management_service.rb
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  def provisioning_uri
+    return nil if user.otp_secret.blank?
+
+    issuer = 'Chatwoot'
+    label = user.email
+    user.otp_provisioning_uri(label, issuer: issuer)
+  end
+  alias two_factor_provisioning_uri provisioning_uri
```
scmmishra: Why do we need the alias? Can we just rename `provisioning_uri` to `two_factor_provisioning_uri`?

app/services/mfa/management_service.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def generate_backup_codes!
+    codes = Array.new(10) { format('%08d', SecureRandom.random_number(100_000_000)) }
```
scmmishra: @sojan-official need your thoughts, shouldn't the backup codes be alphanumeric... 

```suggestion
    codes = Array.new(10) { SecureRandom.hex(4).upcase }
```

scmmishra (commented):
Some minor comments, overall looks good

app/services/mfa/management_service.rb
sojan-official: any specific reasons ?  @tds-1  ? if not lets have them alphanumeirc

sojan-official:
@codex take one more pass

muhsin-k (approved):

PR #12299, opened 2025-08-26
refactor: Consolidate WhatsApp template components and improve naming

muhsin-k:
- Removed duplicate WhatsApp template parser component to eliminate code duplication
- Consolidated template parsing logic to use single source of truth
- Improved component naming to better reflect their roles and avoid confusion


iamsivin (approved):
LGTM

PR #12314, opened 2025-08-27
fix: Remove whatsapp prefix from twilio phone number for QR code

muhsin-k:
Twilio WhatsApp channels store phone numbers in the format `whatsapp:+1234567890`, but the WhatsApp URL for QR codes expects just the phone number without the prefix. This resulted in malformed URLs like `https://wa.me/whatsapp:+1234567890` instead of the correct `https://wa.me/+1234567890`. This PR fixes QR code generation for Twilio WhatsApp inboxes by stripping the "whatsapp:" prefix from phone numbers.


iamsivin (approved):
LGTM

PR #12320, opened 2025-08-28
feat: Assignment service (v2)

tds-1:
## Description

This PR introduces a new robust auto-assignment system for conversations in Chatwoot. The system replaces the existing round-robin assignment with a more sophisticated service-based architecture that supports multiple assignment strategies, rate limiting, and Enterprise features like capacity-based assignment and balanced distribution.

## Type of change

## How Has This Been Tested?

- Unit test cases
- Test conversations getting assigned on status change to open
- Test the job directly via rails console

## Checklist:

---

> [!NOTE]
> Adds a new service-based auto-assignment system with scheduled jobs, rate limiting, enterprise capacity/balanced selection, and wiring via inbox/handler; includes Redis helpers and comprehensive tests.
> 
> - **Auto-assignment v2 (core services)**:
>   - Add `AutoAssignment::AssignmentService` with bulk assignment, configurable conversation priority, RR selection, and per-agent rate limiting via `AutoAssignment::RateLimiter`.
>   - Add `AutoAssignment::RoundRobinSelector` for agent selection.
> - **Jobs & scheduling**:
>   - Add `AutoAssignment::AssignmentJob` (per-inbox bulk assign; env-based limit) and `AutoAssignment::PeriodicAssignmentJob` (batch over accounts/inboxes).
>   - Schedule periodic run in `config/schedule.yml` (`periodic_assignment_job`).
> - **Model/concerns wiring**:
>   - Include `InboxAgentAvailability` in `Inbox`; add `Inbox#auto_assignment_v2_enabled?`.
>   - Update `AutoAssignmentHandler` to trigger v2 job when `auto_assignment_v2_enabled?`, else fallback to legacy.
> - **Enterprise extensions**:
>   - Add `Enterprise::InboxAgentAvailability` (capacity-aware filtering) and `Enterprise::Concerns::Inbox` association `inbox_capacity_limits`.
>   - Extend service via `Enterprise::AutoAssignment::AssignmentService` (policy-driven config, capacity filtering, exclusion rules) and add selectors/services: `BalancedSelector`, `CapacityService`.
> - **Infrastructure**:
>   - Enhance `Redis::Alfred` with `expire`, key scan/count, and extended ZSET helpers (`zadd`, `zcount`, `zcard`, `zrangebyscore`).
> - **Tests**:
>   - Add specs for jobs, core service, rate limiter, RR selector, and enterprise features (capacity, balanced selection, exclusions).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0ebe187c8aea73765b0122a44b18d6f465c2477f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


muhsin-k:
@codex Please review.

config/features.yml
```diff
@@ -203,3 +199,7 @@
   enabled: false
   premium: true
   chatwoot_internal: true
+- name: assignment_v2
+  display_name: Assignment V2
+  enabled: false
+  chatwoot_internal: true
```
pranavrajs: Please keep the order as before. It changed it would behave differently.

The features should keep the order it was defined for the first time.

app/models/concerns/auto_assignment_handler.rb
```diff
@@ -14,7 +14,13 @@ def run_auto_assignment
     return unless conversation_status_changed_to_open?
     return unless should_run_auto_assignment?
 
-    ::AutoAssignment::AgentAssignmentService.new(conversation: self, allowed_agent_ids: inbox.member_ids_with_assignment_capacity).perform
+    if inbox.auto_assignment_enabled?
```
pranavrajs: I would suggest to use something like auto_assignment_v2 to make it explicit.

app/services/auto_assignment/assignment_service.rb
```diff
@@ -0,0 +1,99 @@
...
+  pattr_initialize [:inbox!]
+
+  def perform_for_conversation(conversation)
+    return false unless assignable?(conversation)
+
+    agent = find_available_agent
+    return false unless agent
+
+    assign_conversation(conversation, agent)
+  end
```
pranavrajs: I don't see this method being used anywhere else, why do we need to keep it a public method?

app/services/auto_assignment/assignment_service.rb
```diff
@@ -0,0 +1,99 @@
...
+    assigned_count
+  end
+
+  private
+
+  def assignable?(conversation)
+    inbox.enable_auto_assignment? &&
+      conversation.status == 'open' &&
+      conversation.assignee_id.nil?
+  end
```
pranavrajs, line 30: I'm not sure about this condition. Why do we need this? 

The `inbox.enable_auto_assignment?` is already checked in the perform_bulk_assignment method. We are taking the unassigned open conversations. So, why do we need to check this again when assigning?

app/services/auto_assignment/assignment_service.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  def apply_conversation_priority(scope)
+    case assignment_config['conversation_priority']
+    when 'longest_waiting'
+      scope.order(last_activity_at: :asc, created_at: :asc)
+    else
+      scope.order(created_at: :asc)
+    end
+  end
```
pranavrajs, line 42:
```suggestion
  def unassigned_conversations(limit)
  scope = inbox.conversations.unassigned.open
  
  if assignment_config['conversation_priority'] == 'longest_waiting'
    scope = scope.order(:last_activity_at, :created_at)
  else
    scope = scope.order(:created_at)
  end
  
  scope.limit(limit)
end
```

Keeping it a single function seems more readable.

app/services/auto_assignment/assignment_service.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  def dispatch_assignment_event(conversation, agent)
+    Rails.configuration.dispatcher.dispatch(
+      Events::Types::ASSIGNEE_CHANGED,
+      Time.zone.now,
+      conversation: conversation,
+      user: agent
+    )
+  end
```
pranavrajs, line 75: Any particular reason why this event is manually triggered from here?

app/services/auto_assignment/assignment_service.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    rate_limiter = build_rate_limiter(agent)
+    rate_limiter.track_assignment(conversation)
+
+    dispatch_assignment_event(conversation, agent)
+    true
+  rescue ActiveRecord::RecordInvalid => e
+    Rails.logger.error "AutoAssignment failed for conversation #{conversation.id}: #{e.message}"
+    false
+  end
```
pranavrajs: What is the case where `RecordInvalid` would be thrown?

app/jobs/auto_assignment/periodic_assignment_job.rb
```diff
@@ -0,0 +1,15 @@
...
+  def perform
+    Account.find_each do |account|
+      next unless account.feature_enabled?('assignment_v2')
+
+      account.inboxes.joins(:assignment_policy).find_each do |inbox|
+        next unless inbox.auto_assignment_enabled?
+
+        AutoAssignment::AssignmentJob.perform_later(inbox_id: inbox.id)
+      end
+    end
```
pranavrajs: find_in_batches rather than find_each

app/jobs/auto_assignment/periodic_assignment_job.rb
```diff
@@ -0,0 +1,15 @@
+class AutoAssignment::PeriodicAssignmentJob < ApplicationJob
+  queue_as :scheduled_jobs
+
+  def perform
+    Account.find_each do |account|
+      next unless account.feature_enabled?('assignment_v2')
+
+      account.inboxes.joins(:assignment_policy).find_each do |inbox|
```
pranavrajs: does join help here?

app/services/auto_assignment/rate_limiter.rb
```diff
@@ -0,0 +1,49 @@
+class AutoAssignment::RateLimiter
+  pattr_initialize [:inbox!, :agent!]
+
+  def within_limit?
+    return true unless enabled?
+
+    current_count < limit
+  end
```
pranavrajs, line 8: Let us discuss this over a call.

config/locales/en.yml
```diff
@@ -394,6 +394,11 @@ en:
       %{format_messages}
   agent_capacity_policy:
     inbox_already_assigned: 'Inbox has already been assigned to this policy'
+  jobs:
+    auto_assignment:
+      assignment_job:
+        bulk_assignment_success: 'Assigned %{assigned_count} conversations for inbox %{inbox_id}'
+        bulk_assignment_failed: 'Bulk assignment failed for inbox %{inbox_id}: %{error_message}'
```
pranavrajs: Logs are in english. You don't need a locale file entry for it.

enterprise/app/models/enterprise/assignment_policy.rb
```diff
@@ -0,0 +1,20 @@
...
+  def assignment_order
+    value = read_attribute(:assignment_order)
+    return 'balanced' if value == 1
+
+    super
+  end
+
+  def balanced?
+    self[:assignment_order] == 1
+  end
```
pranavrajs: why do we need this? this is already handled in enterprise/concerns/assignment_policy

app/jobs/auto_assignment/periodic_assignment_job.rb
tds-1: this avoids n+1 queries, when we call inbox.auto_assignment_v2_enabled?

app/services/auto_assignment/assignment_service.rb
tds-1: we directly update the conversation with conversation.update!(assignee: agent), that would not trigger the default one

app/jobs/auto_assignment/periodic_assignment_job.rb
scmmishra, line 1: This is fine for now, but looping over all the accounts first and their inboxes might be suboptimal. There's a point where we can safely assume that if they have an assignment policy for an inbox, the feature is likely enabled.

However, do note that this feedback is for later. We can safely let this through for now, anyway, we run this job every 30 mins.

app/services/auto_assignment/assignment_service.rb
```diff
@@ -0,0 +1,90 @@
+class AutoAssignment::AssignmentService
+  pattr_initialize [:inbox!]
+
+  def perform_bulk_assignment(limit: 100)
+    return 0 unless inbox.enable_auto_assignment
```
scmmishra: Should we be using `auto_assignment_v2_enabled?` instead?

app/services/auto_assignment/assignment_service.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def assignable?(conversation)
+    conversation.status == 'open' &&
+      conversation.assignee_id.nil?
+  end
+
+  def unassigned_conversations(limit)
+    scope = inbox.conversations.unassigned.open
+
+    scope = if assignment_config['conversation_priority'] == 'longest_waiting'
```
scmmishra: Rails already provides a method `earliest_created?` and `longest_waiting?` you could use that instead of a brittle string matching. Similarly for `round_robin?` and `balanced?` 

We also have an option for setting up prefix and suffix for these methods, [ref](https://api.rubyonrails.org/classes/ActiveRecord/Enum.html#:~:text=You%20can%20use%20the%20%3Aprefix%20or%20%3Asuffix%20options%20when%20you%20need%20to%20define%20multiple%20enums%20with%20same%20values.%20If%20the%20passed%20value%20is%20true%2C%20the%20methods%20are%20prefixed/suffixed%20with%20the%20name%20of%20the%20enum.%20It%20is%20also%20possible%20to%20supply%20a%20custom%20value%3A)

We don't need the suffix or prefix for now, but we can use the enum methods anyway

app/services/auto_assignment/assignment_service.rb
```diff
@@ -0,0 +1,90 @@
...
+              scope.order(last_activity_at: :asc, created_at: :asc)
+            else
+              scope.order(created_at: :asc)
```
scmmishra: See if using `reorder` here would be better

app/services/auto_assignment/rate_limiter.rb
```diff
@@ -0,0 +1,49 @@
...
+    @config ||= inbox.auto_assignment_config || {}
+  end
+
+  def assignment_key_pattern
+    "assignment:#{inbox.id}:agent:#{agent.id}:*"
+  end
+
+  def build_assignment_key(conversation_id)
+    "assignment:#{inbox.id}:agent:#{agent.id}:conversation:#{conversation_id}"
+  end
```
scmmishra, line 48: Use `lib/redis/redis_keys.rb`

app/services/auto_assignment/assignment_service.rb
```diff
@@ -0,0 +1,90 @@
...
+    agents = filter_agents_by_rate_limit(inbox.available_agents)
+    return nil if agents.empty?
+
+    round_robin_selector.select_agent(agents)
+  end
+
+  def filter_agents_by_rate_limit(agents)
+    agents.select do |agent_member|
+      rate_limiter = build_rate_limiter(agent_member.user)
+      rate_limiter.within_limit?
```
scmmishra, line 54: Okay, so I have one concern. The `within_limit?` check happens at time T1. 
And the the assignment happens at `T2`, in `def assign_conversation`

There can be a race condition right? Say with a limit of 5, a single processes trying to run this
1. We call `within_limit?` → see count = 4
2. An assignment happens in the background, (manually perhaps)
3. Both work and count becomes 6 (violating limit of 5)

For now, I think it's not an immediate concern, but something to keep in mind. 
CC: @sojan-official

enterprise/app/services/enterprise/auto_assignment/assignment_service.rb
```diff
@@ -0,0 +1,101 @@
...
+    {
+      'conversation_priority' => policy.conversation_priority,
+      'fair_distribution_limit' => policy.fair_distribution_limit,
+      'fair_distribution_window' => policy.fair_distribution_window,
+      'balanced' => policy.balanced?
+    }.compact
+  end
+
+  # Override to check policy first
+  def assignment_enabled?
```
scmmishra: I couldn't find this method in the base service

enterprise/app/services/enterprise/auto_assignment/assignment_service.rb
```diff
@@ -0,0 +1,101 @@
...
+    return policy.enabled? if policy
+
+    super
+  end
+
+  # Extend agent finding to add capacity checks
+  def find_available_agent
+    agents = filter_agents_by_rate_limit(inbox.available_agents)
+    agents = filter_agents_by_capacity(agents) if capacity_filtering_enabled?
```
scmmishra, line 19: I have a feeling this will be a perf bottleneck in the future, since the capacity check does a DB count for each conversation. 

While it will make it complex, we will need a precompute + update mechanism like we do it for rate limiting.  

@sojan-official do you think this will an issue?

scmmishra (commented):
Hey @tds-1 overall this is great. We can test the feature on staging soon. I've added some comments, mostly are nitpicks, or points to note for the future. LMK what you think

CC: @sojan-official tagged you in a few comments, PTAL

scmmishra (approved):
Tested this IRL with @tds-1 and @stonecharioteer 

- Round Robin & Balanced
- Tested Rate Limiting
- Tested Capacity
- Tested Priority Order

@tds-1 to check how V1 assignment works if the flag is not enabled. Just in case. Also check https://github.com/chatwoot/chatwoot/pull/12320#discussion_r2496006093

@tds-1 when updating capacity, the save on blur does not have any feedback. Please work with @iamsivin and @absurdiya to fix this. Preferably in a separate PR.

enterprise/app/models/enterprise/inbox_agent_availability.rb
```diff
@@ -0,0 +1,31 @@
+module Enterprise::InboxAgentAvailability
+  extend ActiveSupport::Concern
+
+  def member_ids_with_assignment_capacity
+    return member_ids unless capacity_filtering_enabled?
+
+    # Get online agents with capacity
+    agents = available_agents
+    agents = filter_by_capacity(agents) if capacity_filtering_enabled?
```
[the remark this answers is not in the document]
scmmishra: @tds-1 please check this before merging

PR #12426, opened 2025-09-12
fix: `display_id` to `id` mapping not handled

scmmishra:
The frontend filtering didn't handle the `id` to `display_id` mapping of conversations. This PR fixes it


iamsivin (approved):
I have tested this PR locally, and it works as expected.
image

muhsin-k (approved):

PR #12450, opened 2025-09-16
feat: Introduce allowed_domains for web widget

### LINKED ISSUE #1985: Website inbox IP or Domain restriction
Hi there, we are trying to setup a custom installation of chatwoot.
Is there any way to restrict the usage of the web snippet to the actual domain / IP of the site we wnat to integrate?

pranavrajs:
We wanted to provide an option for users to specify the domains on which they can show the website. The rest of the sites shouldn't see the widget at all. 

It's not possible generally through Origin because you can't get Origin when loading via an iframe. What I've done is add frame ancestors for the domains specified in allowed domains. I hope this solves most of the problems. 

This is added in a way that it won't affect existing widgets. Only If they have configured allowed domains, it will start blocking. Otherwise, it would follow the previous behavior without any changes.

This change supports called wild card domains as well. You can add a comma‑separated list of domains, either wild card or regular domains.

---

To test, deploy to staging.  Call the following API to update the allowed_domains list.

```
URL: PATCH /api/v1/accounts/<account-id>/inboxes/<inbox-id>

Payload:
{
   "channel": { "allowed_domains": "*.chatwoot.dev,chatwoot.com" }
}

```

Fixes https://github.com/chatwoot/chatwoot/issues/1985

### DIFF
--- a/app/controllers/widgets_controller.rb
+++ b/app/controllers/widgets_controller.rb
@@ -70,7 +70,12 @@ class WidgetsController < ActionController::Base
   end
 
   def allow_iframe_requests
-    response.headers.delete('X-Frame-Options')
+    if @web_widget.allowed_domains.blank?
+      response.headers.delete('X-Frame-Options')
+    else
+      domains = @web_widget.allowed_domains.split(',').map(&:strip).join(' ')
+      response.headers['Content-Security-Policy'] = "frame-ancestors #{domains}"
+    end
   end
 end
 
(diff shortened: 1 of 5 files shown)


muhsin-k (approved):
LGTM
https://www.loom.com/share/18ea1ea2ba854d99844e319459494725

iammattmartin:
Any docs on how to use this, or can this not be done via the UI?

pranavrajs:
@iammattmartin This is only available via API now. It will be updated in the UI in the next releases.

PR #12454, opened 2025-09-17
feat: Add support for grouped file uploads in Slack

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-5646/add-support-for-grouped-file-uploads-in-slack

Previously, when sending multiple attachments to Slack, we uploaded them one by one. For example, sending 5 images would result in 5 separate Slack messages. This created clutter and a poor user experience, since Slack displayed each file as an individual message.
This PR updates the implementation to group all attachments from a message and send them as a single Slack message. As a result, attachments now appear together in one grouped block, providing a much cleaner and more intuitive experience for users.

**Before:** 
Each file uploaded as a separate Slack message.
before

**After:** 
All files from a single message grouped and displayed together in one Slack message (similar to how Slack natively handles grouped uploads).
after

**Changes**

- Upgraded Slack file upload implementation to use the new multiple attachments API available in slack-ruby-client `v2.7.0`.
- Updated attachment handling to upload all files from a message in a single API call.
- Enabled proper attachment grouping in Slack, ensuring related files are presented together.


muhsin-k:
@codex Please review.

sojan-official (approved):

PR #12474, opened 2025-09-19
feat: add independent cache for each tab

scmmishra:


nestordavalos:
Thanks for detailing the issue 👌

I’ve been experimenting with a refactor of the SET_ALL_CONVERSATION mutation that might help address this pagination + cache bug. The core idea is to make the mutation aware of the page number, so it can preserve the server’s intended ordering instead of always appending to the end of the array.

Key changes in this approach:

- On page 1, merge the new conversations with existing ones, but respect the server’s pagination order while retaining properties already loaded (messages, attachments) for the selected chat.
- On subsequent pages, append conversations in the correct order so the list grows consistently.
- Conversations not present in the current page payload are retained, avoiding data loss when switching between tabs.
- This way, the conversation list should stay consistent and ordered correctly, without the “missing until scroll” issue.

It’s just an idea 💡 — if it helps with your fix, feel free to consider it. If not, no worries!

https://github.com/chatwoot/chatwoot/pull/12505

PR #12496, opened 2025-09-23
fix: Restore IPs behind Cloudflare and Traefik

Micorksen:
# Pull Request Template

## Description

Allows the restoring of the client IP behind Cloudflare when you are using Traefik. With this fix, users using NGINX won’t need to setup a config file to get the visitor IP.
Fixes #(no issue at the moment)

## Type of change

## How Has This Been Tested?

We did use this tiny patch in our Docker container by mounting it and opening a test ticket from our website. It now showed the visitor IP instead of Traefik IP.

## Checklist:


Micorksen:
This fix could also be applied to known public Reverse Proxies like Google Cloud, OVHcloud… or provide a `TRUSTED_PROXIES` variable for the user and just let him put whatever he wants in. That could be cleaner than a single file for a single provider.

config/initializers/proxies.rb
stonecharioteer, line 1: Hello, thanks for your contribution. This is helpful.

Now, this is good, but you'd need to add error handling because `.map { |proxy| IPAddr.new(proxy) }` could raise an `IPAddr::InvalidAddressError`

Do this instead.

```ruby
ENV.fetch('TRUSTED_PROXIES', '')
  .split(',"
  .map(&:strip)
  .reject(&:empty?)
  .filter_map do |proxy|
    IPAddr.new(proxy)
  rescue IPAddr::InvalidAddressError => e
    Rails.logger.warn "Invalid IP: #{proxy}"
    nil # return nil for invalid entries
  end
```

This will account for if someone's set `TRUSTED_PROXIES='172.17.0.1,cloudflare,10.0.0.1'`, where `cloudflare` is obviously a bad IP.

Micorksen: Hey!
Nice catch. I'll make the changes before merge.

Micorksen:
> It seems like CircleCI fails due to an audit error, not caused by my code.

It is due in fact by the fact that Gemfile.lock has been edited since my fork was made.
I did a rebase to fix this issue.

vini-fitzner:
Hi @Micorksen — I noticed your PR #12496 addresses a very similar problem around restoring the real client IP behind proxies.

I recently opened #13942, which tackles the same root issue with a slightly broader approach: aligning Rack::Attack IP resolution with `action_dispatch.remote_ip`, adding configurable `TRUSTED_PROXY_CIDRS`, and also improving `/widget` throttling behavior with specs.

Sharing it here in case it helps maintainers compare approaches or consolidate the work. Happy to adapt or split pieces if that makes review easier.

PR #12520, opened 2025-09-25
feat: use resolved contacts as base relation for filtering

scmmishra:
This PR has two changes to speed up contact filtering

### Updated Base Relation

Update the `base_relation` to use resolved contacts scope to improve perf when filtering conversations. This narrows the search space drastically, and what is usually a sequential scan becomes a index scan for that `account_id`

ref: https://github.com/chatwoot/chatwoot/pull/9347
ref: https://github.com/chatwoot/chatwoot/pull/7175/

Result: https://explain.dalibo.com/plan/c8a8gb17f0275fgf#plan

## Selective filtering in Compose New Conversation

We also cost of filtering in compose new conversation dialog by reducing the search space based on the search candidate. For instance, a search term that obviously can’t be a phone, we exclude that from the filter. Similarly we skip name lookups for email-shaped queries.

Removing the phone number took the query times from 50 seconds to under 1 seconds

### Comparison

1. Only Email: https://explain.dalibo.com/plan/h91a6844a4438a6a 
2. Email + Name: https://explain.dalibo.com/plan/beg3aah05ch9ade0
3. Email + Name + Phone: https://explain.dalibo.com/plan/c8a8gb17f0275fgf

PR #12521, opened 2025-09-25
feat: Optimize contact search keys

scmmishra:
This PR reduces the cost of filtering in compose new conversation dialog by reducing the search space based on the search candidate. For instance, a search term that obviously can’t be a phone, we exclude that from the filter. Similarly we skip name lookups for email-shaped queries.


scmmishra:
Included with https://github.com/chatwoot/chatwoot/pull/12520

PR #12522, opened 2025-09-25
chore: Simplify reply message lookup with store-based fetching

iamsivin:
# Pull Request Template

## Description

In this [PR](https://github.com/chatwoot/chatwoot/pull/10024), we introduced the ability to load reply-to messages dynamically when they are missing from the message list. The earlier implementation relied on fetching directly from the API, which sometimes caused multiple redundant calls.

The updated approach shifts this to a store-based logic. While fetching conversation messages, any reply-to messages are also cached in the store. This ensures that when the same reply-to message is needed again (for example, during scroll-up), it is served from the store instead of triggering another API call.

## Checklist:


iamsivin:
Closing this PR for now, will revisit it later.

PR #12542, opened 2025-09-29
fix: Optimize Slack channel fetching to avoid rate limiting issues

muhsin-k:
### Problem
The Slack integration fails when fetching channels from workspaces with many channels due to rate limiting errors. The current implementation makes a single API call requesting both public and private channels simultaneously with `types: 'public_channel,private_channel'`, which causes Slack's API to apply complex filtering and hit rate limits more frequently.

  When testing with a csutomer workspace containing 157 channels:
  - Combined request: Hit rate limits after a few pages, required 22+ API calls with long delays
  - Separate requests: Private channels (1 channel) load instantly, public channels (185 channels) load quickly
  
 
  ### Solution

  Split the channel fetching into two sequential steps:
  1. **Fetch private channels first** with `limit: 1000` (expects very few)
  2. **Fetch public channels second** with pagination as needed

 This approach leverages the fact that Slack's API handles single-type requests much more efficiently than mixed-type requests, avoiding the rate limiting issues entirely while maintaining the same functionality.

### DIFF
--- a/lib/integrations/slack/channel_builder.rb
+++ b/lib/integrations/slack/channel_builder.rb
@@ -24,13 +24,27 @@ class Integrations::Slack::ChannelBuilder
   end
 
   def channels
-    conversations_list = slack_client.conversations_list(types: 'public_channel,private_channel', exclude_archived: true)
+    channel_list = []
+
+    # Step 1: Fetch all private channels in one call (expect very few)
+    private_channels = fetch_channels_by_type('private_channel')
+    channel_list.concat(private_channels)
+
+    # Step 2: Fetch public channels with pagination
+    public_channels = fetch_channels_by_type('public_channel')
+    channel_list.concat(public_channels)
+    channel_list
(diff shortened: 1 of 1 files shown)


sojan-official (approved):

lib/integrations/slack/channel_builder.rb
```diff
@@ -24,13 +24,27 @@ def slack_client
   end
 
   def channels
-    conversations_list = slack_client.conversations_list(types: 'public_channel,private_channel', exclude_archived: true)
+    channel_list = []
+
+    # Step 1: Fetch all private channels in one call (expect very few)
```
sojan-official, line 33: lets add a more detailed comment here. that if we fetch both together, slack won't return the whole list .. so that we will have to do too many calls.

muhsin-k: Done

PR #12618, opened 2025-10-08
fix: Optimize message reindexing to reduce sidekiq job creation

vishnu-narayanan:
# Pull Request Template

## Description

Changes searchkick callback behavior to check `should_index?` before creating reindex jobs, preventing unnecessary job creation for messages that don't need indexing (activity messages, unpaid accounts, etc.).

 Previously, `callbacks: :async` created reindex jobs for all messages (~5,100/min or 7.3M/day in production), which were then filtered by `should_index?` inside the job worker - resulting in 98% wasted jobs, Redis memory pressure, and avoidable p0 alerts.

Now, `should_index?` is checked before job creation via  `after_commit` callback, reducing job creation to actual incoming/outgoing messages from paid accounts.

  Changes:
  - Disable automatic searchkick callbacks
  - Add manual `after_commit` callback with `should_index?` condition
  - Add specs to verify callback behavior

  Expected impact:
  - 98% reduction in sidekiq job creation (~7.3M → ~150K jobs/day)
  - Reduced redis memory usage
  - Same async indexing behavior for eligible messages

## Type of change

## How Has This Been Tested?

- [x] Added specs
- [x] Tested in staging

## Checklist:

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -39,7 +39,7 @@
 #
 
 class Message < ApplicationRecord
-  searchkick callbacks: :async if ChatwootApp.advanced_search_allowed?
+  searchkick callbacks: false if ChatwootApp.advanced_search_allowed?
 
   include MessageFilterHelpers
   include Liquidable
@@ -135,6 +135,7 @@ class Message < ApplicationRecord
   after_create_commit :execute_after_create_commit_callbacks
 
   after_update_commit :dispatch_update_event
+  after_commit :reindex_for_search, if: :should_index?, on: [:create, :update]
 
   def channel_token
     @token ||= inbox.channel.try(:page_access_token)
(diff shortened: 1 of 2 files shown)


app/models/message.rb
```diff
@@ -135,6 +135,7 @@ class Message < ApplicationRecord
   after_create_commit :execute_after_create_commit_callbacks
 
   after_update_commit :dispatch_update_event
+  after_commit :reindex_for_search, if: :should_index?, on: [:create, :update]
```
pranavrajs, line 138: Shouldn't we do this on destroy as well?

pranavrajs (approved):

app/models/message.rb
vishnu-narayanan: Yes, I’m thinking of handling it separately as a `bulk_delete_job` as part of the existing delete_object_job. This would help us avoid an avalanche of `ReindexV2` jobs in case of an account or inbox deletion with a large number of messages.

PR #12627, opened 2025-10-09
fix(filters): correct null matching logic [CW-5741]

scmmishra:
This PR fixes a bug came from assuming the old null check only mattered for the `is_not_present` filter. 

The fix keeps `not_equal_to` working but lets each operator decide what to do with `null`. Presence filters look at a shared `isNullish` flag, text filters still rely on `contains`, and date filters skip conversations with no timestamp. The new spec covers the null-assignee scenario for both `equal_to` and `not_equal_to` so we don’t miss this again.


scmmishra:
@codex please review

scmmishra:
@codex review

iamsivin (approved):
LGTM

PR #12630, opened 2025-10-09
fix: Handle video file types in Slack file shares

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-5752/fix-nomethoderror-when-processing-video-files-in-slack-integration
#### Problem
When users shared video files (like MP4) through Slack, the `file_type` method in `SlackMessageHelper` would return `nil` for unsupported file types. This caused a `NoMethodError (undefined method 'to_sym' for nil)` when the attachment was being processed, as the system expected a symbol value for the `file_type` attribute.

#### Solution
  - Added video file type support in the `file_type` method case statement
  - Added `else` clause to default unknown file types to `:file` instead of returning `nil`
  - This ensures `file_type` always returns a symbol, preventing the `to_sym` error


sojan-official (approved):
Code LGTM

PR #12639, opened 2025-10-10
feat(webhooks): allow localhost URLs on local environment

gabrieljablonski:
## Description

The `url` validator from `@vuelidate/validators` does not accept localhost URLs (`localhost` and `127.0.0.1`) as valid.

This is inconvenient when working with webhooks on a local environment, so allowing them selectively is a big improvement on Developer Experience.

## Type of change

## How Has This Been Tested?

Validator specs were added, and also tested manually with webhook creation form.

| Before | After |
| --- | --- |
| image | image |

## Checklist:


sojan-official:
Thanks for the suggestion.

This is already configurable from the Rails console for self-hosted instances. For now, we don’t plan to add this as a product-level setting since it feels more like an admin/deployment configuration than something most users would need in the UI.

We’ll keep an eye on this and can revisit if we see more demand for making it configurable from the product.

PR #12640, opened 2025-10-10
feat(resend): add resend as ActionMailer delivery method

gabrieljablonski:
# Pull Request Template

## Description

SMTP is not always available to use in certain self-host scenarios ([e.g. DigitalOcean](https://docs.digitalocean.com/support/why-is-smtp-blocked/)), so having an easy-to-setup alternative such as Resend is great.

## Type of change

## How Has This Been Tested?

Added specs for new mailer delivery method, and also tested in production for several months without any issues.

To test, simply create a [Resend](https://resend.com) account and API key, and set Chatwoot env variable `RESEND_API_KEY`.

All transactional and notification emails should now go through Resend.

## Checklist:


app/mailers/application_mailer.rb
```diff
@@ -19,7 +19,7 @@ def global_config
   rescue_from(*ExceptionList::SMTP_EXCEPTIONS, with: :handle_smtp_exceptions)
 
   def smtp_config_set_or_development?
-    ENV.fetch('SMTP_ADDRESS', nil).present? || Rails.env.development?
+    ENV.fetch('SMTP_ADDRESS', nil).present? || ENV.fetch('RESEND_API_KEY', nil).present? || Rails.env.development?
```
gabrieljablonski, line 22: We might consider renaming this method for clearer meaning, but it might not be necessary.

```suggestion
  def mailer_config_set_or_development?
    ENV.fetch('SMTP_ADDRESS', nil).present? || ENV.fetch('RESEND_API_KEY', nil).present? || Rails.env.development?
```

app/views/devise/mailer/reset_password_instructions.html.erb
```diff
@@ -4,5 +4,7 @@
 
 <p><%= link_to 'Change my password', frontend_url('auth/password/edit', reset_password_token: @token) %></p>
 
+<p style="color: #999999; font-size: 12px;">Copy and paste the URL into your browser if the link above doesn't work: <%= frontend_url('auth/password/edit', reset_password_token: @token) %></p>
+
```
gabrieljablonski, line 8: QoL change.

lib/mail/resend_provider.rb
```diff
@@ -0,0 +1,33 @@
+module Mail # rubocop:disable Style/ClassAndModuleChildren
+  class ResendProvider
```
gabrieljablonski, line 2: Based on implementation for `Sendmail`.

https://github.com/mikel/mail/blob/master/lib/mail/network/delivery_methods/sendmail.rb

lib/mail/resend_provider.rb
```diff
@@ -0,0 +1,33 @@
...
+    class DeliveryError < StandardError; end
+
+    attr_accessor :settings
+
+    def initialize(values)
+      @settings = values
+    end
```
gabrieljablonski, line 9: Although we don't use any settings, `def initialize(values)` is required.

We can consider simplifying.

```suggestion
    def initialize(_values); end
```

pranavrajs:
@gabrieljablonski Sorry about the delay in getting back to you.

Unfortunately, we will not be able to accept this change. Taking this in would mean that the ongoing maintenance responsibility falls on us, which is not something we can commit to right now.

The existing SMTP integration is working well for the majority of the community, and we do not see a strong need to introduce and maintain an additional implementation at this point. For that reason, we will not be moving forward with this change.

Also worth noting that Resend already supports SMTP: https://resend.com/docs/send-with-smtp

gabrieljablonski:
@pranavrajs no worries, totally understandable!

As mentioned in the PR description, the idea was to serve as an alternative to providers that do not support SMTP for whatever reason (like Digital Ocean, see https://docs.digitalocean.com/support/why-is-smtp-blocked/).

But I think is a very rare situation which does not affect many providers, so might not justify the effort.

Thanks for the feedback.

PR #12648, opened 2025-10-11
feat: Secure external credentials with database encryption

sojan-official:
## Changelog

 - Added conditional Active Record encryption to every external credential we store (SMTP/IMAP passwords, Twilio tokens,
    Slack/OpenAI hook tokens, Facebook/Instagram tokens, LINE/Telegram keys, Twitter secrets) so new writes are encrypted
    whenever Chatwoot.encryption_configured? is true; legacy installs still receive plaintext until their secrets are
    updated.
  - Tuned encryption settings in config/application.rb to allow legacy reads (support_unencrypted_data) and to extend
    deterministic queries so lookups continue to match plaintext rows during the rollout; added TODOs to retire the
    fallback once encryption becomes mandatory.
  - Introduced an MFA-pipeline test suite (spec/models/external_credentials_encryption_spec.rb) plus shared examples to
    verify each attribute encrypts at rest and that plaintext records re-encrypt on update, with a dedicated Telegram case.
    The existing MFA GitHub workflow now runs these tests using the preconfigured encryption keys.

fixes:  https://linear.app/chatwoot/issue/CW-5453/encrypt-sensitive-credentials-stored-in-plain-text-in-database

## Testing Instructions

 1. Instance without encryption keys
      - Unset ACTIVE_RECORD_ENCRYPTION_* vars (or run in an environment where they’re absent).
      - Create at least one credentialed channel (e.g., Email SMTP).
      - Confirm workflows still function (send/receive mail or a similar sanity check).
      - In the DB you should still see plaintext values—this confirms the guard prevents encryption when keys are missing.
  2. Instance with encryption keys
      - Configure the three encryption env vars and restart.
      - Pick a couple of representative integrations (e.g., Email SMTP + Twilio SMS).
      - Legacy channel check:
          - Use existing records created before enabling keys. Trigger their workflow (send an email / SMS, or hit the
            webhook) to ensure they still authenticate.
          - Inspect the raw column—value remains plaintext until changed.
      - Update legacy channel:
          - Edit one legacy channel’s credential (e.g., change SMTP password).
          - Verify the operation still works and the stored value is now encrypted (raw column differs, accessor returns
            original).
      - New channel creation:
          - Create a new channel of the same type; confirm functionality and that the stored credential is encrypted from
            the start.


scmmishra:
@codex please review

muhsin-k (approved):
LGTM and tested locally.

PR #12664, opened 2025-10-14
feat: single query for reporting event stats

scmmishra:
This PR collapses multiple queries fetching stats from a single table to a single query

```sql
SELECT 
  user_id as user_id,
  COUNT(CASE WHEN name = 'conversation_resolved' THEN 1 END) as resolved_count,
  AVG(CASE WHEN name = 'conversation_resolved' THEN value END) as avg_resolution_time,
  AVG(CASE WHEN name = 'first_response' THEN value END) as avg_first_response_time,
  AVG(CASE WHEN name = 'reply_time' THEN value END) as avg_reply_time 
FROM "reporting_events"
WHERE 
  "reporting_events"."account_id" = <account_id> AND 
  "reporting_events"."created_at" >= '2025-09-14 18:30:00' AND 
  "reporting_events"."created_at" < '2025-10-14 18:29:59'
GROUP BY "reporting_events"."user_id";
```

### Why this works?

Here's why this optimization is faster based on PostgreSQL internals:

- Single Table Scan vs Multiple Scans: Earlier we did 4 sequential scans (or 4 index scans) of the same data, with the same where clause, now in a single scan all 4 `CASE` expressions are evaluated in a single pass. 
- Shared Buffer Cache Efficiency: PostgreSQL's shared buffer cache stores recently accessed pages, with this, pages are loaded once and re-used for all aggregation, earlier with separate queries we were forced to re-read all from the disk each time
- Reduced planning and network overhead (4 vs 1 query)

### How is it tested

1. The specs all pass without making any changes
2. Verified the reports side by side after generating from report seeder

#### How to test

Generate seed data using the following command

```bash
ACCOUNT_ID=1 ENABLE_ACCOUNT_SEEDING=true bundle exec rake db:seed:reports_data
```

Once done download the reports, checkout to this branch and download the reports again and compare them


scmmishra:
@codex please review

stonecharioteer:
Looks good.
Out of curiosity, I asked claude and it said that this pattern of multiple queries exists in both the LabelSummaryBuilder and in the BotMetricsBuilder. Do you think you should change that too?

Ps: I don't yet have the permissions to mark this as approved btw.

scmmishra:
> Looks good. Out of curiosity, I asked claude and it said that this pattern of multiple queries exists in both the LabelSummaryBuilder and in the BotMetricsBuilder. Do you think you should change that too?
> 
> Ps: I don't yet have the permissions to mark this as approved btw.

We recently pushed a bunch of changes to `LabelSummaryBuilder` here: https://github.com/chatwoot/chatwoot/pull/11194
Not sure if Claude got it wrong, but I'll give it another shot 

About `BotMetricsBuilder` yeah that needs work too, I'll address it in a separate PR.

stonecharioteer:
> We recently pushed a bunch of changes to LabelSummaryBuilder here: https://github.com/chatwoot/chatwoot/pull/11194
> Not sure if Claude got it wrong, but I'll give it another shot

Checked again, let me see what it says.

> ● Great question! You're absolutely right to bring that up. PR #11194 did optimize
>   LabelSummaryBuilder, but it solved a different problem. Here's what actually happened:
> 
>   What PR #11194 Actually Fixed (March 2025)
> 
>   Problem: Massive N+1 query issue - looping over each label
>   - 6 labels = 41 queries
>   - 80 labels = 479 queries (!)
> 
>   Solution: Batch query all labels at once
>   - Result: Down to just 5 queries total (not per label)
>   - This was a huge optimization
> 
>   Pattern Used: Batch querying with GROUP BY to get all labels in one go
> 
>   What It DIDN'T Fix
> 
>   PR #11194 still made 3 separate queries to reporting_events:
>   1. fetch_metrics('conversation_resolved')
>   2. fetch_metrics('first_response')
>   3. fetch_metrics('reply')
> 
>   Then Things Got Worse
> 
>   August 2025 - PR #12261 (bug fix): Changed fetch_resolved_counts to query reporting_events
>   instead of conversations.status
>   - Result: Now makes 4 queries to reporting_events (was 3)
>   - Reason: Needed to count resolution events, not current conversation status
> 
>   Current State
> 
>   LabelSummaryBuilder NOW has the exact same pattern PR #12664 fixes:
>   - 4 queries to reporting_events
>   - Same joins, same filters, same date range
>   - Only difference: event name filter
> 
>   Two Different Optimizations
> 
>   | Optimization | PR #11194                  | PR #12664                     |
>   |--------------|----------------------------|-------------------------------|
>   | Type         | Horizontal (across labels) | Vertical (across event types) |
>   | Problem      | N+1 per label              | Multiple queries per table    |
>   | Before       | 5 × N queries              | 4 queries                     |
>   | After        | 5 queries                  | 1 query                       |
>   | Technique    | Batch with GROUP BY        | CASE expressions              |
> 
>   Both are needed! They're complementary optimizations.

The other PR #12261 changed things again?

scmmishra:
> The other PR #12261 changed things again?

Nopes, that was an unrelated fix. 

And yeah, Claude is pretty much on point there. We will have to benchmark it though, if a clubbed query will really be any faster. I'll work on it!

stonecharioteer:
> And yeah, Claude is pretty much on point there. We will have to benchmark it though, if a clubbed query will really be any faster. I'll work on it!

So what I've been doing is asking it to profile your end queries against some dummy data and to use that analyze output to give me insights.

pranavrajs (approved):

PR #12715, opened 2025-10-22
perf: add database index on conversations identifier

muhsin-k:
**Problem**
Slack webhook processing was failing with 500 errors due to database timeouts. The query `Conversation.where(identifier: params[:event][:thread_ts]).first` was performing full table scans and hitting PostgreSQL statement timeout.

**Solution**
Added database index on conversations.identifier and account_id.


sojan-official (approved):

sojan-official:
@muhsin-k  code lgtm, lets run the migration separately on prod before merging.

PR #12716, opened 2025-10-22
chore: Remove linear integration feature flag

muhsin-k:
This PR removes the linear integration feature flag since the integration is pretty much stable and we do display the Linear CTA for users who aren't connected.
Fixes https://linear.app/chatwoot/issue/CW-5819/remove-linear-feature-flag-from-front-end


iamsivin (approved):
LGTM

PR #12721, opened 2025-10-23
fix: activity message not created when handoff tool is used

scmmishra:


scmmishra:
The issue is something else, closing this for now

PR #12739, opened 2025-10-27
fix: callbacks not triggered after handoff

scmmishra:
## Problem

When the Captain agent uses the `HandoffTool` to hand off a conversation to a human agent, no activity message was created, and other status-change-dependent logic (auto-assignment, webhooks, notifications) failed to execute.

This issue only occurred when HandoffTool was used. LLM-initiated handoffs (where the LLM returns `conversation_handoff`) worked correctly.

## Investigation
Initially I suspected that `Current.executed_by` was being lost across threads, as HandoffTool executes in a different thread than `ResponseBuilderJob.` However, with multiple attempts to build a minimal reproduction of this, I found that this theory did not fit.

Eventually discovered the actual issue: **Rails clears dirty tracking information (`previous_changes`, `saved_change_to_status?`) after each save within a transaction**.

The flow was:
1. ResponseBuilderJob starts a transaction
2. AgentRunnerService executes tools inside that transaction
3. HandoffTool calls `conversation.bot_handoff!`
4. `bot_handoff!` calls `open!` which saves the record
5. New messages clears `previous_changes` immediately after save in the conversation model
7. Transaction commits
8. `after_commit` callbacks fire with **empty `previous_changes`**
9. Activity message creation is skipped because `saved_change_to_status?` returns `false`

This is a Rails gotcha. In-fact, there's a [gem](https://github.com/Shopify/ar_transaction_changes/) specifically designed to work through this.

> Note that using `previous_changes` in an `after_commit` hook will only return the attribute changes from the last time the record was saved, which won't include all the changes made to the record in the transaction if the record was saved multiple times in the same transaction.

## Solution

**Defer handoff execution until after the transaction commits.**

### Changes Made

1. **HandoffTool** (`enterprise/lib/captain/tools/handoff_tool.rb`)
   - Sets `Current.handoff_requested = true` flag
   - Does NOT call `bot_handoff!` directly
   - Allows ResponseBuilderJob to handle the actual handoff

2. **ResponseBuilderJob** (`enterprise/app/jobs/captain/conversation/response_builder_job.rb`)
   - Checks for both LLM-initiated handoffs and tool-initiated handoffs (via `Current.handoff_requested`)
   - Defers handoff until AFTER transaction commits
   - Calls `bot_handoff!` outside transaction context via `process_handoff`

3. **Current** (`lib/current.rb`)
   - Added `handoff_requested` flag to track tool-initiated handoffs

<details><summary>The real problem [fine print if you wanna dive deeper]</summary>
<p>

To some degree, we can call ` belongs_to :conversation, touch: true` inside the message model to be a culprit. But not sure how to get around that. 

Here's what happens
  1. `generate_response` runs the agent
  2. Agent creates messages (via tools or responses)
  3. Each `message.save!` triggers `conversation.save!` (due to touch: true)
  4. Conversation gets saved, `previous_changes` is cleared
  5. If we then call `bot_handoff!` in the same transaction:
    - It's a SECOND save 
    - `previous_changes` from the first save is already cleared
    - `after_commit` callbacks see empty `previous_changes`

With our deferred solution:
  1. Inside transaction: messages created → conversation touched/saved multiple times
  2. Transaction ends, previous_changes cleared
  3. Outside transaction: bot_handoff! is called
    - This is the FIRST save for this status change
    - previous_changes = {"status" => ["pending", "open"]}
    - after_commit callbacks work correctly!

There's a rails 7.1 flag that was toggled by default called [`run_commit_callbacks_on_first_saved_instances_in_transaction`](https://guides.rubyonrails.org/configuring.html#config-active-record-run-commit-callbacks-on-first-saved-instances-in-transaction) But I don't think this is directly related, perhaps adjacent or good to know
</p>
</details>

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -23,7 +23,7 @@ Metrics/MethodLength:
     - 'enterprise/lib/captain/agent.rb'
 
 RSpec/ExampleLength:
-  Max: 25
+  Max: 50
 
 Style/Documentation:
   Enabled: false
--- a/app/services/twilio/incoming_message_service.rb
+++ b/app/services/twilio/incoming_message_service.rb
@@ -44,6 +44,12 @@ class Twilio::IncomingMessageService
     twilio_channel.sms? ? params[:From] : params[:From].gsub('whatsapp:', '')
   end
 
+  def normalized_phone_number
+    return phone_number unless twilio_channel.whatsapp?
+
+    Whatsapp::PhoneNumberNormalizationService.new(inbox).normalize_and_find_contact_by_provider("whatsapp:#{phone_number}", :twilio)
+  end
+
   def formatted_phone_number
     TelephoneNumber.parse(phone_number).international_number
   end
@@ -53,8 +59,10 @@ class Twilio::IncomingMessageService
   end
 
   def set_contact
+    source_id = twilio_channel.whatsapp? ? normalized_phone_number : params[:From]
+
     contact_inbox = ::ContactInboxWithContactBuilder.new(
-      source_id: params[:From],
+      source_id: source_id,
       inbox: inbox,
       contact_attributes: contact_attributes
     ).perform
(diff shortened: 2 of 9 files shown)


scmmishra:
Moving this to draft to test with a hotfix: https://github.com/chatwoot/chatwoot/pull/12756

scmmishra:
Closing this PR, found a better fix here: https://github.com/chatwoot/chatwoot/pull/12756

PR #12745, opened 2025-10-28
Add migration notice for v4.6.0 to v4.7.0 upgrade

hienhoceo-dpsmedia:
**Problem:**
- Users upgrading without running migrations get generic login errors  
- account_saml_settings table missing causes authentication failures

**Solution:**
- Detects missing table during login attempts
- Shows helpful error with specific Docker commands:
  * Backup: docker exec chatwoot-db pg_dump -U postgres chatwoot_production > backup_before_migrate_$(date +%F).sql
  * Migrate: docker exec -it chatwoot-rails bundle exec rails db:migrate  
  * Restart: docker restart chatwoot-rails

**Benefits:**
- Prevents login failures during upgrades
- Reduces support tickets  
- Encourages safe database backups
- Provides clear Docker commands
- Supports international users


vishnu-narayanan:
@hienhoceo-dpsmedia Users are expected to run the database migration command during every version upgrade. Rails will execute migrations only if there are any pending ones, so it’s safe to run this command each time.

```
docker compose run --rm rails bundle exec rails db:chatwoot_prepare
```

Ref: https://developers.chatwoot.com/self-hosted/deployment/docker#upgrading

vishnu-narayanan:
Regarding translations, we are managing this via `crowdin`.  Please refer https://developers.chatwoot.com/contributing-guide/translation-guidelines#translate-chatwoot-to-your-language to add/contribute.

1. https://chatwoot.crowdin.com/

sojan-official:
`super_admin/instance_status` already includes a mechanism to indicate whether migrations are completed. There’s no need to expose this information to end users. 

Also running migrations is a mandatory step on all upgrades. we don't need to build logic around a specific migration.  Hence closing this for now. We will reopen later if required in future

PR #12746, opened 2025-10-28
fix: Misleading contact import success notifications

muhsin-k:
#### Summary
Eliminates misleading contact import email notifications that falsely reported "X contacts imported successfully" when most contacts actually failed validation or were rejected as duplicates. This critical fix ensures customers receive honest, transparent feedback about their CSV imports instead of confusing success messages that don't match reality.

#### The Problem
**Before**: Customer uploads 100 contacts → Gets email "100 contacts imported successfully" → Only finds 12 contacts in their account → Confusion and support tickets
**Root Cause**: System counted CSV rows processed rather than actual database insertions, leading to wildly inaccurate success reporting that damaged user trust and created support overhead.

#### The Solution
  **After**: Customer uploads 100 contacts → Gets email "12 contacts imported successfully, 15 contacts created but hidden (no email/phone), 73 contacts failed (validation/duplicates)" → Clear understanding of what happened and why


mazenkhalil:
@muhsin-k , as this is still Work in Progress (WIP), we've identified a significant memory issue in this area: there is no upper limit on the number of imported contacts. We have observed cases where importing a list of 30,000 contacts caused memory consumption to spike, ultimately leading to the worker instance crashing.

The two primary issues that need to be addressed are:
1. Risk of Service Disruption: The current design poses a direct threat to service stability.
2. Poor User Experience: After a service crash, users are not notified of the import's status. This prevents emails or other notifications from being triggered, leading users to repeatedly re-upload the file.

**Suggested Solution:**
* Limit the maximum number of records per import job (e.g., 10,000).
* Implement a queuing mechanism for all import jobs, regardless of the account. Simultaneous imports significantly increase memory usage, posing a risk to overall service stability.

muhsin-k:
Closing for now will reopen when we get back to this.

PR #12747, opened 2025-10-28
fix: timezone offset reports broken by DST transition

stonecharioteer:
## Description

Fixes timezone offset parameter in V2 reports API that was broken by DST transitions. The issue occurred when UK DST ended on October 26, 2025, causing the test to fail starting October 27th.

~~**Initial diagnosis:** The root cause was that `timezone_name_from_offset` used `zone.now.utc_offset` to match timezones, which changes based on the current date's DST status rather than the data being queried.~~

**Actual root cause:** The test was accidentally passing before DST transition. During BST, `timezone_name_from_offset(0)` matched "Azores" (UTC-1) instead of "Edinburgh" (UTC+0), and the -1 hour offset coincidentally split midnight data into [1,5]. After DST ended, it correctly matched "Edinburgh" (UTC+0), but this grouped all conversations into one day [6], exposing that the test data was flawed.

The real issue: Test data created all 6 conversations starting at midnight on a single day, which cannot produce a [1,5] split in true UTC.

Fixes CW-5846

## Type of change

## How Has This Been Tested?

**Test that was failing:**
```bash
bundle exec rspec spec/controllers/api/v2/accounts/reports_controller_spec.rb:25
```

**Changes:**
~~1. Fixed `timezone_name_from_offset` to use January 1st as reference date instead of current date~~
~~2. Converted timezone string to `ActiveSupport::TimeZone` object for `group_by_period` compatibility~~

**Revised approach:**
1. Freeze test time to January 2024 using `travel_to`, making timezone matching deterministic and aligned with test data period
2. Start test conversations at 23:00 instead of midnight to properly span two days and test timezone boundary grouping
3. Keep `zone.now.utc_offset` (correct behavior for real users during DST)

**Why this works:**
- Test runs "in January 2024" → `zone.now.utc_offset` returns January offsets consistently
- Offset `-8` correctly matches Pacific Standard Time (UTC-8 in January)
- Real users in PDT (summer) with offset `-7` → correctly match Pacific Daylight Time
- No production impact, test is deterministic year-round

**Verification:**
- Test now passes consistently regardless of current DST status
- Timezone matching works correctly for real users during DST periods
- Reports correctly group data by timezone offset across all seasons

## Checklist:

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -23,7 +23,7 @@ Metrics/MethodLength:
     - 'enterprise/lib/captain/agent.rb'
 
 RSpec/ExampleLength:
-  Max: 25
+  Max: 50
 
 Style/Documentation:
   Enabled: false
--- a/app/builders/v2/reports/timeseries/base_timeseries_builder.rb
+++ b/app/builders/v2/reports/timeseries/base_timeseries_builder.rb
@@ -41,6 +41,6 @@ class V2::Reports::Timeseries::BaseTimeseriesBuilder
   end
 
   def timezone
-    @timezone ||= timezone_name_from_offset(params[:timezone_offset])
+    @timezone = timezone_name_from_offset(params[:timezone_offset])
   end
 end
--- a/app/builders/v2/reports/timeseries/base_timeseries_builder.rb
+++ b/app/builders/v2/reports/timeseries/base_timeseries_builder.rb
@@ -41,6 +41,6 @@ class V2::Reports::Timeseries::BaseTimeseriesBuilder
   end
 
   def timezone
-    @timezone = timezone_name_from_offset(params[:timezone_offset])
+    @timezone ||= timezone_name_from_offset(params[:timezone_offset])
   end
 end
--- a/spec/controllers/api/v2/accounts/reports_controller_spec.rb
+++ b/spec/controllers/api/v2/accounts/reports_controller_spec.rb
@@ -10,7 +10,7 @@ RSpec.describe Api::V2::Accounts::ReportsController, type: :request do
     context 'when authenticated and authorized' do
       before do
         # Create conversations across 24 hours at different times
-        base_time = Time.utc(2024, 1, 15, 0, 0) # Start at midnight UTC
+        base_time = Time.utc(2024, 1, 14, 23, 0) # Start at 23:00 to span 2 days
 
         # Create conversations every 4 hours across 24 hours
         6.times do |i|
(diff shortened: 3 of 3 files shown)


app/helpers/timezone_helper.rb
```diff
@@ -10,8 +10,11 @@ def timezone_name_from_offset(offset)
     return 'UTC' if offset.blank?
 
     offset_in_seconds = offset.to_f * 3600
+    # Use Jan 1st as a reference date to avoid DST inconsistences
+    reference_date = Time.utc(Time.current.year, 1, 1, 12, 0, 0)
+
     matching_zone = ActiveSupport::TimeZone.all.find do |zone|
-      zone.now.utc_offset == offset_in_seconds
+      zone.at(reference_date).utc_offset == offset_in_seconds
```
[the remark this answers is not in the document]
scmmishra: @stonecharioteer this does raise an important point. The test failure doesn't necessarily mean the code is wrong - it reveals an ambiguity in the API design. We send the timezone offset from the frontend, the real question is "When a client sends timezone_offset: -8, what do they mean?"

The problem in the test is that it hardcodes `-8` expecting Pacific Time behavior. I think we can keep the original code and fix the test to either use `travel_to` and skip the DST issue within the test, or else somehow find a way to make the test timezone-agnostic

I think it's best to use `travel_to` to freeze time during winter when Pacific is actually at -8

stonecharioteer:
The January reference approach would have broken real users during DST. I've revised the fix.

**Root cause:**
Test was accidentally passing. During BST, `timezone_name_from_offset(0)` matched "Azores" (UTC-1), which split midnight data into [1,5]. After DST ended, it correctly matched "Edinburgh" (UTC+0), grouping all conversations into one day [6], exposing that the test data was flawed.

**Revised fix:**
- Use `travel_to Time.utc(2024, 1, 15, 12, 0)` to freeze test time to January 2024
- Start conversations at 23:00 instead of midnight to properly span two days
- Keep `zone.now.utc_offset` unchanged (correct for real users)

**Result:**
- Test runs "in January" so timezone matching is deterministic
- Offset `-8` matches Pacific Standard Time (not Alaska)
- Real users in PDT (summer) with offset `-7` correctly match Pacific Daylight Time
- No production impact

app/helpers/timezone_helper.rb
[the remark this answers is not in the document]
stonecharioteer: Yeah, was just pushing that change. Works now.

scmmishra:
@codex please review

app/builders/v2/reports/timeseries/base_timeseries_builder.rb
```diff
@@ -41,6 +41,6 @@ def group_by
   end
 
   def timezone
-    @timezone ||= timezone_name_from_offset(params[:timezone_offset])
+    @timezone = timezone_name_from_offset(params[:timezone_offset])
```
scmmishra:
```suggestion
    @timezone ||= timezone_name_from_offset(params[:timezone_offset])
```

Is the change intentional?

.rubocop.yml
```diff
@@ -23,7 +23,7 @@ Metrics/MethodLength:
     - 'enterprise/lib/captain/agent.rb'
 
 RSpec/ExampleLength:
-  Max: 25
+  Max: 50
```
scmmishra, line 26: Let's revert this. For the example in the PR specifically we can either extract the constants or params etc outside the example, or just add `rubocop:disable` for that rule on a case-to-case basis

stonecharioteer: I originally made this change on another PR, and it seemed relevant here too, asked Pranav on discord and he said there's no hard rule about this. We'd need to add rubocop disable rules to other tests as well.

app/builders/v2/reports/timeseries/base_timeseries_builder.rb
stonecharioteer: Ah no, I removed my changes from the previous commit, but I think I missed this. Should have reverted instead.

scmmishra (approved):

PR #12750, opened 2025-10-28
fix: handle users being stuck on is_creating billing flow

vishnu-narayanan:
# Pull Request Template

## Description

Fixes https://linear.app/chatwoot/issue/CW-5880/handle-customers-being-stuck-on-biling-page

## Type of change

## How Has This Been Tested?

## Checklist:


enterprise/app/jobs/enterprise/create_stripe_customer_job.rb
```diff
@@ -3,5 +3,19 @@ class Enterprise::CreateStripeCustomerJob < ApplicationJob
 
   def perform(account)
     Enterprise::Billing::CreateStripeCustomerService.new(account: account).perform
+  ensure
+    # Always clear the is_creating_customer flag, even if the job fails
+    # This prevents users from getting stuck on the billing page
+    clear_creating_flag(account)
```
sony-mathew, line 9: @vishnu-narayanan what will happen if the user has paid in stripe and we missed the web-hook event and we clear this flag? How will the customer experience look like? 

Or is this case prominent where customers did not pay and get's redirected back from Stripe?

sony-mathew (approved):
LGTM for mitigation. Added a comment on user experience.

vishnu-narayanan:
@copilot resolve the merge conflicts in this pull request

enterprise/app/jobs/enterprise/create_stripe_customer_job.rb
vishnu-narayanan: @sony-mathew good point - couple things here:

1.There's no user-paid window at this stage. `CreateStripeCustomerService#perform` calls `Stripe::Subscription.create` directly server-to-server and persists right after. User never hits Stripe Checkout before this runs - so "paid + we missed the webhook" isnt reachable in this flow.

2. Made the cleanup race-safe too - ensure now uses atomic postgres `jsonb - 'is_creating_customer'` via `update_all`, so only that one key gets dropped. Any concurrent webhook write to `custom_attributes` (e.g. setting `stripe_customer_id`) is preserved. No read-modify-write, no stale-object clobbering.

The actual prod case we hit today - `Stripe::Customer.create` or `Stripe::Subscription.create` raised, never reached `account.update!`, flag stayed set forever. That shoukd be fixed by the ensure block.

sony-mathew: Makes sense.

sony-mathew (approved):

PR #12753, opened 2025-10-28
fix: Exclude authentication templates from WhatsApp template selection

muhsin-k:
This PR add the changes for excluding the authentication templates from the WhatsApp template selection in the frontend, as these templates are not supported at the moment. Reference: https://www.chatwoot.com/hc/user-guide/articles/1754940076-whatsapp-templates#what-is-not-supported

Fixes https://linear.app/chatwoot/issue/CW-5850/authentication-templates-are-showing-up


iamsivin (approved):
Code LGTM.

PR #12769, opened 2025-10-30
fix: prevent LINE create double conversation

mix5003:
# Pull Request Template

## Description
Prevent double create conversation in LINE channel.

## Type of change

## How Has This Been Tested?
idk, how to reproduce this too. 
i think it upto line webhook. may be user send a large message first. then we received webhook and downloading image. after that line send webhook that just plain text. so we has 2 requests works for create new conversations. that why we has 2 conversations.
 
i just copy code from `app/jobs/webhooks/facebook_events_job.rb`, and test that it still can receive message

## Checklist:


mix5003:
@sojan-official  can you take a look. i found duplicate conversations around 2-3 conversations per month

muhsin-k:
@codex please review

mix5003:
sorry. i am not familiar. this is the best i (and claude) can do. 

if you think this code is not fit in code base. please feel free to reject this PRs. i will create this issue and let you cooks.

PS. i think i still don't known why this happend. may be it not relate to download files. but it up to LINE server that send webhook for same user too fast

mix5003:
if you want i has another solution that not touch a test. 

edit old app/services/line/incoming_message_service.rb
```ruby
  def parse_events
    prepare_conversation

    save_event_message
  end

  def prepare_conversation
    params[:events].each do |event|
      next unless event_type_message?(event)

      minimum_line_contact_info(event)
      next if @line_contact_info['userId'].blank?

      key = format(::Redis::Alfred::LINE_MESSAGE_MUTEX, sender_id: @line_contact_info['userId'], inbox_id: @inbox.id)
      with_lock(key) do
        set_contact
        set_conversation
      end
    end
  end

  def save_event_message
    params[:events].each do |event|
      next unless event_type_message?(event)

      get_line_contact_info(event)
      next if @line_contact_info['userId'].blank?

      set_contact
      set_conversation
      next unless message_created? event

      attach_files event['message']
      @message.save!
    end
  end

  def minimum_line_contact_info(event)
    # for set_contact and set_conversation without request API 
    @line_contact_info = {
      userId: event['source']['userId']
    }
  end
```

and make `Webhooks::LineEventsJob` to `MutexApplicationJob` and add `retry_on`  

in this solution
1. we can safely retry without double LINE conversation/messages from LockAcquisitionError (set_contact and set_conversation can run multiple times without any issue)
2. lock time should not more than 1 secound. because we not do network operation when locked

muhsin-k:
@mix5003 We fixed the race condition for the WhatsApp channel in https://github.com/chatwoot/chatwoot/pull/14060. Could you please review this PR and see if we can apply the same fix for Line as well?

mix5003:
> @mix5003 We fixed the race condition for the WhatsApp channel in #14060. Could you please review this PR and see if we can apply the same fix for Line as well?

@muhsin-k no, it can not do like that. based on [document](https://developers.line.biz/en/reference/messaging-api/#webhook-event-objects:~:text=There%20is%20not%20necessarily%20one%20user%20per%20webhook). it said
```
There is not necessarily one user per webhook. 
A message event from person A and a follow event from person B may be in the same webhook.
```

so we can not change `LineEventsJob` to `MutexApplicationJob`. because it may contain multiple `sender_id` . another way is remove `sender_id` from `redis_key`, but in cloud environment it may performance penalty.

PR #12776, opened 2025-10-31
feat: implement smart chat assignment

nosovk:
Updated chat assignment logic:
New chats are now assigned to the agent with the fewest active chats instead of round-robin.

- If several agents have the same number of chats, the first available one gets the chat.

- Only online agents are considered.

- When an agent closes a chat, their load is updated immediately.

- New online agents start with 0 chats and receive new ones until loads are balanced.

commit from https://github.com/rondelinga/chatwoot/pull/1


nosovk:
@rondelinga tests are failing, could you check pls logs

nosovk:
surpassed by https://github.com/chatwoot/chatwoot/pull/12793

PR #12786, opened 2025-11-03
fix: Optimize Message search_data to prevent OpenSearch field explosion

stonecharioteer:
## Description

Refactored the `Message#search_data` method to prevent exceeding OpenSearch's 1000 field limit during reindex operations.

**Problem:** The previous implementation serialized entire ActiveRecord objects (Inbox, Sender, Conversation) with all their JSONB fields, causing dynamic field explosion in OpenSearch. This resulted in `Searchkick::ImportError` with "Limit of total fields [1000] has been exceeded".

**Solution:** Whitelisted only necessary fields for search and filtering, and flattened JSONB `custom_attributes` into key-value pair arrays to prevent unbounded field creation.

Linked to: CW-5861

## Type of change

## How Has This Been Tested?

- Verified rubocop passes with no offenses
- Code review of search field usage from `enterprise/app/services/enterprise/search_service.rb`
- Analyzed actual search queries to determine required indexed fields

**Still needed:**
- Full reindex test on staging/production environment
- Verify search functionality still works after reindex
- Confirm field count is under 1000 limit

### Before
- Indexed 1000+ fields (entire AR objects with JSONB)
- `inbox` = full Inbox object (23+ fields + JSONB)
- `sender` = full Contact/User/AgentBot object (10+ fields + JSONB)
- `conversation` = full push_event_data
- Dynamic JSONB keys creating unlimited fields

### After
- ~35-40 controlled fields
- Whitelisted search fields: `content`, `attachment_transcribed_text`, `email_subject`
- Filter fields: `account_id`, `inbox_id`, `conversation_id`, `sender_id`, `sender_type`, etc.
- Flattened `custom_attributes`: `[{key, value, value_type}]` format
- Helper methods: `search_conversation_data`, `search_inbox_data`, `search_sender_data`, `search_additional_data`

## Checklist:

## Post-merge Steps

After merging, the following steps are required:

1. **Reindex all messages:**
   ```bash
   bundle exec rails runner "Message.reindex"
   ```

2. **Verify field count:**
   ```bash
   bundle exec rails runner "
     client = Searchkick.client
     index_name = Message.searchkick_index.name
     mapping = client.indices.get_mapping(index: index_name)
     fields = mapping.dig(index_name, 'mappings', 'properties')
     puts 'Total fields: ' + fields.keys.count.to_s
   "
   ```

3. **Test search functionality** to ensure queries still work as expected


stonecharioteer:
Added a nil check for `conversation.contact_inbox` in the `conversation_push_event_data` method to address the pre-existing Sentry issue flagged by the bot.

This prevents the `NoMethodError: undefined method 'source_id' for nil` when `conversation.contact_inbox` is nil.

stonecharioteer:
Addressed Codex feedback about field naming compatibility:

- Maintained nested field structure (`attachments.transcribed_text` and `content_attributes.email.subject`) for backward compatibility with Enterprise search service
- Extracted helper methods (`search_attachments_data` and `search_content_attributes_data`) to keep code clean and pass rubocop checks
- Removed redundant flat `email_subject` field
- Used `filter_map` for better performance

The Enterprise search service will continue to work without any modifications.

app/models/message.rb
```diff
@@ -261,6 +276,80 @@ def should_index?
...
+      contact_id: conversation.contact_id,
+      custom_attributes: flatten_custom_attributes(conversation.custom_attributes),
+      browser: conversation.additional_attributes&.dig('browser'),
+      referer: conversation.additional_attributes&.dig('referer'),
+      initated_at: conversation.additional_attributes&.dig('initated_at')
+    }
+  end
+
+  def search_inbox_data
+    return nil unless index
```
vishnu-narayanan:
```suggestion
    return nil unless inbox
```

vishnu-narayanan: inbox?

app/models/message.rb
```diff
@@ -261,6 +276,80 @@ def should_index?
...
+      id: conversation.id,
+      display_id: conversation.display_id,
+      status: conversation.status,
+      assignee_id: conversation.assignee_id,
+      team_id: conversation.team_id,
+      contact_id: conversation.contact_id,
+      custom_attributes: flatten_custom_attributes(conversation.custom_attributes),
+      browser: conversation.additional_attributes&.dig('browser'),
+      referer: conversation.additional_attributes&.dig('referer'),
+      initated_at: conversation.additional_attributes&.dig('initated_at')
```
vishnu-narayanan:
```suggestion
      initated_at: conversation.additional_attributes&.dig('initiated_at')
```

vishnu-narayanan (commented):
minor comments

app/models/message.rb
```diff
@@ -261,6 +276,80 @@ def should_index?
...
+        value: value.to_s,
+
+        value_type: value.class.name.downcase
+      }
+    end
+  end
+
+  def search_attachments_data
+    attachments.filter_map do |a|
+      { transcribed_text: a.meta&.dig('transcribed_text') }
```
vishnu-narayanan: `transcribed_text` would be `nil` for most of the messages. Should we create the field only if `transcribed_text.present?`

stonecharioteer: Hmm, would be `nil` in this case so that's fine? We would be checking if that field exists anyway, right? This shouldn't add to the keys either.

app/models/message.rb
```diff
@@ -153,21 +154,12 @@ def push_event_data
...
-    data
-  end
-
   def conversation_push_event_data
     {
       assignee_id: conversation.assignee_id,
       unread_count: conversation.unread_incoming_messages.count,
       last_activity_at: conversation.last_activity_at.to_i,
-      contact_inbox: { source_id: conversation.contact_inbox.source_id }
+      contact_inbox: conversation.contact_inbox ? { source_id: conversation.contact_inbox.source_id } : nil
```
pranavrajs: Why was this change added? At the moment, there shouldn't be a case where contact_inbox is nil.

app/models/concerns/message_search_data.rb
```diff
@@ -0,0 +1,103 @@
...
+      contact_id: conversation.contact_id,
+      custom_attributes: flatten_custom_attributes(conversation.custom_attributes),
+      browser: conversation.additional_attributes&.dig('browser'),
+      referer: conversation.additional_attributes&.dig('referer'),
+      initated_at: conversation.additional_attributes&.dig('initiated_at')
+    }
+  end
+
+  def search_inbox_data
+    return nil unless inbox
```
pranavrajs: This is also not required. At the moment, there should not be a case where there is a conversation without the inbox data. Every conversation and message should have an inbox.

app/models/concerns/message_search_data.rb
```diff
@@ -0,0 +1,103 @@
+module MessageSearchData
```
pranavrajs: Can you check the Conversations::EventDataPresenter and try implementing a similar pattern (eg: Message::SearchDataPresenter) and add a spec so that it is easier to test it?

pranavrajs (changes requested):

app/models/message.rb
stonecharioteer: I implemented that because of this comment from sentry. 

https://github.com/chatwoot/chatwoot/pull/12786#issuecomment-3479327361

pranavrajs: Makes sense. The issue exists due to some other reason. It shouldn't be fixed here

stonecharioteer: Okay, I'm removing this and the other bit right now.

app/models/concerns/message_search_data.rb
stonecharioteer: Ah, I need to learn where things plug in.

I've made this change.

pranavrajs (approved):

app/presenters/messages/search_data_presenter.rb
```diff
@@ -0,0 +1,99 @@
...
+      email: {
+        subject: content_attributes.dig(:email, :subject) || content_attributes.dig('email', 'subject')
+      }.compact
+    }.compact
+  end
+
+  def search_conversation_data
+    {
+      id: conversation.id,
+      display_id: conversation.display_id,
```
pranavrajs:
```suggestion
      id: conversation.display_id,
```

PR #12800, opened 2025-11-05
feat: add allow_bots column to campaigns table

scmmishra:
This PR adds an `allow_bots` boolean field to campaigns that controls the initial conversation state when a campaign message is sent. By default (`allow_bots: false`), campaign conversations start in the `:open` state for immediate agent visibility, maintaining the current behavior. This ensures backward compatibility for all existing campaigns.

When `allow_bots` is enabled and the inbox has an active bot or Captain enable, campaign conversations will start in the `:pending` state, allowing bots to handle the conversation first before routing to agents. This gives teams the flexibility to choose whether bots should engage with campaign-initiated conversations.

CleanShot 2025-11-05 at 17 29 03@2x


sojan-official:
@scmmishra  we can go for convention over an additional configuration here. If the campaign sender is bot, lets mark it into pending if the inbox has an active_bot ,  if the sender is a human agent in that case lets continue the current behavior where the conversation remains open.

scmmishra:
Closing this, we found a simpler approach here: https://github.com/chatwoot/chatwoot/pull/12805

Issue #10609, opened 2024-12-21
Spinning Circle at login


SwimGeek:
### Describe the bug

Hi, on one of my browsers, Safari, macOS, when I visit the Chatwoot URL (self hosted), I keep seeing a spinning wheel and the page does not load.

I've tried to delete all the website data / cookies for the site, restarted browser etc.

I'm wondering if it could be a problem where the browser thinks it has a login session open.... but something fails in loading the page.

The javascript console (dev tools) does not seem to work - something is breaking pretty badly.

Is there a logout URL I can try visit to cleanly exit and log back in?

Things work fine when I use another browser on the same laptop, on when I start a 'private browser' in Safari.

### To Reproduce

Visit chatwoot URL.

### Expected behavior

Expecting the chatwoot app to load.

### Environment

Linux VM

### Cloud Provider

Other [please specify in the description]

### Platform

Browser

### Operating system

Chatwoot running on Debian, latest: 3.16

### Browser and version

Safari, macOS, latest

pranavrajs:
@SwimGeek Are there any errors on the browser console which you can share? Doing a [reset on safari](https://support.apple.com/en-gb/guide/safari/sfri11471/mac) should work.

SwimGeek:
Hi, I figured out that if you logged in as a 'super admin' it would get stuck.

So I migrated the super admin rights to my regular user.

You probably need a check - if user is logging in and is only a super admin, without regular account, don't let them log into the regular user area. Something like that.

dhaneshmarch:
Facing the same. 

> Hi, I figured out that if you logged in as a 'super admin' it would get stuck.
> 
> So I migrated the super admin rights to my regular user.
> 
> You probably need a check - if user is logging in and is only a super admin, without regular account, don't let them log into the regular user area. Something like that.

Although this didn'nt work for me at all @SwimGeek 

I see a pattern where /validate_token keeps on trying to complete the request.

pranavrajs:
@stonecharioteer I was working on onboarding, figured it's like a simple fix. I'll raise the PR. If you can review it, that'll be great.

pranavrajs:
Closed via https://github.com/chatwoot/chatwoot/pull/12816

PR #12816, opened 2025-11-07
fix: Handle login when there are no accounts

pranavrajs:
When a SuperAdmin logs into Chatwoot without being part of any accounts, the app currently gets stuck on a loading spinner and hangs the browser instead of displaying an error. This PR fixes that issue #10609. 

While working on the onboarding changes, I noticed this was a simple fix and decided to address it quickly.

Here’s what I changed:
- Added a redirect to a dedicated “No Accounts” page whenever a user without any accounts is routed on the frontend.
- Removed an unnecessary account model that was being loaded on page load — it’s already handled separately in the sidebar, so this was redundant.

Screenshot 2025-11-06 at 5 36 45 PM


app/javascript/dashboard/routes/index.js
```diff
@@ -18,8 +18,17 @@ export const validateAuthenticateRoutePermission = (to, next) => {
     return '';
   }
 
-  if (!to.name) {
-    return next(frontendURL(`accounts/${user.account_id}/dashboard`));
+  const { accounts = [], account_id: accountId } = user;
+
+  if (accounts.length === 0) {
```
pranavrajs:
```suggestion
  if (!accounts.length) {
```

iamsivin (approved):
LGTM
https://www.loom.com/share/c809dda4f666464aa25c7087d922211d

stonecharioteer (approved):

PR #12832, opened 2025-11-10
feat: allow querying reporting events via the API

scmmishra:
This PR API for reporting events. For many power-users where the default reports might not be enough, having access to granular access to the reporting events will be helpful. These endpoints enable retrieval of event data at both the account and conversation levels

### Reporting Events

Retrieve all reporting events for an account with pagination and filtering. This endpoint is admin-only and returns events ordered by most recent first. Supports filtering by date range, inbox, user, and event name.

**Sample curl:**
```bash
curl -X GET "/api/v1/accounts/1/reporting_events?page=1&inbox_id=5&user_id=10" \
  -H "Authorization: Bearer token" \
  -H "Content-Type: application/json"
```

### Conversation Reporting Events

Retrieve reporting events for a specific conversation. Access is determined by conversation permissions (assignee, team membership). Useful for auditing and analyzing conversation-specific activity.

**Sample curl:**
```bash
curl -X GET "/api/v1/accounts/1/conversations/42/reporting_events" \
  -H "Authorization: Bearer token" \
  -H "Content-Type: application/json"
```


enterprise/app/controllers/api/v1/accounts/reporting_events_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  private
+
+  def set_reporting_events
+    @reporting_events = Current.account.reporting_events
+                               .includes(:conversation, :user, :inbox)
+                               .filter_by_date_range(range)
+                               .filter_by_inbox_id(params[:inbox_id])
+                               .filter_by_user_id(params[:user_id])
+                               .filter_by_name(params[:name])
+                               .order(created_at: :desc)
```
stonecharioteer, line 24: Is sorting a necessity for this endpoint? Other endpoints use `Sift` right? Or does it make sense that we're *always* going to be sorting by `created_at: :desc`?

stonecharioteer:
Also, expand the description please. It'll help give context.

enterprise/app/controllers/api/v1/accounts/reporting_events_controller.rb
scmmishra: Default sorting works, reporting/analytics endpoints typically show newest first - this is the expected default behavior. 
Besides we can add it later. Based on what the requirement was, the default sorting is fine I would say

stonecharioteer: Great. Works for me.

stonecharioteer (approved):
LGTM

scmmishra:
@codex please review

PR #12871, opened 2025-11-14
chore: disable worker MemoryHigh throttling in systemd unit

sojan-official:
- set MemoryHigh to infinity in deployment/chatwoot-worker.1.service so the worker is throttled only by the existing
    MemoryMax hard limit
  - prevents cgroup reclaim from slowing Sidekiq under transient spikes while still keeping the hard stop at 1.5 GB


vishnu-narayanan (approved):
good one🔥

PR #12873, opened 2025-11-14
feat: add retry loadWithRetry composable

### LINKED ISSUE #12644: Error for WhatsApp images: This image is no longer available
### Describe the bug

Hi,

I am getting this error after upgrading to 4.6

Image

.0

"This image is no longer available."

### To Reproduce

1. Chatwoot self-hosted
2.  I use railway

### Expected behavior

_No response_

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

None

### Operating system

_No response_

### Browser and version

_No response_

### Docker (if applicable)

_No response_

### Additional context

_No response_

scmmishra:
Attempt to fix: https://github.com/chatwoot/chatwoot/issues/12644

This PR introduces a `useLoadWithRetry` composable that pre-loads images in the background with automatic retry logic. When an image fails to load, it retries up to 3 times with increasing backoff (1s, 2s, 3s). Each retry attempt includes a unique timestamp query parameter to bypass any intermediate caches (proxies, CDNs, service workers). The Image.vue component only renders the `` tag once the background load succeeds, providing a seamless user experience.

The retry mechanism handles transient failures gracefully—whether from Active Storage processing delays, network issues, or temporary CDN problems. After exhausting all retry attempts, the component displays a proper error state to avoid indefinite loading.

### How this works

- The retry mechanism uses a background Image() object to pre-load the image before rendering it in the DOM
- When `loadWithRetry()` is called, it immediately attempts to load the image
- If the load fails (404 or network error), it waits for an exponentially increasing delay (1s, then 2s, then 3s) before retrying
- Each retry is a fresh attempt with the same URL, relying on the  incoming `cache-control: no-cache` header to ensure the browser doesn't serve stale 404s
- If any attempt succeeds, `isLoaded` becomes `true` and the component renders the actual `` tag, which fetches from the browser's cache (since the background Image() already loaded it)
- If all retries exhaust without success, `hasError` is set and the component displays an error message

```
Load Image
      │
      ├─ ✓ Success → Show Image
      │
      ├─ ✗ Fail → Wait 1s → Retry
      │              │
      │              ├─ ✓ Success → Show Image
      │              │
      │              ├─ ✗ Fail → Wait 2s → Retry
      │                         │
      │                         ├─ ✓ Success → Show Image
      │                         │
      │                         └─ ✗ Fail → Show Error 
```

### Why timestampped URLs alone will not work

Active Storage may return 404 Not Found while images are still processing, and because the response includes cache-`control: no-cache`, the browser never caches the 404. Cache breakers (timestamped URLs) only change the URL when the component re-renders, so they don’t automatically retry. If the component stays mounted, the failed image will remain broken even after processing finishes.

The correct fix inevitably is to add a retry mechanism with a short exponential backoff. When the image fails, the client should retry the request a few times with increasing delays and a fresh timestamp each attempt. This lets the UI automatically recover as soon as the image becomes available.


scmmishra:
@codex please review

scmmishra:
@codex please review

stonecharioteer (approved):
LGTM

gkp1:
sorry for ping but I kindly request to consider releasing this fix ASAP as images/audio are still requiring a page reload to appear in the latest release (4.8.0) for anyone using s3 or s3_compatible (EDIT: nevermind, just tested the docker develop image, haha)

+ thank you very much for the fix @scmmishra !

PR #12882, opened 2025-11-14
fix: Change contact_inboxes.source_id to text column

stonecharioteer:
## Description

Fixes CW-5961 where IMAP email processing failed with `ActiveRecord::RecordInvalid: Validation failed: Source is too long (maximum is 255 characters)` error.

This changes the `contact_inboxes.source_id` column from `string` (255 character limit) to `text` (unlimited) to accommodate long email message IDs that were causing validation failures.

Fixes CW-5961

## Type of change

## How Has This Been Tested?

- Added spec test validating `source_id` values longer than 255 characters (300 chars)
- All existing `contact_inbox_spec.rb` tests pass (7 examples, 0 failures)
- Migration applied successfully with reversible up/down methods
- Verified `source_id` column type changed to `text` with `null: false` constraint preserved

## Checklist:

### DIFF
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -40,6 +40,7 @@
 #  index_conversations_on_contact_inbox_id            (contact_inbox_id)
 #  index_conversations_on_first_reply_created_at      (first_reply_created_at)
 #  index_conversations_on_id_and_account_id           (account_id,id)
+#  index_conversations_on_identifier_and_account_id   (identifier,account_id)
 #  index_conversations_on_inbox_id                    (inbox_id)
 #  index_conversations_on_priority                    (priority)
 #  index_conversations_on_status_and_account_id       (status,account_id)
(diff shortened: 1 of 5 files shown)


app/models/conversation.rb
```diff
@@ -40,6 +40,7 @@
 #  index_conversations_on_contact_inbox_id            (contact_inbox_id)
 #  index_conversations_on_first_reply_created_at      (first_reply_created_at)
 #  index_conversations_on_id_and_account_id           (account_id,id)
+#  index_conversations_on_identifier_and_account_id   (identifier,account_id)
```
stonecharioteer, line 43: This change comes from the annotate gem. I'm not sure why it wasn't in the develop branch already. It's unrelated to this though.

pranavrajs (approved):

db/migrate/20251114173609_change_source_id_to_text.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeSourceIdToText < ActiveRecord::Migration[7.1]
+  def up
+    change_column :contact_inboxes, :source_id, :text, null: false
```
sojan-official, line 3: @stonecharioteer  can we run this migration manually on prod  and mark it before deploy to make sure there are no surprises. even though this should be okay.

stonecharioteer: Okay, will do that.

PR #12884, opened 2025-11-16
feat(webhooks): allow selecting inbox

gabrieljablonski:
# Pull Request Template

## Description

The webhook DB model already had an inbox relation from a very old feature that seems to have gotten lost along the way.

This PR makes use of the existing column so we can subscribe to a specific inbox so we don't need to filter events manually on the webhook receiver.

## Type of change

## How Has This Been Tested?

Webhook listener and controller specs, and also tested in production for several months without issue.

| Before | After |
| --- | --- |
|image-2|image-1|
|image-3|image|

## Checklist

### DIFF
--- a/app/controllers/api/v1/accounts/webhooks_controller.rb
+++ b/app/controllers/api/v1/accounts/webhooks_controller.rb
@@ -7,12 +7,12 @@ class Api::V1::Accounts::WebhooksController < Api::V1::Accounts::BaseController
   end
 
   def create
-    @webhook = Current.account.webhooks.new(webhook_params)
+    @webhook = Current.account.webhooks.new(webhook_create_params)
     @webhook.save!
   end
 
   def update
-    @webhook.update!(webhook_params)
+    @webhook.update!(webhook_update_params)
   end
 
   def destroy
@@ -22,8 +22,13 @@ class Api::V1::Accounts::WebhooksController < Api::V1::Accounts::BaseController
 
   private
 
-  def webhook_params
+  def webhook_create_params
     params.require(:webhook).permit(:inbox_id, :name, :url, subscriptions: [])
+  end
+
+  def webhook_update_params
+    # NOTE: Do not allow updating linked inbox to avoid events being sent for unexpected inboxes.
+    params.require(:webhook).permit(:name, :url, subscriptions: [])
   end
 
   def fetch_webhook
--- a/app/javascript/dashboard/routes/dashboard/settings/integrations/Webhooks/WebhookForm.vue
+++ b/app/javascript/dashboard/routes/dashboard/settings/integrations/Webhooks/WebhookForm.vue
@@ -74,6 +74,7 @@ export default {
   computed: {
     inboxesList() {
       if (this.assignedInbox?.id) {
+        // NOTE: Only show "All Inboxes" option when an inbox is already assigned
         return [
           {
             id: 0,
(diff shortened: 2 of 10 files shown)


app/controllers/api/v1/accounts/webhooks_controller.rb
```diff
@@ -22,10 +22,15 @@ def destroy
...
 
-  def webhook_params
+  def webhook_create_params
     params.require(:webhook).permit(:inbox_id, :name, :url, subscriptions: [])
   end
 
+  def webhook_update_params
+    # NOTE: Do not allow updating linked inbox to avoid events being sent for unexpected inboxes.
+    params.require(:webhook).permit(:name, :url, subscriptions: [])
+  end
```
gabrieljablonski, line 32: In my own fork, we also block updating the webhook URL.

The idea is that changing the URL/inbox is a critical action, which shouldn't be done lightly, so deleting and/or creating a new webhook is the way to go.

But as that might be disruptive to existing users, I decided on only blocking updating the new inbox field, but we might want to remove that as well.

app/javascript/dashboard/routes/dashboard/settings/integrations/Webhooks/WebhookForm.vue
```diff
@@ -111,6 +140,29 @@ export default {
...
+          :multiselector-placeholder="
+            $t('INTEGRATION_SETTINGS.WEBHOOK.FORM.INBOX.PLACEHOLDER')
+          "
+          :no-search-result="
+            $t('INTEGRATION_SETTINGS.WEBHOOK.FORM.INBOX.NO_RESULTS')
+          "
+          :input-placeholder="
+            $t('INTEGRATION_SETTINGS.WEBHOOK.FORM.INBOX.INPUT_PLACEHOLDER')
+          "
+          :disabled="isEditing"
```
gabrieljablonski, line 163: As mentioned in previous comment, we might want to remove the "disabled on edit" feature.

app/views/api/v1/accounts/webhooks/_webhook.json.jbuilder
```diff
@@ -7,5 +7,6 @@ if webhook.inbox
   json.inbox do
     json.id webhook.inbox.id
     json.name webhook.inbox.name
+    json.channel_type webhook.inbox.channel_type
```
gabrieljablonski, line 10: Needed for `<InboxName />` component to display the correct icon.

app/listeners/webhook_listener.rb
```diff
@@ -101,6 +101,7 @@ def handle_typing_status(event_name, event)
     payload = {
       event: event_name,
       user: user.webhook_data,
+      inbox_id: inbox.id,
       conversation: conversation.webhook_data,
```
gabrieljablonski, line 105: Inbox id is already available inside `conversation` object, but placing it in the root of the payload makes it cleaner to extract it, and also more scalable when adding future events.

gabrieljablonski, line 105: It should also be noted the `inbox_updated` event (which does not seem to be in use right now) does not include the inbox id in the payload, so this new logic will not apply if `inbox_update` ever comes into play.

Should we mess with the `inbox_update` payload to ensure future compatibility, or maybe just add a comment somewhere?

scmmishra:
Hey @gabrieljablonski you could take a look at Agent Bots for this? That might suffice your use-case, it limits the webhook events to the inbox

gabrieljablonski:
> Hey @gabrieljablonski you could take a look at Agent Bots for this? That might suffice your use-case, it limits the webhook events to the inbox

Just wrote and deleted a comment about how agent bots did not support conversation events, but apparently this isn't the case, must've misread the documentation somewhere 😬

Still, this feature is useful in my own use-case due to convenience of programmatically creating webhooks vs agent bots.

Might be a niche use-case, so I understand if you guys decide not to move forward with this implementation.

PR #12896, opened 2025-11-18
feat: new buttons in the client chat interface

rondelinga:


pranavrajs:
@rondelinga Thanks for the contribution. At the moment, we are unable to accept this PR in its current state.

Could you please raise an issue first with a clear description of what you are trying to achieve, the problem it is solving, and the intended behavior? This will help us understand the context and evaluate whether this change aligns with the project direction and broader community needs.

If you believe this change would be valuable for the community, feel free to reopen the PR after creating the issue and updating the description with the necessary details. That will make it much easier for us to review and provide meaningful feedback.

PR #12905, opened 2025-11-19
feat: Add support for Langfuse LLM Tracing via OTEL

aakashb95:
This PR adds LLM instrumentation on langfuse for ai-editor feature

## Type of change
New feature (non-breaking change which adds functionality)

Needs langfuse account and env vars to be set

## How Has This Been Tested?

I configured personal langfuse credentials and instrumented the app, traces can be seen in langfuse. 
each conversation is one session. 
image
image

## Checklist:

- [x ] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my code
- [ x] I have commented on my code, particularly in hard-to-understand areas
- [ x] My changes generate no new warnings


bin/vite
```diff
@@ -1,27 +1,29 @@
 #!/usr/bin/env ruby
-# frozen_string_literal: true
-
 #
-# This file was generated by Bundler.
+# This file was generated by RubyGems.
```
aakashb95: not sure how this got altered

config/initializers/opentelemetry.rb
```diff
@@ -0,0 +1,34 @@
...
+    exporter_config = {
+      endpoint: ENV.fetch('OTEL_EXPORTER_OTLP_ENDPOINT', 'https://cloud.langfuse.com/api/public/otel/v1/traces'),
+      headers: {
+        'Authorization' => "Basic #{Base64.strict_encode64("#{ENV.fetch('LANGFUSE_PUBLIC_KEY', nil)}:#{ENV.fetch('LANGFUSE_SECRET_KEY', nil)}")}"
+      }
+    }
+
+    # Only disable SSL verification in non-production environments
+    # Production should use proper SSL verification for security
+    exporter_config[:ssl_verify_mode] = OpenSSL::SSL::VERIFY_NONE unless Rails.env.production?
```
aakashb95: should enable in staging too i guess for proper testing?

bin/spring
```diff
@@ -1,16 +1,29 @@
 #!/usr/bin/env ruby
+#
```
pranavrajs: This is an unrelated change. Can we revert this?

aakashb95: yes, i also saw `bundler/gems/azure-storage-ruby-9957cf899d33` and  `bundler/gems/devise-secure_password-adcc85fe1bab`

should those be removed too?

aakashb95:
@codex review

app/javascript/dashboard/api/integrations/openapi.js
aakashb95, line 1: @iamsivin is this ok?

pranavrajs (commented):
A few comments from my side:
1. We don’t have any specs for the LLM instrumentation right now. Would be good to add at least a basic set of specs for the module.

2. We should keep the upcoming Ruby-LLM migration in mind. Once we switch, we’ll have multiple providers and the prompt/usage/token formats will differ. If possible, let’s start separating those concerns now. If it’s too much for this PR, we can split it out, but let’s not hard-code things that will become painful later.

3. OTEL + env vars can move to installation config.
Similar to how AI Agents works, you shouldn’t need to change env variables every time. Just read from the DB and update from there. That would simplify things a lot.

4. I’d prefer something like OTEL_PROVIDER = langfuse instead of relying on OTEL_ENABLED = true. I know that’s not the main goal here, but structuring it so that it’s easy to swap providers would help.


5. Usage event names (gen_ai.usage.*).These should ideally be constants. Either a separate constants file or at least defined at the top of the instrumentation file so we have one place to see what’s being tracked.

6. Comments only where needed. There are a lot of comments which is not really useful. The code should explain what it does by itself. If there is a complex logic that is visible, a decision that is not part of the code, that can be documented. Otherwise, all of that would get stale pretty quickly.

aakashb95:
Thanks for the feedback.

>We don’t have any specs for the LLM instrumentation right now. Would be good to add at least a basic set of specs for the module.

added with claude's help

>We should keep the upcoming Ruby-LLM migration in mind. Once we switch, we’ll have multiple providers and the prompt/usage/token formats will differ. If possible, let’s start separating those concerns now. If it’s too much for this PR, we can split it out, but let’s not hard-code things that will become painful later.

This is what i am trying right now, just to be clear, we are only moving the editor ai features to Ruby-LLM, rest will continue with ai-agents right? Ruby-LLM doesn't have observability baked in. Figuring this out.

>OTEL + env vars can move to installation config.
Similar to how AI Agents works, you shouldn’t need to change env variables every time. Just read from the DB and update from there. That would simplify things a lot.

makes sense, this is done

>I’d prefer something like OTEL_PROVIDER = langfuse instead of relying on OTEL_ENABLED = true. I know that’s not the main goal here, but structuring it so that it’s easy to swap providers would help.

Aligned, added.

>Usage event names (gen_ai.usage.*).These should ideally be constants. Either a separate constants file or at least defined at the top of the instrumentation file so we have one place to see what’s being tracked.

Done. Will have to monitor the otel attributes associated with genai since some are under development. But langfuse seems 
to be supporting older and newer variants.

>Comments only where needed. There are a lot of comments which is not really useful. The code should explain what it does by itself. If there is a complex logic that is visible, a decision that is not part of the code, that can be documented. Otherwise, all of that would get stale pretty quickly.

My bad, most of this PR is indeed claude code, and me trying to understand and implement. Will ensure fewer comments

spec/lib/integrations/openai/processor_service_spec.rb
```diff
@@ -76,7 +78,9 @@
           .to_return(status: 200, body: openai_response, headers: {})
 
         result = subject.perform
-        expect(result).to eq({ :message => 'This is a reply from openai.' })
+        expect(result[:message]).to eq('This is a reply from openai.')
+        expect(result[:request_messages]).to be_an(Array)
```
aakashb95: @pranavrajs these exist because earlier we were just sending the llm output, now we are sending the messages array [{'role': 'user', 'content': ...}] and the usage so langfuse can tell us cost of that call.

since these are stub_requests, we expect usage to be nil

pranavrajs (approved):

Issue #1887, opened 2021-03-09
emoji support for slack integration


psifertex:
**Is your enhancement request related to a problem? Please describe.**
Currently when sending emoji via slack to a web-user, the emoji will come through with the slack text name, such as `:frowning:`

This is true even if the slack user pastes a native emoji and doesn't use the slack picker. 

Interestingly, emoji coming the other way work just fine. 

**Describe the solution you'd like**
I'd like the emoji to be shown as the visual emoji.

**Describe alternatives you've considered**
Only alternate I can think of is implementing this feature myself. Might take a look at the source and see. :-) 

**Additional context**
<img width="413" alt="Screen Shot 2021-03-09 at 5 17 56 PM" src="https://user-images.githubusercontent.com/140215/110545640-65eb5400-80fb-11eb-8442-b71b039c3758.png">

pranavrajs:
@psifertex Are you aware of any standard by which we can transform it to emoji? It is not defined here https://unicode.org/emoji/charts/full-emoji-list.html

psifertex:
Yup, they publish it here: https://github.com/iamcal/emoji-data

Here's a write up on someone else doing the same thing: https://aaronparecki.com/2017/02/05/8/day-47-slack-emoji

psifertex:
I've found a bunch of other libraries that make this trivial, for example:

https://github.com/jrobertson/emoji2020

I don't know ruby but it /looks/ like adding that library and replacing: https://github.com/chatwoot/chatwoot/blob/develop/lib/integrations/slack/incoming_message_builder.rb#L86

With:

```
      content: Emoji2020.new(params[:event][:text]).to_s,
```

should basically solve this with a one line change!

psifertex:
Now that I think about it some more, it might actually be better to render as an html image which would require something slightly more complicated such as using the first link I pasted above and having the images available in a directory. Still relatively straight forward and would be much less susceptible to font issues on various websites.

sojan-official:
@psifertex  Thanks for providing all that info.  it's super helpful. We might not be able to get to this for the 1.14. release. But we will try to take this up for our subsequent one.

psifertex:
No worries, thanks for the update. Let me know if I can help with anything else. If I get some more time I might take a look at trying some of the libraries out and submitting a PR but it would probably take more time to figure out the deployment/setup process than actually writing the code. 🤣

sojan-official:
[...]

PR #12928, opened 2025-11-21
fix: render slack emoji shortcodes as unicode characters

Lomuzord:
This PR fixes an issue where Slack emojis are rendered as text shortcodes (e.g. :rocket:) instead of the actual emoji characters in Chatwoot messages.

It introduces a new EmojiFormatter class that uses the emoji-data mapping to convert shortcodes to unicode characters.


Lomuzord:
Fixes #1887

lib/integrations/slack/emoji_formatter.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'json'
+
+class Integrations::Slack::EmojiFormatter
+  EMOJI_MAPPING = JSON.parse(File.read(File.join(__dir__, 'emojis.json'))).freeze
```
stonecharioteer: Is it possible to get the slack emojis from somewhere upstream instead?

Lomuzord: Yes, it is possible. The most popular upstream source is [https://github.com/github/gemoji](https://github.com/github/gemoji) , which is maintained by GitHub.

stonecharioteer: I'd recommend doing that, please. Updating this file would be a full-time task otherwise.

sony-mathew (approved):
LGTM. Thank you the contribution @Lomuzord !

PR #12944, opened 2025-11-25
fix: Prevent SLA deletion timeouts by moving to async job

muhsin-k:
This PR fixes the HTTP 500 timeout errors occurring when deleting SLA policies that have large volumes of historical data.
The fix moves the deletion workflow to asynchronous background processing using the existing `DeleteObjectJob`.
By offloading heavy cascaded deletions (applied SLAs, SLA events, conversation nullifications) from the request cycle, the API can now return immediately while the cleanup continues in the background avoiding the `Rack::Timeout::RequestTimeoutException`. This ensures that SLA policies can be deleted reliably, regardless of data size.

### Problem
Deleting an SLA policy via `DELETE /api/v1/accounts/{account_id}/sla_policies/{id}` fails consistently with `Rack::Timeout::RequestTimeoutException (15s)` for policies with large amounts of related data.

Because the current implementation performs all dependent deletions **synchronously**, Rails processes:

- `has_many :applied_slas, dependent: :destroy` (thousands)
- Each `AppliedSla#destroy` → triggers destruction of many `SlaEvent` records
- `has_many :conversations, dependent: :nullify` (thousands)

This processing far exceeds the Rack timeout window and consistently triggers HTTP 500 errors for users.

### Solution

This PR applies the same pattern used successfully in Inbox deletion.

**Move deletion to async background jobs**

- Uses `DeleteObjectJob` for centralized, reliable cleanup.
- Allows the DELETE API call to respond immediately.

**Chunk large datasets**

- Records are processed in **batches of 5,000** to reduce DB load and avoid job timeouts.

Fixes https://linear.app/chatwoot/issue/CW-5871/unable-to-delete-sla-policy


stonecharioteer (approved):
Looks good. I'm curious, how many SLAs do we expect usually that we need to bulk delete this way?

enterprise/app/jobs/enterprise/delete_object_job.rb
```diff
@@ -1,10 +1,16 @@
 module Enterprise::DeleteObjectJob
+  def self.prepended(base)
```
muhsin-k: Convert the constant HEAVY_ASSOCIATIONS to a method in base class and override.
cc @sojan-official

sojan-official (approved):

PR #12956, opened 2025-11-26
perf: use account.contacts directly in search to reduce DB load

vishnu-narayanan:
- Use resolved contacts instead of accounts.contacts for search


pranavrajs:
Note: This is done to reduce the query complexity on production systems post the incident related to contact search.

pranavrajs (approved):

PR #12961, opened 2025-11-26
feat: migrate editor to ruby-llm

aakashb95:
## Description
Migrate editor ai features to ruby_llm gem

## Type of change
Breaking change (fix or feature that would cause existing functionality not to work as expected)

## How Has This Been Tested?

Tested all editor features on local and the corresponding instrumentation on Langfuse. Currently only tested with Openai models. But it is written in a way to support more models via RubyLLM + hooks going forward.

image

## Checklist:


lib/integrations/llm_instrumentation.rb
```diff
@@ -24,6 +24,14 @@ module Integrations::LlmInstrumentation
   ATTR_LANGFUSE_SESSION_ID = 'langfuse.session.id'
   ATTR_LANGFUSE_TAGS = 'langfuse.trace.tags'
 
+  PROVIDER_PREFIXES = {
```
aakashb95, line 11: need to check how to get these dynamically from rubyllm, although we will support only these initially

aakashb95:
@codex please review

lib/integrations/llm_base_service.rb
```diff
@@ -82,64 +81,84 @@ def event_is_cacheable?
...
-    return build_error_response(response, messages) unless response.success?
+  def send_conversation_messages(chat, messages)
+    conversation_messages = messages.reject { |m| m['role'] == 'system' }
+
+    return chat.ask(conversation_messages.first['content']) if conversation_messages.length == 1
 
-    parsed_response = JSON.parse(response.body)
-    build_success_response(parsed_response, messages)
+    add_conversation_history(chat, conversation_messages[0...-1])
+    chat.ask(conversation_messages.last['content'])
```
scmmishra, line 133: `conversation_messages` can be empty too. In that case it can break. See if we can handle this case

lib/integrations/llm_base_service.rb
```diff
@@ -82,64 +81,84 @@ def event_is_cacheable?
...
 
-  def execute_api_request(body, messages)
-    Rails.logger.info("OpenAI API request: #{body}")
-    response = HTTParty.post(api_url, headers: api_headers, body: body)
-    Rails.logger.info("OpenAI API response: #{response.body}")
+    Llm::Config.with_api_key(hook.settings['api_key'], api_base: api_base) do |context|
+      chat = context.chat(model: model)
+      setup_chat_with_messages(chat, messages)
+    end
+  rescue RubyLLM::Error => e
```
scmmishra: I am assuming any other error type will be bubbled up, but handled safely upstream

lib/integrations/llm_base_service.rb
```diff
@@ -82,64 +81,84 @@ def event_is_cacheable?
...
+    build_error_response_from_exception(e, messages)
```
scmmishra, line 109: It seems like this error is important information... Let's also capture this exception, use the `ChatwootExceptionTracker` we can also set an alert on this. 

Let me know if I am missing something

aakashb95: makes sense, should catch any and all errors at this place i think, have made the change

lib/integrations/llm_base_service.rb
aakashb95: good catch, should raise an error in this case right

aakashb95:
@scmmishra i tested this PR on [staging](https://us.cloud.langfuse.com/project/cmi7uwal4014yad07et0pmu8y/traces?filter=name%3BstringOptions%3B%3Bnone+of%3Bllm.label_suggestion), it is working fine. 
image

i think we can merge this

lib/integrations/llm_instrumentation.rb
```diff
@@ -2,9 +2,19 @@
...
   include Integrations::LlmInstrumentationConstants
+  include Integrations::LlmInstrumentationHelpers
+
+  PROVIDER_PREFIXES = {
+    'openai' => %w[gpt- o1 o3 o4 text-embedding- whisper- tts-],
+    'anthropic' => %w[claude-],
+    'google' => %w[gemini-],
+    'mistral' => %w[mistral- codestral-],
+    'deepseek' => %w[deepseek-]
+  }.freeze
```
scmmishra, line 17: Just be careful about one thing, RubyLLM has model name verification, so you need to periodically refresh it's model list. We can disable the behaviour, but it comes with some caveats. 

Read more: https://rubyllm.com/models/

scmmishra (approved):
Just one comment to take note of. The rest is good

lib/integrations/llm_instrumentation.rb
aakashb95: yeah agree, i thought of addressing it in a later PR when we support different providers. There we will have to think about this, and ensuring tool calling related functionality is performed by select providers and models, etc..

PR #12787, opened 2025-11-03
feat: Add support for sending CSAT surveys via templates (Whatsapp Cloud)

muhsin-k:
This PR enables sending CSAT surveys on WhatsApp using approved WhatsApp message templates, ensuring survey delivery even after the 24-hour session window.

The system now automatically creates, updates, and monitors WhatsApp CSAT templates without manual intervention.

approved

#### Why this change

Previously, WhatsApp CSAT messages failed outside the 24-hour customer window.

With this update:

- CSAT surveys are delivered reliably using WhatsApp templates
- Template creation happens automatically in the background
- Users can modify survey content and recreate templates easily
- Clear UI states show template approval status

#### Screens & States

<details>
<summary>Default — No template configured yet</summary>
default
</details>
<details>
<summary>Pending — Template submitted, awaiting Meta approval</summary>
pending
</details>
<details>
<summary>Approved — Survey will be sent when conversation resolves</summary>
approved
</details>
<details>
<summary>Rejected — Template rejected by Meta</summary>
rejected

</details>
<details>
<summary>Not Found — Template missing in Meta Platform</summary>
not-exist
</details>
<details>
<summary>Edit Template — Delete & recreate template on change</summary>
edit-survey
</details>

#### Test Cases

**1. First-time CSAT setup on WhatsApp inbox**

- Enable CSAT
- Enter message + button text
- Save
- Expected: Template created automatically, UI shows pending state

**2. CSAT toggle without changing text**

- Existing approved template
- Toggle CSAT OFF → ON (no text change)
- Expected: No confirmation alert, no template recreation

**3. Editing only survey rules**

- Modify labels or rule conditions only
- Expected: No confirmation alert, template remains unchanged

**4. Template text change**

- Change survey message or button text
- Save
- Expected:
    - Confirmation dialog shown
    - On confirm → previous template deleted, new one created
    - On cancel → revert to previous values

**5. Language change**

- Change template language (e.g., en → es)
- Expected: Confirmation dialog + new template on confirm

 **6. Sending survey**

- Template approved → always send template
- Template pending → send free-form within 24 hours only
- Template rejected/missing → fallback to free-form (if within window)
- Outside 24 hours & no approved template → activity log only

**7. Non-WhatsApp inbox**

- Enable CSAT for email/web inbox
- Expected: No template logic triggered

Fixes https://linear.app/chatwoot/issue/CW-6188/support-for-sending-csat-surveys-via-approved-whatsapp

PR #12984, opened 2025-12-01
feat: Add backend changes for whatsapp csat template

muhsin-k:
This PR add the backend changes for the feature [sending CSAT surveys via WhatsApp message templates ](https://github.com/chatwoot/chatwoot/pull/12787)


muhsin-k:
@codex Please review

app/controllers/api/v1/accounts/inbox_csat_templates_controller.rb
```diff
@@ -0,0 +1,154 @@
+class Api::V1::Accounts::InboxCsatTemplatesController < Api::V1::Accounts::BaseController
```
muhsin-k, line 1: Created a separate controller because the inbox controller was growing too large.

app/controllers/api/v1/accounts/inbox_csat_templates_controller.rb
```diff
@@ -0,0 +1,157 @@
...
+  end
+
+  private
+
+  def fetch_inbox
+    @inbox = Current.account.inboxes.find(params[:inbox_id])
+    authorize @inbox, :show?
+  end
+
+  def validate_whatsapp_channel
```
stonecharioteer, line 45: It would be a good idea to record that we don't support creation of templates for the TwilioSms whatsapp channel?

app/services/whatsapp/csat_template_service.rb
```diff
@@ -0,0 +1,144 @@
...
+  rescue StandardError => e
+    Rails.logger.error "Error fetching template status: #{e.message}"
+    { success: false, error: e.message }
+  end
+
+  private
+
+  def generate_template_name(base_name)
+    # Get current template to check if we already have a version number
+    template = @whatsapp_channel.inbox.csat_config&.dig('template') || {}
```
stonecharioteer: In our implementation, the whatsapp templates are scoped to the business account, right? So what happens when users of 2 separate inboxes try to use the same template name? Users could have 2 whatsapp channels with different phone numbers, right?

app/controllers/api/v1/accounts/inbox_csat_templates_controller.rb
muhsin-k: We have a plan to support twilio whatsapp channel as well.

app/services/whatsapp/csat_template_service.rb
muhsin-k: @stonecharioteer Good catch. Changed the implementation to  `customer_satisfaction_survey_{inbox_id}_{version_number}` for template name so each inbox gets a unique template name.

stonecharioteer:
@codex please review

app/controllers/api/v1/accounts/inbox_csat_templates_controller.rb
```diff
@@ -0,0 +1,160 @@
...
+  def create_template_via_provider(template_params)
+    template_config = {
+      message: template_params[:message],
+      button_text: template_params[:button_text] || DEFAULT_BUTTON_TEXT,
+      base_url: ENV.fetch('FRONTEND_URL', 'http://localhost:3000'),
+      language: template_params[:language] || DEFAULT_LANGUAGE,
+      template_name: Whatsapp::CsatTemplateNameService.csat_template_name(@inbox.id)
+    }
+
+    @inbox.channel.provider_service.create_csat_template(template_config)
```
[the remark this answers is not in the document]
muhsin-k: We no longer support 360dialog, so we can keep the existing logic as is.

stonecharioteer (approved):
LGTM, any thoughts on the timeline for the twilio x whatsapp channel?

muhsin-k:
I'll pick this up after the whatsapp cloud front-end changes go live.

PR #12988, opened 2025-12-02
feat: Add AI credit topup flow for Stripe

tds-1:


enterprise/app/controllers/enterprise/api/v1/accounts_controller.rb
```diff
@@ -55,6 +55,24 @@ def toggle_deletion
     end
   end
 
+  def topup_options
```
scmmishra: Let's keep the topup options in the frontend only, no need for a network call also. We can add the catalog later. On the backend we can validate from a hard coded array. Some duplication is fine here. Easier to write specs too, and the frontend will be simpler as well

enterprise/app/services/enterprise/billing/handle_stripe_event_service.rb
```diff
@@ -101,10 +124,39 @@ def enable_features_for_current_plan
...
+      limits: current_limits.merge(
+        'captain_responses_monthly' => monthly_credits,
+        'captain_responses_topup' => current_topup,
+        'captain_responses' => total_credits,
+        'captain_documents' => plan_credits[:documents]
+      )
+    )
+
+    # Reset the usage counter for the new billing period
     account.reset_response_usage
```
scmmishra: Will the reset usage also reset the usage of their credits the customers purchased? I think we can remove this call

enterprise/app/services/enterprise/billing/topup_checkout_service.rb
```diff
@@ -0,0 +1,115 @@
...
+      },
+      quantity: 1
+    }
+  end
+
+  def session_metadata(credits, topup_option)
+    {
+      account_id: account.id.to_s,
+      credits: credits.to_s,
+      amount_cents: (topup_option[:amount] * 100).to_i.to_s,
```
scmmishra:
```suggestion
      amount_cents: (topup_option[:amount] * 100).to_s,
```

The `to_i` is redundant

enterprise/app/services/enterprise/billing/topup_fulfillment_service.rb
```diff
@@ -0,0 +1,60 @@
...
+  def initialize(account:)
+    @account = account
+  end
+
+  def fulfill(credits:, amount_cents:, currency:)
+    account.with_lock do
+      credit_grant = create_stripe_credit_grant(credits, amount_cents, currency)
+      update_account_credits(credits)
+
+      { success: true, credit_grant_id: credit_grant.id, credits_added: credits }
```
scmmishra: The pattern of returning an object is not great. Let's return what information we actually need. And throw errors for any other issues.

enterprise/app/services/enterprise/billing/topup_fulfillment_service.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+  rescue Stripe::StripeError => e
+    Rails.logger.error("Failed to create credit grant for account #{account.id}: #{e.message}")
+    { success: false, message: "Stripe error: #{e.message}" }
```
scmmishra: Unless we're doing something useful with the error here, it's best to let it bubble up and handle it at the consumer side. Passing this `{ success: false, message: "Stripe error: #{e.message}" }` is not very Ruby like and also is lossy, since the error object is more useful to the consumer

In general, let's not return hashes, it's not very idiomatic. This pattern works really well if we have strong types, but in Ruby it's not the case.

scmmishra: Also, let's use ChatwootExceptionTracker at the top wherever we're handling it. Billing errors should be monitored closely

scmmishra (dismissed):

enterprise/app/services/enterprise/billing/handle_stripe_event_service.rb
tds-1: This function makes the usage to 0, and resets for the next month or subscription.

scmmishra: Yeah, that makes credits purchased not `one-time` use. 

Alex buys 5000 credits -> Uses 4000 credits -> Usage resets next month, they have 5000 credits again

scmmishra: I think reset should reduce the usage by the default credits entitled

tds-1: Understood

enterprise/app/services/enterprise/billing/handle_stripe_event_service.rb
```diff
@@ -73,6 +77,26 @@ def process_subscription_deleted
...
+
+    credits = metadata['credits'].to_i
+    amount_cents = metadata['amount_cents'].to_i
+    currency = metadata['currency'] || 'usd'
+    Enterprise::Billing::TopupFulfillmentService.new(account: topup_account).fulfill(
+      credits: credits,
+      amount_cents: amount_cents,
+      currency: currency
+    )
+  end
```
scmmishra, line 118: Let's catch any error here instead of the `fulfill` method. And track the exception here, and include any details we have from stripe as well. Maybe using Rails.logger.error. That way we have enough data to fix any data after the fact

scmmishra (commented):
One small change

app/javascript/dashboard/i18n/locale/en/settings.json
```diff
@@ -400,14 +400,28 @@
...
     },
-    "NO_BILLING_USER": "Your billing account is being configured. Please refresh the page and try again."
+    "NO_BILLING_USER": "Your billing account is being configured. Please refresh the page and try again.",
+    "TOPUP": {
+      "BUY_CREDITS": "Buy more credits",
+      "MODAL_TITLE": "Purchase Credits",
+      "MODAL_DESCRIPTION": "Select a credit package to add to your account.",
+      "CREDITS": "credits",
+      "POPULAR": "Popular",
+      "NOTE": "Credits are added to your account immediately after purchase and never expire.",
```
scmmishra:
```suggestion
      "NOTE": "Credits are added to your account immediately after purchase and expire in 6 months.",
```

pranavrajs: Please use the exact text in the Figma. https://www.figma.com/make/YlrympEwG6kDlRGpeaA6Ab/Improve-Billing-Flow?node-id=0-1&p=f&t=kXqfCNUX0VAsCn7y-0

enterprise/app/services/enterprise/billing/topup_checkout_service.rb
```diff
@@ -0,0 +1,115 @@
...
+  def create_checkout_session(credits:)
+    topup_option = validate_and_find_topup_option(credits)
+    session = create_stripe_session(topup_option, credits)
+    session.url
+  end
+
+  private
+
+  def validate_and_find_topup_option(credits)
+    raise Error, 'Invalid credits amount' unless credits.to_i.positive?
```
pranavrajs: use i18n, this is a message sent to the UI.

pranavrajs: all the items below.

enterprise/app/services/enterprise/billing/topup_checkout_service.rb
```diff
@@ -0,0 +1,115 @@
...
+    topup_option
+  end
+
+  def validate_plan_eligibility!
+    plan_name = account.custom_attributes&.[]('plan_name')
+
+    return if plan_name.present? && plan_name.downcase != 'hacker'
+
+    raise Error, 'Top-ups are only available for Startup, Business, and Enterprise plans. Please upgrade your plan first.'
+  end
```
pranavrajs: Use billing helper, default_plan?(account) is enough.

enterprise/app/services/enterprise/billing/topup_checkout_service.rb
```diff
@@ -0,0 +1,115 @@
...
+      # Create invoice for this payment so it appears in customer portal
+      invoice_creation: build_invoice_creation_data(credits, topup_option)
+    )
+  end
+
+  def build_invoice_creation_data(credits, topup_option)
+    {
+      enabled: true,
+      invoice_data: {
+        description: "Credit Topup: #{credits} credits",
```
pranavrajs:
```suggestion
        description: "AI Credit Topup: #{credits} credits",
```

enterprise/app/services/enterprise/billing/topup_checkout_service.rb
```diff
@@ -0,0 +1,115 @@
...
+      }
+    }
+  end
+
+  def build_line_item(topup_option, credits)
+    {
+      price_data: {
+        currency: topup_option[:currency],
+        unit_amount: (topup_option[:amount] * 100).to_i,
+        product_data: { name: "Credit Topup: #{credits} credits" }
```
pranavrajs:
```suggestion
        product_data: { name: "AI Credit Topup: #{credits} credits" }
```

enterprise/app/services/enterprise/billing/topup_fulfillment_service.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    account.update!(
+      limits: current_limits.merge(
+        'captain_responses' => new_total
+      )
+    )
+  end
+
+  def stripe_customer_id
+    account.custom_attributes&.[]('stripe_customer_id')
```
pranavrajs: why do we do this everywhere? we can just use account.custom_attributes[:stripe_customer_id] directly.

enterprise/app/services/enterprise/billing/handle_stripe_event_service.rb
```diff
@@ -101,8 +144,25 @@ def enable_features_for_current_plan
...
+
+    topup_deduct = [previous_usage[:responses] - previous_usage[:monthly], 0].max
+    total_credits = total_credits - topup_deduct - previous_usage[:monthly] + new_plan_credits
+
+    account.update!(limits: current_limits.merge('captain_responses' => total_credits))
+  end
+
+  def get_plan_credits(plan_name)
+    config = InstallationConfig.find_by(name: CAPTAIN_CLOUD_PLAN_LIMITS)&.value
+    return { responses: 0, documents: 0 } if config.blank? || plan_name.blank?
```
pranavrajs: There is no need for this check. This would be available in production.

enterprise/app/services/enterprise/billing/handle_stripe_event_service.rb
```diff
@@ -101,8 +144,25 @@ def enable_features_for_current_plan
...
+
+    plan_quota = JSON.parse(config)
```
pranavrajs: config.value would give the required information.

enterprise/app/services/enterprise/billing/topup_checkout_service.rb
```diff
@@ -0,0 +1,115 @@
...
+  TOPUP_OPTIONS = [
+    { credits: 1000, amount: 20.0, currency: 'usd' },
+    { credits: 2500, amount: 50.0, currency: 'usd' },
+    { credits: 5000, amount: 100.0, currency: 'usd' },
+    { credits: 10_000, amount: 200.0, currency: 'usd' }
+  ].freeze
+
+  attr_reader :account
+
+  def initialize(account:)
```
pranavrajs: use pattr_intialize

enterprise/app/services/enterprise/billing/topup_checkout_service.rb
```diff
@@ -0,0 +1,115 @@
...
+    }
+  end
+
+  def success_url
+    "#{base_url}/app/accounts/#{account.id}/settings/billing?topup=success"
+  end
+
+  def cancel_url
+    "#{base_url}/app/accounts/#{account.id}/settings/billing"
+  end
```
pranavrajs: Use rails helpers. Check twitter callbacks controller for an example.

enterprise/app/services/enterprise/billing/topup_checkout_service.rb
```diff
@@ -0,0 +1,115 @@
...
+
+  def base_url
+    ENV.fetch('FRONTEND_URL', 'http://localhost:3000')
+  end
```
pranavrajs: this is not required if you are using rails helper.

enterprise/app/services/enterprise/billing/topup_checkout_service.rb
```diff
@@ -0,0 +1,115 @@
...
+
+  def stripe_customer_id
+    account.custom_attributes&.[]('stripe_customer_id')
+  end
```
pranavrajs, line 94: use account.custom_attributes[:stripe_customer_id]  directly.

enterprise/app/services/enterprise/billing/topup_fulfillment_service.rb
```diff
@@ -0,0 +1,52 @@
+class Enterprise::Billing::TopupFulfillmentService
+  attr_reader :account
+
+  def initialize(account:)
+    @account = account
+  end
+
```
pranavrajs: Use pattr_intialize.

enterprise/app/controllers/enterprise/api/v1/accounts_controller.rb
```diff
@@ -55,6 +55,16 @@ def toggle_deletion
     end
   end
 
+  def topup_checkout
+    return render json: { error: 'Credits amount is required' }, status: :unprocessable_entity if params[:credits].blank?
```
pranavrajs: Same here. use i18n.

enterprise/app/controllers/enterprise/api/v1/accounts_controller.rb
```diff
@@ -55,6 +55,16 @@ def toggle_deletion
...
+
+    service = Enterprise::Billing::TopupCheckoutService.new(account: @account)
+    redirect_url = service.create_checkout_session(credits: params[:credits].to_i)
+    render json: { redirect_url: redirect_url }
+  rescue Enterprise::Billing::TopupCheckoutService::Error => e
```
pranavrajs, line 64: log this error. donot throw it to sentry.

enterprise/app/controllers/enterprise/api/v1/accounts_controller.rb
```diff
@@ -55,6 +55,16 @@ def toggle_deletion
...
+    render json: { error: e.message }, status: :unprocessable_entity
```
pranavrajs: use the helper for sending errors.  There is a method `render_could_not_create_error`

enterprise/app/services/enterprise/billing/handle_stripe_event_service.rb
```diff
@@ -101,8 +144,25 @@ def enable_features_for_current_plan
...
-    account.reset_response_usage
+  # Recalculates credits on renewal: deducts usage exceeding free allowance, then adds new plan credits.
+  # Formula: new_credits = current - max(usage - monthly_free, 0) - monthly_free + new_plan_allowance
+  def handle_subscription_credits(plan, previous_usage)
+    new_plan_credits = get_plan_credits(plan['name'])[:responses]
+    current_limits = account.limits || {}
+    total_credits = current_limits['captain_responses'].to_i
+
+    topup_deduct = [previous_usage[:responses] - previous_usage[:monthly], 0].max
+    total_credits = total_credits - topup_deduct - previous_usage[:monthly] + new_plan_credits
```
pranavrajs: Shall we rename it better and remove the comment?

```rb
    current_credits = (account.limits || {})['captain_responses'].to_i
    new_plan_credits = get_plan_credits(plan['name'])[:responses]

    consumed_topup_credits = [previous_usage[:responses] - previous_usage[:monthly], 0].max
    updated_credits = current_credits - consumed_topup_credits - previous_usage[:monthly] + new_plan_credits
```

pranavrajs (changes requested):
Please take a look at the comments.

pranavrajs (approved):

PR #12997, opened 2025-12-03
feat: Add support for shared post and story attachment types in Instagram messages

muhsin-k:
When users share Instagram posts or stories via DM, Instagram sends webhooks with type `ig_post` and `ig_story` attachments. The system was failing on these types because they weren't defined in the file_types. This PR fixes the issue by handling all shared types and rendering them on the front end.

**Shared post**

CleanShot 2025-12-03 at 16 29 14@2x

**Shared status**
CleanShot 2025-12-03 at 16 10 25@2x

Fixes https://linear.app/chatwoot/issue/CW-5441/argumenterror-ig-story-is-not-a-valid-file-type-argumenterror


stonecharioteer:
LGTM, the codex comment about i18n seems relevant, the other comment not so much.

@muhsin-k

stonecharioteer (approved):

sojan-official (approved):
Code LGTM

PR #13003, opened 2025-12-04
fix: Filter out unsupported ephemeral message attachments

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-6070/argumenterror-ephemeral-is-not-a-valid-file-type-argumenterror

**Problem**
The instagram webhooks containing ephemeral (disappearing) message were causing ArgumentError exceptions because this attachment type is not supported and was not in the enum validation. 

**Solution**
- Added ephemeral to the unsupported_file_type? filter
- Ephemeral attachments are now silently filtered out before processing, following the same pattern as existing unsupported types (template, unsupported_type)


stonecharioteer (approved):
LGTM, is this a new change from the Instagram end?

muhsin-k:
> is this a new change from the Instagram end?

No, It was already there.

PR #13013, opened 2025-12-05
feat: Add automatic favicon fetching for companies

stonecharioteer:
## Summary

This Enterprise-only feature automatically fetches a favicon for companies created with a domain, and adds a batch task to backfill missing avatars for existing companies. The flow only targets companies that do not already have an attached avatar, so existing avatars are left untouched.

## Demo 

https://github.com/user-attachments/assets/d050334e-769f-4e46-b6e7-f7423727a192

## What changed

- Added `Avatar::AvatarFromFaviconJob` to build a Google favicon URL from the company domain and fetch it through `Avatar::AvatarFromUrlJob`
- Triggered favicon fetching from `Company` with `after_create_commit`
- Added `Companies::FetchAvatarsJob` to batch existing companies that are missing avatars
- Added `companies:fetch_missing_avatars` under `enterprise/lib/tasks`
- Kept the company-specific implementation inside the Enterprise boundary
- Stubbed the new favicon request in unrelated specs that now hit this callback indirectly
- Updated a couple of CI-sensitive specs that were failing due to callback side effects / reload-safe exception assertions

## How to verify

1. Create a company in Enterprise with a valid domain and no avatar.
2. Confirm that a favicon-based avatar gets attached shortly after creation.
3. Create another company with a domain and an avatar already attached.
4. Confirm that the existing avatar is not replaced.
5. Run `companies:fetch_missing_avatars`.
6. Confirm that existing companies without avatars get one, while companies that already have avatars remain unchanged.

## Notes

- This change does not refresh or overwrite existing company avatars
- Favicon fetching only runs for companies with a present domain
- The branch includes the latest `develop`

### DIFF
--- a/app/jobs/avatar/avatar_from_favicon_job.rb
+++ b/app/jobs/avatar/avatar_from_favicon_job.rb
@@ -3,7 +3,7 @@ class Avatar::AvatarFromFaviconJob < ApplicationJob
 
   def perform(company)
     return if company.domain.blank?
-    return if company.avatar_url.present?
+    return if company.avatar.attached?
 
     favicon_url = "https://www.google.com/s2/favicons?domain=#{company.domain}&sz=256"
     Avatar::AvatarFromUrlJob.perform_now(company, favicon_url)
--- a/spec/enterprise/jobs/avatar/avatar_from_favicon_job_spec.rb
+++ b/spec/enterprise/jobs/avatar/avatar_from_favicon_job_spec.rb
@@ -4,25 +4,12 @@ RSpec.describe Avatar::AvatarFromFaviconJob do
   let(:company) { create(:company, domain: 'wikipedia.org') }
   let(:favicon_url) { 'https://www.google.com/s2/favicons?domain=wikipedia.org&sz=256' }
 
-  it 'enqueues the job' do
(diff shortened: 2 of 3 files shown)


app/jobs/avatar/avatar_from_favicon_job.rb
```diff
@@ -0,0 +1,10 @@
+class Avatar::AvatarFromFaviconJob < ApplicationJob
+  queue_as :purgable
+  def perform(company)
+    return if company.domain.blank?
+    return if company.avatar_url.present?
+
+    favicon_url = "https://www.google.com/s2/favicons?domain=#{company.domain}&sz=256"
```
scmmishra, line 8: @stonecharioteer we might have to test the rate limit for this, since this is an undocumented API

app/jobs/avatar/avatar_from_favicon_job.rb
```diff
@@ -0,0 +1,10 @@
...
+    Avatar::AvatarFromUrlJob.perform_later(company, favicon_url)
```
scmmishra: Since this is already in a job, would be a good idea to do `perform_now` instead of `perform_later`

sojan-official:
@codex review plz

PR #13089, opened 2025-12-17
feat(inbox): add E keyboard shortcut to delete notification and advance

salmonumbrella:
## Summary

Adds Superhuman-style keyboard shortcut for faster inbox processing:

- **Press E** to delete the current notification and automatically navigate to the next one
- Works in both conversation view (`/inbox-view/conversation/:id`) and sidebar list
- Disabled when cursor is in a text input (reply box, search, etc.)
- Includes analytics tracking and user feedback toast

## Motivation

Power users processing many notifications benefit from keyboard-driven workflows. This matches patterns from Superhuman, Notion Mail, and similar productivity tools where E = archive/delete and advance.

## Implementation

- Uses existing `useKeyboardEvents` composable (same pattern as L key for labels)
- Captures notification ID before delete to avoid race conditions with reactive store updates
- Falls back to inbox view when on the last notification
- Properly separated error handling: if delete fails, don't navigate

## Test plan

- [ ] Press E while viewing a conversation in inbox view → deletes and goes to next
- [ ] Press E while on last notification → deletes and shows inbox empty state
- [ ] Press E while typing in reply box → does nothing (shortcut disabled)
- [ ] Press E in sidebar list with notification selected → deletes and advances
- [ ] Verify toast notification appears after delete
- [ ] Check keyboard shortcuts modal shows new shortcut

## Screenshots

N/A - keyboard shortcut, no visual changes

### DIFF
--- a/app/javascript/dashboard/components/widgets/modal/constants.js
+++ b/app/javascript/dashboard/components/widgets/modal/constants.js
@@ -92,4 +92,10 @@ export const SHORTCUT_KEYS = [
     displayKeys: [KEYS.ALT, 'M'],
     keySet: ['Alt+KeyM'],
   },
+  {
+    id: 15,
+    label: 'DELETE_NOTIFICATION',
+    displayKeys: ['E'],
+    keySet: ['KeyE'],
+  },
 ];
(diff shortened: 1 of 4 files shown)


salmonumbrella:
Thanks for catching this! Fixed in 15d88f78d - InboxList.vue now checks `if (currentConversationId.value) return;` to skip handling when a conversation is open, letting InboxView.vue handle it instead.

salmonumbrella:
Hey @scmmishra — this adds an 'E' keyboard shortcut to delete a notification and advance to the next one, similar to Gmail-style workflow. It's a small quality-of-life improvement. Could you review when you get a moment? Thanks\!

PR #13091, opened 2025-12-17
fix: Prioritize SDK `enableFileUpload` flag when explicitly set

muhsin-k:
### Problem
Currently, the attachment button visibility in the widget uses both the SDK's `enableFileUpload` flag AND the inbox's attachment settings with an AND condition. This creates an issue for users who want to control attachments solely through inbox settings, since the SDK flag defaults to `true` even when not explicitly provided.

  **Before:**
  - SDK not set → `enableFileUpload: true` (default) + inbox setting = attachment shown only if both true
  - SDK set to false → `enableFileUpload: false` + inbox setting = attachment always hidden
  - SDK set to true → `enableFileUpload: true` + inbox setting = attachment shown only if both true
  
  This meant users couldn't rely solely on inbox settings when the SDK flag wasn't explicitly provided.

  ### Solution

  Changed the logic to prioritize explicit SDK configuration when provided, and fall back to inbox settings when not provided:

  **After:**
  - SDK not set → `enableFileUpload: undefined` → use inbox setting only
  - SDK set to false → `enableFileUpload: false` → hide attachment (SDK controls)
  - SDK set to true → `enableFileUpload: true` → show attachment (SDK controls)


muhsin-k:
@codex please review

muhsin-k:
@codex please review

iamsivin (approved):
Tested locally, LGTM.

https://www.loom.com/share/5dc1c218e90d4c0585260a4cecc5ad41

scmmishra (approved):

PR #13093, opened 2025-12-17
chore(dev): add cleanup flow to force_run in Makefile

stonecharioteer:
## Summary

Improve local dev restart reliability by enhancing `make force_run` to run cleanup before starting Overmind.

## How To Reproduce

During local development, if `make run` is interrupted (for example with Ctrl-C), stale state can remain (`.overmind.sock`, PID files, and processes on ports `3000`/`3036`), which can block or complicate the next restart.

## Changes

Updated `force_run` in `Makefile` to:
- print cleanup start/end messages
- kill processes on ports `3036` and `3000` (best-effort)
- remove `.overmind.sock`
- remove `tmp/pids/*.pid`
- then start `Procfile.dev` via Overmind

No other files are changed in this PR.

## Testing

- Verified branch diff against `develop` only touches `Makefile`.
- Ran `make -n force_run` to validate the command sequence and startup flow.


Makefile
```diff
@@ -32,6 +32,14 @@ server:
 burn:
 	bundle && pnpm install
 
+cleanup:
```
sojan-official: we can club this to the force_run command

vishnu-narayanan:
@stonecharioteer The trick is to never user `make run` - always use `make force_run` ;)

PS: As @sojan-official suggested, should we combine `cleanup` with `force_run`.

sojan-official:
Addressed review feedback: removed the standalone `cleanup` target and merged its cleanup steps into `make force_run` in the Makefile.

PR #13095, opened 2025-12-17
fix: bot should respond only if the chat status is pending

rondelinga:
# Pull Request Template

## Description

Fixes # (issue)

## Type of change

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/app/builders/messages/message_builder.rb
+++ b/app/builders/messages/message_builder.rb
@@ -21,6 +21,11 @@ class Messages::MessageBuilder
   end
 
   def perform
+    if @user.is_a?(AgentBot) && @conversation.status != 'pending'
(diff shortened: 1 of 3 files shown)


pranavrajs:
@rondelinga

I don’t think we can accept this. This is an implementation-level concern.

If the agent bot shouldn’t respond to conversations that are in a pending state, that logic can be handled at the application layer when the bot is implemented. The conversation status is already exposed via webhooks and across almost all APIs, so developers have enough signals to handle this on their side.

We don’t want to enforce this behavior at the platform level, especially since we’re planning to move away from the pending status. We’ll be introducing an explicit assignment for agent bots in the future, which should address this more cleanly.

I’m going to close this for now. If you have a different angle or a concrete use case that we should reconsider, feel free to follow up.

PR #13096, opened 2025-12-17
fix: Use original email timestamp for imported messages

stonecharioteer:
## Description

When importing historical emails via IMAP, preserve the original email send timestamp instead of using the current time. This ensures that imported emails display with their actual dates in the conversation view.

This is essential for customer data migrations where historical email conversations need to maintain their original chronology.

## Type of change

## How Has This Been Tested?

Tested locally by:
1. Connecting a Gmail inbox via IMAP
2. Running `Inboxes::FetchImapEmailsJob.perform_now(channel, 90)` to import 90 days of historical emails
3. Verified messages display original email timestamps in the conversation UI
4. Verified conversation list shows correct "last message" times
5. Confirmed historical emails maintain proper chronological order

## Changes

- Set message `created_at` to the email's original date (used by frontend for display)
- Set conversation `initiated_at` timestamp to the email's original date (used for metadata)  
- Both fields fall back to current time if email lacks a date header
- Uses `processed_mail.date&.utc` which reads from the email's Date header (RFC 2822)
- Added RuboCop disable for `Metrics/AbcSize` as the complexity increase is justified

## Checklist:

### DIFF
--- a/app/javascript/entrypoints/sdk.js
+++ b/app/javascript/entrypoints/sdk.js
@@ -76,7 +76,7 @@ const runSDK = ({ baseUrl, websiteToken }) => {
     welcomeDescription: chatwootSettings.welcomeDescription || '',
     availableMessage: chatwootSettings.availableMessage || '',
     unavailableMessage: chatwootSettings.unavailableMessage || '',
-    enableFileUpload: chatwootSettings.enableFileUpload ?? true,
+    enableFileUpload: chatwootSettings.enableFileUpload,
     enableEmojiPicker: chatwootSettings.enableEmojiPicker ?? true,
     enableEndConversation: chatwootSettings.enableEndConversation ?? true,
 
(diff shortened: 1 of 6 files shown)


stonecharioteer:
In the frontend, when ingesting older emails, it should look like this now. 
Screenshot 2025-12-17 at 19 29 19

Previously, it showed the ingestion date at both spots.

app/mailboxes/mailbox_helper.rb
```diff
@@ -1,6 +1,7 @@
 module MailboxHelper
   private
 
+  # rubocop:disable Metrics/AbcSize
   def create_message
     Rails.logger.info "[MailboxHelper] Creating message #{processed_mail.message_id}"
     return if @conversation.messages.find_by(source_id: processed_mail.message_id).present?
```
scmmishra, line 7: If we move this to a separate method, I think we can fix the `Metrics/AbcSize` issue

app/mailboxes/mailbox_helper.rb
```diff
@@ -13,13 +14,15 @@ def create_message
       message_type: 'incoming',
       content_type: 'incoming_email',
       source_id: processed_mail.message_id,
+      created_at: processed_mail.date&.utc || Time.current,
```
pranavrajs, line 17: Mutating created_at breaks Rails conventions and makes the record’s lifecycle ambiguous.
To preserve the email’s original originate time, we should add an original_created_at column, default it to created_at, and set it explicitly when ingesting emails. Whenever it is displayed on the UI, the created time can be shown.

pranavrajs:

PR #13106, opened 2025-12-18
feat: reply editor follow-up

scmmishra:
https://github.com/user-attachments/assets/47b20ef7-6a56-436e-91ac-25fcf2f33eae

AI-8


enterprise/app/controllers/api/v1/accounts/captain/tasks_controller.rb
```diff
@@ -39,6 +39,16 @@ def label_suggestion
     render_result(result)
   end
 
+  def follow_up
+    result = Captain::FollowUpService.new(
+      account: Current.account,
+      follow_up_context: params[:follow_up_context]&.to_unsafe_h,
+      user_message: params[:message]
```
aakashb95: what do you think about sending `conversation_display_id` from frontend here, that way we can attach followups to the same session_id in langfuse (`<account_id>_<conversation_id>`) and we will have everything that happened in a conversation in one place - assistant, copilot, editor and now follow-ups?

scmmishra: Good idea, I'll include that

scmmishra: @aakashb95 done

PR #13121, opened 2025-12-20
fix(sidebar): improve active child route matching logic

gabrieljablonski:
## Description

This PR fixes a visual bug in the sidebar where selecting a conversation inside a custom view (folder) would incorrectly highlight a different folder in the sidebar.

### The Problem
When navigating to a conversation within a custom view (e.g., folder with ID 10), the sidebar would incorrectly highlight the folder with ID 1. Similarly, folder ID 20 would highlight folder ID 2, and so on.

### Root Cause
The `activeChild` computed property in `SidebarGroup.vue` uses a fallback path matching mechanism with `startsWith()`. This caused `/custom_view/1` to incorrectly match routes like `/custom_view/10/conversations/123` because the string "10" starts with "1".

### The Fix
Updated the path matching logic to require either:
1. An exact path match, OR
2. The route path starts with the child path followed by a `/`

This ensures `/custom_view/1` only matches routes like `/custom_view/1/...` and not `/custom_view/10/...`.

### Key Changes
- Fixed path prefix matching in `SidebarGroup.vue` to use proper path boundary detection

## Type of change

## How Has This Been Tested?

| Before | After |
| --- | --- |
|image|image|

1. Create multiple custom views/folders (e.g., with IDs 1, 10, 11, 20)
2. Navigate to folder 10 and select a conversation
3. Verify that folder 10 remains highlighted in the sidebar (not folder 1)
4. Navigate to folder 1 and select a conversation
5. Verify that folder 1 is correctly highlighted
6. Repeat for other ID combinations (11, 12, 20, 21, etc.)

### DIFF
--- a/app/javascript/dashboard/components-next/sidebar/SidebarGroup.vue
+++ b/app/javascript/dashboard/components-next/sidebar/SidebarGroup.vue
@@ -158,9 +158,11 @@ const activeChild = computed(() => {
     return rankedPage ?? activeOnPages[0];
   }
 
-  return navigableChildren.value.find(
-    child => child.to && route.path.startsWith(resolvePath(child.to))
-  );
+  return navigableChildren.value.find(child => {
+    if (!child.to) return false;
+    const childPath = resolvePath(child.to);
+    return route.path === childPath || route.path.startsWith(childPath + '/');
+  });
 });
 
 const hasActiveChild = computed(() => {
--- a/app/javascript/dashboard/components-next/sidebar/SidebarGroup.vue
+++ b/app/javascript/dashboard/components-next/sidebar/SidebarGroup.vue
@@ -161,7 +161,7 @@ const activeChild = computed(() => {
   return navigableChildren.value.find(child => {
     if (!child.to) return false;
     const childPath = resolvePath(child.to);
-    return route.path === childPath || route.path.startsWith(childPath + '/');
+    return route.path === childPath || route.path.startsWith(`${childPath}/`);
   });
 });
 


iamsivin:
@gabrieljablonski Thanks for the PR. I wasn’t able to reproduce the exact issue you shared. If possible, could you share a screencast and confirm if the issue is still happening?

image

gabrieljablonski:
@iamsivin it might not be reproducible due to 10 being your first folder on the list, and it comes before `1`, which would be the incorrectly selected one due to the bug.

Try creating another one (probably ID 11 or later) so it goes to the bottom of the list, and try from there. Still reproducible on my end, testing on up-to-date `develop` branch.

https://github.com/user-attachments/assets/8e11f946-9e15-4726-9fc6-a9b583037102

app/javascript/dashboard/components-next/sidebar/SidebarGroup.vue
```diff
@@ -158,9 +158,11 @@ const activeChild = computed(() => {
     return rankedPage ?? activeOnPages[0];
   }
 
-  return navigableChildren.value.find(
-    child => child.to && route.path.startsWith(resolvePath(child.to))
-  );
+  return navigableChildren.value.find(child => {
+    if (!child.to) return false;
+    const childPath = resolvePath(child.to);
+    return route.path === childPath || route.path.startsWith(childPath + '/');
```
iamsivin: Let us use template literals.
```suggestion
    return route.path === childPath || route.path.startsWith(`${childPath}/`);
```

iamsivin (commented):
@gabrieljablonski Thanks for the quick response and for the PR. I’m able to reproduce the issue now. I’ve added a minor comment, please take a look.

**Before**
Image

**After**

Image

iamsivin (approved):

PR #13127, opened 2025-12-21
fix: improve handling of empty custom attributes list in settings

### LINKED ISSUE #13120: Custom Attributes page tabs not visible when no attributes exist for default tab (v4.9.0)
### Describe the bug

After upgrading to v4.9.0, the Custom Attributes page (/settings/custom-attributes/list) does not show the Conversation/Contact tabs if there are no attributes for the default "Conversation" tab, even if Contact attributes exist.

### To Reproduce

1. Have only contact_attribute type custom attributes (no conversation_attribute)
2. Navigate to Settings → Custom Attributes
3. Page shows "There are no custom attributes created" with no tabs visible

### Expected behavior

Tabs should always be visible so users can switch between Conversation and Contact attributes.

### Environment

Docker

### Cloud Provider

### Platform

Browser

### Operating system

_No response_

### Browser and version

_No response_

### Docker (if applicable)

_No response_

### Additional context

_No response_

gabrieljablonski:
## Description

This PR fixes a UX bug in the Custom Attributes settings page where switching tabs becomes impossible when the currently selected tab has no attributes.

Closes #13120

### The Problem
When viewing the Custom Attributes settings page, if one tab (e.g., Conversation) has no attributes, users could not switch to the other tab (e.g., Contact) which might have attributes.

### Root Cause
The `SettingsLayout` component receives `no-records-found` prop which, when true, hides the entire body content including the TabBar. Since the TabBar was inside the body slot, it would be hidden whenever the current tab had no attributes, preventing users from switching tabs.

### The Fix
- Removed the `no-records-found` and `no-records-message` props from `SettingsLayout`
- Moved the empty state message inline within the body, displayed below the TabBar
- The TabBar is now always visible regardless of whether there are attributes in the selected tab

### Key Changes
- Modified `Index.vue` to handle empty state inline while keeping TabBar accessible

## Type of change

## How Has This Been Tested?

1. Navigate to Settings > Custom Attributes
2. Ensure only one tab (e.g., Contact) has custom attributes
3. Switch to the empty tab (e.g., Conversation)
4. Verify the TabBar remains visible and the empty state message is shown
5. Switch back to the tab with attributes
6. Verify attributes are displayed correctly
7. Repeat with both tabs empty and both tabs with attributes

### DIFF
--- a/app/javascript/dashboard/routes/dashboard/settings/attributes/Index.vue
+++ b/app/javascript/dashboard/routes/dashboard/settings/attributes/Index.vue
@@ -150,8 +150,6 @@ const derivedAttributes = computed(() =>
   <SettingsLayout
     :is-loading="uiFlags.isFetching"
     :loading-message="$t('ATTRIBUTES_MGMT.LOADING')"
-    :no-records-found="!attributes.length"
-    :no-records-message="$t('ATTRIBUTES_MGMT.LIST.EMPTY_RESULT.404')"
   >
     <template #header>
       <BaseSettingsHeader
@@ -177,7 +175,7 @@ const derivedAttributes = computed(() =>
           class="max-w-xl"
           @tab-changed="onClickTabChange"
         />
-        <div class="grid gap-3">
+        <div v-if="derivedAttributes.length" class="grid gap-3">
           <AttributeListItem
             v-for="attribute in derivedAttributes"
             :key="attribute.id"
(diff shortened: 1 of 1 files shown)


pranavrajs (approved):

pranavrajs:
@gabrieljablonski Thanks for the PR, this helps. ❤️

PR #13139, opened 2025-12-23
feat(CW-6187): include headers from incoming emails

scmmishra:
Forwarded email flows (Google Groups, aliases, relays) commonly rewrite headers so the `Reply-To` header points at the list or group address. 

In Chatwoot, `MailPresenter#from` intentionally prefers Reply-To when building the `from` field because that is the correct reply target for end users. The downside is that in a forwarded flow, the original technical sender (for example `reply@idealista.pt`) gets obscured, even though it is still present in the raw headers as `X-Original-From` or `X-Original-Sender`.

https://github.com/chatwoot/chatwoot/blob/65bdf99d5627c0260bc19ec52acfc57eab367eec/app/presenters/mail_presenter.rb#L127-L130

Any inbox that receives messages via mailing lists, shared addresses, or domain forwarding can lose sender identity if only normalized `from` is preserved. Persisting a small, curated headers map (for example `X-Original-From`, `X-Original-Sender`, and `X-Forwarded-For`) in `serialized_data` keeps the original evidence without changing conversation semantics or reply behavior.

This issue came to light when a Chatwoot user was trying to process emails via webhooks. The webhook service read `content_attributes.email` from the serialized email data, but we do not persist raw forwarding headers today. The result is that `content_attributes.email.from` reflects the forwarding address (the group) and the `headers` object is missing, so downstream integrations cannot distinguish between senders that matter for routing or any other task.

> Note: A follow-up for this can be where we use the X-Original-Sender in the from address if we're able to detect that the email is from a mailing list

### DIFF
--- a/app/presenters/mail_presenter.rb
+++ b/app/presenters/mail_presenter.rb
@@ -95,6 +95,7 @@ class MailPresenter < SimpleDelegator
       content_type: content_type,
       date: date,
       from: from,
+      headers: headers_data,
       html_content: html_content,
       in_reply_to: in_reply_to,
       message_id: message_id,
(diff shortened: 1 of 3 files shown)


scmmishra:
@codex please review

sojan-official (approved):
code LGTM

PR #13150, opened 2025-12-25
fix: Double notification email after importing contacts

mazenkhalil:
Eliminate the second email notification sent to the account admins after processing the contacts import.


pranavrajs:
@mazenkhalil Thanks for the PR, this is helpful.

PR #13178, opened 2026-01-05
fix: captain assistant handoff

aakashb95:
## Description

Whenever Captain needs to handoff to human agent, it looks for `conversation_handoff` in the response json. 
We allow users to add custom instructions to the Captain assistant prompt, and many users direct Captain to let customers know they are being transferred to a human. In these scenarios (which are quite frequent), Captain doesn't say `conversation_handoff` at all and the handoff never happens.

This is because we expect `conversation_handoff` inside the `response` field. That field has now two jobs instead of one.

I suggest we let the `response` key have the actual LLM response and add a new `action` key for the next course of action decided by the LLM. If it says `continue`, Captain assistant remains active. If it says `handoff`, a user friendly response goes to the user while handoff happens behind the scenes

## Type of change

## How Has This Been Tested?

Using Specs and my Duolingo Spanish 

image

image

## Checklist:


aakashb95:
@greptileai review

aakashb95:
v2 handles this. will revisit if needed.

PR #13179, opened 2026-01-05
feat: Add call-to-action template support for Twilio

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-6228/add-call-to-action-template-support-for-twilio-whatsapp-templates

Adds support for Twilio WhatsApp call-to-action templates, enabling customers to use URL button templates with variable inputs.

CleanShot 2026-01-05 at 16 25 55@2x


sojan-official (approved):
Code LGTM

PR #13181, opened 2026-01-06
feat: allow agent bot and captain responses to reset waiting since

scmmishra:
When AgentBot responds to customer messages, the `waiting_since` timestamp is not reset, causing inflated reply time metrics when a human agent eventually responds. This results in inaccurate reporting that incorrectly includes periods when customers were satisfied with bot responses.

### Timeline from Production Data

```
Dec 12, 16:20:14 - Customer sends message (ID: 368451924)
                   ↓ waiting_since = Dec 12, 16:20:14

Dec 12, 16:20:17 - AgentBot replies (ID: 368451960)
                   ↓ waiting_since STILL = Dec 12, 16:20:14 ❌
                   ↓ (Bot response doesn't clear it)

14-day gap        - Customer satisfied, no messages
                   ↓ waiting_since STILL = Dec 12, 16:20:14 ❌

Dec 26, 22:25:45 - Customer sends new message (ID: 383522275)
                   ↓ waiting_since STILL = Dec 12, 16:20:14 ❌
                   ↓ (New message doesn't reset it)

Dec 26-27         - More AgentBot interactions
                   ↓ waiting_since STILL = Dec 12, 16:20:14 ❌

Dec 27, 07:36:53 - Human agent finally replies (ID: 383799517)
                   ↓ Reply time calculated: 1,268,404 seconds
                   ↓ = 14.7 DAYS ❌
```
## Root Cause

The core issues is in `app/models/message.rb`, where **AgentBot messages does not clear `waiting_since`** - The `human_response?` method only returns true for `User` senders, so bot replies never trigger the clearing logic. This means once `waiting_since` is set, it stays set even when customers send new messages after receiving bot responses.

The solution is to simply reset `waiting_since` **after a bot has responded**. This ensures reply time metrics reflect actual human agent response times, not bot-handled periods.

### What triggers the rest

This is an intentional "gotcha", that only `AgentBot` and `Captain::Assistant` messages trigger the waiting time reset. Automation and campaign messages maintain current behavior (no reset). This is because interactive bot assistants provide conversational help that might satisfy customers. Automation and campaigns are one-way communications and shouldn't affect waiting time calculations.

## Related Work

Extends PR #11787 which fixed `waiting_since` clearing on conversation resolution. This PR addresses the bot interaction scenario which was not covered by that fix.

Scripts to clean data: https://gist.github.com/scmmishra/bd133208e219d0ab52fbfdf03036c48a

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -322,15 +322,19 @@ class Message < ApplicationRecord
   end
 
   def update_waiting_since
-    if human_response? && !private && conversation.waiting_since.present?
-      Rails.configuration.dispatcher.dispatch(
-        REPLY_CREATED, Time.zone.now, waiting_since: conversation.waiting_since, message: self
-      )
-      conversation.update(waiting_since: nil)
-    end
+    waiting_present = conversation.waiting_since.present?
 
-    # Bot responses also clear waiting_since (simpler than checking on next customer message)
-    conversation.update(waiting_since: nil) if bot_response? && !private && conversation.waiting_since.present?
+    if waiting_present && !private
+      if human_response?
+        Rails.configuration.dispatcher.dispatch(
+          REPLY_CREATED, Time.zone.now, waiting_since: conversation.waiting_since, message: self
+        )
+        conversation.update(waiting_since: nil)
+      elsif bot_response?
+        # Bot responses also clear waiting_since (simpler than checking on next customer message)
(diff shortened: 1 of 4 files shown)


scmmishra:
@codex please review

sojan-official (approved):
Code LGTM

PR #13217, opened 2026-01-09
feat: Use amplitude for Cloud Analytics

pranavrajs:
Migrates our analytics integration on Cloud from PostHog to Amplitude. This change updates the core AnalyticsHelper class to use the Amplitude SDK while maintaining the same tracking interface. Rest of all existing analytics calls throughout the codebase continue to work without modification.

**Changes:**
- Replace PostHog analytics with Amplitude SDK
- Rename ANALYTICS_TOKEN to CLOUD_ANALYTICS_TOKEN for clarity
- Fix bug in page() method signature that was causing malformed payloads


iamsivin (approved):
LGTM

PR #13235, opened 2026-01-12
feat: new Captain Editor

scmmishra:


lib/integrations/openai/openai_prompts/reply.liquid
aakashb95, line 1: this needs to be updated. will check.

lib/captain/follow_up_service.rb
```diff
@@ -48,6 +48,7 @@ def build_follow_up_system_prompt(session_data)
       You just performed a #{action_context} action for a customer support agent.
       Your job now is to help them refine the result based on their feedback.
       Be concise and focused on their specific request.
+      Output only the reply, no preamble, tags, or explanation.
```
aakashb95, line 51: noticed preamble case while testing

Image

enterprise/config/premium_features.yml
```diff
@@ -3,5 +3,4 @@
 - audit_logs
 - response_bot
 - sla
-- captain_integration
```
scmmishra, line 6: @sojan-official I'm removing captain from premium features, since the copilot editor is allowed for all anyway. @pranavrajs increasingly it seems like having copilot on editor as a separate feature flag would be a good idea. 

CC: @aakashb95

app/javascript/dashboard/composables/useCaptain.js
```diff
@@ -8,18 +8,40 @@ import { usePolicy } from 'dashboard/composables/usePolicy';
...
-
+  const { hasPremiumEnterprise } = usePolicy();
   const captainEnabled = computed(() => {
     return isCloudFeatureEnabled(FEATURE_FLAGS.CAPTAIN);
   });
 
-  const showCopilotOnEditor = computed(() =>
-    isPremiumFeatureVisible('captain_integration')
-  );
+  const showCopilotOnEditor = computed(() => {
```
scmmishra: @pranavrajs this function determines the visibility of the copilot features on the editor

aakashb95 (approved):

PR #13236, opened 2026-01-12
fix: PDF errors not loading in CI

scmmishra:
## Problem

CI failed with `NameError: uninitialized constant CustomExceptions::PdfUploadError` in `spec/enterprise/services/captain/llm/pdf_processing_service_spec.rb`. 

Locally it often passed because the constant was already loaded (Spring/zeitwerk cache or prior spec order). That hid the autoload issue. In a clean CI boot, Zeitwerk tries to autoload `CustomExceptions::PdfUploadError` and can’t find a matching file, because the class lived in `lib/custom_exceptions/pdf_processing_error.rb` instead of a file that matches the constant name.

See: https://app.circleci.com/pipelines/github/chatwoot/chatwoot/107061/workflows/f877b180-6690-4477-990f-2dd4840116f4/jobs/123770

CleanShot 2026-01-12 at 14 17 12@2x

## Solution

PDF exceptions now live under `CustomExceptions::Pdf` in `lib/custom_exceptions/pdf.rb`, which matches Zeitwerk’s expectations. Call sites were updated to use `CustomExceptions::Pdf::UploadError` and `CustomExceptions::Pdf::FaqGenerationError`. The unused `ProcessingError` class was removed, so each PDF error inherits directly from `CustomExceptions::Base`. This makes autoloading deterministic, so CI and local runs behave the same.


scmmishra:
@codex please review

tds-1 (approved):

aakashb95 (approved):

PR #13274, opened 2026-01-15
fix: TypeError: can't access property "emitsOptions", r is null

iamsivin:
# Pull Request Template

## Description

This PR fixes a Sentry error: **`TypeError: can't access property "emitsOptions", r is null`** originating from the Vue runtime core.

### Cause

The issue was caused by a race condition triggered by an **event storm**. When multiple `InboxCard` components mounted at the same time (for example, during inbox navigation), each card executed `onBeforeMount(contextMenuActions.close)`, which emitted a `contextMenuClose` event.
This resulted in a burst of simultaneous events, causing Vue to process updates on components that were not yet fully initialized, eventually leading to null component instances when accessing `emitsOptions`.

### Solution

Updated `onBeforeMount` to reset the local context menu state directly (`isContextMenuOpen.value = false`) instead of emitting the `contextMenuClose` event.
This avoids the event storm during initialization while preserving the intended behavior and state consistency.

### Reproducibility

I wasn’t able to reproduce this issue locally. Hopefully, this fix resolves the problem.

Fixes https://linear.app/chatwoot/issue/CW-6330/typeerror-cant-access-property-emitsoptions-r-is-null

## Type of change

## Checklist:

### DIFF
--- a/app/javascript/dashboard/routes/dashboard/inbox/InboxList.vue
+++ b/app/javascript/dashboard/routes/dashboard/inbox/InboxList.vue
@@ -234,8 +234,7 @@ onMounted(() => {
       />
       <div
         ref="notificationList"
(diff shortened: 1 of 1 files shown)


iamsivin:
@codex review

PR #13282, opened 2026-01-15
fix: mark conversation open on human msg

aakashb95:
# Pull Request Template

## Description
When Captain or Agent bot are involved in the conversation, the status is PENDING. If a human comes in between and assigns it to themself, the status still remains PENDING unless they click the UI to mark it as OPEN.
Ideally, we should mark conversation as OPEN on human agent message.

## Type of change

## How Has This Been Tested?

Specs and locally

Before:
image

Midway through the fix, template messages also fired AFTER human message:
image

After: (conversation: open, template messages suppressed on human message)
image

## Checklist:


app/models/message.rb
```diff
@@ -389,6 +390,13 @@ def reopen_conversation
     reopen_resolved_conversation if conversation.resolved?
   end
 
+  def open_conversation_on_human_response
+    return unless human_response?
+    return unless conversation.pending?
```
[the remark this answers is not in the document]
aakashb95: fair play, good catch!

app/models/message.rb
```diff
@@ -310,10 +310,11 @@ def ensure_content_type
   def execute_after_create_commit_callbacks
     # rails issue with order of active record callbacks being executed https://github.com/rails/rails/issues/20911
     reopen_conversation
+    open_conversation_on_human_response
     set_conversation_activity
     dispatch_create_events
     send_reply
-    execute_message_template_hooks
+    execute_message_template_hooks unless human_response? || activity?
```
pranavrajs: there are too many if cases in the files, is it possible to simplify the logic somehow?

app/services/message_templates/hook_execution_service.rb
```diff
@@ -49,6 +49,11 @@ def email_collect_was_sent?
 
   # TODO: we should be able to reduce this logic once we have a toggle for email collect messages
   def should_send_email_collect?
+    # Don't send email collect if a human agent has already replied
+    # It breaks UX if an automated template shows up during a human conversation
+    # The agent can manually collect email if needed
+    return false if conversation.messages.outgoing.exists?(private: false, sender_type: 'User')
```
aakashb95, line 55: this code is necessary because it fixes a UX problem
email collect fires when customer sends a follow up. 

Image

app/models/message.rb
aakashb95: added `if incoming?`

aakashb95:
will revisit this later

PR #13285, opened 2026-01-15
fix: Respect survey label rules for WhatsApp CSAT template

muhsin-k:
Ensure CSAT survey label rules are evaluated once in CsatSurveyService before any channel-specific sending (including WhatsApp/Twilio templates), remove the duplicated rule check from the template builder, and cover the blocking-label scenario in service specs while simplifying the template specs accordingly.


muhsin-k:
@codex please review

sojan-official (approved):
Code LGTM

PR #13289, opened 2026-01-15
feat(ee): Review Notes for CSAT Reports

pranavrajs:
CSAT scores are helpful, but on their own they rarely tell the full story. A drop in rating can come from delayed timelines, unclear expectations, or simple misunderstandings, even when the issue itself was handled correctly.

Review Notes for CSAT let admins/report manager roles add internal-only context next to each CSAT response. This makes it easier to interpret scores properly and focus on patterns and root causes, not just numbers.

image

### Why this matters

* Capture the real context behind individual CSAT ratings
* Clarify whether a low score points to a genuine service issue or a process gap
* Spot recurring themes across conversations and teams
* Make CSAT reviews more useful for leadership reviews and retrospectives

### How Review Notes work

**View CSAT responses**
Open the CSAT report to see overall metrics, rating distribution, and individual responses.

**Add a Review Note**
For any CSAT entry, managers can add a Review Note directly below the customer’s feedback.

**Document internal insights**
Use Review Notes to capture things like:

* Why a score was lower or higher than expected
* Patterns you are seeing across similar cases
* Observations around communication, timelines, or customer expectations

Review Notes are visible only to administrators and people with report access only. We may expand visibility to agents in the future based on feedback. However, customers never see them. 

Each note clearly shows who added it and when, making it easy to review context and changes over time.


sojan-official (approved):

sojan-official (approved):
LGTM

PR #13340, opened 2026-01-22
feat: search documentation tool for reply suggestions

aakashb95:
# Pull Request Template

## Description
Adds `search_documentation` to the editor reply_suggestion feature for EE only.

## Type of change

## How Has This Been Tested?

With tool call enabled for reply_suggestion

image

Without tool call

image

## Checklist:


aakashb95:
@codex review

enterprise/app/services/captain/tools/search_reply_documentation_service.rb
```diff
@@ -0,0 +1,42 @@
+class Captain::Tools::SearchReplyDocumentationService < RubyLLM::Tool
```
aakashb95, line 1: added a new tool so this remains different from assistant

enterprise/app/services/captain/tools/base_tool.rb
```diff
@@ -1,4 +1,6 @@
 class Captain::Tools::BaseTool < RubyLLM::Tool
+  prepend Captain::Tools::Instrumentation
```
scmmishra, line 2: Do we need a separate prepend here? Can we not just bring the execute function from `enterprise/app/services/captain/tools/instrumentation.rb`

enterprise/lib/enterprise/captain/reply_suggestion_service.rb
```diff
@@ -0,0 +1,28 @@
+module Enterprise::Captain::ReplySuggestionService
+  def make_api_call(model:, messages:)
+    return super unless use_search_tool?
+
+    make_api_call_with_tools(
+      model: model,
+      messages: messages,
```
scmmishra: Where does the `messages` value come from?

lib/captain/base_task_service.rb
```diff
@@ -55,6 +56,23 @@ def make_api_call(model:, messages:)
     end
   end
 
+  def make_api_call_with_tools(model:, messages:, tools:)
```
scmmishra: We can unify this with `make_api_call` let's take tool as an optional parameter. Seems like most of the code is the same for either functions. That would be much cleaner. 

Same feedback applies to `execute_ruby_llm_request_with_tools`, but that might not be straightforward so I leave that to your discretion

scmmishra (changes requested):

lib/captain/base_task_service.rb
aakashb95: yeah i think i got slightly carried away by the demo aspect of it and tried to make it work first. let me take a look

enterprise/app/services/captain/tools/base_tool.rb
aakashb95: this was for instrumentation so other classes' execute doesn't replace this one

enterprise/lib/enterprise/captain/reply_suggestion_service.rb
aakashb95: from the perform method
```ruby
class Captain::ReplySuggestionService < Captain::BaseTaskService
  pattr_initialize [:account!, :conversation_display_id!, :user!]

  def perform
    make_api_call(
      model: GPT_MODEL,
      messages: [
        { role: 'system', content: system_prompt },
        { role: 'user', content: formatted_conversation }
      ]
    )
  end
  ```

File: enterprise/app/services/captain/tools/base_tool.rb (hunk 2 above)
Comment (aakashb95) [reply enterprise/app/services/captain/tools/base_tool.rb:2]: with prepend:
image

without prepend:
image

**Comment by scmmishra** [comment]
@codex please review

File: enterprise/app/services/captain/tools/search_reply_documentation_service.rb [hunk 5]
```diff
@@ -0,0 +1,42 @@
...
+  param :query, desc: 'Search Query', required: true
+
+  def initialize(account:, assistant: nil)
+    @account = account
+    @assistant = assistant
+    super()
+  end
+
+  def name
+    'search_documentation'
```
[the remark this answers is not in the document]
aakashb95: @scmmishra 
from rubyllm documentation:

```
The tool’s class name is automatically converted to a snake_case name used in the API call (e.g., WeatherLookup becomes weather_lookup). This is how the LLM would call it. You can override this by defining a name method in your tool class:
```

i think i can revert the self.name in other places..

lib/captain/base_task_service.rb
```diff
@@ -43,23 +43,15 @@ def make_api_call(model:, messages:, tools: [])
...
-                   execute_ruby_llm_request(model: model, messages: messages)
-                 end
-               end
-
-    # Build follow-up context for client-side refinement, when applicable
-    if build_follow_up_context? && response[:message].present?
-      response.merge(follow_up_context: build_follow_up_context(messages, response))
-    else
-      response
+    response = send(instrumentation_method, instrumentation_params) do
```
aakashb95, line 48: nice. TIL!

aakashb95:
When `search_documentation` returns no results:
image

scmmishra (approved):

PR #13371, opened 2026-01-27
feat: Handle external echo messages from native apps

muhsin-k:
When businesses use WhatsApp Business App (co-existence mode) or Instagram App or TikTok alongside Chatwoot, messages sent from the native apps were not synced properly back to Chatwoot. This left agents with an incomplete conversation history and no visibility into responses sent outside the dashboard. Additionally, if these echo messages did arrive, they appeared as "Sent by: Bot" in the UI since they had no sender, making it confusing for agents.

This PR subscribes to WhatsApp `smb_message_echoes` webhook events and routes them through the existing service with an `outgoing_echo` flag, mirroring how Instagram already handles echoes.  On the Instagram side, echo messages now also carry the `external_echo` content attribute and `delivered` status. 

On the frontend, messages with `externalEcho` are distinguished from bot messages showing a "Native app" avatar and an advisory note encouraging agents to reply from Chatwoot to maintain the service window.

CleanShot 2026-01-29 at 13 37 57@2x

Fixes https://linear.app/chatwoot/issue/CW-4204/display-messages-not-sent-from-chatwoot-in-case-of-outgoing-echo Fixes https://linear.app/chatwoot/issue/PLA-33/incoming-from-me-messages-from-whatsapp-business-app-are-not-falling


muhsin-k:
@codex please review

jvitorr:
Do we have a date for this release?

muhsin-k:
@codex please review

sojan-official:
few notes from codex:  please see if these make sense 

 - Medium: subscribe_waba_webhook now always includes smb_message_echoes. If WHATSAPP_API_VERSION is older or the WABA
    isn’t enabled for SMB echoes, Meta can reject the subscription and onboarding fails. Consider gating by API version/
    feature flag or falling back to messages only on error. app/services/whatsapp/facebook_api_client.rb:63
  - Low: Echo messages still enqueue SendReplyJob, and the only loop protection is source_id being present. If an echo
    payload ever lacks id, the job will try to send the echo back out. Consider skipping SendReplyJob for
    content_attributes['external_echo'] or adding an explicit guard in send services. app/services/whatsapp/
    incoming_message_base_service.rb:177, app/models/message.rb:310
  - Low: external_echo is set for Instagram echoes too, but the UI copy is WhatsApp‑specific (“message window”). If
    Instagram shouldn’t show this warning, gate the banner by channel or use channel‑specific copy. app/builders/messages/
    instagram/base_message_builder.rb:158, app/javascript/dashboard/components-next/message/Message.vue:536, app/
    javascript/dashboard/i18n/locale/en/conversation.json:255

  Questions / Assumptions

  - Are you expecting existing WABA installations to re‑subscribe automatically? If not, they won’t receive echo events
    until re‑onboarding.
  - Is the “message window” advisory intended to apply to Instagram echoes as well?

  Change Summary

  - Adds SMB echo handling for WhatsApp Cloud and marks echo messages as outgoing with external_echo flags, plus UI
    affordances for native‑app messages.

meaqese:
Is there any plan to support such messages for the API Channel?

sojan-official:
@meaqese The API channel already supports creating both incoming and outgoing messages, so developers can implement this behavior on their side.

sojan-official:
@jvitorr This will be included in the next release.

meaqese:
> @meaqese The API channel already supports creating both incoming and outgoing messages, so developers can implement this behavior on their side.

@sojan-official 

Yes, you’re right. However, when an outgoing message is created and sent via the native application, it is also delivered to the API webhook, so the message ends up being sent twice. Is it possible to add some kind of marker/attribute to outgoing messages so that the webhook gateway ignores them? Without using Redis or external solutions

muhsin-k:
> Please see if these make sense
> 
> * Medium: subscribe_waba_webhook now always includes smb_message_echoes. If WHATSAPP_API_VERSION is older or the WABA
>   isn’t enabled for SMB echoes, Meta can reject the subscription and onboarding fails. Consider gating by API version/

Meta ignores unrecognized subscribed_fields rather than rejecting the subscription — I've verified this during testing. Adding a fallback-on-error would be defensive but adds complexity for a scenario that doesn't happen today.

> Low: Echo messages still enqueue SendReplyJob

This is already handled.

> UI copy is WhatsApp-specific

Already addressed. The advisory says "service window" which applies to both WhatsApp and Instagram (both have messaging policy windows).

cc @sojan-official

PR #13431, opened 2026-02-03
feat: conversation summaries

aakashb95:
# Pull Request Template

## Description

Generates summaries of conversations for a quick glance for customer support agent
## Type of change

## How Has This Been Tested?

## Checklist:


aakashb95:
@pranavrajs 
one edge case: if inside an org there is agent level locale, one person english, another spanish and if org level language is english, the spanish agent will see english summary. I don't know how common this is going to be so did not address it.

aakashb95:
@pranavrajs some points codex highlighted in the frontend:

Full review comments:

- [P3] Preserve summarize abort signal compatibility — 
`app/javascript/dashboard/api/captain/tasks.js`:45-45
 Since summarize now expects an options object, existing callers that still pass an AbortSignal directly (e.g. useCaptain.summarizeConversation uses TasksAPI.summarize(conversationId, options.signal)) will no longer forward the signal. The destructuring treats the signal as the options object, so signal becomes undefined and the request can’t be aborted when users navigate away or cancel. Consider updating call sites or supporting the old signature to avoid breaking   cancellation.
- [P2] Allow retries after list summary fetch errors — `app/javascript/dashboard/components/widgets/conversation/ConversationSummary.vue`:83-85
   If the summary request fails, error stays set and the toggleSummary guard !error.value prevents any subsequent fetch attempts. Because the error isn’t cleared when collapsing/reopening, the conversation list summary gets stuck in an error state unless the page reloads. Clearing error on open or removing the error gate would let users retry after transient failures.

aakashb95:
This needs to be spec'd out. Will revisit.

PR #13432, opened 2026-02-03
perf: optimize meta endpoint with status-specific counter caches

tds-1:
## Description

Optimizes /conversations/meta endpoint by implementing status-specific counter caches, eliminating expensive COUNT(*) queries for the most common  use case.

Fixes: Meta API slowness and load on CPU 

## Type of change

## Checklist:

---

> [!NOTE]
> **Medium Risk**
> Touches database schema and introduces counter-cache maintenance on `Conversation` updates; incorrect counter updates or flag gating could lead to wrong meta counts in some scenarios.
> 
> **Overview**
> Adds status-specific conversation counters on `accounts` and wires them up via the `counter_culture` gem so counts stay updated as `Conversation.status` changes.
> 
> Optimizes the `/conversations/meta` counting path by conditionally returning `*_conversations_count` for admin users when only a status filter is applied (behind the internal `counter_cache_optimization` feature flag), falling back to live `COUNT(*)` queries for other filter combinations or `status=all`. Includes migrations to add and backfill the new counter columns.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit efc58eef9a81347d1a51c8910e780d9b78fd7c55. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


db/migrate/20260203125200_backfill_status_conversation_counters.rb
scmmishra, line 1: This migration will have to be run separately. Let's move it to a service that can be called outside and link it here. That way we can manually mark this migration as done

Also, some open questions

We might have to worry about drift, say we update the counter caches, what happens if during the processing new conversations come in (a race condition sort of situation), will that be handled. 

Also we're running this for all accounts, despite the `counter_cache_optimization` check. Is that necessary

scmmishra: Perhaps, we can try using `counter_culture_fix_counts`

scmmishra (commented):
Some points added. 

In general I don't really like the idea of DB persisted counter caches, as there are no correctness guarantees, drifts happen outside the callback, susceptible to race conditions. PG triggers would have been a better place to implement this, or redis backed that is periodically evicted. But this concern of mine is not grounded with data, so let's consider this a nitpick

This PR is good enough to go through once the open questions are addressed, it's definitely worth trying it out. In case this doesn't turn out to be good enough we can revert. I'd say let @sojan-official take a look too.

db/migrate/20260203125200_backfill_status_conversation_counters.rb
tds-1: Moved backfill to a service, deleted the backfill migration entirely. The logic lives in `Internal::BackfillConversationCountersService` which can be called manually. It's triggered automatically via an after_save callback on Account whenever counter_cache_optimization is enabled for an account.

I tried `counter_culture_fix_counts` but it requires adding a column_names hash to the counter_culture model definition (needed for the SQL reconciler since it can't invoke the Ruby proc). That felt like unnecessary model complexity for a one-time backfill. Went with a simple update loop in service instead.

PR #13435, opened 2026-02-03
fix(super-admin): render all account limit fields

sojan-official:
## Bug Explanation
- The Super Admin limits form renders inputs by iterating the keys of `account.limits`.
- When `account.limits` was present, `AccountLimitsField#to_s` returned only that hash (no defaults).
- On save, `SuperAdmin::AccountsController` compacts the limits hash, removing blank keys.
- Result: if only one key (e.g., `agents`) was saved, the other keys were missing from the hash and their fields disappeared on the next render.

## Fix
- Always start from a defaults hash of all expected limit keys and merge in any saved overrides.
- This keeps the UI stable and ensures all limit inputs remain visible even when the stored hash is partial.
- Upgraded meta_request to `0.8.5` to stop a dev‑only `SystemStackError` caused by JSON‑encoding ActiveRecord::Transaction in Rails 7.2. No production behavior changes.

## Reproduction Steps
1. In Super Admin, edit an account and set only `agents` in the limits; leave other limit fields blank and save.
2. Re-open the same account in Super Admin.
3. Observe that only `agents` is rendered and other limit fields are missing.

## Testing
- Tested on UI


muhsin-k (approved):
LGTM and verified locally

PR #13440, opened 2026-02-04
feat: allow hiding metrics from reports

scmmishra:
Some report metrics are computationally expensive to query, particularly `incoming_messages_count`, `outgoing_messages_count`, and `reply_time`. For accounts with large data volumes, these queries can take so long that they cause the entire reports API to time out. When this happens, users see no data at all rather than partial data.

This PR implements a stopgap solution to unblock affected accounts. Rather than showing users a blank reports page due to timeouts, we can now disable the problematic metrics for specific accounts. The remaining metrics will load normally, giving users access to at least some of their reporting data while we work on optimizing the underlying queries.

This is an internal configuration intended to be set by developers via Rails console, not exposed to end users.

### Implementation Notes

The filtering happens at multiple layers. On the backend, the summary endpoint skips computing hidden metrics entirely via `MetricBuilder`, avoiding the expensive database queries that cause timeouts. For other endpoints, the `Reports::MetricFilter` concern strips hidden metrics from responses. On the frontend, `WootReports.vue` filters the metric keys before making any API calls, so hidden metrics are never requested in the first place.

One subtlety worth noting: the summary builders use `avg_reply_time` as the key name, while the timeseries endpoint uses `reply_time`. The concern normalizes these via an alias map so that hiding `reply_time` in the account settings correctly hides both variants.

### Configuration

```ruby
account = Account.find(ID)
account.update!(report_hidden_metrics: ['reply_time', 'incoming_messages_count', 'outgoing_messages_count'])
```

### DIFF
--- a/app/builders/v2/reports/conversations/metric_builder.rb
+++ b/app/builders/v2/reports/conversations/metric_builder.rb
@@ -1,10 +1,5 @@
 class V2::Reports::Conversations::MetricBuilder < V2::Reports::Conversations::BaseReportBuilder
-  # Mapping from metric keys to canonical hidden metric names
-  METRIC_TO_HIDDEN_KEY = {
-    'incoming_messages_count' => 'incoming_messages_count',
-    'outgoing_messages_count' => 'outgoing_messages_count',
-    'reply_time' => 'reply_time'
-  }.freeze
+  HIDEABLE_METRICS = %w[incoming_messages_count outgoing_messages_count reply_time].freeze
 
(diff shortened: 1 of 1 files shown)


scmmishra:
@codex please review

pranavrajs:
Screenshot 2026-02-04 at 5 08 12 PM

There is something wrong with the implementation. I'm seeing Nan seconds. There's also a codex comment. Please check if it's relevant.

PR #13455, opened 2026-02-05
fix: add loading and silent retry to summary reports

scmmishra:
For large accounts, summary report queries can take several seconds to complete, often times hitting the 15-second production request timeout. The existing implementation silently swallows these failures and provides no feedback during loading. Users see stale data with no indication that a fetch is in progress, and if they interact with filters while a request is in flight, they trigger race conditions that can result in mismatched data being displayed.

This is a UX-level fix for what is fundamentally a performance problem. While the underlying query performance is addressed separately, users need proper feedback either way

## Approach

The PR adds three things: 

1. A loading overlay on the table, to provide feedback on loading state
2. Disabled filter inputs during loading so that the user does not request new information that can cause race conditions in updating the store
3. Silent retry before showing an error.

The retry exists because these queries often succeed on the second attempt—likely due to database query caching. Rather than immediately showing an error and forcing the user to manually retry, we do it automatically. If the second attempt also fails, we show a toast so the user knows something went wrong.

The store previously caught and discarded errors entirely. It now rethrows them after resetting the loading flag, allowing components to handle failures as they see fit.

#### Double Retry and Error

https://github.com/user-attachments/assets/c189b173-8017-44b7-9493-417d65582c95

#### Loading State

https://github.com/user-attachments/assets/9f899c20-fbad-469b-93cc-f0d05d0853b0


iamsivin (approved):
LGTM

iamsivin:
@scmmishra  added a fade transition to smooth out the loading overlay.


https://github.com/user-attachments/assets/fc656a13-61a9-4c99-bc53-61c2f62ee95f

PR #13469, opened 2026-02-06
fix: Silence RubyLLM legacy acts_as deprecation warning

JorgeAlan:
## Summary

- Opts in to RubyLLM's new (v1.7+) `acts_as` API by setting `use_new_acts_as = true`
- Silences the deprecation warning emitted on every Rails boot:
  ```
  !!! RubyLLM's legacy acts_as API is deprecated and will be removed in RubyLLM 2.0.0.
  ```
- Future-proofs against the removal of the legacy API in RubyLLM 2.0

## Context

Since Chatwoot does not use any of RubyLLM's ActiveRecord persistence helpers (`acts_as_chat` / `acts_as_message` / `acts_as_tool_call`), this is effectively a no-op. The only behavioral difference is that the new `ActsAs` module is included instead of `ActsAsLegacy` — but since neither is called, there is no functional impact.

The configuration is placed in `config/application.rb` before the `Application` class definition, as [recommended by the RubyLLM migration guide](https://rubyllm.com/upgrading/), to ensure it is evaluated before the railtie's `on_load :active_record` callback fires.

## Test plan

- [ ] Verify Rails boots without the deprecation warning
- [ ] Verify Captain/LLM features still work correctly (chat, tools, etc.)
- [ ] Run existing test suite

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,6 +94,7 @@ yarn-debug.log*
 .vscode
 .claude/settings.local.json
 .cursor
+.codex/
 CLAUDE.local.md
 
 # Histoire deployment
@@ -101,3 +102,4 @@ CLAUDE.local.md
 .histoire
 .pnpm-store/*
 local/
+Procfile.worktree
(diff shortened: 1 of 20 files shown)


sojan-official:
@codex review plz

aakashb95:
Hi @JorgeAlan,

Thanks for the PR.

The change is cosmetic for us and doesn’t affect behavior (the warning only appears on boot). We’ll handle this in a broader RubyLLM upgrade/migration pass closer to 2.0.

Thanks again for contributing.

PR #13489, opened 2026-02-09
fix: Captain not responding to campaign conversations

scmmishra:
Fixes: CW-6461

Customers reported that Captain would not respond when a customer replied to a campaign message, even though the conversation was in `pending` status — which is exactly when Captain should be handling it.

The root cause is a blanket early return in `HookExecutionService#perform`: `return if conversation.campaign.present?`. This guard was originally added to prevent greeting, out-of-office, and email collect templates from firing on campaign conversations (since campaigns have their own initial message). However, the enterprise Captain logic is triggered from `trigger_templates`, which is called *after* that guard — so Captain never gets a chance to run for campaign conversations at all.

The fix moves the campaign check out of `perform` and into each individual `should_send_*` method (`should_send_greeting?`, `should_send_out_of_office_message?`, `should_send_email_collect?`). This way, `trigger_templates` is still called for campaign conversations, allowing the enterprise override to schedule Captain, while each OSS template method independently decides to suppress itself. This is safe because there are two layers of protection: the enterprise `captain_handling_conversation?` check already suppresses templates when Captain is active, and the newly added `conversation.campaign.present?` guards in each method handle the OSS / no-Captain case.

### Behavior matrix for campaign conversations

| Scenario | Before | After |
|---|---|---|
| Campaign reply + Captain configured | Captain never responds (blocked by early return) | Captain responds normally |
| Campaign reply + Captain quota exceeded | Nothing happens | Handoff to agent, no OOO sent |
| Campaign reply + Captain errors/requests handoff | Nothing happens | Handoff to agent, no OOO sent |
| Campaign reply + no Captain (OSS) | No templates sent | No templates sent (unchanged) |
| Campaign reply + greeting enabled | Greeting suppressed | Greeting suppressed (unchanged) |
| Campaign reply + outside business hours | OOO suppressed | OOO suppressed (unchanged) |
| Campaign reply + email collect enabled | Email collect suppressed | Email collect suppressed (unchanged) |
| Non-campaign conversation | All templates work normally | All templates work normally (unchanged) |

### Test plan

- Captain response job is scheduled for pending campaign conversations with an incoming message
- Greeting, out-of-office, and email collect templates are still suppressed on campaign conversations (both with and without Captain configured)
- OOO is suppressed in all three handoff paths (quota-exceeded, Captain-requested, error-triggered)
- Existing template behavior for non-campaign conversations is unchanged


aakashb95 (approved):

scmmishra:
@codex please review

pranavrajs (approved):

PR #13490, opened 2026-02-09
fix: escape special characters in Linear GraphQL queries 

scmmishra:
Creating a Linear issue from Chatwoot fails with a GraphQL parse error when the title, description, or search term contains double quotes. For example, a description like `the sender is "Bot"` produces this broken query:

```graphql
issueCreate(input: { description: "the sender is "Bot"" })
```

Linear's API rejects this with `Syntax Error: Expected ":", found String`. This affects issue creation, issue linking, and issue search — any flow where user-provided text is interpolated into a GraphQL query.

The `graphql_value` helper was only escaping newlines (`\n`) but not quotes, backslashes, or other characters that are meaningful inside a GraphQL string literal. On top of that, `issue_link` and `search_issue` bypassed `graphql_value` entirely, using raw string interpolation instead.

The fix replaces the manual `gsub` escaping with Ruby's `to_json`, which produces a properly escaped, double-quoted string that handles all special characters. This is a minimal, well-understood substitution — `to_json` on a Ruby string returns a valid JSON string literal, which is also a valid GraphQL string literal since GraphQL uses the same escaping rules. The `issue_link` mutation and `search_issue` query are updated to route their parameters through `graphql_value` instead of raw interpolation.

The `team_entities_query` and `linked_issues` methods in `queries.rb` also use raw interpolation, but their inputs are system-generated IDs and URLs rather than user-provided text, so they're left as-is to keep this change focused.

Fixes https://linear.app/chatwoot/issue/CW-5548/cannot-create-a-linear-ticket

### DIFF
--- a/lib/linear/mutations.rb
+++ b/lib/linear/mutations.rb
@@ -2,13 +2,10 @@ module Linear::Mutations
   def self.graphql_value(value)
     case value
     when String
-      # Strings must be enclosed in double quotes
-      "\"#{value.gsub("\n", '\\n')}\""
+      value.to_json
     when Array
-      # Arrays need to be recursively converted
       "[#{value.map { |v| graphql_value(v) }.join(', ')}]"
     else
-      # Other types (numbers, booleans) can be directly converted to strings
       value.to_s
     end
   end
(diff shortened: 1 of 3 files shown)


scmmishra:
@codex please review

tds-1 (approved):

PR #13528, opened 2026-02-12
fix: Prevent user enumeration on password reset endpoint

### LINKED ISSUE #13527: User Enumeration via Password Recovery Screen
### Describe the bug

The application's password recovery functionality allows for the enumeration of valid users by returning distinct responses depending on whether or not the email/username entered in the system exists. When a valid email address is submitted, the application returns a different message than the one displayed when the address does not exist in the database.

### To Reproduce

1. Access any ChatWoot
2. Click for password recovery
3. Insert any email
4. The returns are different for users who exist

### Expected behavior

A commom return to not allow user enumeration

### Environment

### Cloud Provider

### Platform

None

### Operating system

All

### Browser and version

All

### Docker (if applicable)

_No response_

### Additional context

_No response_

joao-baza:
## Description

The current password reset endpoint returns different HTTP status codes and messages depending on whether the email exists in the system (200 for existing emails, 404 for non-existing ones). This allows attackers to enumerate valid email addresses via the password reset form.

### `app/controllers/devise_overrides/passwords_controller.rb`
- Removed the `if/else` branch that returned different responses based on email existence
- Now always returns a generic `200 OK` response with the same message regardless of whether the email exists
- Uses safe navigation operator (`&.`) to send reset instructions only if the user exists

### `config/locales/en.yml`
- Consolidated `reset_password_success` and `reset_password_failure` into a single generic `reset_password` key
- New message does not reveal whether the email exists in the system

## Security Impact
- **Before**: An attacker could determine if an email was registered by observing the HTTP status code (200 vs 404) and response message
- **After**: All requests receive the same 200 response with a generic message, preventing user enumeration

This follows [OWASP guidelines for authentication error messages](https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#authentication-responses).

Fixes #13527

### DIFF
--- a/app/controllers/devise_overrides/passwords_controller.rb
+++ b/app/controllers/devise_overrides/passwords_controller.rb
@@ -6,12 +6,8 @@ class DeviseOverrides::PasswordsController < Devise::PasswordsController
 
   def create
     @user = User.from_email(params[:email])
-    if @user
-      @user.send_reset_password_instructions
-      build_response(I18n.t('messages.reset_password_success'), 200)
-    else
-      build_response(I18n.t('messages.reset_password_failure'), 404)
-    end
+    @user&.send_reset_password_instructions
+    build_response(I18n.t('messages.reset_password'), 200)
   end
 
   def update
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -41,8 +41,7 @@ en:
       invalid_email: 'Please enter a valid email address'
       authentication_failed: 'Authentication failed. Please check your credentials and try again.'
   messages:
-    reset_password_success: Woot! Request for password reset is successful. Check your mail for instructions.
-    reset_password_failure: Uh ho! We could not find any user with the specified email.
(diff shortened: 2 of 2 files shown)


vishnu-narayanan (approved):
@joao-baza Thank you for the PR :)

pranavrajs:
@joao-baza @vishnu-narayanan I have a slightly different take on the security concern here. User enumeration is often brought up as concern seeing our code. However, we already have rate limits in place for IP as well as email address (check rack_attack.rb) and it makes large-scale probing difficult in practice. 

From a UX perspective, this error message could actually help legitimate users. A more common issue is forgetting which email variant was used to sign in. For example, I often use email+tag@company.com to differentiate between the accounts and removing the spammers. If I try a different variation and get no useful signal, it becomes harder to recover access.

I agree with getting the security practices right, but I also think we should balance that with real-world usability.

PR #13533, opened 2026-02-13
fix: deleted inboxes reappearing after page refresh

scmmishra:
When an inbox is deleted, it disappears from the UI immediately — but reappears if the user refreshes the page. The inbox stays visible until someone manually resets the cache. This PR fixes the two caching gaps that cause this.

### Why this happens

Chatwoot uses a two-layer cache for inboxes: a Redis-backed cache key (timestamp) on the backend, and IndexedDB on the frontend. On page load, the frontend compares its local cache key against the backend's. If they match, it trusts the IndexedDB data and skips the network fetch.

The delete flow has two gaps in this system. First, the controller queues inbox deletion via `DeleteObjectJob.perform_later` on the `:low` queue but never updates the Redis cache key itself — it relies on the `after_commit :destroy` hook on the Inbox model, which only fires when the background job eventually runs `object.destroy!`. Until then, the cache key is stale and the frontend considers its local data valid. Second, the frontend delete action removes the inbox from the Vuex store (in-memory) but never touches IndexedDB. So on refresh, the stale IndexedDB data is loaded back into state.

These two gaps compound: the backend says the cache is still valid (because the key hasn't changed), and the frontend has stale data ready to serve (because IndexedDB was never cleared).

### What this PR does

On the frontend, after a successful delete API call, we now call `clearDataInCache()` to wipe the inbox data from IndexedDB. This is a new method on `CacheEnabledApiClient` that clears the IndexedDB object store for the given model. On the next page load, `getFromCache()` finds an empty local store and falls through to `refetchAndCommit()`, fetching fresh data from the API regardless of cache key state.

### DIFF
--- a/app/javascript/dashboard/components/widgets/WootWriter/Editor.vue
+++ b/app/javascript/dashboard/components/widgets/WootWriter/Editor.vue
@@ -830,7 +830,13 @@ useEmitter(BUS_EVENTS.INSERT_INTO_RICH_EDITOR, insertContentIntoEditor);
 </script>
 
 <template>
-  <div ref="editorRoot" class="relative w-full">
+  <div
+    ref="editorRoot"
+    class="relative w-full"
+    :class="{
+      'opacity-50 cursor-not-allowed pointer-events-none': disabled,
+    }"
+  >
     <TagAgents
       v-if="showUserMentions && isPrivate"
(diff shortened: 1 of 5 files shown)


iamsivin:
@codex review

PR #13560, opened 2026-02-17
fix: call authorization_error! on IMAP auth failures

tds-1:
## Notion document
https://www.notion.so/chatwoot/Email-IMAP-Issue-30aa5f274c928062aa6bddc2e5877a63?showMoveTo=true&saveParent=true

## Description

PLAIN IMAP channels (non-OAuth) were silently retrying failed authentication every minute, forever. When credentials are wrong/expired, Net::IMAP::NoResponseError was caught and logged but channel.authorization_error! was never called — so the Redis error counter never incremented, reauthorization_required? was never set, and admins were never notified. OAuth channels already had this handled correctly via the Reauthorizable concern.
Additionally, Net::IMAP::ResponseParseError (raised by non-RFC-compliant IMAP servers) was falling through to the StandardError catch-all, flooding
Estimated impact before fix: ~70–75 broken IMAP inboxes generating ~700k–750k wasted Sidekiq jobs/week.

## Type of change

## Checklist:


vishnu-narayanan (commented):
@tds-1   Couple of questions

Channel::Email already includes the Reauthorizable concern, which has a Redis-based error counter and prompt_reauthorization! flow. It looks like the core bug is that channel.authorization_error! is never called in the IMAP_EXCEPTIONS rescue, so the existing mechanism never kicks in.

Was there a specific reason the new DB-based backoff was needed over wiring up the existing Reauthorizable flow? For example, something like:

```
rescue *ExceptionList::IMAP_EXCEPTIONS => e
Rails.logger.error "Authorization error for email channel - #{channel.inbox.id} : #{e.message}"
channel.authorization_error!

```

I might be missing context from the Notion doc (don't have access).

A couple of other things I noticed:

- Net::IMAP::NoResponseError (the actual auth failure when LOGIN is rejected) is currently in the second rescue block alongside transient errors like EOFError and
ResponseParseError. Might be worth moving it into IMAP_EXCEPTIONS (or a separate auth-specific list) so it gets the right treatment.
- The second rescue block applies apply_imap_backoff! to transient network/protocol errors too - wouldn't a temporary server issue or malformed response escalate towards ting reauthorization?

tds-1:
@vishnu-narayanan 
The DB-based backoff is intentional — auth failures from some providers are transient (rate limiting), and `authorization_error!`. The 5-retry exponential backoff gives it room to self-heal before escalating to `prompt_reauthorization!`.

**`NoResponseError` placement** — moved it into `IMAP_EXCEPTIONS`. 

**Transient errors escalating to reauth** — keeping `apply_imap_backoff!` for transient errors too. These errors (SSL misconfiguration, malformed responses) are often persistent, not truly transient — removing the backoff would just hammer the server on every scheduled run with no throttling and no escalation. The exponential backoff gives ~30 min to self-heal before `prompt_reauthorization!` fires. `clear_imap_backoff!` on the next successful fetch resets the counter, so genuinely brief outages don't accumulate.

vishnu-narayanan:
@tds-1 Thanks for the context. A couple more comments 

1. NoResponseError isn't always an auth error

The first comment from codex seems valid. 

`Net::IMAP::NoResponseError` is raised for any IMAP NO response, not just authentication failures. A NO on SELECT, SEARCH, or FETCH is a server/mailbox error, not bad credentials. With the current approach (catching at the job level), we can't distinguish them.

A more precise approach would be to catch it at the source, the authenticate call in build_imap_client:

```
  def build_imap_client
    imap = Net::IMAP.new(channel.imap_address, port: channel.imap_port, ssl: true)
    begin
      imap.authenticate(authentication_type, channel.imap_login, imap_password)
    rescue Net::IMAP::NoResponseError => e
      raise Imap::AuthenticationError, e.message
    end
    imap.select('INBOX')
    imap
  end
```
Then the job can rescue Imap::AuthenticationError specifically and call authorization_error! only for actual auth failures, while other NoResponseError cases (mailbox issues, server errors) stay in the transient error bucket.

2. Consider building backoff into Reauthorizable

Reauthorizable is used by 7 consumers (Email, WhatsApp, Facebook, Instagram, TikTok, Integrations::Hook, AutomationRule). The rate-limiting problem isn't unique to IMAP, any of these channels could benefit from backoff before escalation. Adding it to the concern (Redis-based, no migration needed) would avoid a parallel DB mechanism and benefit all channels:
```
  # In Reauthorizable
  def authorization_error!
    ::Redis::Alfred.incr(authorization_error_count_key)
    if authorization_error_count >= self.class::AUTHORIZATION_ERROR_THRESHOLD
      prompt_reauthorization!
    elsif respond_to?(:backoff_enabled?) && backoff_enabled?
      wait = BACKOFF_BASE_SECONDS * (2**authorization_error_count)
      ::Redis::Alfred.setex(backoff_key, wait, (Time.current + wait).to_i)
    end
  end
```
      
Channels opt in with def backoff_enabled? = true. Redis TTL handles expiration automatically, no migration needed.

3. `backoff` check missing from the parent job

FetchImapEmailInboxesJob#should_fetch_emails? already checks `reauthorization_required?` before enqueuing, but doesn't check `in_backoff?`. So during a backoff window, the scheduler will enqueue a job every minute.

tds-1:
@codex please review

tds-1:
@codex please review

vishnu-narayanan (approved):
@tds-1 

Looks good, approving.

Two items:

1. Please add specs for Backoffable (backoff ramp, exhaustion, clear) and update FetchImapEmailsJob specs for the new rescue logic. reauthorizable_shared.rb is a good pattern to follow.
2. Follow-up suggestion for another PR: consider adding a last_synced_at column on channel_email updated on successful fetch, and persisting connection/auth status in DB so it survives Redis flushes. Would make it easy to spot stale inboxes and could tie into an inbox health view similar to something we have with whatsapp/meta inboxes.

PR #13573, opened 2026-02-18
feat: add per-webhook secret with backfill migration

scmmishra:
Chatwoot's outgoing webhooks send plain JSON with no authentication headers. Receivers have no way to verify that a payload actually came from Chatwoot—anyone who discovers an endpoint URL can send a forged request that looks identical to a real one. Similarly, an attacker who captures a legitimate webhook can replay it later and the receiver has no basis to reject it. This makes building secure integrations on top of Chatwoot unnecessarily difficult: teams either skip verification entirely or bolt on fragile workarounds like IP allowlisting.

This PR gives receivers two standard primitives to solve both problems.

## What's added

Every outgoing webhook now carries three additional headers:

- `X-Chatwoot-Signature: sha256=<hex>` — an HMAC-SHA256 of the request body, signed with a per-webhook secret. Receivers can recompute the signature and reject anything that doesn't match.
- `X-Chatwoot-Timestamp: <unix seconds>` — the time the request was sent. The timestamp is included in the signed string (not just added as a header), so an attacker can't swap in a fresh timestamp without invalidating the signature. Receivers should reject requests where the timestamp is more than ~5 minutes old.
- `X-Chatwoot-Delivery: <uuid>` — a stable identifier for each delivery attempt. Sidekiq retries reuse the same UUID (generated at enqueue time, not inside `perform`), so receivers can deduplicate without treating a retry as a new event.

Each webhook record gets its own secret, exposed via the API so account owners can configure their receivers.

CleanShot 2026-02-18 at 16 02 26@2x

## Implementation notes

**Secret generation and storage** follow the same pattern already used throughout the codebase (`hmac_token` on `Channel::WebWidget`, `access_token` on `Channel::Api`, etc.): `has_secure_token` generates a random Base58 token on `before_create`, and `encrypts :secret` is applied conditionally on `Chatwoot.encryption_configured?` to match how every other sensitive column is handled.

**The backfill migration** uses `update_column` rather than a normal `update` call. This is intentional: `update` would go through Active Record encryption and write a ciphertext that plaintext-only installs (those without encryption keys configured) couldn't read back. `update_column` writes raw values that are readable everywhere, and `support_unencrypted_data: true` (already set globally in `config/application.rb`) ensures they continue to work correctly once encryption is enabled later.

**The signed string is `"#{timestamp}.#{body}"`**, not just `body`. This is the Stripe convention and it matters: if the timestamp were only an unsigned header, an attacker could replay an old request by replacing it with a current one while keeping the original signature. By including the timestamp in the HMAC input, the signature covers both the payload and when it was sent.

**`Channel::Api` inboxes** receive a `X-Chatwoot-Delivery` header but no signature. They don't have a model-backed secret (they use a shared `hmac_token` for a different purpose), and extending that here felt out of scope. The delivery ID alone still gives those receivers idempotency.

## Receiver verification

<details>
<summary>Ruby</summary>

```ruby
ts  = request.headers['X-Chatwoot-Timestamp'].to_i
sig = request.headers['X-Chatwoot-Signature']

raise "Request too old" if (Time.now.to_i - ts).abs > 300

body     = request.body.read
expected = "sha256=#{OpenSSL::HMAC.hexdigest('SHA256', webhook_secret, "#{ts}.#{body}")}"

raise "Invalid signature" unless ActiveSupport::SecurityUtils.secure_compare(expected, sig)
```

</details>

<details>
<summary>Python</summary>

```python
import hashlib
import hmac
import time

def verify_webhook(request, secret: str) -> None:
    ts  = int(request.headers["X-Chatwoot-Timestamp"])
    sig = request.headers["X-Chatwoot-Signature"]

    if abs(time.time() - ts) > 300:
        raise ValueError("Request too old")

    body = request.get_data()
    signed_payload = f"{ts}.{body.decode()}"
    expected = "sha256=" + hmac.new(
        secret.encode(), signed_payload.encode(), hashlib.sha256
    ).hexdigest()

    if not hmac.compare_digest(expected, sig):  # constant-time comparison
        raise ValueError("Invalid signature")
```

</details>

<details>
<summary>Go</summary>

```go
import (
    "crypto/hmac"
    "crypto/sha256"
    "encoding/hex"
    "errors"
    "fmt"
    "io"
    "math"
    "net/http"
    "strconv"
    "time"
)

func verifyWebhook(r *http.Request, secret string) error {
    tsStr := r.Header.Get("X-Chatwoot-Timestamp")
    sig   := r.Header.Get("X-Chatwoot-Signature")

    ts, err := strconv.ParseInt(tsStr, 10, 64)
    if err != nil {
        return errors.New("missing or invalid timestamp")
    }
    if math.Abs(float64(time.Now().Unix()-ts)) > 300 {
        return errors.New("request too old")
    }

    body, err := io.ReadAll(r.Body)
    if err != nil {
        return err
    }

    mac := hmac.New(sha256.New, []byte(secret))
    mac.Write([]byte(fmt.Sprintf("%d.%s", ts, body)))
    expected := "sha256=" + hex.EncodeToString(mac.Sum(nil))

    if !hmac.Equal([]byte(expected), []byte(sig)) {
        return errors.New("invalid signature")
    }
    return nil
}
```

</details>

## Future work

Two webhook surfaces are not covered by this PR. Agent bot webhooks and `Channel::Api` inbox webhooks both currently lack a model-backed secret, so they can only receive the `X-Chatwoot-Delivery` header. The right fix for each is to introduce a dedicated secret field on `AgentBot` and `Channel::Api` respectively — the signing logic in `Webhooks::Trigger` already accepts a `secret:` kwarg, so wiring them up would be straightforward once the secrets exist.


sojan-official:
@codex review plz

sojan-official (approved):
@scmmishra The backend changes look good to me. Please proceed with the final round of testing and move forward with the merge.

@tds-1 @muhsin-k Could one of you also take a quick look when you get a chance? That would be helpful.

tds-1 (approved):

Issue #9809, opened 2024-07-19
Whatsapp Cloud: Unable to receive images or attachments (500)


rgon:
### Describe the bug

When sending a whatsapp message with a photo or any other attachment, Chatwoot is unable to download it, after being triggered by the webhook. The `Webhooks::WhatsappEventsJob` is enqueued, but errors out with the following traceback:

<details>
<summary>Error log @ photo message webhook:</summary>
<pre>
rails-1     | I, [2024-07-19T18:52:52.302242 #1]  INFO -- : [cff9be71-2068-4d1b-9d9e-ad6d9d6da8fb] Started POST "/webhooks/whatsapp/+MyPhoneNumber" for my.ip.address.xx at 2024-07-19 18:52:52 +0000
rails-1     | I, [2024-07-19T18:52:52.311326 #1]  INFO -- : [cff9be71-2068-4d1b-9d9e-ad6d9d6da8fb] Processing by Webhooks::WhatsappController#process_payload as */*
[...]

PR #13610, opened 2026-02-21
fix(whatsapp): resilient media download via Graph API with timeouts and error handling

mikelito859:
# Fix: Resilient WhatsApp Cloud API media download

## Summary

Makes incoming WhatsApp Cloud media (image, document, audio, video) handling resilient to intermittent download failures (connection reset, 500) so that the message is still created and only the attachment is skipped when the download fails.

## Problem

When a contact sends an image or other attachment via WhatsApp Cloud, the webhook triggers `Webhooks::WhatsappEventsJob` and `Whatsapp::IncomingMessageWhatsappCloudService#download_attachment_file`. The file is downloaded from the URL returned by the Graph API (lookaside.fbsbx.com). This download can intermittently fail with:

- **`Down::ConnectionError`** (e.g. "Connection reset by peer")
- **`Down::ServerError`** (500 Internal Server Error)

When that happens, the job raises and the **entire message** is never created in Chatwoot, so the conversation shows no trace of the contact's message.

Related context: [Meta Cloud API Media docs](https://developers.facebook.com/docs/whatsapp/cloud-api/reference/media/) — the download URL is temporary and the endpoint can be unreliable under load.

## Solution

1. **Explicit Graph API flow**  
   Obtain the download URL only via `GET /{media_id}` on the Graph API. We do not use any `url` field from the webhook payload (ephemeral and not intended for backend use).

2. **Timeouts**  
   Use `open_timeout: 20` and `read_timeout: 20` on the download request to avoid hanging on slow or stuck connections.

3. **Graceful degradation**  
   Rescue `Down::Error` and `IOError` in `download_attachment_file`, log a warning with `media_id`, and return `nil`. The base service already treats a blank result as "no attachment": the **message is still created**, only the attachment is missing. The conversation continues without breaking the job.

4. **Payload key handling**  
   Support both symbol and string keys for `attachment_payload["id"]` / `attachment_payload[:id]` for robustness with different param structures.

5. **Documentation**  
   Comments clarify that the Cloud API requires the `Authorization` header on the final download request and that we expect the full attachment hash from the base service.

## Testing

- [ ] Receive an image via WhatsApp Cloud → message and attachment appear.
- [ ] (Optional) Simulate or wait for a download failure → message is created, attachment missing; job does not raise; warning is logged.

## Refs

- Same code path as **#9809** (there the cause was token permissions; here we harden against transient network/server errors).

### DIFF
--- a/.github/workflows/build-fix-whatsapp-media.yml
+++ b/.github/workflows/build-fix-whatsapp-media.yml
@@ -1,5 +1,5 @@
 # Build and push Chatwoot image with WhatsApp media download fix.
-# Image: mikelito859/chatwoot:fix-whatsapp-media-download
+# Image: mikelito85/chatwoot:fix-whatsapp-media-download
 #
 # Required repo secret: DOCKERHUB_TOKEN (Docker Hub Access Token from https://hub.docker.com/settings/security)
 
@@ -12,7 +12,7 @@ on:
   workflow_dispatch:
 
 env:
-  DOCKER_REPO: mikelito859/chatwoot
+  DOCKER_REPO: mikelito85/chatwoot
   DOCKER_TAG: fix-whatsapp-media-download
 
 jobs:
@@ -33,7 +33,7 @@ jobs:
       - name: Login to Docker Hub
         uses: docker/login-action@v3
         with:
-          username: mikelito859
+          username: mikelito85
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
       - name: Build and push
--- a/.github/workflows/build-fix-whatsapp-media.yml
+++ b/.github/workflows/build-fix-whatsapp-media.yml
@@ -1,0 +1,46 @@
+# Build and push Chatwoot image with WhatsApp media download fix.
+# Image: mikelito859/chatwoot:fix-whatsapp-media-download
+#
+# Required repo secrets: DOCKERHUB_USERNAME (mikelito859), DOCKERHUB_TOKEN (Docker Hub access token).
+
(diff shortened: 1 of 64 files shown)


muhsin-k:
@mikelito859 Could you please fix conflicts?

Administration-LGL:
Hi @mikelito859 — drive-by community review, not a maintainer.

The Ruby/Vue change in `incoming_message_whatsapp_cloud_service.rb` etc. looks reasonable on the merits (Down.download with timeouts and a fallback bubble — nice). Two things I'd suggest cleaning up before this can land:

1. **`.github/workflows/build-fix-whatsapp-media.yml`** appears to be your personal CI for building a Docker image to your own Docker Hub (`mikelito85/chatwoot:fix-whatsapp-media-download` using `secrets.DOCKERHUB_TOKEN`). It triggers on `push` to a specific branch name, and it's not `pull_request_target`, so it doesn't pose a real CI-poisoning risk for upstream — but it's also not something upstream wants to ship. Worth removing from this PR.
2. `Rails.logger.info("...body=#{response.body}")` in the service logs the full Graph API response body. The body contains signed media URLs that are short-lived but still credentials-shaped. I'd drop the `body=` part of the log line.
3. There's an unrelated test refactor in `spec/enterprise/services/enterprise/billing/topup_checkout_service_spec.rb` that should probably go in a separate PR.

Otherwise the actual fix looks good — would be a shame for it to bounce off review for the workflow file.

---

*(This review was prepared with the assistance of an AI tool as part of a broader audit of open pull requests.)*

PR #13624, opened 2026-02-24
fix(slack): Show correct sender name and avatar for Slack replies

muhsin-k:
When an agent replied to a customer conversation from Slack, the message appeared as "Bot" with a bot avatar in both the dashboard and the chat widget because the system tried to match the Slack user's email against Chatwoot accounts when no match was found, sender was saved as nil and both UIs treated any nil-sender outgoing message as a bot. 

This PR fixes the issue by storing the Slack user's display name and profile image in additional attributes when no Chatwoot account match is found, and updating both the dashboard and widget to use those values before falling back to "Bot".

 | Scenario | Before | After |
  |---|---|---|
  | Slack email matches Chatwoot agent | Shows agent name ✓ | Shows agent name ✓ |
  | Slack email doesn't match | Shows "Bot" ✗ | Shows Slack display name + avatar ✓ |


muhsin-k:
@codex please review

scmmishra (approved):

PR #13644, opened 2026-02-25
feat: add campaign context to Captain v2 prompts

scmmishra:
When a conversation originates from a campaign (e.g., a WhatsApp broadcast or website trigger), `conversation.campaign_id` links to a `Campaign` record. Until now, Captain v2's agent runner had no awareness of this — the AI couldn't tell whether a customer was responding to a specific campaign message or starting a fresh conversation.

This adds campaign metadata (title, type, message, description) to the prompt context so the AI agent can tailor its responses accordingly. For example, if a customer replies to a "Summer Sale" broadcast, the agent now knows what offer was sent and can respond in context instead of asking the customer to repeat themselves.

The data flows through the same path as conversation and contact context: `build_state` extracts it from the association, `agentable` merges it into the template context, and a new `campaign.liquid` snippet renders it into the prompt. The snippet is guarded on `campaign.id` rather than `campaign` because an empty hash is truthy in Liquid, which would render an empty campaign block for non-campaign conversations.

Fixes: AI-123


aakashb95:
code LGTM!

aakashb95 (approved):

aakashb95:
gpt-5.3-codex high found this:

Findings


[P1] Campaign context will not render; prompt emits Liquid errors instead.
The new campaign path uses {% render 'campaign' %} in both templates [assistant.liquid:24](app://-/index.html#), [scenario.liquid:24](app://-/index.html#), but Captain::PromptRenderer still renders without a Liquid file system/registers [prompt_renderer.rb:8](app://-/index.html#).
I verified this on the review branch with rails runner: output contains Liquid error: This liquid context does not allow includes. (three times when campaign is present), so the feature goal (“include campaign in prompt”) is not achieved in runtime output.

aakashb95:
image

So the prompt renderer issue looks to be pre-existing. will push a fix.

Issue #11013, opened 2025-03-03
Issue with image loading in the chat


ivanmirandaf:
### Describe the bug

When sending or receiving an image in the chat, the image does not load correctly in the conversation. The image content is only displayed after manually refreshing the page.

![Image](https://github.com/user-attachments/assets/e26dc90f-88f6-424d-8198-9a45c61effe1)

### To Reproduce

1. Send/Receive an image through the chat.
2. Observe that the image is not loaded in the conversation.
3. Manually refresh the page to view the image.

### Expected behavior

The image should load and display automatically in the conversation as soon as it is sent or received, without the need to refresh the page.

### Environment

Docker

### Cloud Provider

Other [please specify in the description]

### Platform

Browser

### Operating system

Ubuntu 20.04.6 LTS

### Browser and version

Google Chrome Versão 133.0.6943.142 (Versão oficial) 64 bits

### Docker (if applicable)

- Docker Engine - Community
Version:    28.0.1
- Docker Compose (Docker Inc.)
Version:  v2.33.1

### Additional context

- I am using the latest version of Chatwoot.

- File storage is managed by MinIO.

- When analyzing the browser requests, I noticed that Chatwoot returns a 404 error when trying to load the image from the MinIO URL. However, if the same URL is copied and pasted directly into the browser, the image loads normally.
![Image](https://github.com/user-attachments/assets/0ac062f6-2f15-4072-b465-858c8004c491)

gkp1:
edit: Fix made by another person: https://pastebin.com/0cLkzkpd (I applied this and built the docker image and it did not work)


- Basically, I need to upgrade from `4.2.0 to latest`, but v3 UI was apparently removed in latest version, and I'd be forced to use v4 UI, while images and audios fail to load when we're chatting with a user, which makes it sort of unusable 


This has been present in ' V4' since 4.1.0 when I started using chatwoot. This has left me stuck in the V3 interface. I use cloudflare R2 with 's3_compatible' env vars, works normally in chatwoot v3, but on chatwoot v4 it thinks _every media_ that arrives does not exist (404), then you need to switch conversations and switch back. or press f5.

Same thing for Audios or any file. Audios don't load until you f5. It is unusable unless I stay on v3

To Reproduce error:

1. use cloudflare r2 in s3_compat .env vars
2. open a conversation screen
3. receive a media message (image) from contact
[...]

PR #13653, opened 2026-02-25
fix: Prevent AudioTranscriptionJob from crashing on OpenAI 401 error

eloijrseganfredo:
Describe the bug
In v4.8.0, when an audio message is received, the system enqueues Messages::AudioTranscriptionJob even if OpenAI and Captain are disabled. This causes a Faraday::UnauthorizedError (401) which crashes the Sidekiq job and breaks the pipeline for that message.

To Reproduce
Disable OpenAI/Captain integrations.

Send an audio message to an inbox.

Check Sidekiq logs and observe the 401 crash in AudioTranscriptionService.

What this PR does
Adds a rescue Faraday::UnauthorizedError block inside AudioTranscriptionService#perform. Instead of crashing the worker, it logs a warning and gracefully exits, allowing the job to complete successfully.

Note: This fixes the backend crash. However, there is still a frontend reactivity issue where the audio player UI requires an F5 to load the media, which has been reported in Issue #11013.


sojan-official:
@aakashb95  could you take a look ?

aakashb95:
Minor nit: 
> To Reproduce
Disable OpenAI/Captain integrations.

We need to add an incorrect key to reproduce this. That said, LGTM. Tested locally.

aakashb95 (approved):

eloijrseganfredo:
> Minor nit:
> 
> > To Reproduce
> > Disable OpenAI/Captain integrations.
> 
> We need to add an incorrect key to reproduce this. That said, LGTM. Tested locally.

Thanks for the review and approval, @aakashb95! Glad I could help fix the backend crash.

Just a quick follow-up/reminder regarding the frontend side of this scenario: while this PR successfully keeps the Sidekiq worker alive, the UI audio player still freezes at 00:00 (requiring a manual F5) when the message arrives. This seems to be a race condition between the message.created WebSocket event and ActiveStorage finishing the file processing.

I left more details about this UI behavior on Issue #11013. Hopefully, the team can take a look at the Vue components for the upcoming releases so the audio loads dynamically. Thanks again!

PR #13662, opened 2026-02-26
fix(super-admin): prefill confirmed_at in new user form

sojan-official:
On self-hosted instances without email configured, users created from Super Admin can get stuck in an unconfirmed state. This PR implements the default at the Super Admin frontend form layer, not in backend creation logic.

What changed:
- Added a custom `ConfirmedAtField` for Super Admin user forms.
- Prefills `confirmed_at` with current time on the **New User** form (`GET /super_admin/users/new`).
- Kept backend create behavior unchanged (`resource_class.new(resource_params)`), so API/manual payloads still behave normally.

Behavior:
- In Super Admin UI, `confirmed_at` is prefilled by default.
- If someone wants an unconfirmed user, they can clear the `confirmed_at` field before saving.
- If `confirmed_at` is omitted from payload entirely, the created user remains unconfirmed.

Scope note: external signup flows are intentionally unchanged in this PR (`/api/v1/accounts`, `/api/v2/accounts`, and social/omniauth signup behavior are not modified).

## Demo 

https://github.com/user-attachments/assets/436abbb0-d4cf-49a6-a1b8-4b6aa85aa09f


sojan-official:
@codex review

pranavrajs (approved):

PR #13665, opened 2026-02-26
feat(facebook): Mark Messenger native-app echoes as external echo message

muhsin-k:
When agents send replies from the native Facebook Messenger app (not Chatwoot), echo events were created without external_echo metadata and could be misrepresented in the UI. This change updates Messenger echo message creation to:

- set content_attributes.external_echo = true for outgoing_echo messages
- set echo message status to delivered
- keep sender as nil for echo messages (existing behavior)

CleanShot 2026-02-26 at 16 32 04@2x


muhsin-k:
@codex please review

sojan-official (approved):
code LGTM

PR #13674, opened 2026-02-27
fix: Normalize URL before executing HTTP request

akylkb:
Support for Unicode characters, such as Cyrillic

### DIFF
--- a/enterprise/lib/captain/tools/http_tool.rb
+++ b/enterprise/lib/captain/tools/http_tool.rb
@@ -13,9 +13,10 @@ class Captain::Tools::HttpTool < Agents::Tool
 
   def perform(tool_context, **params)
     url = @custom_tool.build_request_url(params)
+    normalized_url = Addressable::URI.parse(url).normalize.to_s
     body = @custom_tool.build_request_body(params)
 
-    response = execute_http_request(url, body, tool_context)
+    response = execute_http_request(normalized_url, body, tool_context)
     @custom_tool.format_response(response.body)
   rescue StandardError => e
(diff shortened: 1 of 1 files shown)


nestordavalos:
Hi, I reviewed this PR and noticed that the case it's trying to solve is already covered in the current codebase.

`SafeEndpointValidatable` has a [`validate_no_unicode_chars`](https://github.com/chatwoot/chatwoot/blob/develop/enterprise/app/models/concerns/safe_endpoint_validatable.rb#L78-L83) method that rejects URLs with non-ASCII characters in the hostname at save time. A domain with Cyrillic characters would never reach `execute_http_request`.

The only scenario where Unicode could appear in the final URL would be through Liquid templates (when the LLM injects values into path segments), but that's already handled by the `rescue StandardError` in [`perform`](https://github.com/chatwoot/chatwoot/blob/develop/enterprise/lib/captain/tools/http_tool.rb#L20-L22) without causing a crash.

Could you maybe share a concrete use case where this actually fails? That way we can better understand if there's a scenario I'm not considering.

CC: [aakashb95](https://github.com/aakashb95)

sojan-official:
Thanks for the contribution!

From the current code it looks like URLs with non-ASCII characters in the hostname are already rejected at save time via `validate_no_unicode_chars`, so they shouldn’t reach the HTTP execution path.

Could you share a concrete scenario where this fails today? If possible, it would help to include:

- Reproduction steps  
- Example URL / payload that triggers the issue  
- Expected vs actual behavior

We’ll close this for now, but if you’re able to provide a reproducible case we’d be happy to revisit it.

Also, please take a look at our contribution guidelines for future PRs:  
https://developers.chatwoot.com/contributing-guide

PR #13675, opened 2026-02-27
fix: retry loading audio attachments

muhsin-k:
Fixes https://linear.app/chatwoot/issue/CW-4112/issue-with-image-loading-in-the-chat

This PR extends the `useLoadWithRetry` composable (introduced in #12873 for images) to also handle audio attachments.


muhsin-k:
@codex please review

iamsivin:
@codex review?

iamsivin:
@codex review?

filipeclacerda:
While investigating #14644, I reproduced a complementary playback-state problem in `message/chips/Audio.vue` on develop at `e87cea84d2e1728c4395f1d6f97cfb63c1623e60`. The same `playOrPause` block is unchanged in this PR's current head (`1a16b2901b908c210964e68215f33817b08fd820`): `play()` is called without handling its promise, and `isPlaying` is set immediately.

I tested the **unmodified develop component** in an isolated Vue/Chromium harness with a generated WAV and local HTTP redirects shaped like Active Storage URLs:

- Start the valid WAV, then call the underlying media element's `pause()` (simulating an external media control). The element reports `paused: true`, but the chip still shows the pause icon.
- Delay the WAV response, click Play, then immediately Pause. Chromium produces an unhandled `AbortError: The play() request was interrupted by a call to pause()`.
- Inject a rejected `play()` promise with `NotAllowedError`. It is unhandled and the chip still shows the pause icon even though the element remains paused. This case is deliberately fault-injected, not a claim that ordinary user-initiated playback is blocked.
- With a 404 response, the nested `source` emits an error; clicking Play leaves the player at 00:00 with the pause icon. This loading case overlaps your retry work, so I would coordinate it rather than duplicate it.

I can contribute a small fix and regression tests for rejected/interrupted play requests and native playback-state synchronization, preserving pending-start cancellation, the existing Active Storage URL flow, transcripts, and downloads. The reproduction is component-level; I have not run full-stack Chatwoot E2E or upstream CI.

Would you prefer this as a contribution to the existing work or a separate follow-up against develop linked to #14644? Following the contributing guide, I am waiting for scope agreement/issue assignment before implementing the fix; I have not opened a competing PR.

PR #13691, opened 2026-03-02
fix: Handle Facebook reel attachment type

muhsin-k:
Fixes https://linear.app/chatwoot/issue/PLA-96/argumenterror-reel-is-not-a-valid-file-type-argumenterror 
Fixes a crash in `when a user shares a Facebook reel via Messenger. Facebook sends these attachments with `type: "reel"`, which isn't a valid `file_type` enum value on the Attachment model (only `ig_reel` exists, matching Instagram's format).


muhsin-k:
@codex please review

sojan-official:
@codex review plz

sojan-official (approved):
Code LGTM

PR #13700, opened 2026-03-03
fix(ui): show delivered state for Instagram external echo messages

muhsin-k:
Instagram external echo messages were being saved with status: delivered, but the message meta UI did not treat Instagram as a channel eligible for delivered-state rendering. As a result, these messages fell back to progress and showed as “Sending”. This change updates the message status mapping in the new message UI to include Instagram in the delivered-state condition.


muhsin-k:
@codex please review

iamsivin (approved):
Code LGTM

PR #13713, opened 2026-03-04
fix: Prevent duplicate conversations in conversation list

muhsin-k:
Agents using API channel inboxes (e.g., WhatsApp Automate) reported seeing the same conversation appear twice in their conversation list — one showing the last message preview and the other showing "No Messages". Backend investigation confirmed no duplicate conversations exist in the database, making this purely a frontend issue.

The root cause is a race condition in WebSocket event delivery. When a conversation is created via the API with auto-assignment, the backend enqueues multiple ActionCable broadcast jobs (`conversation.created`,  `assignee.changed`, `team.changed`) within milliseconds of each other. In production with multi-threaded Sidekiq workers, these events can arrive at the frontend out of order. If `assignee.changed` arrives before `conversation.created`, the `UPDATE_CONVERSATION` mutation pushes the conversation into the store (since it doesn't exist yet), and then `ADD_CONVERSATION` blindly pushes it again — resulting in a duplicate entry.

The fix adds a uniqueness check in the `ADD_CONVERSATION` mutation to skip the push if a conversation with the same ID already exists in the store, matching the dedup pattern already used by `SET_ALL_CONVERSATION`.


muhsin-k:
@codex please review

app/javascript/dashboard/store/modules/conversations/index.js
```diff
@@ -228,7 +228,10 @@ export const mutations = {
   },
 
   [types.ADD_CONVERSATION](_state, conversation) {
-    _state.allConversations.push(conversation);
+    const exists = _state.allConversations.some(c => c.id === conversation.id);
```
iamsivin, line 231: See if we can use `getConversationById`.

iamsivin (approved):
LGTM

iamsivin (approved):
I’ve tested and verified this issue locally, and it’s now resolved. Please see the Loom video showing the before and after.


### Loom video
**Before**
https://www.loom.com/share/a821049f021c4cae9be22549d46b3ba0


**After**
https://www.loom.com/share/a65fce87093a425189ccf2e8c3a612ea

PR #13740, opened 2026-03-09
fix: make conversation transcript rate limit configurable

vishnu-narayanan:
## Summary
- The conversation transcript endpoint rate limit is hardcoded at 30 requests/hour per account with no way to override it
- Self-hosted users with active accounts hit this limit and get 429 errors across all channels
- Add `RATE_LIMIT_CONVERSATION_TRANSCRIPT` env var (default: `1000`) to make it configurable, consistent with other throttles like `RATE_LIMIT_CONTACT_SEARCH` and `RATE_LIMIT_REPORTS_API_ACCOUNT_LEVEL`


config/initializers/rack_attack.rb
```diff
@@ -185,7 +185,8 @@ def path_without_extentions
   ###-----------------------------------------------###
 
   ## Prevent Abuse of Converstion Transcript APIs ###
-  throttle('/api/v1/accounts/:account_id/conversations/:conversation_id/transcript', limit: 30, period: 1.hour) do |req|
+  throttle('/api/v1/accounts/:account_id/conversations/:conversation_id/transcript',
+           limit: ENV.fetch('RATE_LIMIT_CONVERSATION_TRANSCRIPT', '30').to_i, period: 1.hour) do |req|
```
sojan-official: Maybe we should keep a slightly higher default limit in the code and control the effective threshold via environment configuration in Cloud if needed.
Will reduce the friction for self hosted instances.

vishnu-narayanan: Makes sense. Will make the change.

vishnu-narayanan: done

pranavrajs (approved):

PR #13743, opened 2026-03-10
refactor: extract custom attribute methods from FilterService

scmmishra:
- Extracted 6 custom attribute methods (`custom_attribute_query`, `attribute_model`, `attribute_data_type`, `build_custom_attr_query`, `custom_attribute`, `not_in_custom_attr_query`) into a new `Filters::CustomAttributeFilterHelper` module.
- Added an inline `rubocop:disable` for the intentional `Lint/ShadowedException` in `coerce_lt_gt_value` — `Date::Error` is a subclass of `ArgumentError`, but both are listed explicitly for clarity.

## Why `app/services/filters/`

The existing `Filters::FilterHelper` lives in `app/helpers/filters/`, but that location triggers `Rails/HelperInstanceVariable` for any module that uses instance variables. The extracted methods share state with `FilterService` via instance variables (`@attribute_key`, `@account`, `@custom_attribute`, etc.), so placing them in `app/helpers/` would require a cop disable.

`app/services/filters/` is a better fit because:
- The module is a service mixin, not a view helper — it's only included by `FilterService` and its subclasses (`Conversations::FilterService`, `Contacts::FilterService`, `AutomationRules::ConditionsFilterService`).
- It sits alongside the services that use it.
- No cop disables needed.

### DIFF
--- a/app/services/filter_service.rb
+++ b/app/services/filter_service.rb
@@ -2,6 +2,7 @@ require 'json'
 
 class FilterService
   include Filters::FilterHelper
+  include Filters::CustomAttributeFilterHelper
   include CustomExceptions::CustomFilter
 
   ATTRIBUTE_MODEL = 'conversation_attribute'.freeze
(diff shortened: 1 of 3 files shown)


app/views/fields/confirmed_at_field/_show.html.erb
```diff
@@ -0,0 +1,3 @@
+<% if field.data %>
```
vishnu-narayanan, line 1: Do we need this file?

app/services/filter_service.rb
```diff
@@ -173,44 +156,10 @@ def coerce_lt_gt_value(raw_value, attribute_data_type, attribute_key)
     else
       raise CustomExceptions::CustomFilter::InvalidValue.new(attribute_name: attribute_key)
     end
-  rescue Date::Error, ArgumentError, FloatDomainError, TypeError
+  rescue Date::Error, ArgumentError, FloatDomainError, TypeError # rubocop:disable Lint/ShadowedException
```
vishnu-narayanan:
```suggestion
  rescue  ArgumentError, FloatDomainError, TypeError
```

For the ShadowedException, can we remove Date::Error from the rescue instead of disabling the cop -ArgumentError shoulde cover it

vishnu-narayanan (approved):
Looks good, minor comments.

app/views/fields/confirmed_at_field/_show.html.erb
scmmishra: There was a CI failing, related to a previous commit.

PR #13745, opened 2026-03-10
fix(signup): normalize account signup config checks

scmmishra:
This makes account signup enforcement consistent when signup is disabled at the installation level. Email signup and Google signup now stay blocked regardless of whether the config value is stored as a string or a boolean. 

This effectively covers the config-loader path, where `YAML.safe_load` reads `value: false` from `installation_config.yml` as a native boolean and persists it that way.

- Normalized the account signup check so disabled signup is handled consistently across config value types.
- Reused the same check across API signup and Google signup entry points.
- Added regression coverage for the disabled-signup cases in the existing controller specs.


vishnu-narayanan (approved):
LGTM

PR #13759, opened 2026-03-10
fix: optimize message query with account_id filter

aakashb95:
## Description

This PR optimizes message queries by explicitly filtering with `account_id` so the database can use the existing indexes more efficiently.

Changes:
- Add `account_id` to message query filters to improve index utilization.
- Update `last_incoming_message` query to include `account_id`.
- Avoid unnecessary preloading of `contact_inboxes` where it is not required.
- Update specs to ensure `account_id` is set correctly in message-related tests.

These changes reduce query cost and improve performance for message lookups, especially on large accounts.


sojan-official (approved):
LGTM

PR #13760, opened 2026-03-11
docs(swagger): document account label endpoints

sojan-official:
Documents the missing account-level label CRUD endpoints in Chatwoot's Swagger output so label management is discoverable alongside the existing contact and conversation label APIs.

Fixes: none
Closes: none

Why
The account-level label API already exists in Chatwoot, but it was missing from the published Swagger spec. That made label management harder to discover even though contact and conversation label assignment endpoints were already documented.

What this change does
- adds a `Labels` tag to the application Swagger docs
- adds the label resource and create/update payload schemas
- documents `GET/POST /api/v1/accounts/{account_id}/labels`
- documents `GET/PATCH/DELETE /api/v1/accounts/{account_id}/labels/{id}`
- regenerates the compiled Swagger JSON artifacts

Validation
- rebuilt the Swagger JSON from the source YAML
- verified the new label endpoints appear in `swagger/swagger.json`
- verified the new label endpoints appear in `swagger/tag_groups/application_swagger.json`
- started the local Rails server and confirmed `/swagger` and `/swagger/swagger.json` return `200 OK`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,8 +93,8 @@ jobs:
               exit 1
             fi
             mkdir -p ~/tmp
-            curl -L https://repo1.maven.org/maven2/org/openapitools/openapi-generator-cli/6.3.0/openapi-generator-cli-6.3.0.jar > ~/tmp/openapi-generator-cli-6.3.0.jar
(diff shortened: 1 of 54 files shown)


sojan-official:
@codex review plz

pranavrajs (approved):

Issue #10402, opened 2024-11-11
Help Center search is very slow - perhaps because of bloated response?


brundles:
### Describe the bug

I think the Help Center (HC) seems very slow search - especially if there are multiple languages. Could it be that the search is not only searching articles of the currently selected language somehow, and generally just including WAY more data than is required? 

It is also returning A LOT of data. So it takes 14-15 secs to do a search, and even though it only finds approx 20 articles (out of 50 or so), it uses approx 2000 lines in the JSON per result as it includes the data from the article - including the full content (not just the first 50 words or whatever) and also for some reason all  categories in all languages in the "categories" key (1800 lines!). So if you have 20 languages it will show all of these.
The result is also 1.1 mb when i check in Chrome inspector 😳

### To Reproduce

Do a search like in https://www.chatwoot.com/hc/user-guide/en and check the articles.json response

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

Browser

pranavrajs:
@brundles Thanks for reporting this, we will take a look.

brundles:
Thanks @pranavrajs - my guess is that by just strictly excluding non-relevant information (information about categories in all languages, full article text, etc). it would be considerably faster.

pranavrajs:
@brundles Can you please share the number of articles on the help center or the URL where we can test the behavior?

brundles:
@pranavrajs 

I saw it on another instance, but if you just check your own at https://www.chatwoot.com/hc/user-guide/en and search for something - like "chatwoot", then you will see an almost 2 second loading time, and a response that can take up to 6-7 seconds and loads almost 11k, despite in theory it should be enough to serve approx 2-3% of this.

<img width="1166" alt="image" src="https://github.com/user-attachments/assets/79577d3b-2831-4a18-b69f-4319d1d96e24">

The below JSON is a snippet from your own site - see how much bloated category information there is. Then imagine having 20 categories and 20 languages = 400 categories listed for each single article it finds. So its not only that you're printing the full article, its also the "categories" array for each that is quite massive :-) :)

`{
            "id": 454,
            "category_id": 193,
            "title": "Chatwoot Glossary",
[...]

PR #13761, opened 2026-03-11
fix: slim help center search results

sojan-official:
Fixes help center public article search so query responses stay compact and locale-scoped. Whitespace-only queries are now treated as empty in both the portal UI and the server-side search path, and search suggestions stay aligned with the trimmed query.

Fixes: https://github.com/chatwoot/chatwoot/issues/10402
Closes: https://github.com/chatwoot/chatwoot/issues/10402

## Why

The public help center search endpoint reused the full article serializer for query responses, which returned much more data than the search suggestions UI needed. That made responses heavier than necessary and also surfaced nested portal and category data that made the results look cross-locale.

Whitespace-only searches could also still reach the backend search path, and in Enterprise that meant embedding search could be invoked for a blank query.

## What changed

- return a compact search-specific payload for article query responses
- keep the existing full article serializer for normal article listing responses
- preserve current-locale search behavior for the portal search flow
- trim whitespace-only search terms on the client so they do not open suggestions or trigger a request
- reuse the normalized query on the backend so whitespace-only requests are treated as empty searches in both OSS and Enterprise paths
- pass the trimmed search term into suggestions so highlighting matches the actual query being sent
- add request and frontend regression coverage for compact payloads, locale scoping, and whitespace-only search behavior

## Validation

1. Open `/hc/:portal/:locale` in the public help center.
2. Enter only spaces in the search box and confirm suggestions do not open.
3. Search for a real term and confirm suggestions appear.
4. Verify the results are limited to the active locale.
5. Click a suggestion and confirm it opens the correct article page.
6. Inspect the query response and confirm it returns the compact search payload instead of the full article serializer.

### DIFF
--- a/app/mailers/conversation_reply_mailer.rb
+++ b/app/mailers/conversation_reply_mailer.rb
@@ -35,9 +35,8 @@ class ConversationReplyMailer < ApplicationMailer
   end
 
   def email_reply(message)
-    return unless smtp_config_set_or_development?
-
     init_conversation_attributes(message.conversation)
+    return unless smtp_config_set_or_development? || email_smtp_enabled? || (email_imap_enabled? && email_oauth_enabled?)
 
     @message = message
     prepare_mail(true)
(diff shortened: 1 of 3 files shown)


sojan-official:
@codex review plz

sojan-official:
@codex take another pass plz

sojan-official:
@codex review plz

PR #13793, opened 2026-03-13
fix: restrict existing user sign-in to account members

scmmishra:
SAML sign-in now only links an existing user when that user already belongs to the account that initiated SSO. New users can still be created for SAML-enabled accounts, and invited members can continue to sign in through their IdP, but SAML will no longer auto-attach an unrelated existing user record during login.

**What changed**
- Added an account-membership check before SAML reuses an existing user by email.
- Kept first-time SAML user creation unchanged for valid new users.
- Added builder and request specs covering the allowed and rejected login paths.


scmmishra:
@codex please review

sojan-official (approved):
Code LGTM

PR #13842, opened 2026-03-18
fix(imap): fall back to LOGIN when server does not advertise AUTH=PLAIN

suoaao:
IMAP email channels configured with PLAIN authentication fail to connect to servers that do not advertise `AUTH=PLAIN` in their CAPABILITY response (e.g. Alibaba Mail / AliMail). The `AUTHENTICATE PLAIN` command is rejected with an error, preventing email fetching.

This change checks the server's CAPABILITY list before choosing the authentication method. When `AUTH=PLAIN` is absent, it uses the standard IMAP `LOGIN` command instead of the SASL `AUTHENTICATE PLAIN` command. Servers that do advertise `AUTH=PLAIN` continue to use SASL as before.

## How to reproduce

1. Configure an email channel with IMAP using a provider that does not support SASL PLAIN (e.g. Alibaba Mail: `imap.qiye.aliyun.com`).
2. Observe that email syncing fails with an authentication error in Sidekiq logs.

## What changed

- Extracted `imap_authenticate` in `Imap::BaseFetchEmailService` to inspect `imap.capability` before authenticating.
- When `authentication_type` is `PLAIN` and the server CAPABILITY does not include `AUTH=PLAIN`, uses `imap.login` instead of `imap.authenticate`.
- Added unit tests covering both code paths (LOGIN fallback and SASL PLAIN).

### DIFF
--- a/app/services/imap/base_fetch_email_service.rb
+++ b/app/services/imap/base_fetch_email_service.rb
@@ -107,9 +107,19 @@ class Imap::BaseFetchEmailService
 
   def build_imap_client
     imap = Net::IMAP.new(channel.imap_address, port: channel.imap_port, ssl: true)
-    imap.authenticate(authentication_type, channel.imap_login, imap_password)
+    imap_authenticate(imap)
     imap.select('INBOX')
     imap
+  end
+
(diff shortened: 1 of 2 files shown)


suoaao:
The latest version allows the option to use login authentication for IMAP authentication, and this PR is closed.

PR #13843, opened 2026-03-18
fix: patch Devise confirmable race condition vulnerability

scmmishra:
Devise 4.9.x has a race condition in the reconfirmable flow where concurrent email change requests can desynchronize the confirmation token from `unconfirmed_email`, letting an attacker confirm an email they don't own. We use `:confirmable` with `reconfirmable = true`, so we're directly exposed.

The upstream fix is in Devise 5.0.3, but we can't upgrade — `devise-two-factor` only supports Devise 5 from v6.4.0, which also raised its Rails minimum to 7.2+. No released version supports both Devise 5 and Rails 7.1.

This PR ports the Devise 5.0.3 fix locally by overriding `postpone_email_change_until_confirmation_and_regenerate_confirmation_token` on the User model to persist the record before regenerating the token. This is a stopgap — remove it once the dependency chain allows upgrading to Devise 5.

### How to test

Sign in as a confirmed user and change your email. The app should send a confirmation to the new address while keeping the current email unchanged until confirmed.

### DIFF
--- a/.bundler-audit.yml
+++ b/.bundler-audit.yml
@@ -1,3 +1,4 @@
 ---
 ignore:
   - CVE-2021-41098 # https://github.com/chatwoot/chatwoot/issues/3097 (update once azure blob storage is updated)
+  - GHSA-57hq-95w6-v4fc # Devise confirmable race condition — patched locally in User model (remove once on Devise 5+)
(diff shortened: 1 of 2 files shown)


scmmishra:
@codex please review

sojan-official (approved):

PR #13844, opened 2026-03-18
chore(i18n): add missing pt_BR locale imports for companies, mfa, snooze, webhooks and more

dollyzn:
## Description

Add missing JSON imports and spread exports for `companies`, `contentTemplates`, `mfa`, `snooze`, `webhooks`, and `yearInReview` so these translations are properly loaded in the pt_BR locale. Without these imports, those sections of the UI were falling back to English for Brazilian Portuguese users.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings

PR #13856, opened 2026-03-20
fix(i18n): auto-load dashboard locale modules

sojan-official:
This refactor makes dashboard locale loading deterministic and removes the fragile per-locale import chain, so split locale JSON files are loaded from disk without needing an extra JS wiring change for each language.

ref: https://github.com/chatwoot/chatwoot/pull/13844

## Why
The previous dashboard i18n setup depended on hand-maintained locale index files. That made it easy for translation files to exist on disk but never get bundled, which is what caused the recent pt_BR regression. Auditing the locale tree showed that this gap was systemic.

## What this change does
- replaces manual dashboard locale imports with an allowlisted `import.meta.glob` loader
- keeps the bundled locale list and module list explicit in one place
- adds explicit English fallback for dashboard, v3 app, and Histoire i18n boot
- adds regression coverage for locale-folder parity, English module parity, duplicate top-level keys, actual file loading, and fallback behavior
- removes the now-unused per-locale dashboard `index.js` files

## Validation
- verified the loader now covers all 56 dashboard locale folders on disk
- verified 55 locales have full module coverage and only `zh` relies on English fallback for missing files
- added focused i18n coverage that exercises real missing-key fallback and protects the new loader contract from drifting


iamsivin:
@codex review?

PR #13868, opened 2026-03-21
fix: show all statuses in the All assignee tab

### LINKED ISSUE #13865: Resolved conversations disappear from the "All" conversations tab
### Describe the bug

When viewing conversations under the "All" assignee tab, resolved conversations are never fetched from the API. The `conversationFilters` computed property in `ChatList.vue` always passes `activeStatus` (which defaults to `'open'`) as the status filter, regardless of which assignee tab is selected.

This means the "All" tab behaves identically to the "Mine" and "Unassigned" tabs — only open conversations appear. Users have no way to see resolved conversations in the main conversation list without applying a manual filter.

We run Chatwoot in production for our healthcare organization (v4.11.1, Docker on AWS) and our support team flagged this — they rely on the All tab to review both open and resolved conversations in a single view.

### To Reproduce

1. Open the Chatwoot dashboard
2. Resolve one or more conversations
3. Switch to the "All" assignee tab
4. Observe that resolved conversations are not shown

### Expected behavior

The "All" tab should display conversations across all statuses (open, resolved, pending), not just open conversations.

### Root cause

In `app/javascript/dashboard/components/ChatList.vue`, the `conversationFilters` computed always uses `activeStatus` for the `status` field:

```js
const conversationFilters = computed(() => {
  return {
    // ...
    status: activeStatus.value,  // defaults to 'open' — never 'all'
  };
});
```

Additionally, `filterByStatus` in `helpers.js` doesn't handle array values, which prevents multi-status filtering from working correctly.

### Environment

- Self-hosted (Docker on AWS EC2)
- Chatwoot v4.11.1
- Browser: Chrome

ak-padma:
## Description

The "All" assignee tab only shows open conversations because `conversationFilters` always passes `activeStatus` (defaults to `'open'`) as the status filter. Resolved conversations silently disappear from the list.

This changes the All tab to always fetch with `status: 'all'` so resolved, pending, and open conversations all appear. Mine/Unassigned tabs continue using the user-controlled status filter.

Also fixes `filterByStatus` in the conversations store to handle array values correctly.

Fixes #13865

## Type of change

## How Has This Been Tested?

Running in production (v4.11.1, Docker on AWS) for our healthcare support team. Verified:
- All tab shows open, resolved, and pending conversations
- Mine and Unassigned tabs still default to open-only
- Status badge counts reflect the effective filter

## Checklist:

### DIFF
--- a/app/javascript/dashboard/components/ChatList.vue
+++ b/app/javascript/dashboard/components/ChatList.vue
@@ -264,10 +264,13 @@ const conversationListPagination = computed(() => {
 });
 
 const conversationFilters = computed(() => {
+  // All tab always shows every status so resolved conversations don't disappear.
+  // Mine/Unassigned keep the user-controlled status filter (defaults to 'open').
+  const isAllTab = activeAssigneeTab.value === wootConstants.ASSIGNEE_TYPE.ALL;
   return {
     inboxId: props.conversationInbox ? props.conversationInbox : undefined,
     assigneeType: activeAssigneeTab.value,
-    status: activeStatus.value,
+    status: isAllTab ? wootConstants.STATUS_TYPE.ALL : activeStatus.value,
     sortBy: activeSortBy.value,
     page: conversationListPagination.value,
     labels: props.label ? [props.label] : undefined,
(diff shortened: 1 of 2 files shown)


sojan-official:
Screenshot 2026-03-23 at 7 26 06 PM

This is the intended product behavior.

The “All” tab is designed to show only active conversations (open), which represent items currently being worked on. Other statuses like pending or resolved are intentionally hidden in the default view.

If needed, those conversations can still be accessed by applying the status filter explicitly.

stemper-dev:
@sojan_official This is really confusing! 'All' should really mean 'all'. Also, as you said, add a new filter called 'Worked On' for instance

sojan-official:
@0xPatryk Thanks for sharing this feedback. We’ll keep an eye on it and see if we get more similar inputs before prioritizing any changes.

PR #13877, opened 2026-03-23
fix: remove bulk_auto_assignment_job cron schedule

tds-1:
## Description

Remove the `bulk_auto_assignment_job` cron entry that runs every 15 minutes. This job does not respect agent capacity limits and team boundaries, causing conversations to be over-assigned or assigned outside the designated team. The newer `periodic_assignment_job` (Assignment V2) already handles bulk auto-assignment correctly.

## Type of change

## How Has This Been Tested?

Verified on a self-hosted instance by disabling the job from the Super Admin Sidekiq Cron panel — assignment behavior worked as expected with only the V2 periodic job active.

## Checklist:                                                                                                                                            
                                                                                                                                                         

sojan-official:
@tds-1  the PR removes the job from schedule.yml, but load_from_hash (which we use in config/initializers/sidekiq.rb) only upserts — it doesn't clean up jobs that are no longer in the YAML. So on existing deployments, the bulk_auto_assignment_job cron entry will persist in Redis and keep firing (even though the job class is gone, which will cause errors).

Can you look into switching to load_from_hash! with source: 'schedule' in the initializer? There's already a TODO for this. That way any job removed from the YAML gets automatically cleaned up from Redis on deploy.

app/jobs/inboxes/bulk_auto_assignment_job.rb
```diff
@@ -1,47 +0,0 @@
-class Inboxes::BulkAutoAssignmentJob < ApplicationJob
```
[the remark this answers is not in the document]
tds-1: The concern is theoretically valid but practically negligible for a cron job with a 15 minute cadence.

config/initializers/sidekiq.rb
```diff
@@ -34,7 +34,18 @@ def call(_worker, job, queue)
...
+  # jobs that share the same source tag but are no longer in the file.
+  # This ensures deleted schedule entries are cleaned up on deploy.
+  if File.exist?(schedule_file) && Sidekiq.server?
+    schedule = YAML.load_file(schedule_file)
+
+    # Cron entries removed from schedule.yml but possibly still in Redis
+    # with source:'dynamic' (predating the source tag). load_from_hash!
+    # only cleans up source:'schedule' entries, so these need explicit removal.
+    # Remove names from this list once they've been through a deploy cycle.
+    %w[bulk_auto_assignment_job].each { |name| Sidekiq::Cron::Job.destroy(name) }
```
scmmishra, line 47: Should we move this to a migration?

scmmishra: Or rather, should we run it manually, having this in the codebase does not make sense after the clearing is done

tds-1: We should not run it manually, since it is required to be done at self hosted instances as well

tds-1: > Should we move this to a migration?

No, this is a Redis cleanup, not a DB change, so it does not belong in the migration pipeline. The initializer runs every deploy (not just once) and is idempotent, so it'll catch the entry even if Redis is restored from a backup. We can drop the line after a deploy cycle or two.

scmmishra (approved):

PR #13897, opened 2026-03-24
feat: reclaim response_bot flag for custom_tools

aakashb95:
Repurpose the deprecated response_bot feature flag slot for custom_tools.

Migration disables the flag on any accounts that had response_bot enabled so the repurposed slot starts in its default-off state.

Pre-deploy: run the disable script on production using the old flag name (response_bot) before deploying this migration.


aakashb95:
decided against feature flag for tools

db/migrate/20260324102005_repurpose_response_bot_flag_for_custom_tools.rb
```diff
@@ -0,0 +1,11 @@
+class RepurposeResponseBotFlagForCustomTools < ActiveRecord::Migration[7.1]
+  def up
+    # The response_bot flag (deprecated) has been renamed to custom_tools.
+    # Disable it on any accounts that had response_bot enabled so the repurposed
+    # flag starts in its intended default-off state.
+    Account.feature_custom_tools.find_each(batch_size: 100) do |account|
+      account.disable_features(:custom_tools)
+      account.save!(validate: false)
```
pranavrajs, line 8: why do we write this? i mean what is reason behind saving without validation.

aakashb95: we want to start the new flag with default off, regardless of any unrelated errors

scmmishra:
@codex please review

scmmishra (approved):

PR #13904, opened 2026-03-25
fix: resolve V2 capacity bypass in team assignment

tds-1:
## Description

When Assignment V2 is enabled, the V2 capacity policies (AgentCapacityPolicy / InboxCapacityLimit) are not respected during team-based assignment paths. The system falls back to the legacy V1 max_assignment_limit, and since V1 is deprecated and typically unconfigured in V2 setups, agents receive unlimited assignments regardless of their V2 capacity.

Root cause: Inbox class directly defined member_ids_with_assignment_capacity, which shadowed the Enterprise::InboxAgentAvailability module override in  Ruby's method resolution order (MRO). This made the V2 capacity check unreachable (dead code) for any code path using member_ids_with_assignment_capacity.                                                                                                                  

## Type of change

## How Has This Been Tested?

⏺ Before the fix                                                                                                                                         
                                                                                                                                                      
1. Enable assignment_v2 + advanced_assignment on account                                                                                               
2. Create AgentCapacityPolicy with InboxCapacityLimit = 1 for an inbox                                                                                 
3. Assign the policy to an agent (e.g., John)                                                                                                          
4. Create 1 open conversation assigned to John (now at capacity)            
5. Create a new unassigned conversation in the same inbox                                                                                              
6. Assign a team (containing John) to that conversation                                                                                                
7. Result: John gets assigned despite being at capacity                                                                                                
                                                                                                                                                        
⏺ After the fix                                                                                                                                          
                                                                                                                                                        
Same steps 1–6.                                                                                                                                        
7. Result: John is NOT assigned — conversation stays unassigned (no agents with capacity available)                                                    


## Checklist:

### DIFF
--- a/enterprise/app/models/enterprise/inbox.rb
+++ b/enterprise/app/models/enterprise/inbox.rb
@@ -1,6 +1,7 @@
 module Enterprise::Inbox
   def member_ids_with_assignment_capacity
     return super unless enable_auto_assignment?
+    return filter_by_capacity(available_agents).map(&:user_id) if auto_assignment_v2_enabled?
 
     max_assignment_limit = auto_assignment_config['max_assignment_limit']
     overloaded_agent_ids = max_assignment_limit.present? ? get_agent_ids_over_assignment_limit(max_assignment_limit) : []
--- a/enterprise/app/models/enterprise/inbox_agent_availability.rb
+++ b/enterprise/app/models/enterprise/inbox_agent_availability.rb
@@ -21,7 +21,7 @@ module Enterprise::InboxAgentAvailability
   end
 
   def capacity_filtering_enabled?
-    account.feature_enabled?('assignment_v2') &&
+    account.feature_enabled?('advanced_assignment') &&
       account.account_users.joins(:agent_capacity_policy).exists?
   end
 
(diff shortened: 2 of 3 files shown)


tds-1:
@codex please review

tds-1:
@codex please review

tds-1:
@codex please review

sojan-official (approved):
the fix looks good to me. but there is a concern 

I think the underlying issue is that `member_ids_with_assignment_capacity` is currently defined in too many layers. `Enterprise::Inbox` now contains the real V2 fix, but `Enterprise::InboxAgentAvailability#member_ids_with_assignment_capacity` still looks like an active override even though `Inbox` dispatch never really reaches it for `Inbox` instances.

Would it make sense to remove that public method from `Enterprise::InboxAgentAvailability` and keep only the private helpers there? That would make it clearer that the real dispatch now lives in `Enterprise::Inbox`.

> The method remaining there would be confusing down the lane.

tds-1:
@sojan-official  After this fix it went from four layers of overriding to three, which is better but still more than ideal. I will make the change to remove the public method from InboxAgentAvailability in a separate PR to keep this one focused.

PR #13906, opened 2026-03-25
chore: mark conversation notifications as read on visit

iamsivin:
# Pull Request Template

## Description

This PR marks all conversation notifications as read when the user visits it (i.e., when the last seen timestamp is updated).
It introduces `mark_conversation_notifications_read` to handle this and keep notification states in sync with user activity.

Fixes https://linear.app/chatwoot/issue/PLA-105/unread-mark-does-not-change-when-navigating-between-conversation-and

## Type of change

### Loom video
https://www.loom.com/share/db1460c3bdd944a787a600feb644ee66

## Checklist:


sojan-official:
@iamsivin I moved this from a controller concern to a small service because the logic here is really business/domain logic, not shared controller behavior.

A concern is most useful when multiple controllers need the same controller-level code. In this case, the behavior is specifically “mark this conversation’s notifications as read,” which fits better as a dedicated service class. That keeps the controller thin, makes the intent easier to understand, and gives us a clean place to reuse this later if another flow needs the same behavior.

app/services/notification/mark_conversation_read_service.rb
```diff
@@ -0,0 +1,25 @@
+class Notification::MarkConversationReadService
+  pattr_initialize [:user!, :account!, :conversation!]
+
+  def perform
+    return unless user.is_a?(User)
+
+    notifications.find_each do |notification|
+      notification.update!(read_at: read_at)
```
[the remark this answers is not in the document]
sojan-official: We will need frontend updates when notification is updated.. so update all is not and option..

sojan-official (approved):
Hey @iamsivin, this touches a hot write path and the conversations endpoints are already heavy on writes, so let’s keep a close eye during deploy.

Would be good to monitor DB write load, query timings, and any spikes in latency or errors right after rollout.

Code looks good to me otherwise 👍

app/services/notification/mark_conversation_read_service.rb
```diff
@@ -0,0 +1,25 @@
+class Notification::MarkConversationReadService
+  pattr_initialize [:user!, :account!, :conversation!]
+
+  def perform
+    return unless user.is_a?(User)
```
pranavrajs, line 5: why is this needed?

app/controllers/api/v1/accounts/conversations_controller.rb
```diff
@@ -116,6 +116,8 @@ def update_last_seen
     # High-traffic accounts generate excessive DB writes when agents frequently switch between conversations.
     # Throttle last_seen updates to once per hour when there are no unread messages to reduce DB load.
     # Always update immediately if there are unread messages to maintain accurate read/unread state.
+    # Visiting a conversation should clear any unread inbox notifications for this conversation.
+    Notification::MarkConversationReadService.new(user: current_user, account: current_account, conversation: @conversation).perform
```
pranavrajs: current_user or Current.user?

app/services/notification/mark_conversation_read_service.rb
sojan-official: Agent Bots can also access conversations

app/controllers/api/v1/accounts/conversations_controller.rb
sojan-official: Current would be better.

muhsin-k (approved):

PR #13916, opened 2026-03-26
feat: Adds the ability to resize the editor

iamsivin:
# Pull Request Template

## Description

This PR adds support for resizing the reply editor up to nearly half the screen height. It also deprecates the old modal-based pop-out reply box, clicking the same button now expands the editor inline. Users can adjust the height using the slider or the expand button.

## Type of change

### Loom video
https://www.loom.com/share/be27e1c06d19475ab404289710b3b0da

## Checklist:


pranavrajs:
@iamsivin The experience is mostly fine, but there are a couple of issues. Right now it takes up 100% of the available height, which feels odd since the messages are not visible.

We should default this to around 50% of the screen height. Users can expand it further if needed, but starting at full height isn’t ideal. It doesn’t have to be capped strictly at 50%, just a smaller, more reasonable default with the ability to resize.

Also, when expanding, it goes below the action buttons, making parts of it inaccessible. That needs to be fixed so the full view remains usable.
Screenshot 2026-04-10 at 6 28 32 PM

iamsivin:
> @iamsivin The experience is mostly fine, but there are a couple of issues. Right now it takes up 100% of the available height, which feels odd since the messages are not visible.
> 
> We should default this to around 50% of the screen height. Users can expand it further if needed, but starting at full height isn’t ideal. It doesn’t have to be capped strictly at 50%, just a smaller, more reasonable default with the ability to resize.
> 
> Also, when expanding, it goes below the action buttons, making parts of it inaccessible. That needs to be fixed so the full view remains usable.

@pranavrajs Updated the PR, on clicking the expand button, the view shifts up to 50% of the message list height, and can then be resized up to ~80% of the list height.

sony-mathew (approved):
LGTM

sony-mathew (approved):
LGTM

PR #13937, opened 2026-03-29
fix(widget): unread preview in real time and without-bubble offset

### LINKED ISSUE #12395: Widget without bubble: bug in unread badge height and scroll behavior
### Describe the bug

When using the widget in without bubble mode, I noticed two issues:

### **Bug 1: Scroll behavior (Unsolved)**
A scroll bar is shown even though it should not appear.
Instead, the three message bubbles should be fully visible by default.
Currently, this only works correctly after refreshing the page (F5).
➝ Couldn’t find a direct fix, seems to be an internal widget issue.

### **Bug 2: Unread badge height (Solved)**
The unread counter shows with incorrect height.
➝ Easy fix: adjusting the CSS height resolves the problem.

Image

### To Reproduce

1. Open the widget in without bubble mode.
2. Have a support agent send a message to the customer.
3. Observe that:
    - The unread badge height is incorrect.
    - A scroll bar appears and the three bubbles are not fully visible (unless the page is refreshed).

### **Bug 1: Scroll behavior (Unsolved)**
A scroll bar is shown even though it should not appear.
Instead, the three message bubbles should be fully visible by default.

Currently, this only works correctly after refreshing the page (F5).

I couldn’t find a direct fix, but I identified that the widget iframe keeps the style:

`style="height: 219px !important"`

After a page refresh, the height is automatically adjusted to the correct value that fits all three bubbles, and then the scroll bar disappears.
However, in live usage this adjustment doesn’t happen, so the scroll bar remains visible.

In addition to the fix above, I would also suggest limiting the number of characters per bubble.
That way, when three very long messages are sent, each bubble would be truncated with ... at the end.

This ensures that the total widget height (640px) is respected, resolving the scroll bar issue both for:
    - three short messages, and
    - three very long messages.

**Unlike Bug 2, this scroll behavior issue happens in all widget modes, including those with bubble, since site widgets also receive replies and end up showing the unwanted scroll bar.**

---

### **Bug 2: Unread badge height (Solved)**
The unread counter shows with incorrect height.

➝ Easy fix: adjusting the CSS height resolves the problem.

app/javascript/sdk/sdk.js

```
 .woot-widget-holder.has-unread-view {
    border-radius: 0 !important;
    min-height: 80px !important;
    height: auto;
 +  transform: translateY(-30px);
    bottom: 94px;
    box-shadow: none !important;
    border: 0;
 }
```

Resultd:

Image

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

None

### Operating system

_No response_

### Browser and version

_No response_

### Docker (if applicable)

_No response_

### Additional context

_No response_

vini-fitzner:
Fixes #12395

## Summary
This PR fixes the widget issue where the unread preview looked correct after refresh but broke during realtime updates.

It also fixes the positioning of the widget in without-bubble mode.

## Changes
- emit `ON_AGENT_MESSAGE_RECEIVED` after store updates on `message.updated`
- rework unread preview routing and iframe height recalculation
- keep only the last 3 unread preview messages with stable keys
- clamp preview message text to avoid layout breakage
- prevent overflow issues in the unread preview layout
- adjust `.woot-widget--without-bubble` bottom offset from `20px` to `62px`

## Testing
- verified unread preview after refresh
- verified unread preview during realtime updates
- verified without-bubble positioning

### Without bubble mode
image
<br />
<br />

### Bubble mode
image

### DIFF
--- a/app/javascript/sdk/sdk.js
+++ b/app/javascript/sdk/sdk.js
@@ -189,7 +189,7 @@ export const SDK_CSS = `
 }
 
 .woot-widget--without-bubble {
-  bottom: 20px !important;
+  bottom: 62px !important;
 }
 .woot-widget-holder.woot--hide{
   transform: translateY(40px);
--- a/app/javascript/sdk/sdk.js
+++ b/app/javascript/sdk/sdk.js
@@ -189,7 +189,15 @@ export const SDK_CSS = `
 }
 
 .woot-widget--without-bubble {
-  bottom: 20px !important;
+  bottom: 62px !important;
+}
+
+@media only screen and (max-width: 767px) {
+  .woot-widget-holder.has-unread-view.woot-widget--without-bubble {
+    --woot-unread-bottom-offset: clamp(0px, calc(100vh - 464px), 76px);
+    bottom: var(--woot-unread-bottom-offset) !important;
+    max-height: calc(100vh - var(--woot-unread-bottom-offset)) !important;
+  }
 }
 .woot-widget-holder.woot--hide{
   transform: translateY(40px);
--- a/app/javascript/widget/App.vue
+++ b/app/javascript/widget/App.vue
@@ -135,11 +135,13 @@ export default {
     },
     setIframeHeight(isFixedHeight) {
       this.$nextTick(() => {
-        const extraHeight = getExtraSpaceToScroll();
-        IFrameHelper.sendMessage({
-          event: 'updateIframeHeight',
-          isFixedHeight,
-          extraHeight,
+        requestAnimationFrame(() => {
+          const extraHeight = getExtraSpaceToScroll();
+          IFrameHelper.sendMessage({
+            event: 'updateIframeHeight',
+            isFixedHeight,
+            extraHeight,
+          });
         });
       });
     },
@@ -219,12 +221,28 @@ export default {
       } else if (
         this.isIFrame &&
         unreadMessageCount > 0 &&
-        !this.isWidgetOpen
+        (!this.isWidgetOpen ||
+          ['unread-messages', 'campaigns'].includes(this.$route.name))
       ) {
-        this.router.replace({ name: 'unread-messages' }).then(() => {
-          this.setIframeHeight(true);
-          IFrameHelper.sendMessage({ event: 'setUnreadMode' });
-        });
+        const applyPreviewChrome = () => {
+          const skipSetUnreadMode =
+            this.$route.name === 'unread-messages' && this.isWidgetOpen;
+          if (!skipSetUnreadMode) {
+            IFrameHelper.sendMessage({ event: 'setUnreadMode' });
+          }
+          this.$nextTick(() => {
+            requestAnimationFrame(() => {
+              this.setIframeHeight(true);
+            });
+          });
+        };
+        if (this.$route.name === 'unread-messages') {
+          applyPreviewChrome();
+        } else {
+          this.router.replace({ name: 'unread-messages' }).then(() => {
+            applyPreviewChrome();
+          });
+        }
         this.handleUnreadNotificationDot();
       }
     },
--- a/app/javascript/widget/App.vue
+++ b/app/javascript/widget/App.vue
@@ -135,11 +135,13 @@ export default {
     },
     setIframeHeight(isFixedHeight) {
       this.$nextTick(() => {
-        const extraHeight = getExtraSpaceToScroll();
-        IFrameHelper.sendMessage({
-          event: 'updateIframeHeight',
-          isFixedHeight,
-          extraHeight,
+        requestAnimationFrame(() => {
+          const extraHeight = getExtraSpaceToScroll();
+          IFrameHelper.sendMessage({
+            event: 'updateIframeHeight',
+            isFixedHeight,
+            extraHeight,
+          });
         });
       });
     },
@@ -161,13 +163,27 @@ export default {
       }
     },
     registerUnreadEvents() {
-      emitter.on(ON_AGENT_MESSAGE_RECEIVED, () => {
-        const { name: routeName } = this.$route;
-        if ((this.isWidgetOpen || !this.isIFrame) && routeName === 'messages') {
-          this.$store.dispatch('conversation/setUserLastSeen');
+      emitter.on(
+        ON_AGENT_MESSAGE_RECEIVED,
+        ({ source } = { source: 'created' }) => {
+          const { name: routeName } = this.$route;
+
+          if (source === 'updated') {
+            if (routeName !== 'unread-messages') return;
+            this.setUnreadView();
+            return;
+          }
+
+          if (
+            (this.isWidgetOpen || !this.isIFrame) &&
+            routeName === 'messages'
+          ) {
+            this.$store.dispatch('conversation/setUserLastSeen');
+          }
+
+          this.setUnreadView();
         }
-        this.setUnreadView();
-      });
+      );
       emitter.on(ON_UNREAD_MESSAGE_CLICK, () => {
         this.router
           .replace({ name: 'messages' })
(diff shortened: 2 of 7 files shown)


iamsivin:
@codex review?

app/javascript/widget/components/UnreadMessageList.vue
```diff
@@ -67,8 +80,8 @@ export default {
     </div>
     <div class="unread-messages">
       <UnreadMessage
-        v-for="(message, index) in messages"
-        :key="message.id"
+        v-for="(message, index) in previewMessages"
+        :key="unreadBubbleKey(message, index)"
```
iamsivin: I don't think the custom key is needed here. The message content is already reactive since it's passed as a prop. Also, because this key changes whenever the text changes, Vue will recreate the entire bubble component and replay its animation instead of just updating the content. A simpler key should be sufficient:
```suggestion
        :key="message.id ?? message.campaignId ?? index"
```

app/javascript/widget/App.vue
```diff
@@ -219,12 +221,28 @@ export default {
...
+          const skipSetUnreadMode =
+            this.$route.name === 'unread-messages' && this.isWidgetOpen;
+          if (!skipSetUnreadMode) {
+            IFrameHelper.sendMessage({ event: 'setUnreadMode' });
+          }
+          this.$nextTick(() => {
+            requestAnimationFrame(() => {
+              this.setIframeHeight(true);
+            });
+          });
```
iamsivin: I think we're doing this twice here. `setIframeHeight` already wraps its logic with `$nextTick` and `requestAnimationFrame`.

app/javascript/widget/components/UnreadMessageList.vue
```diff
@@ -51,6 +54,16 @@ export default {
...
+    unreadBubbleKey(message, index) {
+      const id = message.id ?? message.campaignId ?? `idx-${index}`;
+      const text = this.getMessageContent(message);
+      const mod = 1_000_000_007;
+      let h = 0;
+      for (let i = 0; i < Math.min(text.length, 256); i += 1) {
+        h = (h * 31 + text.charCodeAt(i)) % mod;
+      }
+      return `${id}-${text.length}-${h}`;
+    },
```
iamsivin: We can remove the `unreadBubbleKey` method as well.

app/javascript/widget/components/UnreadMessageList.vue
```diff
@@ -26,9 +26,12 @@ export default {
       unreadMessageCount: 'conversation/getUnreadMessageCount',
       widgetColor: 'appConfig/getWidgetColor',
     }),
+    previewMessages() {
+      return this.messages.slice(-3);
+    },
```
iamsivin: Just a question: is this additional `slice(-3)` intentional as a safeguard? `getUnreadTextMessages` already limits the unread messages to 3, so I wanted to understand the extra check here.

iamsivin (commented):
Thanks for the PR 🙌
I've also left a few review comments. Could you please take a look and update the PR branch with the latest `develop` changes?

app/javascript/widget/components/UnreadMessageList.vue
vini-fitzner: Thanks for pointing this out. I switched this to the simpler stable key:

`message.id ?? message.campaignId ?? index`

This should avoid remounting the bubble when only the message content changes.

app/javascript/widget/App.vue
vini-fitzner: Good catch. I removed the extra `$nextTick` / `requestAnimationFrame` wrapper and now call `setIframeHeight(true)` directly, since `setIframeHeight` already handles that internally.

app/javascript/widget/components/UnreadMessageList.vue
vini-fitzner: Done. Removed the `unreadBubbleKey` method as suggested.

app/javascript/widget/components/UnreadMessageList.vue
vini-fitzner: It was added as a defensive safeguard, but you're right that it's redundant since `getUnreadTextMessages` already caps the list at 3 using `Math.min(unreadCount, 3)`. I removed `previewMessages` and now iterate over `messages` directly.

app/javascript/widget/assets/scss/views/_conversation.scss
```diff
@@ -123,10 +123,10 @@
 }
 
 .unread-messages {
-  @apply flex flex-col flex-nowrap mt-0 overflow-y-auto w-full pb-2;
+  @apply flex flex-col flex-nowrap mt-0 overflow-hidden w-full min-w-0 pb-2;
```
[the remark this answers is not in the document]
vini-fitzner: Good catch. I updated the unread messages container to preserve vertical scrolling in constrained viewports while still preventing horizontal overflow.

It now uses overflow-x-hidden overflow-y-auto and keeps min-w-0, so previews and action buttons remain reachable on small/mobile landscape screens without reintroducing horizontal overflow.

app/javascript/sdk/sdk.js
```diff
@@ -189,7 +189,7 @@ export const SDK_CSS = `
 }
 
 .woot-widget--without-bubble {
-  bottom: 20px !important;
+  bottom: 62px !important;
```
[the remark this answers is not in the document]
vini-fitzner: Thanks for flagging this. I kept the existing SDK mobile breakpoint unchanged and scoped the offset adjustment to the without-bubble unread preview only.

The regular embedded widget behavior remains unchanged. The without-bubble unread preview now gets a narrow-layout offset so it does not sit too low in the in-app layout, while the mobile internal alignment fix is scoped to .is-mobile.is-unread-preview.is-bubble-hidden.

app/javascript/widget/helpers/actionCable.js
```diff
@@ -85,7 +88,9 @@ class ActionCableConnector extends BaseActionCableConnector {
       });
     }
 
-    this.app.$store.dispatch('conversation/addOrUpdateMessage', data);
+    this.app.$store
+      .dispatch('conversation/addOrUpdateMessage', data)
+      .then(() => emitter.emit(ON_AGENT_MESSAGE_RECEIVED));
```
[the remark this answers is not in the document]
vini-fitzner: Fixed. message.created keeps the existing behavior and can open the unread preview, while message.updated now only refreshes the unread preview when it is already visible. This prevents generic message updates from reopening a dismissed preview or marking messages as seen.

app/javascript/sdk/sdk.js
```diff
@@ -262,6 +268,10 @@ export const SDK_CSS = `
     padding: 0 8px;
   }
 
+  .woot-widget-holder.has-unread-view.woot-widget--without-bubble {
+    bottom: 76px !important;
```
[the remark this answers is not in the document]
vini-fitzner: Fixed. The hidden-bubble unread holder now uses a responsive bottom offset and subtracts that offset from `max-height`, so the preview stays within the available viewport instead of allowing a fixed 76px offset with a 100vh holder.

On shorter mobile/tablet viewports, the offset gradually reduces, which avoids over-constraining the unread preview while still keeping it above the button when there is enough vertical space.

vini-fitzner:
Fixed and pushed.

I attached two short viewport-resize tests for context.

**Video 1 — Regular bubble mode**

Shows the baseline behavior of the unread preview with the regular widget bubble on narrow/mobile viewports.

https://github.com/user-attachments/assets/651ac9fc-c3d0-446f-a4c9-2f2643f793da

**Video 2 — Hidden-bubble / dashboard mode**

Shows the updated hidden-bubble behavior after the fix, including the responsive bottom offset.

https://github.com/user-attachments/assets/18299a1c-10cc-4851-afa7-7ac329b18c6c

While testing this, I used the regular bubble mode as a reference for how the unread preview behaves on constrained mobile viewports.

For the hidden-bubble/dashboard case, there were a few extra constraints:

- The dashboard menu button starts collapsing around the `767px` range.
- The SDK only reaches its narrower mobile layout at `667px`.
- Because of that, this case cannot rely only on the existing `667px` mobile rules.
- The fixed `76px` bottom offset was being combined with a holder that could still use up to `100vh` height.
- On shorter viewports, that can over-constrain the unread preview and make the CTA hard to reach.

The updated CSS now uses a responsive bottom offset and subtracts that same offset from `max-height`.

This preserves the clearance above the dashboard menu button when there is enough vertical space, while gradually reducing the offset on shorter viewports so the preview stays within the available space and the CTA remains reachable.

The existing internal scroll behavior on short mobile viewports is preserved.

This keeps the change limited to the hidden-bubble unread holder and does not change unread message rendering or unread count behavior.

PR #13948, opened 2026-03-30
ci: add API schema drift check for swagger definitions

### LINKED ISSUE #13871: Suggestion: CI check for API schema drift
The recent fix in #13693 caught a mismatch between the OpenAPI schema definitions and the actual jbuilder responses for the Public API. That kind of drift — where the spec says one thing but the API returns another — can break SDK consumers silently.

Since the swagger definitions are checked into the repo and actively maintained, a CI check could catch these drifts automatically on every PR that touches the schema.

What it would catch:
- Fields removed from response definitions
- Type mismatches between spec and implementation  
- Required parameters added without updating consumers
- Enum values changed

We built an open-source Action for this — it diffs the base branch swagger against the PR and posts a comment with any breaking changes:

```yaml
name: API Schema Check
on:
  pull_request:
    paths:
      - 'swagger/**'

jobs:
  check:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0
      - uses: delimit-ai/delimit-action@v1
        with:
          spec: swagger/swagger.json
```

Example of what the PR comment looks like: https://github.com/delimit-ai/delimit-action-demo/pull/1

No config or API keys needed. Advisory mode by default — posts a comment but doesn't block merges.

infracore:
## Description
Adds a CI workflow that checks for breaking API changes whenever a PR touches `swagger/**`. It diffs the base branch schema against the PR and posts an advisory comment identifying:

- Breaking vs non-breaking changes (field removals, type mismatches, required param additions, enum changes)
- Semver impact classification
- Migration guidance for breaking changes

Advisory mode only — posts a comment but **never blocks merges**. No API keys or configuration needed.

### How it works
The workflow uses [delimit-action](https://github.com/delimit-ai/delimit-action) to diff `swagger/swagger.json` against the base branch version. Example output: [demo PR comment](https://github.com/delimit-ai/delimit-action-demo/pull/1).

### Context
Issue #13693 caught a mismatch between the swagger definitions and actual jbuilder responses. This workflow would catch that class of drift automatically on future PRs.

Closes #13871

## Type
- [x] Chore

### DIFF
--- a/.github/workflows/api_schema_check.yml
+++ b/.github/workflows/api_schema_check.yml
@@ -1,0 +1,22 @@
+name: API Schema Check
+
+on:
+  pull_request:
+    paths:
+      - 'swagger/**'
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  schema-drift:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: delimit-ai/delimit-action@v1
+        with:
+          spec: swagger/swagger.json
--- a/.github/workflows/api_schema_check.yml
+++ b/.github/workflows/api_schema_check.yml
@@ -2,9 +2,16 @@
 # that touches swagger definitions. Posts an advisory comment — does not block merges.
 # ref: https://github.com/chatwoot/chatwoot/issues/13871
 #
-# Note: This diffs the source-of-truth files under swagger/ directly.
-# If swagger/swagger.json is generated from the fragments in definitions/
-# and paths/, consider adding a build step to regenerate it before the diff.
+# Build process: Chatwoot compiles swagger/swagger.json from YAML fragments
+# (swagger/index.yml, swagger/paths/, swagger/definitions/) using the json_refs
+# gem: `bundle exec rake swagger:build`. Contributors must run this locally
+# before committing so that swagger.json stays in sync with the fragments.
+#
+# This workflow diffs swagger.json between the base and PR branches, so it will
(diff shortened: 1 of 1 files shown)


infracore:
Good catch on the fragmented spec scenario. Added a comment in the workflow noting that if `swagger/swagger.json` is generated from the fragments in `definitions/` and `paths/`, a build step to regenerate it before the diff would prevent false negatives.

Happy to add that regeneration step if the team can point me to how `swagger.json` is currently built — didn't want to assume the toolchain.

infracore:
Researched the swagger build process to make sure this workflow covers it correctly.

Chatwoot compiles `swagger/swagger.json` from the YAML fragments in `swagger/paths/` and `swagger/definitions/` using `bundle exec rake swagger:build` (via the `json_refs` gem). Contributors run this locally before committing.

This workflow diffs `swagger/swagger.json` between the base and PR branches, so it catches breaking changes whenever contributors follow the existing convention of regenerating the file before pushing. The `paths: swagger/**` trigger fires on both fragment edits and `swagger.json` changes.

If the team wants to enforce regeneration in CI (so fragment-only edits without a rebuild are also caught), a Ruby setup + `bundle install` + `rake swagger:build` step could be added before the delimit-action step. Updated the workflow comments to document this.

scmmishra:
Hey @infracore thanks for the PR, we already have setup [Skooma](https://github.com/skryukov/skooma) for validations, and are [progressively enabling](https://github.com/search?q=repo%3Achatwoot%2Fchatwoot%20conform_schema&type=code) it across controllers.

PR #13954, opened 2026-03-31
fix(sentry): syntaxError: No error message

iamsivin:
# Pull Request Template

## Description

This PR fixes a `SyntaxError: No error message` reported in Sentry.

**Root cause**
When a user logs in, the app fetches account data and sets the i18n locale from it. If the account API call fails (e.g., network issues or timeouts), the error is silently ignored and execution continues. As a result, the locale becomes `undefined`, which causes `vue-i18n` to throw a `NOT_SUPPORT_LOCALE_TYPE` error in any component using translations.

**When this can happen**

* Intermittent network issues during login
* Slow or failed responses from the account API
* Any scenario where account data is unavailable when `initializeAccount` completes

**What changed**
Locale updates now guard against falsy values. `setLocale` only updates `i18n.locale` when the incoming value is valid. If the account API fails and returns `undefined`, the update is skipped, preserving the existing locale set during initialization (default, or the `'en'` fallback from `createI18n`). This ensures the app never enters an invalid locale state.

**Reproduction steps**

1. Log in to the dashboard
2. Open DevTools → Network tab
3. Block requests matching `/api/v1/accounts`
4. Refresh the page
5. Observe the `SyntaxError: cannot support locale type` in the console, triggered by components using `t()`

**Screenshot**
image

Fixes https://linear.app/chatwoot/issue/CW-4128/syntaxerror-no-error-message

## Type of change

## How Has This Been Tested?

1. Log in to the dashboard
2. Open DevTools → Network tab
3. Block requests matching `/api/v1/accounts`.
4. Refresh the page
5. Observe any `SyntaxError: cannot support locale type` error in the console.

## Checklist:

### DIFF
--- a/app/javascript/dashboard/App.vue
+++ b/app/javascript/dashboard/App.vue
@@ -98,7 +98,7 @@ export default {
       mql.onchange = e => setColorTheme(e.matches);
     },
     setLocale(locale) {
-      this.$root.$i18n.locale = locale;
+      this.$root.$i18n.locale = locale || 'en';
     },
     async initializeAccount() {
       await this.$store.dispatch('accounts/get');
--- a/app/javascript/dashboard/App.vue
+++ b/app/javascript/dashboard/App.vue
@@ -98,7 +98,9 @@ export default {
       mql.onchange = e => setColorTheme(e.matches);
     },
     setLocale(locale) {
-      this.$root.$i18n.locale = locale || 'en';
+      if (locale) {
+        this.$root.$i18n.locale = locale;
+      }
     },
     async initializeAccount() {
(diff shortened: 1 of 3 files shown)


iamsivin:
@codex review

pranavrajs:
Does setting a fallback fix the issue? https://vue-i18n.intlify.dev/guide/essentials/fallback.html

iamsivin:
> Does setting a fallback fix the issue? https://vue-i18n.intlify.dev/guide/essentials/fallback.html

I don’t think fallback will help here. It only handles missing translation keys, not invalid locale values.

muhsin-k (approved):

PR #13958, opened 2026-03-31
fix: validate support_email format and handle parse errors in mailer

tds-1:
## Description

ConversationReplyMailer#parse_email calls Mail::Address.new(email_string).address without error handling. When an account's support_email contains a non-email string (e.g., "Smith Smith"), the mail gem raises Mail::Field::IncompleteParseError, crashing conversation transcript emails.

This has caused 1,056 errors on Sentry (EXTERNAL-CHATINC-JX) since Feb 25, all from a single account that has a name stored in the support_email field instead of a valid email address.

Closes https://linear.app/chatwoot/issue/CW-6687/mailfieldincompleteparseerror-mailaddresslist-can-not-parse-orsmith

## Type of change

## Checklist:


tds-1:
@codex please review

vishnu-narayanan (changes requested):
@tds-1 The mailer-level rescue is a good safety net, but the root cause here is that `support_email` on Account has no format validation. Any arbitrary string (like "Smith Smith") can be saved through the API without complaint.

1. Could we also add a model validation to reject bad input at save time?
2. One shared parser/fallback helper instead of duplicating the logic in two places
3. Add specs - at least one spec for malformed stored support_email values and one for supported display-name format

vishnu-narayanan (approved):

tds-1:
@codex please review

tds-1:
@codex please review

app/models/concerns/email_address_parseable.rb
```diff
@@ -0,0 +1,15 @@
...
+  private
+
+  def parse_email(email_string)
+    Mail::Address.new(email_string).address.presence || default_sender_email_address
+  rescue Mail::Field::ParseError, Mail::Field::IncompleteParseError
+    default_sender_email_address
+  end
+
+  def default_sender_email_address
+    Mail::Address.new(ENV.fetch('MAILER_SENDER_EMAIL', 'accounts@chatwoot.com')).address
```
sony-mathew, line 13: Should we change the default to `accounts+noreply@chatwoot.com` or so? Or are we expecting replies on this email?

spec/models/account_spec.rb
```diff
@@ -256,6 +256,29 @@
...
+        expect(account).to be_valid
+      end
+
+      it 'allows display-name format' do
+        account.support_email = 'Support Team <support@example.com>'
+        expect(account).to be_valid
+      end
+
+      it 'allows blank values' do
+        account.support_email = ''
```
sony-mathew, line 271: Do we allow support_email to be blank?

app/models/concerns/account_settings_schema.rb
```diff
@@ -0,0 +1,47 @@
+module AccountSettingsSchema
```
sony-mathew, line 1: Good abstraction! 👏🏾

sony-mathew (approved):
Approved with minor comments.

spec/models/account_spec.rb
tds-1: Yes. Validation returns early on blank and the getter falls back to MAILER_SENDER_EMAIL, which means blank is expected to have mailer_sender_email

app/models/concerns/email_address_parseable.rb
tds-1: There are few more places, where this default is used, we can create a separate task to evaluate change them all if required.

PR #13968, opened 2026-04-01
fix: log only on system api key failures

aakashb95:
Removes sentry flooding of unnecessary rubyllm logs of wrong API key. Logs only system api key error since it would be P0.


lib/llm/credential.rb
scmmishra, line 1: This does not have to be a separate file. You can define it inline. 
Also, the data structure is small enough to be handled inline as a hash, we don't need a separate `Data.define` declaration for this

lib/integrations/llm_base_service.rb
```diff
@@ -100,13 +101,14 @@ def make_api_call(body)
...
+    credential = llm_credential
 
-    Llm::Config.with_api_key(hook.settings['api_key'], api_base: api_base) do |context|
+    Llm::Config.with_api_key(credential.api_key, api_base: api_base) do |context|
       chat = context.chat(model: model)
       setup_chat_with_messages(chat, messages)
     end
   rescue StandardError => e
-    ChatwootExceptionTracker.new(e, account: hook.account).capture_exception
+    capture_llm_exception(e, credential: credential)
```
vishnu-narayanan, line 111: @aakashb95 This currently does an early return and we will loose any visibility into failures modes from `llm_base_service`. Could you log the error before doing an early return?

lib/llm/exception_trackable.rb
```diff
@@ -0,0 +1,9 @@
+module Llm::ExceptionTrackable
+  private
+
+  def capture_llm_exception(error, credential:)
+    return unless credential&.system?
+
+    ChatwootExceptionTracker.new(error, account: exception_tracking_account).capture_exception
```
vishnu-narayanan: This will silently skip any errors from llm_base_service. Chatwoot exception tracker did both logging and sentry event capture. Should we log the error to have some visibility?

Something like
```suggestion
    if credential&.system?
      ChatwootExceptionTracker.new(error, account: exception_tracking_account).capture_exception
    else
      Rails.logger.error(error)
    end
```

vishnu-narayanan (commented):
@aakashb95 Minor comments.

lib/integrations/llm_base_service.rb
aakashb95: shifted it to `capture_llm_exception`
so if system key error -> sentry, any other key error -> logger

enterprise/lib/captain/conversation_completion_service.rb
```diff
@@ -56,12 +56,10 @@ def default_incomplete_response(reason)
...
 
-  # Prefer the system API key over the account's OpenAI hook key.
   # This is an internal operational evaluation, not a customer-triggered feature,
-  # so it should not consume the customer's OpenAI credits on hosted platforms.
-  # Falls back to the account hook for self-hosted deployments without a system key.
-  def api_key
-    @api_key ||= system_api_key.presence || openai_hook&.settings&.dig('api_key')
+  # so it should always use the installation key.
+  def llm_credential
+    @llm_credential ||= system_llm_credential
```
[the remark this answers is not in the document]
aakashb95: this is a system level job which should always use system key

lib/llm/exception_trackable.rb
```diff
@@ -0,0 +1,11 @@
+module Llm::ExceptionTrackable
+  private
+
+  def capture_llm_exception(error, credential:)
+    if credential && credential[:source] == :system
+      ChatwootExceptionTracker.new(error, account: exception_tracking_account).capture_exception
+    else
+      Rails.logger.error(error)
```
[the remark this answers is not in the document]
aakashb95: this is intentional with this PR only raising sentry errors for system key. hook based errors would be addressed in a following PR.

vishnu-narayanan (approved):

PR #13972, opened 2026-04-01
feat(whatsapp): use unified webhook URL for all WhatsApp Cloud numbers

matheus-caldeira:
## Description

Use a single `/webhooks/whatsapp` endpoint as the default callback URL for all WhatsApp Cloud API channels, instead of per-number URLs. This simplifies multi-number WABA setups by sharing one webhook across all phone numbers under the same Business Account.

The legacy `/webhooks/whatsapp/:phone_number` route remains functional for backward compatibility with existing configurations and non-cloud providers (e.g., 360dialog).

### What changed

- **`config/routes.rb`**: Added generic `/webhooks/whatsapp` routes (GET + POST) without `:phone_number` param.
- **`whatsapp_controller.rb`**: `valid_token?` now supports both modes — lookup by `phone_number` param (legacy) or by matching `webhook_verify_token` across all `whatsapp_cloud` channels (default).
- **`webhook_setup_service.rb`**: `build_callback_url` always returns `/webhooks/whatsapp` (no phone number). `resolve_verify_token` reuses the verify token from an existing sibling channel on the same WABA.
- **`inbox.rb`** / **`health_service.rb`**: Webhook URL always uses the generic endpoint.
- **`webhook_teardown_service.rb`**: Skips teardown if sibling channels still exist on the same WABA.
- **`channel/whatsapp.rb`**: Added `waba_sibling_exists?` to detect other channels sharing the same `business_account_id`.

## Type of change

## How Has This Been Tested?

- **Controller spec** (`spec/controllers/webhooks/whatsapp_controller_spec.rb` — 12 tests):
  - Legacy route (`/webhooks/whatsapp/:phone_number`): verify with valid/invalid tokens, POST dispatching, inactive number filtering.
  - Default route (`/webhooks/whatsapp`): verify by token match across cloud channels, POST dispatching, ensures non-cloud providers are excluded, inactive number check is skipped when no phone_number param.

- **Setup service spec** (`spec/services/whatsapp/webhook_setup_service_spec.rb` — 27 tests):
  - Shared examples run all webhook URL scenarios for both default and legacy URLs.
  - Covers: phone registration, reauthorization flow, PIN reuse, sibling verify token reuse, error handling (API down, health service failure, registration failure), parameter validation.

- **Manual testing**: Received a live WhatsApp message via the generic `/webhooks/whatsapp` endpoint — contact created, conversation opened, message processed, notifications dispatched successfully.

## Checklist:

### DIFF
--- a/app/controllers/webhooks/whatsapp_controller.rb
+++ b/app/controllers/webhooks/whatsapp_controller.rb
@@ -15,9 +15,21 @@ class Webhooks::WhatsappController < ActionController::API
   private
 
   def valid_token?(token)
-    channel = Channel::Whatsapp.find_by(phone_number: params[:phone_number])
-    whatsapp_webhook_verify_token = channel.provider_config['webhook_verify_token'] if channel.present?
-    token == whatsapp_webhook_verify_token if whatsapp_webhook_verify_token.present?
+    return false if token.blank?
+
+    if params[:phone_number].present?
+      channel = Channel::Whatsapp.find_by(phone_number: params[:phone_number])
+      if channel.present?
+        expected = channel.provider_config['webhook_verify_token']
+        return false if expected.blank?
+
+        return ActiveSupport::SecurityUtils.secure_compare(token.to_s, expected.to_s)
+      end
+    else
+      return Channel::Whatsapp.where(provider: 'whatsapp_cloud')
+                              .exists?(["provider_config->>'webhook_verify_token' = ?", token])
+    end
+    false
   end
 
   def inactive_whatsapp_number?
--- a/app/jobs/webhooks/whatsapp_events_job.rb
+++ b/app/jobs/webhooks/whatsapp_events_job.rb
@@ -9,6 +9,11 @@ class Webhooks::WhatsappEventsJob < ApplicationJob
       return
     end
 
+    if inactive_whatsapp_number?(channel)
+      Rails.logger.warn("Rejected webhook for inactive WhatsApp number: #{channel.phone_number}")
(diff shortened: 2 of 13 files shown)


sojan-official:
@tds-1  @muhsin-k could you take a look when you get a chance?

nestordavalos:
I agree with the direction (single webhook for multi-number WABA), but I see a couple of important gaps before merge:

1. Blocking: `INACTIVE_WHATSAPP_NUMBERS` is bypassed on the new unified endpoint  
   - In `/webhooks/whatsapp`, `params[:phone_number]` is absent, so `inactive_whatsapp_number?` always returns `false`.  
   - This effectively disables the inactive-number kill switch for cloud channels using the new route.

2. Channel resolution should prioritize `phone_number_id` directly  
   - Current flow still derives from `display_phone_number` first and only then validates `phone_number_id`.  
   - For a unified webhook model, matching directly by `provider_config['phone_number_id']` is safer and avoids formatting mismatches on display numbers.

3. `resolve_verify_token` should guard against blank sibling tokens  
   - If sibling token is blank/nil (legacy/manual data), we currently persist/use it and can subscribe with an invalid token.  
   - Suggested fallback: if sibling token is blank, keep current channel token (or regenerate if needed), and do not persist blank values.

4. Missing coverage around sibling teardown behavior  
   - Since teardown semantics changed (`!waba_sibling_exists?`), we should add explicit specs for:
     - skip unsubscribe when siblings exist
     - unsubscribe when last sibling is removed

Issue #61, opened 2019-10-02
Fix "Rubocop/Metrics/MethodLength" issue in app/models/conversation.rb


pranavrajs:
Method has too many lines. [14/10]

https://codeclimate.com/github/chatwoot/chatwoot/app/models/conversation.rb#issue_5d87480c87cf1900010003c7

MatheusRich:
I'm refactoring this file! So I'll tackle the issues related to `conversation.rb`

Issue #56, opened 2019-10-02
Fix "Rubocop/Metrics/MethodLength" issue in app/builders/messages/message_builder.rb


pranavrajs:
Method has too many lines. [17/10]

https://codeclimate.com/github/chatwoot/chatwoot/app/builders/messages/message_builder.rb#issue_5d87480b87cf190001000226

sojan-official:
fixed in #100

Issue #52, opened 2019-10-02
Fix "Rubocop/Metrics/AbcSize" issue in app/controllers/passwords_controller.rb


pranavrajs:
Assignment Branch Condition size for set_headers is too high. [25.71/15]

https://codeclimate.com/github/chatwoot/chatwoot/app/controllers/passwords_controller.rb#issue_5d87480b87cf190001000335

jash-kothari:
Can i pick this up?

pranavrajs:
@jash-kothari Please go ahead, assign this issue to yourself and please start working on it, if you are interested. 

Please read the contributing guide as well. https://github.com/chatwoot/chatwoot/blob/master/docs/contributing.md

AlexisChavez97:
Is this still open?

PR #13978, opened 2026-04-02
feat: account enrichment using context.dev [UPM-27]

scmmishra:
## Account branding enrichment during signup

This PR does the following

### Replace Firecrawl with Context.dev

Switches the enterprise brand lookup from Firecrawl to Context.dev for better data quality, built-in caching, and automatic filtering of free/disposable email providers. The service interface changes from URL to email input to match Context.dev's email endpoint. OSS still falls back to basic HTML scraping with a normalized output shape across both paths.

The enterprise path intentionally does not fall back to HTML scraping on failure — speed matters more than completeness. We want the user on the editable onboarding form fast, and a slow fallback scrape is worse than letting them fill it in.

Requires `CONTEXT_DEV_API_KEY` in Super Admin → App Config. Without it, falls back to OSS HTML scraping.

### Add job to enrich account details

After account creation, `Account::BrandingEnrichmentJob` looks up the signup email and pre-fills the account name, colors, logos, social links, and industry into `custom_attributes['brand_info']`.

The job signals completion via a short-lived Redis key (30s TTL) + an ActionCable broadcast (`account.enrichment_completed`). The Redis key lets the frontend distinguish "still running" from "finished with no results."


app/controllers/api/v1/accounts_controller.rb
```diff
@@ -69,6 +70,13 @@ def update_active_at
 
   private
 
+  def enqueue_branding_enrichment
+    return if account_params[:email].blank?
+
+    Account::BrandingEnrichmentJob.perform_later(@account.id, account_params[:email])
+    Redis::Alfred.set(format(Redis::Alfred::ACCOUNT_ONBOARDING_ENRICHMENT, account_id: @account.id), '1', ex: 30)
```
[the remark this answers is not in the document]
scmmishra: Acceptable risk

enterprise/app/services/enterprise/website_branding_service.rb
```diff
@@ -1,112 +1,67 @@
...
+      address: brand['address']&.symbolize_keys,
+      colors: deep_symbolize_array(brand['colors']),
+      logos: deep_symbolize_array(brand['logos']),
+      socials: deep_symbolize_array(brand['socials']),
+      links: brand['links']&.symbolize_keys,
+      email: @email,
+      industries: deep_symbolize_array(brand.dig('industries', 'eic')),
+      stock: brand['stock']&.symbolize_keys,
+      is_nsfw: brand['is_nsfw'] || false
     }
```
sony-mathew: Instead of doing symbolize_keys on each line #52, #56, #59, can we do `deep_symbolize_keys` on this line #61 - at the end of this hash? What do you think? Will there be perf impact?

app/services/website_branding_service.rb
```diff
@@ -39,16 +44,45 @@ def fetch_page
...
-    doc.at_css('html')&.[]('lang')&.split('-')&.first&.downcase
+  def extract_colors(doc)
+    color = doc.at_css('meta[name="theme-color"]')&.[]('content')
+    return [] if color.blank?
+
+    [{ hex: color, name: nil }]
+  end
+
+  def extract_logos(doc)
+    favicon = doc.at_css('link[rel*="icon"]')&.[]('href')
```
sony-mathew, line 63: Does this take the first one if there are multiple elements available with this css query selection?

sony-mathew: I can see multiple elements selected on chatwoot website with this selector:
Image

spec/services/website_branding_service_spec.rb
```diff
@@ -2,19 +2,18 @@
...
     let(:url) { 'https://example.com' }
     let(:html_body) do
       <<~HTML
         <html lang="en">
         <head>
           <title>Acme Corp | Home</title>
           <meta property="og:site_name" content="Acme Corp" />
-          <meta property="og:image" content="https://example.com/og-image.png" />
           <meta name="theme-color" content="#FF5733" />
           <link rel="icon" href="/favicon.ico" />
```
sony-mathew, line 14: Can we add multiple elements here for icon to make sure selection works fine? Same for social links at the bottom of the page. Websites can repeat that in multiple places.

sony-mathew (commented):
Added some minor comments.

app/services/website_branding_service.rb
scmmishra: Yep, Nokigiri `at_css` gives the first item, it is equivalent to `css(...).first`

spec/services/website_branding_service_spec.rb
scmmishra: Done

enterprise/app/services/enterprise/website_branding_service.rb
scmmishra: Done, perf impact is negligible

sony-mathew (approved):
LGTM

app/jobs/account/branding_enrichment_job.rb
```diff
@@ -0,0 +1,32 @@
+class Account::BrandingEnrichmentJob < ApplicationJob
+  queue_as :low
+
+  def perform(account_id, email)
+    result = WebsiteBrandingService.new(email).perform
+    return if result.blank?
+
+    account = Account.find(account_id)
+    account.name = result[:title] if result[:title].present?
```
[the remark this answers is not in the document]
scmmishra: This is intentional. We start with an account name by default, this overrides that, we expect the encrichment to go through before that

PR #12835, opened 2025-11-11
feat: allow configuring attachment upload limit

sojan-official:
## Summary
- add a configurable MAXIMUM_FILE_UPLOAD_SIZE installation setting and surface it through super admin and global config payloads
- apply the configurable limit to attachment validations and shared upload helpers on dashboard and widget
- introduce a reusable helper with unit tests for parsing the limit and extend attachment specs for configurability

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_6912644786b08326bc8dee9401af6d0a)

PR #14012, opened 2026-04-07
fix: harden fetching on upload endpoint

scmmishra:


scmmishra:
@codex please review

lib/safe_fetch.rb
```diff
@@ -0,0 +1,87 @@
+require 'ssrf_filter'
+
+module SafeFetch
+  DEFAULT_MAX_BYTES = 10.megabytes
```
sony-mathew: Should we take this from ENV variable: `MAXIMUM_FILE_UPLOAD_SIZE`?

lib/safe_fetch.rb
```diff
@@ -0,0 +1,87 @@
...
+  DEFAULT_ALLOWED_CONTENT_TYPE_PREFIXES = %w[image/ video/].freeze
+  DEFAULT_OPEN_TIMEOUT = 5
+  DEFAULT_READ_TIMEOUT = 10
```
sony-mathew: Is this timeout values in seconds? If it's in milli seconds, then this seems very low. If this is in seconds, looks decent for default values. (Maybe reduce open timeout to 2 seconds and increase read timeout to 20 seconds)? I assume read time can vary based on the attachment size as well.

lib/safe_fetch.rb
```diff
@@ -0,0 +1,87 @@
...
+  class FileTooLargeError < Error; end
+  class UnsupportedContentTypeError < Error; end
+
+  def self.fetch(url,
+                 max_bytes: DEFAULT_MAX_BYTES,
+                 allowed_content_type_prefixes: DEFAULT_ALLOWED_CONTENT_TYPE_PREFIXES)
+    raise ArgumentError, 'block required' unless block_given?
+
+    uri = parse_and_validate_url!(url)
+    filename = File.basename(uri.path).presence || 'download'
```
sony-mathew: Shall we change the  default name from 'download' to more random & dynamic like "download-#{Time.now}"? In case if we ever need to debug, this might be helpful.

lib/safe_fetch.rb
```diff
@@ -0,0 +1,87 @@
...
+    tempfile = Tempfile.new('chatwoot-safe-fetch', binmode: true)
```
sony-mathew, line 27: Instead of 'chatwoot-safe-fetch', shall we pass the same `filename` variable as above as first parameter to Tempfile.new?

sony-mathew (changes requested):
Added minor comments.

lib/safe_fetch.rb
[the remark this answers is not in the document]
vishnu-narayanan: This is valid and already available as a GlobalConfig. We should reuse this.
https://github.com/chatwoot/chatwoot/pull/12835

vishnu-narayanan: Probably reuse #12835? `MAXIMUM_FILE_UPLOAD_SIZE`

scmmishra: Yep, on it

app/controllers/api/v1/accounts/upload_controller.rb
```diff
@@ -19,32 +19,12 @@ def create_from_file
...
-      create_and_save_blob(file, File.basename(uri.path), file.content_type)
+    SafeFetch.fetch(params[:external_url].to_s) do |result|
+      create_and_save_blob(result.tempfile, result.filename, result.content_type)
     end
-  rescue OpenURI::HTTPError => e
+  rescue SafeFetch::HttpError => e
     render_error("Failed to fetch file from URL: #{e.message}", :unprocessable_entity)
-  rescue SocketError
+  rescue SafeFetch::Error
     render_error('Invalid URL provided', :unprocessable_entity)
```
vishnu-narayanan: Everyone will get an `invalid URL provided 422`. Should we handle some know cases from safe_fetch separately?

vishnu-narayanan (commented):
Minor nitpicks

lib/safe_fetch.rb
scmmishra: It's in seconds. The open timeout of at-least 5 seconds seems better, in real-world, p90 would be well within a second, but for the p99s the 5 second gives enough buffer, default is 60 seconds at the core Net::HTTP library. 

I'll make the read timeout 20 seconds. That sounds good

lib/safe_fetch.rb
scmmishra: Done

lib/safe_fetch.rb
scmmishra: It can potentially create weird names like `image.png20260407-28034-sswsdr`
Tempfile by default will generate a file name with no extension and suffix the PID and some random strings to ensure there's no conflict, so this is safe as it is

scmmishra: Example: 
CleanShot 2026-04-07 at 16 36 25@2x

app/controllers/api/v1/accounts/upload_controller.rb
scmmishra: Fixed

scmmishra:
@codex please review

vishnu-narayanan (approved):
lgtm

sony-mathew (approved):
LGTM

PR #13994, opened 2026-04-05
feat: add WhatsApp BSUID and username support for contacts

nestordavalos:
Add support for WhatsApp Business Suite User ID (BSUID) and username as alternative contact identifiers. This enables managing WhatsApp contacts that may not have accessible phone numbers.

Changes:
- Add whatsapp_bsuid and whatsapp_username columns to contacts table
- Update contact merging to include WhatsApp fields
- Modify contact lookup to support finding by BSUID
- Enable WhatsApp source_id resolution using BSUID as fallback
- Add WhatsApp fields to contact search functionality
- Include WhatsApp attributes in permitted parameters
- Add database indexes and triggers for efficient searching

This enhancement allows the application to handle WhatsApp contacts more flexibly, particularly for business accounts where phone numbers may not be directly available.

image

# Pull Request Template

## Description

Please include a summary of the change and issue(s) fixed. Also, mention relevant motivation, context, and any dependencies that this change requires.
Fixes # (issue)

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #14025, opened 2026-04-08
feat(whatsapp): add coexistence app syncing for history and contacts

zubenkoivan:
Ports [chatwoot/chatwoot#12149](https://github.com/chatwoot/chatwoot/pull/12149) — a temporary integration of the WhatsApp Business App coexistence feature until it lands in upstream. Enables contact state sync, message echo syncing, and full conversation history import during embedded signup onboarding.

Closes N/A (temporary until upstream merges chatwoot/chatwoot#12149)

## What changed

- **New `Whatsapp::ContactSyncService`** — handles `smb_app_state_sync` webhooks to add/remove contacts from the WhatsApp Business App
- **New `Whatsapp::ConversationSyncService`** — handles `smb_message_echoes` (outgoing echo messages) and `history` (bulk conversation import) webhooks
- **`WhatsappEventsJob`** — routing extended: `smb_app_state_sync`, `smb_message_echoes`, and `history` fields are now dispatched to dedicated sync services instead of the generic message handler
- **`ActionCableListener`** — historical sync messages are excluded from real-time frontend broadcasts to prevent duplication
- **`EmbeddedSignupService` / `ChannelCreationService`** — when `is_business_app_onboarding` is true, automatically triggers state sync and history sync requests after channel creation
- **`FacebookApiClient`** — adds `request_state_sync` and `request_history_sync` via the `smb_app_data` endpoint
- **`WebhookSetupService`** — skips phone number registration step during Business App onboarding
- **`WhatsappEmbeddedSignup.vue`** — sets `is_business_app_onboarding: true` when the signup event is `FINISH_WHATSAPP_BUSINESS_APP_ONBOARDING`

## How to test

1. Set up a WhatsApp Cloud inbox using the embedded signup flow with a Business App account (event `FINISH_WHATSAPP_BUSINESS_APP_ONBOARDING`)
2. Verify contacts are synced from the Business App after channel creation
3. Verify existing conversation history appears as historical messages (not broadcast live)
4. Send a message from the WhatsApp Business App and verify it appears in Chatwoot as an outgoing echo message

## Notes

- One intentional deviation from the upstream PR: the controller keeps flat params (`params.permit(...)`) rather than requiring a nested `authorization` wrapper — the upstream diff had a bug where the frontend would have broken
- The `subscribed_fields` in `FacebookApiClient#override_waba_callback` does not yet include `smb_app_state_sync` and `history`; those will need to be added when enabling this feature end-to-end


nestordavalos:
In addition to this, we must also take into account this branch in order to proceed: https://github.com/chatwoot/chatwoot/pull/13994

PR #14026, opened 2026-04-08
feat: Add unified Call model for voice calling

muhsin-k:
Adds a Call model to track voice call state across providers (Twilio, WhatsApp). This replaces storing call data in conversation.additional_attributes and provides a foundation for call analytics multi-call-per-conversation support, and future voice providers.

Fixes https://linear.app/chatwoot/issue/CW-6850/add-unified-call-model-for-voice-calling


db/migrate/20260408170902_create_calls.rb
```diff
@@ -0,0 +1,33 @@
+class CreateCalls < ActiveRecord::Migration[7.0]
```
tds-1, line 1: Should we also store the “from” and “to” numbers? While they can be inferred from the call direction along with inbox or contact details, deriving them would require conditional joins, which could make querying call history more complex.

tds-1 (approved):

db/migrate/20260408170902_create_calls.rb
```diff
@@ -0,0 +1,33 @@
...
+    add_call_indexes
+  end
+
+  private
+
+  def add_call_indexes
+    add_index :calls, [:provider, :provider_call_id], unique: true
+    add_index :calls, [:account_id, :conversation_id]
+    add_index :calls, :message_id
+    add_index :calls, :meta, using: :gin
```
sony-mathew: @muhsin-k  Where/How are we planning to use this index on `meta`?

enterprise/app/models/call.rb
```diff
@@ -0,0 +1,24 @@
+class Call < ApplicationRecord
+  # All valid call statuses
+  STATUSES = %w[ringing in_progress completed no_answer failed].freeze
+  # Statuses where the call is finished and won't change again
+  TERMINAL_STATUSES = %w[completed no_answer failed].freeze
+
+  enum :provider, { twilio: 0, whatsapp: 1 }
+  enum :direction, { incoming: 0, outgoing: 1 }
+
+  belongs_to :account
```
sony-mathew, line 40: Should we add a belongs to contact also here?

sony-mathew (commented):
Added minor comments.

db/migrate/20260408170902_create_calls.rb
tds-1: To get the from_phone_number we have to query Inbox model if the direction is outgoing, else we have to query the contact table. 
Right now there is no use case to query calls based on phone numbers, but in case if there is any future use case, the query will be less complex

db/migrate/20260408170902_create_calls.rb
muhsin-k: We don't need this for now. Removed.

enterprise/app/models/call.rb
muhsin-k: We could it accessible via call.conversation.contact?. Right?

db/schema.rb
```diff
@@ -261,6 +261,28 @@
...
+    t.string "status", default: "ringing", null: false
+    t.integer "duration_seconds"
+    t.string "end_reason"
+    t.jsonb "meta", default: {}
+    t.text "transcript"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["account_id", "conversation_id"], name: "index_calls_on_account_id_and_conversation_id"
+    t.index ["message_id"], name: "index_calls_on_message_id"
+    t.index ["meta"], name: "index_calls_on_meta", using: :gin
```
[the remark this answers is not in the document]
muhsin-k: https://github.com/chatwoot/chatwoot/pull/14026/commits/eea203acf51262dd8fe2c35e152c648a68b2ea98

db/schema.rb
```diff
@@ -261,6 +261,28 @@
...
+    t.bigint "inbox_id", null: false
+    t.bigint "conversation_id", null: false
+    t.bigint "message_id"
+    t.bigint "accepted_by_agent_id"
+    t.string "provider_call_id", null: false
+    t.integer "provider", default: 0, null: false
+    t.integer "direction", null: false
+    t.string "status", default: "ringing", null: false
+    t.integer "duration_seconds"
+    t.string "end_reason"
```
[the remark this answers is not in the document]
muhsin-k: https://github.com/chatwoot/chatwoot/pull/14026/commits/eea203acf51262dd8fe2c35e152c648a68b2ea98

muhsin-k:
@codex please review

enterprise/app/models/call.rb
sony-mathew: Yes true, but if we were to list all the calls from a customer under the profile as a separate tab, we might need to go through a join with conversations. Wouldn't it be better to have the contact mapping also here? Might be helpful later also to have voice specific dashboards.

sony-mathew (approved):
Approved with a minor comment.

enterprise/app/models/call.rb
muhsin-k: Done https://github.com/chatwoot/chatwoot/pull/14026/commits/c5cebccf023de909e4bcbfa6c04c096168c137ce

PR #14040, opened 2026-04-10
fix: prevent lost custom_attributes updates from concurrent jsonb writes

tds-1:
## Linear ticket
https://linear.app/chatwoot/issue/CW-6834/billing-upgrade-didnt-work

## Description
A `customer.subscription.updated` Stripe webhook for account 76162 returned 200 OK but did not persist the new `subscribed_quantity`. Root cause: a race condition between the webhook handler and `increment_response_usage` (Captain usage counter), both doing read-modify-write on the `custom_attributes` JSONB column. The webhook wrote `quantity: 6`, then a concurrent `save` from `increment_response_usage` overwrote the entire hash with stale data — restoring `quantity: 5`.

Fix: use atomic `jsonb_set` so usage counter updates only touch the single key they care about, instead of rewriting the whole `custom_attributes` hash. `increment_custom_attribute` also performs the increment in SQL, making concurrent increments correct as well.

## Type of change

## How Has This Been Tested?

- New regression spec in `handle_stripe_event_service_spec.rb` that simulates concurrent webhook + `increment_response_usage` and asserts both `subscribed_quantity` and `captain_responses_usage` survive
- Existing account, billing, captain, and topup specs all pass locally

## Checklist:

### DIFF
--- a/enterprise/app/models/enterprise/account/plan_usage_and_limits.rb
+++ b/enterprise/app/models/enterprise/account/plan_usage_and_limits.rb
@@ -16,20 +16,15 @@ module Enterprise::Account::PlanUsageAndLimits # rubocop:disable Metrics/ModuleL
   end
 
   def increment_response_usage
-    current_usage = custom_attributes[CAPTAIN_RESPONSES_USAGE].to_i || 0
-    custom_attributes[CAPTAIN_RESPONSES_USAGE] = current_usage + 1
-    save
+    increment_custom_attribute(CAPTAIN_RESPONSES_USAGE)
   end
 
   def reset_response_usage
-    custom_attributes[CAPTAIN_RESPONSES_USAGE] = 0
-    save
+    update_custom_attribute(CAPTAIN_RESPONSES_USAGE, 0)
   end
 
   def update_document_usage
-    # this will ensure that the document count is always accurate
-    custom_attributes[CAPTAIN_DOCUMENTS_USAGE] = captain_documents.count
-    save
+    update_custom_attribute(CAPTAIN_DOCUMENTS_USAGE, captain_documents.count)
   end
 
   def email_transcript_enabled?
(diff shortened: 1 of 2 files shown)


tds-1:
@codex please review

enterprise/app/models/enterprise/account/plan_usage_and_limits.rb
```diff
@@ -130,6 +125,30 @@ def get_limits(limit_name)
...
   end
 
+  # Atomic jsonb_set to avoid clobbering concurrent writes to other custom_attributes keys.
+  def update_custom_attribute(key, value)
+    sql = <<~SQL.squish
+      UPDATE accounts
+      SET custom_attributes = jsonb_set(COALESCE(custom_attributes, '{}'), ARRAY[?], ?::jsonb)
+      WHERE id = ?
+    SQL
+    Account.connection.exec_update(Account.sanitize_sql_array([sql, key, value.to_json, id]))
```
sony-mathew: Hey @tds-1 , I think we will run into problems later if we introduce raw SQL statements like the above one, if and when we want to introduce sharding. 

It would be better to do something like the following:
```
Account.where(id: id).update_all([
      "custom_attributes = jsonb_set(COALESCE(custom_attributes, '{}'), ARRAY[:key], :value::jsonb)",
      { key: key, value: value.to_json }
    ])
```

This would help us with shard routing if and when we do that later.

enterprise/app/models/enterprise/account/plan_usage_and_limits.rb
```diff
@@ -130,6 +125,30 @@ def get_limits(limit_name)
...
+  def increment_custom_attribute(key)
+    sql = <<~SQL.squish
+      UPDATE accounts
+      SET custom_attributes = jsonb_set(
+        COALESCE(custom_attributes, '{}'), ARRAY[?],
+        (COALESCE((custom_attributes ->> ?)::int, 0) + 1)::text::jsonb
+      )
+      WHERE id = ?
+    SQL
+    Account.connection.exec_update(Account.sanitize_sql_array([sql, key, key, id]))
```
sony-mathew: Hey @tds-1 , I think we will run into problems later if we introduce raw SQL statements like the above one, if and when we want to introduce sharding. 

It would be better to do something like the following:
```
Account.where(id: id).update_all([
      <<-SQL.squish,
        custom_attributes = jsonb_set(
          COALESCE(custom_attributes, '{}'), 
          ARRAY[:key], 
          (COALESCE((custom_attributes ->> :key)::int, 0) + 1)::text::jsonb
        )
      SQL
      { key: safe_key }
    ])
```

This would help us with shard routing if and when we do that later.

sony-mathew (changes requested):
@tds-1 Added some comments around using the raw SQL and implication on sharding/partitioning.

enterprise/app/models/enterprise/account/plan_usage_and_limits.rb
tds-1: I have switched to Account.where(id: id).update_all(...) with named bindings as you suggested. One thing to flag: this trips Rails/SkipsModelValidations, so I had to add a # rubocop:disable Rails/SkipsModelValidations block.

enterprise/app/models/enterprise/account/plan_usage_and_limits.rb
tds-1: I have switched to Account.where(id: id).update_all(...) with named bindings as you suggested. One thing to flag: this trips Rails/SkipsModelValidations, so I had to add a # rubocop:disable Rails/SkipsModelValidations block.

sony-mathew (approved):
LGTM

PR #14059, opened 2026-04-14
fix(boot): resolve Zeitwerk double-load of enterprise concerns directory

ak-padma:
Self-hosted deploys with `config.eager_load = true` (production / staging) crash on boot with:

    uninitialized constant Captain::Assistant::Concerns (NameError)
      include Concerns::CaptainToolsHelpers

## Root cause

The glob on line 45 of `config/application.rb` adds every subdirectory under `enterprise/app` to the eager load paths — including `enterprise/app/models/concerns`. Zeitwerk registers the concern files with a namespace context derived from the directory tree, so `captain_tools_helpers.rb` becomes `Captain::Assistant::Concerns::CaptainToolsHelpers` rather than the intended top-level `::Concerns::CaptainToolsHelpers`.

When `Captain::Assistant` does `include Concerns::CaptainToolsHelpers`, Ruby's constant lookup searches `Captain::Assistant::Concerns` first, hits the glob-registered version, and raises `NameError`.

Rails avoids this for `app/models/concerns` by treating it as a root namespace automatically. The enterprise overlay doesn't get that treatment because it's wired through the glob, not through `config.paths`.

## What changed

- Exclude `/concerns` directories from the `enterprise/app/**` glob so Zeitwerk doesn't double-register them with a nested namespace.
- Add `enterprise/app/models/concerns` as an explicit root-namespace eager load path — the same treatment Rails gives `app/models/concerns`.

## How to reproduce

1. Clone with the enterprise directory present.
2. Set `config.eager_load = true` (or deploy to any environment that eager-loads).
3. Boot the Rails app — crashes immediately with the `NameError` above.

After this fix the concerns resolve as `::Concerns::CaptainToolsHelpers`, matching the `include` statements in `Captain::Assistant` and `Captain::Scenario`.

## Type of change

## Checklist:

### DIFF
--- a/enterprise/app/models/captain/assistant.rb
+++ b/enterprise/app/models/captain/assistant.rb
@@ -18,8 +18,10 @@
 #
 class Captain::Assistant < ApplicationRecord
   include Avatarable
-  include Concerns::CaptainToolsHelpers
-  include Concerns::Agentable
+  # Use absolute (::) reference to avoid Ruby resolving Concerns relative
+  # to the Captain namespace — which would look for Captain::Concerns and fail.
+  include ::Concerns::CaptainToolsHelpers
+  include ::Concerns::Agentable
 
   self.table_name = 'captain_assistants'
 
--- a/enterprise/app/models/captain/custom_tool.rb
+++ b/enterprise/app/models/captain/custom_tool.rb
@@ -28,8 +28,10 @@ class Captain::CustomTool < ApplicationRecord
 
   MAX_PER_ACCOUNT = 15
 
-  include Concerns::Toolable
-  include Concerns::SafeEndpointValidatable
+  # Use absolute (::) reference to avoid Ruby resolving Concerns relative
+  # to the Captain namespace — which would look for Captain::Concerns and fail.
+  include ::Concerns::Toolable
+  include ::Concerns::SafeEndpointValidatable
 
   self.table_name = 'captain_custom_tools'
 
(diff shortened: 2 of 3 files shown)


ak-padma:
## Updated approach — simpler, more targeted fix

**Previous approach:** Modified the `Dir` glob in `config/application.rb` to exclude `/concerns` directories and re-added `enterprise/app/models/concerns` as an explicit root. This changed the autoloader topology and could affect other enterprise extensions.

**New approach:** Added `::` prefix to `include Concerns::*` in the 3 Captain models (`Captain::Assistant`, `Captain::Scenario`, `Captain::CustomTool`). This is the minimal fix — 3 files, 6 lines changed, zero config changes.

### Root cause

Inside the `Captain::` namespace, `include Concerns::CaptainToolsHelpers` triggers Ruby's constant lookup to search `Captain::Concerns` first. When Zeitwerk has the `enterprise/app/models/concerns/` directory registered under the `enterprise/app/models` root, it tries to resolve that relative path and raises:

```
uninitialized constant Captain::Assistant::Concerns (NameError)
```

The `::` prefix forces Ruby to resolve from the top-level namespace directly, bypassing the `Captain::` namespace search. This is the standard Ruby pattern for disambiguating nested constant references.

### Why this is better
- No changes to `config/application.rb` — autoloader topology stays exactly as-is
- Only touches the 3 files that actually have the bug
- Same pattern Ruby developers use everywhere for this class of issue

sony-mathew:
@ak-padma , we are unable to reproduce this. We already have this `eager_load: true` in the configs as you can see here: https://github.com/chatwoot/chatwoot/blob/develop/config/environments/production.rb#L11

Can  you help us understand the problem a bit better? Can you also share the environment that you are trying to do the setup?

PR #14063, opened 2026-04-14
feat: Wire Twilio voice to unified Call model

muhsin-k:
Twilio voice calls now persist as rows in the unified `calls` table rather than state stuffed into `conversation.additional_attributes` and `conversation.identifier`. Each call gets its own `Call` record and voice_call message bubble, making call history, analytics, and multi-call-per-conversation flows natural.


muhsin-k:
@codex please review

PR #14068, opened 2026-04-15
fix: validate OpenAI hook credentials

aakashb95:
# Pull Request Template

## Description

- Validates openai key while configuring hooks
- added backfill logic

Fixes # (issue)

## Type of change

## How Has This Been Tested?

locally

CleanShot 2026-04-15 at 16 15 02@2x

CleanShot 2026-05-11 at 19 22 53@2x

## Checklist:


app/models/integrations/hook.rb
```diff
@@ -95,6 +96,26 @@ def validate_settings_json_schema
     errors.add(:settings, ': Invalid settings data') unless JSONSchemer.schema(app.params[:settings_json_schema]).valid?(settings)
   end
 
+  # TODO: When adding credential validation for other integrations (dialogflow, dyte, etc.),
+  # extract this into an app-level config flag in apps.yml instead of hardcoding app_id checks.
+  def validate_openai_api_key?
+    app_id == 'openai' && enabled? && openai_api_key_changed?
```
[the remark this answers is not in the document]
aakashb95: Cannot happen. In the UI there is a disconnect option which will remove the hook, and when they try to connect again, they will need to enter correct API key.

lib/integrations/openai/key_validator.rb
```diff
@@ -0,0 +1,26 @@
...
+    connection = Faraday.new do |f|
+      f.options.timeout = TIMEOUT_SECONDS
+      f.options.open_timeout = TIMEOUT_SECONDS
+    end
+
+    response = connection.get("#{api_base}/models") do |req|
+      req.headers['Authorization'] = "Bearer #{api_key}"
+    end
+
+    response.status != 401
```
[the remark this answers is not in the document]
aakashb95: 404 can only happen if the endpoint captured in super-admin is a wrong one - Out of scope of this PR
403 doesn't mean the key is wrong but the country/region isn't supported. https://developers.openai.com/api/docs/guides/error-codes

can be ignored

app/jobs/migration/validate_openai_hooks_job.rb
```diff
@@ -0,0 +1,22 @@
...
+    scope = Integrations::Hook.where(app_id: 'openai', status: 'enabled')
+    scope = scope.where(account_id: account.id) if account
+
+    stats = { checked: 0, destroyed: 0 }
+
+    scope.find_each do |hook|
+      stats[:checked] += 1
+      next if Integrations::Openai::KeyValidator.valid?(hook.settings&.dig('api_key'))
+
+      hook.destroy!
```
vishnu-narayanan, line 15: @aakashb95 Any specific reason to prefer `destroy`. Should we go with `disable` or `prompt_reauthorization!?` If we run this as-is, users will have no idea where their configured integration went. The user gets a "reconnect" prompt in the UI and knows exactly what to fix, vs their integration silently disappearing. Feels more consistent with the rest of the codebase.

app/jobs/internal/validate_openai_hooks_job.rb
```diff
@@ -0,0 +1,22 @@
+class Internal::ValidateOpenaiHooksJob < ApplicationJob
+  queue_as :low
```
vishnu-narayanan:
```suggestion
class Migration::ValidateOpenaiHooksJob < ApplicationJob
  queue_as :async_database_migration
```
Since this is a one-off task, it'd fit better as a migration job. The file should also move to app/jobs/migration/ to match the existing backfill convention (e.g.,Migration::BackfillCompaniesContactsCountJob).

app/models/integrations/hook.rb
```diff
@@ -29,6 +29,7 @@ class Integrations::Hook < ApplicationRecord
   validates :inbox_id, presence: true, if: -> { hook_type == 'inbox' }
   validate :validate_settings_json_schema
   validate :ensure_feature_enabled
+  validate :validate_openai_api_key, if: :validate_openai_api_key?
```
vishnu-narayanan, line 32: Can we scope this to fire only when api_key actually changed? Right now every save on an enabled openai hook hits openai with a 5s timeout, including status toggles and label_suggestion flips.

And when it does run, could we show something like "Validating with OpenAI..." near the submit button? The generic spinner is the same one we show for every integration, so the user has no idea we're waiting on openai and the UI just looks frozen for 5 seconds.

vishnu-narayanan (commented):
Looks good. Minor comments.

app/models/integrations/hook.rb
vishnu-narayanan: @iamsivin  what do you think?
cc: @sony-mathew

app/jobs/migration/validate_openai_hooks_job.rb
aakashb95: there's no in-place key update UI. the normal flow when a user wants to swap a key is Disconnect (DELETE) → Connect (CREATE). So `destroy!` matches what the user would do manually.

app/jobs/internal/validate_openai_hooks_job.rb
aakashb95: makes sense, will fix

app/jobs/migration/validate_openai_hooks_job.rb
aakashb95: We could wire up email to admins saying openai integration hook was `disconnected due to <reason>, please reconnect`. What do you think?

app/models/integrations/hook.rb
iamsivin: @vishnu-narayanan Sounds good. Alternatively, we can update the submit button text to show “Validating” with a disabled state and a loader.

app/jobs/migration/validate_openai_hooks_job.rb
vishnu-narayanan: > We could wire up email to admins saying openai integration hook was disconnected due to <reason>, please reconnect. What do you think?

+1

vishnu-narayanan: > there's no in-place key update UI. the normal flow when a user wants to swap a key is Disconnect (DELETE) → Connect (CREATE). So destroy! matches what the user would do manually.

True but from the UI that's an explicit user triggered event 😅

aakashb95: Image

how does this look?

app/models/integrations/hook.rb
aakashb95: Image

something like this?

app/jobs/migration/validate_openai_hooks_job.rb
```diff
@@ -0,0 +1,23 @@
+class Migration::ValidateOpenaiHooksJob < ApplicationJob
+  queue_as :async_database_migration
```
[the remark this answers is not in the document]
aakashb95: added a migration

app/models/integrations/hook.rb
sony-mathew: Yes, this last gif looks good.

app/jobs/migration/validate_openai_hooks_job.rb
sony-mathew: Looks good.

app/mailers/administrator_notifications/integrations_notification_mailer.rb
```diff
@@ -9,4 +9,10 @@ def dialogflow_disconnect
     subject = 'Your Dialogflow integration was disconnected'
     send_notification(subject)
   end
+
+  def openai_disconnect
+    subject = 'Your OpenAI integration was disconnected'
```
sony-mathew, line 14: Do we need to do I18n translation for this subject?

app/models/integrations/hook.rb
sony-mathew: @vishnu-narayanan  for this:

> Can we scope this to fire only when api_key actually changed? Right now every save on an enabled openai hook hits openai with a 5s timeout, including status toggles and label_suggestion flips.

I'm not sure if this is right, it can be revoked from openAI side also anytime right, so we can't rely on the API key changes alone in my opinion. what do you think?

config/locales/en.yml
```diff
@@ -123,6 +123,8 @@ en:
         sdp_offer_required: 'sdp_offer is required'
         contact_phone_required: 'Contact phone number is required'
         permission_request_failed: 'Failed to send call permission request'
+    openai:
+      invalid_api_key: 'OpenAI API key is invalid or revoked. Please check your key in your OpenAI dashboard and save again.'
```
sony-mathew: Instead of this "Please check your key in your OpenAI dashboard and save again.", can we just say "Please check your key in your OpenAI dashboard." - "save again" feels little off.

spec/jobs/migration/validate_openai_hooks_job_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'rails_helper'
```
sony-mathew, line 1: Do we generally write tests for migrations?

sony-mathew (approved):
LGTM. Approved with minor comments.

app/mailers/administrator_notifications/integrations_notification_mailer.rb
aakashb95: I don't see it everywhere, and AFAIK we rely on community to add translations, and this is sort of a one time job so it is okay to be in English?

config/locales/en.yml
aakashb95: Yeah makes sense, i wanted them to fix the issue and come and reconfigure with the latest key, but your version is cleaner.

spec/jobs/migration/validate_openai_hooks_job_spec.rb
aakashb95: More of a behaviour spec i had gpt write, we have written it for backfill and delete jobs before

app/mailers/administrator_notifications/integrations_notification_mailer.rb
sony-mathew: Sure, I can see other places where there is no translation in the same file. Tracking this holistically here: https://linear.app/chatwoot/issue/CW-7126/translations-for-subjects-in-this-file-appmailersadministrator

For now, this is fine.

PR #14077, opened 2026-04-16
fix: prevent Ctrl+Enter adding extra line break on send (Windows)

iamsivin:
# Pull Request Template

## Description

This PR includes,
On Windows, pressing **Ctrl+Enter** in the reply editor was inserting an unintended line break before sending. This led to two issues:

* **Unexpected blank lines**
  After adding a line break with Shift+Enter and removing it with Backspace, the editor looked correct. However, sending with Ctrl+Enter reintroduced a hidden break, resulting in an extra blank line in the final message.

* **Selected text being replaced**
  When text was selected and Ctrl+Enter was pressed, the selection was replaced with a line break instead of being sent.

Fixes https://linear.app/chatwoot/issue/CW-6840/newline-bug-in-the-editor

### **Cause**

Two keyboard handlers responded to **Ctrl+Enter** on Windows:

* ProseMirror (`Mod-Enter`) inserted a hard break
* ReplyBox (`$mod+Enter`) triggered send

The existing guard only checked `metaKey` (Cmd), so it never worked on Windows. As a result, a line break was inserted just before sending.

### **Solution**

Make the modifier check platform-aware so the editor correctly intercepts the send shortcut:

* Added `detectOS`, `isMac`, and `OS` constants
* Introduced `hasPressedMod` (uses `metaKey` on macOS, `ctrlKey` elsewhere)

This ensures Ctrl+Enter sends the message without modifying content, while keeping existing behavior unchanged.

**NB:** macOS behavior with Cmd+Enter remains unchanged

## Type of change

## How Has This Been Tested?

**Case 1: line break**

1. Type `hello`
2. Press Shift+Enter, then Backspace
3. Press Ctrl+Enter
   → Message contains an unexpected blank new line

**Case 2: Selection replaced**

1. Type two lines using Shift+Enter
2. Select text on the second line
3. Press Ctrl+Enter
   → Selected text is replaced and not sent

### Screencast

**Before**

https://github.com/user-attachments/assets/d6d285a9-260b-4711-8bbd-d0c8519e8d20

**After**

https://github.com/user-attachments/assets/c0ace1f7-5d22-44a2-8e08-22190ee21e61

## Checklist:


sony-mathew (approved):
LGTM. (Tested on windows and it's working well)

PR #14078, opened 2026-04-16
fix: prevent Country from being set to Zimbabwe in the contacts list

### LINKED ISSUE #7822: [CW-2663] Sending country_code on `Set the user in the widget` but not able to display on contant list
### Describe the bug

Hello,

I try to send attiribute with related method: window.$chatwoot.setUser(`<unique-identifier-key-of-the-user>`, {
...
}); 

Even if I added country_code I am still not getting country data on contact list 

### To Reproduce

Here is the payload from browser 
https://chatwoot-staging.letshego.com/api/v1/widget/contact/set_user
```
{
    "identifier": "5454123",
    "email": "qwweqwer@gmail.com",
    "name": "taleya",
    "phone_number": "+264676767222",
    "additional_attributes": {
        "country_code": "NA"
    }
}
```
Here is the screenshot from contact page /app/accounts/1/contacts?page=1
Capture

You can also find the response of contact list api for related user https://chatwoot-staging.letshego.com/api/v1/accounts/1/contacts?include_contact_inboxes=false&page=1&sort=-last_activity_a
```
{
    "additional_attributes": {
        "country_code": "NA"
    },
    "availability_status": "offline",
    "email": "mailmail@gmail.com",
    "id": 518,
    "name": "customNNNNN",
    "phone_number": "+264676767222",
    "identifier": "9019013333",
    "thumbnail": "",
    "custom_attributes": {
        "country": "NA",
        "textkey": "NA"
    },
    "conversations_count": 1,
    "last_activity_at": 1693299841,
    "created_at": 1693299829
}
```

### Expected behavior

_No response_

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

None

### Operating system

_No response_

### Browser and version

_No response_

### Docker (if applicable)

_No response_

### Additional context

_No response_

<sub>[CW-2663](https://linear.app/chatwoot/issue/CW-2663/sending-country-code-on-set-the-user-in-the-widget-but-not-able-to)</sub>

ko-vasilev:
## Description

This PR resolves an issue that was causing Zimbabwe country being selected in the Contacts list.
The root cause is objects in `countries.json` file are missing the `code` property which was causing the countries map to always have the last country in the list also map to an `undefined` key. In turn, this was causing an error when contact's `country` was undefined.

Additional thing to keep in mind: I am not sure if there is a case when contact's `country` property is used or if it can be removed as well, but in my Chatwoot installation contacts only have a `countryCode` property, and there is no way to set `country` from the widget sdk.

Fixes #7822

## Type of change

## How Has This Been Tested?

Simply running the vue application locally targeting my live chatwoot API.

## Checklist:

### DIFF
--- a/app/javascript/dashboard/components-next/Contacts/ContactsCard/ContactsCard.vue
+++ b/app/javascript/dashboard/components-next/Contacts/ContactsCard/ContactsCard.vue
@@ -51,7 +51,6 @@ const isFormInvalid = computed(() => contactsFormRef.value?.isFormInvalid);
 
 const countriesMap = computed(() => {
   return countries.reduce((acc, country) => {
-    acc[country.code] = country;
     acc[country.id] = country;
     return acc;
   }, {});


Mrsandeep27 (commented):
Good catch on the root cause — the reduce was writing `acc[undefined] = country` on every iteration because `countries.json` entries have `id: 'AF'` but no `code` field, so the last country (Zimbabwe) always won the undefined key and surfaced as the fallback for any unknown `country` / `countryCode` lookup.

**One thing to flag before merge**: the same exact bug exists in a second file that this PR doesn't touch.

`app/javascript/dashboard/modules/search/components/SearchResultContactItem.vue:52`
```js
    acc[country.code] = country;   // ← same undefined-key bug
    acc[country.id] = country;
    return acc;
  }, {});
});
```

Same symptom will happen in the search results card when the contact's country/countryCode is missing or unknown — Zimbabwe will appear. Worth removing the dead `acc[country.code]` line there in this PR too so the fix is complete.

**Nice-to-have follow-up (not in this PR)**: both of these `countriesMap` computations are effectively copy-pastes of each other. Extracting the reduce into a shared helper in `shared/constants/countries.js` (e.g. `export const byId = countries.reduce(...)` computed once at module load) would both fix this class of bug permanently and save the reduce cost per component mount.

Everything else looks correct — small, targeted change. Just include the `SearchResultContactItem.vue` line and it should be good to merge.

sojan-official (approved):
Thanks for working on this and contributing the fix. Really appreciate it.

PR #14093, opened 2026-04-18
fix: fall back to path-only attachment URL when FRONTEND_URL is blank (#14088)

### LINKED ISSUE #14088: bug: Upload Image not loading due to incorrect raiuls URL
### Describe the bug

When using a reverse proxy without configuring a URL address (FRONTEND_URL and ASSET_CDN_HOST are empty strings), users in Live Chat find that uploaded images fail to load. Similarly, in the Agent's conversation view, the images also fail to load.

After checking the Network tab via F12, I found that the issue is the missing FRONTEND_URL.
Upon inspection, the actual URL being accessed is:
`http:///rails/active_storage/disk/==--51175364a281c7f1682235bc5906e953a0511636/avatar_face_manga_large.png`

While configuring FRONTEND_URL would solve this, but the goal is to keep it unconfigured. 
It is missing the domain part. In the following code:
```
// code at /app/javascript/dashboard/components-next/message/bubbles/Image.vue
onMounted(() => {
  if (attachment.value?.dataUrl) {
    loadWithRetry(attachment.value.dataUrl);
  }
});
```
At this point, attachment.value.dataUrl is already `http:///rails/active_storage/xxxxx`.

### To Reproduce

Use an Nginx/Caddy proxy in front of Chatwoot
not set custom FRONTEND_URL \ ASSET_CDN_HOST 
Web user upload live chat image
Both web user and agent user cannot access the image for `http:///rails/xxxxx`

### Expected behavior

Can we simply process dataUrl here to use `/rails/active_storage/xxxxx` instead? If so, the frontend should be able to load it correctly.

### Environment

Docker

### Cloud Provider

None

### Platform

None

### Operating system

_No response_

### Browser and version

_No response_

### Docker (if applicable)

_No response_

### Additional context

_No response_

sanchitkaul:
## Summary
Fixes #14088. When Chatwoot is deployed behind a reverse proxy without \`FRONTEND_URL\` (and thus \`Rails.application.routes.default_url_options[:host]\`) configured, \`Attachment#file_url\` and \`Attachment#thumb_url\` call \`url_for(file)\` with \`host: nil\`. Rails emits \`http:///rails/active_storage/...\` (empty host), which the dashboard image loader and the live-chat widget cannot resolve.
## Fix
Route ActiveStorage URL generation in \`Attachment\` through a small private helper that returns a path-only URL when no host is configured. When \`FRONTEND_URL\` is set, behavior is unchanged. When it is blank, \`data_url\` becomes an origin-relative path (e.g. \`/rails/active_storage/disk/...\`), which the browser resolves against the current origin — correct for both the dashboard (same origin) and the widget iframe.
Fixes the reported problem in every consumer of \`dataUrl\` in one place: \`Image.vue\`, \`Video.vue\`, \`Audio.vue\`, \`File.vue\`, \`Embed.vue\`, \`GalleryView.vue\`, and the widget's \`AgentMessage.vue\` / \`UserMessage.vue\`, plus \`thumb_url\` which had the same bug.
Consumers that need absolute URLs (email mailers, Line sender, Captain OpenAI message builder) still require \`FRONTEND_URL\` — unchanged. \`ASSET_CDN_HOST\` is unaffected because it only applies to \`action_controller.asset_host\`, not ActiveStorage.
## Test plan
- [x] Added model specs covering configured and blank host cases for \`file_url\`, \`thumb_url\`, and \`push_event_data[:data_url]\`.
- [ ] Reproduce issue with \`FRONTEND_URL=""\`, confirm dashboard and widget image upload loads.
- [ ] Confirm with \`FRONTEND_URL=https://chatwoot.example.com\`, absolute URLs are still produced.
- [ ] Run \`bundle exec rspec spec/models/attachment_spec.rb\`.

### DIFF
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -51,7 +51,7 @@ class Attachment < ApplicationRecord
 
   # NOTE: the URl returned does a 301 redirect to the actual file
   def file_url
-    file.attached? ? url_for(file) : ''
+    file.attached? ? url_for_active_storage(file) : ''
   end
 
   # NOTE: for External services use this methods since redirect doesn't work effectively in a lot of cases
@@ -64,7 +64,7 @@ class Attachment < ApplicationRecord
     return '' unless file.attached? && image?
 
     begin
-      url_for(file.representation(resize_to_fill: [250, nil]))
+      url_for_active_storage(file.representation(resize_to_fill: [250, nil]))
     rescue ActiveStorage::UnrepresentableError => e
       Rails.logger.warn "Unrepresentable image attachment: #{id} (#{file.filename}) - #{e.message}"
       ''
@@ -76,6 +76,27 @@ class Attachment < ApplicationRecord
   end
 
   private
+
+  # Generates the URL for an ActiveStorage attachment or representation.
+  #
+  # Falls back to a path-only URL when FRONTEND_URL (and therefore
+  # `Rails.application.routes.default_url_options[:host]`) is blank.
+  # Without this fallback, `url_for` emits a malformed absolute URL like
+  # `http:///rails/active_storage/...` (empty host) which breaks image
+  # loading in the dashboard and in the live chat widget iframe when
+  # Chatwoot is deployed behind a reverse proxy without a configured
+  # FRONTEND_URL.
+  #
+  # Path-only URLs are resolved by the browser against the current
+  # document origin, which is the same origin as the Rails backend for
+  # both the dashboard and the widget iframe, so they load correctly.
+  def url_for_active_storage(resource)
+    url_for(resource, only_path: default_url_host.blank?)
+  end
+
+  def default_url_host
+    Rails.application.routes.default_url_options[:host]
+  end
 
   def metadata_for_file_type
     case file_type.to_sym
--- a/spec/models/attachment_spec.rb
+++ b/spec/models/attachment_spec.rb
@@ -32,6 +32,56 @@ RSpec.describe Attachment do
     end
   end
 
+  describe 'file_url host handling' do
+    let(:attachment) do
+      message.attachments.create!(account_id: message.account_id, file_type: :image).tap do |a|
+        a.file.attach(io: Rails.root.join('spec/assets/avatar.png').open, filename: 'avatar.png', content_type: 'image/png')
+      end
+    end
+
+    context 'when default_url_options[:host] is configured' do
+      before do
+        allow(Rails.application.routes).to receive(:default_url_options).and_return(host: 'chatwoot.example.com')
+      end
+
+      it 'returns an absolute file_url' do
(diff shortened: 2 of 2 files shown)


Mrsandeep27 (commented):
Nice, surgical fix — and really appreciate the commentary on the helper explaining *why* this fallback is safe. Read through the diff and the spec and have a few notes (all non-blocking):

1. **Correctness of the path-only fallback**
   Confirmed the behaviour matches what's described — `url_for(..., only_path: true)` emits `/rails/active_storage/...` which the browser resolves against the current document origin, and both the dashboard and the widget `iframe` load from the same Rails host, so the fix resolves cleanly for every consumer listed in the PR description (`Image.vue`, `Video.vue`, `Audio.vue`, `File.vue`, `Embed.vue`, `GalleryView.vue`, widget `AgentMessage.vue` / `UserMessage.vue`).

2. **Out-of-scope callers** (worth calling out in the PR body / a code comment on the helper)
   Every caller that serialises `data_url` into something the **browser** reads is fine. Callers that embed URLs into content consumed off-site — mailers, Line sender, OpenAI / Captain, push/webhook outbound bodies — still need an absolute URL and will silently start shipping path-only URLs if someone were to route them through this helper in the future. Worth explicitly documenting on `url_for_active_storage` that it's for same-origin consumption only, to stop future refactors from accidentally reusing it.

3. **`push_event_data[:data_url]`**
   Websocket/ActionCable consumers that live on another origin (some mobile SDK integrations, native desktop clients) would also start receiving path-only URLs. That's arguably correct — they should configure FRONTEND_URL — but it's a behaviour change worth noting in the PR description / changelog so self-hosted admins know the expected remediation.

4. **Minor test nit**
   `start_with('/rails/active_storage/')` will stay green even if the implementation ever accidentally emits `/rails/active_storage//...` or `/rails/active_storage/disk/blob_id/...` with a different signing format. If you want slightly stronger assertions:
   ```ruby
   expect(url).to match(%r{\A/rails/active_storage/(blobs|disk|representations)/})
   ```
   Totally optional — the existing check is fine for the specific regression.

5. **Tiny style**
   `default_url_host` is inlined once and used exactly once (`default_url_host.blank?`). Could be inlined back into the helper as `Rails.application.routes.default_url_options[:host].blank?`, but the named method reads clearly so no change needed.

LGTM overall — the regression spec is exactly what this class needs.

sojan-official:
Thanks for the detailed fix and tests here.

We’re going to close this one since `FRONTEND_URL` is a required environment variable for Chatwoot deployments. Falling back to relative attachment URLs when it is missing could make an invalid deployment configuration appear partially supported, which may create confusing behavior across the dashboard, widget, email, channel integrations, and other places that still require absolute URLs.

The right fix for this class of issue is to ensure `FRONTEND_URL` is configured correctly in the deployment environment.

Thanks again for investigating this and for the thoughtful implementation.

PR #14098, opened 2026-04-19
fix: prevent country defaulting to Zimbabwe in search result contacts

Mrsandeep27:
## Context

Same root-cause as #7822 / #14078, but in a second file that PR #14078 doesn't touch:
\`app/javascript/dashboard/modules/search/components/SearchResultContactItem.vue\`.

\`countries.json\` entries only have \`id\` (e.g. \`"AF"\`) — there is no \`code\` field. So:

\`\`\`js
const countriesMap = countries.reduce((acc, country) => {
  acc[country.code] = country;   // country.code is undefined → acc[undefined] = country
  acc[country.id] = country;
  return acc;
}, {});
\`\`\`

…overwrites \`acc[undefined]\` on every iteration, leaving whichever country comes last in the list (Zimbabwe, \`id: "ZW"\`). Later, the \`countryDetails\` lookup falls back to that value whenever the contact's \`country\` / \`countryCode\` is missing or unknown, and Zimbabwe is displayed incorrectly.

## Fix

One-line delete — drop the dead \`acc[country.code] = country\` write. Lookups by ISO code continue to work via \`country.id\`.

## Scope

Only the search-results card. The Contacts card is already being fixed in #14078 with the same one-line delete. It's worth patching this surface too so the same symptom doesn't reappear when the same contact is accessed via \`Ctrl+K\` search.

## Test plan

- [ ] Reproduce: search for a contact with \`country: null\` / \`countryCode: null\` — before this patch the flag renders as Zimbabwe; after, it renders as no country (current expected fallback).
- [ ] Search for a contact with a valid ISO \`countryCode\` (e.g. \`"IN"\`) — country still resolves correctly.
- [ ] Contacts list page (\`ContactsCard.vue\`, fixed in #14078) — no regression.

## Follow-up (not in this PR)

Both components keep rebuilding the same \`countriesMap\` per mount. A small \`shared/constants/countries.js\` export (\`export const byId = countries.reduce(...)\`, computed once at module load) would save the per-mount cost and centralise the shape so this bug can't return.


sojan-official (approved):
Thanks for working on this and contributing the fix. Really appreciate it.

PR #14100, opened 2026-04-19
fix(widget/campaign): fire campaigns when widget was opened programmatically (closes #14022)

### LINKED ISSUE #14022: Campaign not triggered when widget is opened via `toggle("open")`
### Describe the bug

**Root cause:**

```js
if (!isWidgetOpen) {
  // set active campaign
}
```

### To Reproduce

**Reproduction:**

```js
});
```

### Expected behavior

### Environment

Kubernetes

### Cloud Provider

None

### Platform

None

### Operating system

_No response_

### Browser and version

_No response_

### Docker (if applicable)

_No response_

### Additional context

_No response_

Mrsandeep27:
## Context

Closes #14022.

When an integrator calls \`window.\$chatwoot.toggle('open')\` inside their \`chatwoot:ready\` handler to pre-open the widget for the visitor, any campaign configured for that page never fires. The visitor lands, the widget is already open, the campaign timer eventually elapses, and \`startCampaign\` silently skips with:

\`\`\`js
// app/javascript/widget/store/modules/campaign.js
if (!isWidgetOpen) {          // ← true, because toggle('open') already flipped it
  ...
}
\`\`\`

But the visitor hasn't actually clicked the widget — the integrator's script did.

## Root cause

The flag \`isWidgetOpen\` conflates two fundamentally different states:

| Scenario | \`isWidgetOpen\` | Should campaign fire? |
|---|---|---|
| Visitor has ignored the widget | \`false\` | Yes |
| Visitor clicked the bubble, is actively chatting | \`true\` | No — would be annoying |
| Integrator pre-opened via SDK \`toggle('open')\` | \`true\` | **Yes** — visitor hasn't engaged yet |

The current guard treats case 2 and case 3 identically.

## Change

Differentiate user-initiated opens from SDK-initiated opens end-to-end and guard campaigns on the finer signal.

1. **\`app/javascript/sdk/bubbleHelpers.js\`**
   - \`onBubbleClick(props)\` — when invoked with no \`toggleValue\` it's the bubble DOM click listener (real user click); when invoked with an explicit boolean it comes from \`IFrameHelper.events.toggleBubble\` (SDK \`toggle()\`). Derives \`isUserInitiated = toggleValue === undefined\` and propagates it through \`handleBubbleToggle\`.

2. **\`app/javascript/sdk/IFrameHelper.js\`**
   - \`events.onBubbleToggle(isOpen, isUserInitiated = false)\` forwards the flag in the \`toggle-open\` \`postMessage\`.

3. **\`app/javascript/widget/App.vue\`**
   - \`toggle-open\` handler dispatches \`appConfig/toggleWidgetOpen\` with \`{ isWidgetOpen, isUserInitiated }\` instead of the bare boolean.

4. **\`app/javascript/widget/store/modules/appConfig.js\`**
   - New state \`isWidgetOpenedByUser\` (default \`false\`).
   - New getter \`getIsWidgetOpenedByUser\`.
   - \`toggleWidgetOpen\` action now accepts either the legacy boolean **or** \`{ isWidgetOpen, isUserInitiated }\` (backward-compat is preserved for any external caller I may have missed).
   - \`TOGGLE_WIDGET_OPEN\` mutation: sets \`isWidgetOpenedByUser = true\` only on user-initiated opens, clears it on every close.

5. **\`app/javascript/widget/store/modules/campaign.js\`**
   - \`startCampaign\` guard switches from \`isWidgetOpen\` to \`isWidgetOpenedByUser\`. Campaigns now fire when the widget was only pre-opened programmatically.

## Tests

\`app/javascript/widget/store/modules/specs/campaign/actions.spec.js\`:

- Migrated the two existing specs to the renamed flag.
- Added a regression spec that reproduces the issue — \`isWidgetOpen: true\`, \`isWidgetOpenedByUser: false\` — and asserts \`setActiveCampaign\` is committed.

## Test plan

- [ ] \`pnpm test app/javascript/widget/store/modules/specs/campaign\` green, including the new regression spec.
- [ ] Manually reproduce the issue steps:
  \`\`\`html
  <script>
      window.$chatwoot.toggle('open');
    });
  </script>
  \`\`\`
  With an active campaign configured for the page, confirm the campaign appears after its delay.
- [ ] Manually verify the original suppression still works: visitor clicks the bubble to open → wait past a campaign's delay → campaign should **not** fire.
- [ ] Close the widget and re-open via user click → \`isWidgetOpenedByUser\` goes back to \`true\`; no regression.
- [ ] Close the widget and re-open via \`toggle('open')\` → \`isWidgetOpenedByUser\` stays \`false\`; campaign would fire if delay hadn't been consumed yet.

### DIFF
--- a/app/javascript/sdk/IFrameHelper.js
+++ b/app/javascript/sdk/IFrameHelper.js
@@ -244,8 +244,8 @@ export const IFrameHelper = {
       removeUnreadClass();
     },
 
-    onBubbleToggle: isOpen => {
-      IFrameHelper.sendMessage('toggle-open', { isOpen });
+    onBubbleToggle: (isOpen, isUserInitiated = false) => {
+      IFrameHelper.sendMessage('toggle-open', { isOpen, isUserInitiated });
       if (isOpen) {
         IFrameHelper.pushEvent('webwidget.triggered');
       }
--- a/app/javascript/sdk/bubbleHelpers.js
+++ b/app/javascript/sdk/bubbleHelpers.js
@@ -71,8 +71,8 @@ export const createBubbleHolder = hideMessageBubble => {
   body.appendChild(bubbleHolder);
 };
 
-const handleBubbleToggle = newIsOpen => {
-  IFrameHelper.events.onBubbleToggle(newIsOpen);
+const handleBubbleToggle = (newIsOpen, isUserInitiated) => {
+  IFrameHelper.events.onBubbleToggle(newIsOpen, isUserInitiated);
 
   if (newIsOpen) {
     dispatchWindowEvent({ eventName: CHATWOOT_OPENED });
@@ -87,6 +87,11 @@ export const onBubbleClick = (props = {}) => {
   const { isOpen } = window.$chatwoot;
   if (isOpen === toggleValue) return;
 
+  // When `toggleValue` is undefined, this invocation came from the DOM
+  // click listener on the bubble (real user click). When it's an explicit
+  // boolean, the call originated from the SDK (e.g. window.$chatwoot.toggle).
+  const isUserInitiated = toggleValue === undefined;
+
   const newIsOpen = toggleValue === undefined ? !isOpen : toggleValue;
   window.$chatwoot.isOpen = newIsOpen;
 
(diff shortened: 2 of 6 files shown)


Mrsandeep27:
Hi team! Just following up — you merged #14098 and #14099 recently, so wanted to check if there's any feedback needed on this one. Happy to adjust anything!

PR #14122, opened 2026-04-21
fix: index email subject from conversation for outbound messages

vishnu-narayanan:
## Summary

Outbound email messages never populate `content_attributes.email.subject`. The subject lives only on the parent conversation's `additional_attributes.mail_subject`. The search index doc built by `Messages::SearchDataPresenter` pulls from the message-level field only, so outbound email subjects are unsearchable for accounts on the advanced_search (Elasticsearch) path.

This change makes the presenter fall back to `conversation.additional_attributes.mail_subject` when the message-level subject is blank. Inbound email messages keep their existing behavior (message-level subject takes precedence).

Closes https://linear.app/chatwoot/issue/CW-6877

## What changes

- `Messages::SearchDataPresenter#content_attributes_data` now falls back to the conversation's `mail_subject` when the message-level subject is blank.
- Added specs covering the fallback, precedence, and the neither-set case.

## What does not change

- No schema changes, no migrations, no backfill of existing OpenSearch documents.
- Searchkick's `after_commit :reindex_for_search` on `Message` will pick up the new field for all newly created or updated messages via the existing indexing path.
- Postgres search path (free accounts, fallback) is untouched. Broader subject search for those users is a separate follow-up.

### DIFF
--- a/app/presenters/messages/search_data_presenter.rb
+++ b/app/presenters/messages/search_data_presenter.rb
@@ -39,7 +39,8 @@ class Messages::SearchDataPresenter < SimpleDelegator
   end
 
   def content_attributes_data
-    email_subject = content_attributes.dig(:email, :subject)
+    email_subject = content_attributes.dig(:email, :subject).presence ||
+                    conversation.additional_attributes&.dig('mail_subject').presence
     return {} if email_subject.blank?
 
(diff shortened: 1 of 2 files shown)


sony-mathew (approved):
LGTM

PR #14140, opened 2026-04-23
fix: html-escape captured values in helpcenter article markdown embeds

vishnu-narayanan:
Embed templates interpolate regex captures from user-authored article URLs into HTML attribute values. CommonMark's angle-bracket link destination syntax allows characters that the capture regexes don't filter, so the unescaped substitution could produce malformed attribute output. Escaping at substitution time keeps the render deterministic regardless of the URL.

### How was this tested?
Added specs.

Fixes [CW-6934](https://linear.app/chatwoot/issue/CW-6934/)


sony-mathew:
@codex please review.

sony-mathew:
@vishnu-narayanan I'm not able to reproduce this issue in local. @iamsivin can you try once to reproduce the issue please?

sony-mathew:
Hey @vishnu-narayanan  on further tries, I was able to test this. The fix works.

sony-mathew (approved):
LGTM

iamsivin (approved):
@sony-mathew I can able to reproduce, and the fix works.
<details>
image
</details>

PR #14164, opened 2026-04-24
fix: use correct company_name key in contact CSV import

RamblesShambles:
Contact CSV import stores company data under `additional_attributes[:company]`, but the entire codebase (Contact model scopes, Vue frontend components, Enterprise services, Slack integration) uses `additional_attributes[:company_name]`. This means imported company data never appears in the UI.

**What changed:** Changed the key from `:company` to `:company_name` in `DataImport::ContactManager#update_contact_attributes`. The CSV column header remains `company` (the read is correct), only the storage key is fixed.

**How to reproduce:** Import a contacts CSV with a `company` column. View the imported contact — the company field will be empty in the UI. Check `additional_attributes` in the DB and you'll see `company` instead of `company_name`.

Closes #14096

### DIFF
--- a/app/services/data_import/contact_manager.rb
+++ b/app/services/data_import/contact_manager.rb
@@ -61,7 +61,7 @@ class DataImport::ContactManager
   def update_contact_attributes(params, contact)
     contact.name = params[:name] if params[:name].present?
     contact.additional_attributes ||= {}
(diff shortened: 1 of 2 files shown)


RamblesShambles:
Closing in favor of a different approach. Thanks for the review feedback.

PR #14166, opened 2026-04-24
fix: allow retry of failed messages by checking external_echo flag

### LINKED ISSUE #14120: Retrying a failed WhatsApp message silently succeeds locally but never re-sends to the provider
### Describe the bug

`POST /api/v1/accounts/:account_id/conversations/:conversation_display_id/messages/:id/retry` resets the message UI state to `sent` and clears `content_attributes.external_error`, but the subsequent `SendReplyJob` does not actually re-send the message to the underlying channel provider. Originally observed in production on WhatsApp Cloud; reproduced on 360dialog via a spec. The bug is in provider-agnostic code (`Base::SendOnChannelService`), so any channel that inherits from it is affected.

Agent-visible symptom: the message shows "single tick" (sent), no error, yet the recipient never receives it. Meta / the provider emits no further status webhooks because the retry never left Chatwoot.

### To Reproduce

1. Send a WhatsApp template message to a recipient where Meta will reject delivery. Easiest trigger is Meta error `131049` ("This message was not delivered to maintain healthy ecosystem engagement"). Any code path that leads to a `failed` status webhook works.
2. The outbound record is created with `status: sent` and a `source_id` (the wamid returned by Meta's send API).
3. Shortly after, Meta fires a `failed` status webhook; Chatwoot flips the message to `status: failed` with an `external_error` populated.
4. Agent clicks "Retry" in the UI → `POST /api/v1/accounts/:id/conversations/:cid/messages/:mid/retry`.
5. UI shows the message back at `sent` (single tick), error gone.
6. No HTTP call is made to the provider; `source_id` on the record is unchanged from the first failed attempt; no new status webhooks arrive.

### Expected behavior

After `/retry`, `SendReplyJob` hits the provider's send endpoint, a fresh external id (wamid) is stored on the message, and delivery is attempted again. If the provider rejects again, a new `failed` webhook should update the record with the new error.

### Environment

Docker

### Cloud Provider

### Platform

Browser

### Operating system

_Not relevant — backend bug independent of client OS._

### Browser and version

_Not relevant — backend bug independent of client browser._

### Docker (if applicable)

_Not relevant — bug is in application code, not environment-specific._

## Where the problem is

`app/controllers/api/v1/accounts/conversations/messages_controller.rb` — `#retry` clears `content_attributes` but not `source_id`, so the stale id from the failed first attempt is preserved on the record:

```ruby
def retry
  return if message.blank?
  service = Messages::StatusUpdateService.new(message, 'sent')
  service.perform
  message.update!(content_attributes: {})
  ::SendReplyJob.perform_later(message.id)
rescue StandardError => e
  render_could_not_create_error(e.message)
end
```

`app/services/base/send_on_channel_service.rb` — the send-path loop guard treats any `source_id.present?` as "originated from the channel, don't re-send," which short-circuits the retry:

```ruby
def perform
  validate_target_channel
  return unless outgoing_message?
  return if invalid_message?     # short-circuits the retry
  perform_reply
end

def outgoing_message_originated_from_channel?
  message.source_id.present?
end

def invalid_message?
  message.private? || outgoing_message_originated_from_channel?
end
```

This guard was intended to stop re-sending echo messages created from inbound webhooks (which also carry a `source_id`). It cannot distinguish those from failed-and-retried outgoing messages, so the retry becomes a no-op. The existing `# TODO: we need to refactor this logic as more integrations comes by` comment on the method acknowledges the same limitation.

Relevant adjacent file — echo messages are marked with `content_attributes['external_echo'] = true` in `app/services/whatsapp/incoming_message_base_service.rb` (`create_message`).

## Impact

- Agents see a successful retry in the UI; recipients never receive the message.
- Business-critical template flows that fail on delivery (error codes `131049`, `131026`, `131047`, quality rate limits, etc.) cannot be recovered through the retry endpoint.
- Applies to every channel that inherits from `Base::SendOnChannelService`.

## Reproducing with a spec

The bug can be reproduced deterministically. Adding the following example to `spec/services/whatsapp/send_on_whatsapp_service_spec.rb` inside the existing `when a valid message` context reproduces it — it asserts that the provider's send endpoint is hit, which currently fails because `SendReplyJob`'s send path short-circuits on the stale `source_id`:

```ruby
it 'resends to the provider on retry even when the failed attempt left a stale source_id' do
  message = create(:message,
                   message_type: :outgoing,
                   status: :sent,
                   source_id: 'wamid.STALE_FROM_FAILED_SEND',
                   additional_attributes: { template_params: template_params },
                   content: 'Your package will be delivered in 3 business days.',
                   conversation: conversation,
                   account: conversation.account)

  stub = stub_request(:post, 'https://waba.360dialog.io/v1/messages')
         .with(headers: headers, body: template_body.to_json)
         .to_return(status: 200, body: success_response, headers: { 'content-type' => 'application/json' })

  described_class.new(message: message).perform

  expect(stub).to have_been_requested
  expect(message.reload.source_id).to eq('123456789')
end
```

Current failure:

```
The request POST https://waba.360dialog.io/v1/messages ...
was expected to execute 1 time but it executed 0 times
```

RamblesShambles:
When a WhatsApp (or any channel) message fails and a user clicks retry, the guard clause in `Base::SendOnChannelService#outgoing_message_originated_from_channel?` blocks the re-send because `source_id` is still set from the failed attempt. The retry silently no-ops.

**Root cause:** The guard used `message.source_id.present?` to detect echo messages (messages from channel webhooks that shouldn't be re-sent). But failed outgoing messages also have a `source_id`, so retries get caught by the same guard.

**What changed:**
- `SendOnChannelService`: Changed guard to check `content_attributes['external_echo']` instead of `source_id`. The `external_echo` flag is already set by WhatsApp, Facebook, Instagram, and TikTok channel providers on echo messages.
- `Twitter::DirectMessageParserService`: Added `external_echo: true` to echo messages for consistency with other providers.

This is safe because failed outgoing messages never have the `external_echo` flag, and all channels that create echo messages already set it (Twitter now included).

**How to reproduce:** Send a WhatsApp message that fails. Click retry. Before this fix, nothing happens. After, the message is re-sent.

Closes #14120


muhsin-k:
The retry controller already clears the stale provider source_id before enqueueing delivery. That fix was merged on August 13 and is included in this PR’s current head. It addresses the original issue, [#14120](https://github.com/chatwoot/chatwoot/issues/14120), which is closed.

PR #14171, opened 2026-04-24
fix: include private messages in conversation event webhook payload

### LINKED ISSUE #14023: Private message webhook sends incorrect content

cahrll:
## Description

When an agent adds a private note to a conversation, the `message_created` webhook fires correctly (because `Message#webhook_sendable?` permits private messages), but the attached `conversation.messages` array silently substitutes the previous *public* message. Consumers listening for private-note events therefore act on the wrong content — a pain point for routing, audit logging, and summarization pipelines.

The root cause is the `Message.chat` scope (`where.not(message_type: :activity).where(private: false)`) applied in `Conversations::EventDataPresenter#webhook_push_messages`. The private filter is appropriate for in-app chat rendering but wrong for outbound webhook payloads.

This PR drops only the private filter from the webhook presenter, keeping the activity filter so system messages continue to be skipped. The sibling `push_messages` path (ActionCable) is intentionally unchanged — it broadcasts to both agent and contact tokens, and widening it could leak private content to contact sockets. The scope of the fix is limited to the one webhook call site.

Fixes #14023

## Type of change

## How Has This Been Tested?

**Automated.** Added two RSpec cases to the existing `#webhook_data` describe block: one asserting that the latest private message is returned in the payload with `private: true`; another regression-guarding that activity messages are still excluded. Pre-existing webhook specs continue to pass.

**Manual.** Reproduced on `develop` by configuring an outbound webhook, sending a public reply, then adding a private note. Before the fix the webhook payload's `conversation.messages[0].content` is the earlier public reply; after the fix it is the private note, with `private: true`.

## Checklist:

### DIFF
--- a/app/presenters/conversations/event_data_presenter.rb
+++ b/app/presenters/conversations/event_data_presenter.rb
@@ -33,7 +33,9 @@ class Conversations::EventDataPresenter < SimpleDelegator
   end
 
   def webhook_push_messages
-    [messages.where(account_id: account_id).chat.last&.webhook_push_event_data].compact
+    # `.chat` hides private messages, but webhooks must include them (see Message#webhook_sendable?).
+    last_message = messages.where(account_id: account_id).where.not(message_type: :activity).last
+    [last_message&.webhook_push_event_data].compact
   end
 
   def push_meta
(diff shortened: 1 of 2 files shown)


sony-mathew:
Thanks for the contribution. We already have an earlier open PR tracking this same issue: [14148](https://github.com/chatwoot/chatwoot/pull/14148).

PR #14186, opened 2026-04-25
fix(admin): properly disable unchecked feature flags on account update

### LINKED ISSUE #14064: Super Admin account feature update does not disable unchecked features
### Describe the bug

When updating account features from the Super Admin account edit screen, checked features are sent in the enabled_features payload as feature_x => "true", but unchecked features are omitted from the request entirely. In practice, this can cause previously enabled features to remain active after saving, even when they were unchecked in the form. This behavior was reproduced with multiple features, not just one specific feature.

### To Reproduce

1 - Open Super Admin > Accounts > Edit for an existing account.
2 - Uncheck one or more features that are currently enabled.
3 - Save the account.
4 - Inspect the PATCH request payload and/or verify the account behavior afterward.

Observed result
The PATCH request includes only the still-checked features inside enabled_features. Unchecked features are absent from the payload rather than being sent as disabled. In the reproduced request, several unchecked features disappeared from the submitted enabled_features hash.

### Expected behavior

The saved account feature set should exactly match the selected checkboxes in the form. If a feature is unchecked, it should be disabled/removed from the account after save.

### Environment

Linux VM

### Cloud Provider

### Platform

Browser

### Operating system

4.12.0

### Browser and version

Chrome - latest

### Docker (if applicable)

Chatwoot: self-hosted
Deployment: Linux VM
Cloud provider: OVH Cloud
Docker: 24+
Docker Compose: v2
Platform used for reproduction: browser-based Super Admin UI

### Additional context

This looks like a mismatch between standard HTML checkbox behavior and backend feature-set recomposition. Checked boxes are submitted, but unchecked boxes are not. If the controller only processes submitted keys and does not rebuild the final feature set from the full managed list, disabled features may remain enabled after update. The reproduction logs show that checked features are present in enabled_features, while unchecked ones are simply absent from the payload.

RamblesShambles:
## Summary

Unchecking feature flags in the Super Admin account edit screen had no effect. Previously enabled features remained active even after being unchecked and saved.

**Root cause**: Two related issues:
1. The controller only passed checked features to the model (`if params[:enabled_features].present?`), so unchecked checkboxes (which HTML omits from form submissions) were never communicated as "disabled"
2. The `selected_feature_flags=` setter was only defined in the Enterprise module with a `super` call to a non-existent base method

**Fix**:
- Added `selected_feature_flags=` to the `Featurable` concern that disables all features first, then enables only the selected ones
- Changed the controller to always pass the feature list using `params.fetch(:enabled_features, {})` instead of conditional assignment

Closes #14064

## How to reproduce

1. Open Super Admin > Accounts > Edit for an existing account
2. Enable several features and save
3. Uncheck some of those features and save again
4. **Before fix**: unchecked features remain enabled
5. **After fix**: unchecked features are properly disabled

## What changed

- `app/models/concerns/featurable.rb` — Added `selected_feature_flags=` method that resets all features then enables only the provided list
- `app/controllers/super_admin/accounts_controller.rb` — Changed from conditional assignment to `params.fetch` so empty submissions (all unchecked) are handled correctly

## Trade-offs

The fix disables all features before re-enabling the selected ones, which means every flag is written on every save. This is a conscious choice: the alternative (computing a diff of checked vs. unchecked) would require knowing the previous state and is more error-prone. Since feature flag updates are infrequent admin operations, the overhead is negligible.

### DIFF
--- a/app/controllers/super_admin/accounts_controller.rb
+++ b/app/controllers/super_admin/accounts_controller.rb
@@ -36,7 +36,7 @@ class SuperAdmin::AccountsController < SuperAdmin::ApplicationController
   def resource_params
     permitted_params = super
     permitted_params[:limits] = permitted_params[:limits].to_h.compact
-    permitted_params[:selected_feature_flags] = params.fetch(:enabled_features, {}).keys.map(&:to_sym)
+    permitted_params[:selected_feature_flags] = params.fetch(:enabled_features, {}).select { |_k, v| v == "true" }.keys.map(&:to_sym)
     permitted_params
   end
 
--- a/app/controllers/super_admin/accounts_controller.rb
+++ b/app/controllers/super_admin/accounts_controller.rb
@@ -36,7 +36,7 @@ class SuperAdmin::AccountsController < SuperAdmin::ApplicationController
   def resource_params
     permitted_params = super
     permitted_params[:limits] = permitted_params[:limits].to_h.compact
-    permitted_params[:selected_feature_flags] = params[:enabled_features].keys.map(&:to_sym) if params[:enabled_features].present?
+    permitted_params[:selected_feature_flags] = params.fetch(:enabled_features, {}).keys.map(&:to_sym)
     permitted_params
   end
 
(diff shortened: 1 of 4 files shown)


sony-mathew:
Thanks for the contribution. We already have an earlier open PR tracking this same issue: #14179.

Issue #9907, opened 2024-08-07
Incorrect "company" field name in contact import handling


dukhaSlayer:
### Describe the bug

It looks like Contacts import functionality uses wrong attribute name for import at DataImport::ContactManager#update_contact_attributes.
It goes:
```
    contact.additional_attributes[:company] = params[:company] if params[:company].present?
```
While the rest of the product uses `additional_attributes[:company_name]`.

### To Reproduce

1. Import CSV file with a contact containing both `company` and `company_name` fields
2. Neither of these fields end up shoing in the `Company` field on the Chatwoot UI

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

None

stonecharioteer:
This issue is being closed due to lack of activity for over a year. [For more details, visit this discussion.](https://github.com/orgs/chatwoot/discussions/12810)

PR #14193, opened 2026-04-25
fix: normalize company field key in contact CSV import (#14096)

Mpho-Mulaudzi:
## Fix: Normalize company field key during contact CSV import

Closes #14096

---

### What was wrong

`DataImport::ContactManager#update_contact_attributes` stored the company
value under `additional_attributes[:company]`, but the rest of the
application (UI serializer, contact show view) reads
`additional_attributes[:company_name]`. This caused imported company names
to be silently written to a key that nothing reads, making the Company field
always appear blank after import.

This was a pre-existing inconsistency first reported in #9907.

---

### What this PR does

- Writes the company value to `additional_attributes[:company_name]` — the
  canonical key used throughout the app.
- Accepts **both** `company_name` and `company` as CSV column names, with
  `company_name` taking precedence if both are present. This is a backwards-
  compatible change: existing CSVs using the `company` column now work, and
  CSVs using the correct `company_name` column continue to work.

---

### What I considered but didn't do

- **Migrating existing records**: contacts already imported with the wrong
  key will still have stale data. A data migration could fix this but is
  outside the scope of this bug fix and carries risk. I've left a comment in
  the code noting this.
- **Changing the CSV template**: the sample CSV could be updated to use
  `company_name`. That's a docs/template change that can be handled
  separately.

---

### Testing

Added four spec cases covering: legacy `company` key, canonical
`company_name` key, both present (precedence check), and verifying no stray
`company` key is written.

Run: `bundle exec rspec spec/services/data_import/contact_manager_spec.rb`

### DIFF
--- a/app/services/data_import/contact_manager.rb
+++ b/app/services/data_import/contact_manager.rb
@@ -58,11 +58,15 @@ class DataImport::ContactManager
 
   private
 
-  def update_contact_attributes(params, contact)
+  def update_contact_attributes
     contact.name = params[:name] if params[:name].present?
-    contact.additional_attributes ||= {}
-    contact.additional_attributes[:company] = params[:company] if params[:company].present?
-    contact.additional_attributes[:city] = params[:city] if params[:city].present?
-    contact.assign_attributes(custom_attributes: contact.custom_attributes.merge(params.except(:identifier, :email, :name, :phone_number)))
-  end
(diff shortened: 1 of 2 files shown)


sojan-official:
closing as this is a duplicate of : https://github.com/chatwoot/chatwoot/pull/14099

PR #14194, opened 2026-04-25
fix: trigger widget campaign when already open

### LINKED ISSUE #14022: Campaign not triggered when widget is opened via `toggle("open")`

Siruyy:
## Summary
- Fix `startCampaign` so it still activates a campaign when the widget is already open.
- This restores triggered campaigns for sites that call `window.$chatwoot.toggle('open')` on `chatwoot:ready` before a scheduled campaign fires.
- Add a regression test for `isWidgetOpen: true`.

Fixes #14022

## Test plan
- [x] `pnpm test app/javascript/widget/store/modules/specs/campaign/actions.spec.js`
- [ ] Manual QA (not performed in this PR)

### DIFF
--- a/app/javascript/widget/store/modules/campaign.js
+++ b/app/javascript/widget/store/modules/campaign.js
@@ -102,23 +102,14 @@ export const actions = {
       );
     }
   },
-  startCampaign: async (
-    {
-      commit,
-      rootState: {
-        appConfig: { isWidgetOpen },
-      },
-    },
-    { websiteToken, campaignId }
-  ) => {
-    // Disable campaign execution if widget is opened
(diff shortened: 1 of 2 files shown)


sony-mathew:
Thanks for the contribution. We already have an earlier open PR tracking this same issue: #14100.

Issue #11961, opened 2025-07-16
availability_status parameter ignored on /api/v1/accounts/{account_id}/agents/{id} PATCH endpoint


bravin99:
### Describe the bug

When interacting with the endpoint
PATCH /api/v1/accounts/{account_id}/agents/{id}
on my Chatwoot deployment, including the body parameter
"availability_status": "available"
does not update the agent’s availability status.

Investigation:

When updating the logged-in user’s availability status via the UI, the request is sent to a different endpoint:
PATCH /api/v1/profile/availability

This endpoint accepts a different body parameter ("availability": "available") and updates the status successfully.

Trying the same parameter i.e ("availability": "available") and not ("availability_status": "available") on /api/v1/accounts/{account_id}/agents/{id} works well.

### To Reproduce

Try making the request below to update status for an account agent:
curl --request PATCH \
  --url https://app.chatwoot.com/api/v1/accounts/{account_id}/agents/{id} \
  --header 'Content-Type: application/json' \
  --header 'api_access_token: <api-key>' \
  --data '{
  "role": "agent",
  "availability_status": "available",
  "auto_offline": true
}'

### Expected behavior

Either the /api/v1/accounts/{account_id}/agents/{id} endpoint should support updating 'availability_status', or

The API documentation should clarify that availability updates must be done using 'availability'.

### Environment

Docker

### Cloud Provider

None

### Platform

None

Joewus12345:
> ### Describe the bug
> When interacting with the endpoint PATCH /api/v1/accounts/{account_id}/agents/{id} on my Chatwoot deployment, including the body parameter "availability_status": "available" does not update the agent’s availability status.
> 
> Investigation:
> 
> When updating the logged-in user’s availability status via the UI, the request is sent to a different endpoint: PATCH /api/v1/profile/availability
> 
> This endpoint accepts a different body parameter ("availability": "available") and updates the status successfully.
> 
> Trying the same parameter i.e ("availability": "available") and not ("availability_status": "available") on /api/v1/accounts/{account_id}/agents/{id} works well.
> 
> ### To Reproduce
[...]

PR #14205, opened 2026-04-25
Support availability_status when updating agents

kabamechack:
## Summary
Fixes agent availability updates when the API receives `availability_status`.

Previously, the agents update endpoint allowed `availability`, but not `availability_status`. Because of that, when API users sent `availability_status`, Rails ignored the parameter and the agent availability did not update.

## What changed
- Added `availability_status` to permitted agent params
- Added `availability_status` to account user update attributes
- Added a request spec that sends `availability_status: 'busy'` and checks that the response and account user record are updated correctly

## Why this fixes the issue
The controller now permits `availability_status`, so the value is passed into the account user update instead of being ignored.

The new request spec covers the exact case from the issue: updating an agent using `availability_status`.

## Testing
- Ran `git diff --check`
- Result: no whitespace errors

Note:
I could not run RSpec locally because Bundler is not available in my Windows environment.

### DIFF
--- a/app/controllers/api/v1/accounts/agents_controller.rb
+++ b/app/controllers/api/v1/accounts/agents_controller.rb
@@ -68,11 +68,11 @@ class Api::V1::Accounts::AgentsController < Api::V1::Accounts::BaseController
   end
 
   def account_user_attributes
-    [:role, :availability, :auto_offline]
+    [:role, :availability, :availability_status, :auto_offline]
   end
 
   def allowed_agent_params
-    [:name, :email, :role, :availability, :auto_offline]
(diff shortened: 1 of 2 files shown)


sojan-official:
Thanks for taking the time to work on this. We reviewed the API contract and the earlier maintainer decision in #11961 / #14319.

We are going to keep `availability_status` as a read-only response field because it represents the effective status derived from configured availability, auto-offline, and current presence. The writable field remains `availability`.

I opened #14533 to correct the Swagger docs so integrations use the right request field. Closing this PR in favor of that docs/API contract fix.

PR #14215, opened 2026-04-26
fix(integrations): redact sensitive hook settings in dashboard API

### LINKED ISSUE #14042: GET /api/v1/accounts/:id/integrations/apps exposes raw integration secrets to the frontend
### Describe the bug

The authenticated dashboard endpoint `GET /api/v1/accounts/:id/integrations/apps` returns raw secret values in the JSON response sent to the browser.

In our self-hosted instance, the response includes sensitive values inside `payload[].hooks[].settings`, including:
- OpenAI `api_key`
- Google Translate service account credentials
- Google service account `private_key`

These values are visible from the browser network tab and can be retrieved by any authenticated user who can access this endpoint. Secrets should not be serialized into frontend API responses.

### To Reproduce

1. Use a self-hosted Chatwoot instance with at least one configured integration that stores secrets, for example OpenAI or Google Translate.
2. Sign in to the dashboard with an authenticated user account.
3. Open the browser devtools network tab.
4. Trigger or manually request:
   `GET /api/v1/accounts/:account_id/integrations/apps`
5. Inspect the JSON response body.
6. Observe that `payload[].hooks[].settings` contains raw secret values such as `api_key` and service account credentials.

Example request:
```bash
curl 'https://your-chatwoot-instance/api/v1/accounts/1/integrations/apps' \
  -H 'uid: <uid>' \
  -H 'client: <client>' \
  -H 'access-token: <token>' \
  -H 'token-type: Bearer' \
  -H 'expiry: <expiry>' \
  -H 'Accept: application/json'
```

Example problematic response shape:
```json
{
  "payload": [
    {
      "id": "openai",
      "hooks": [
        {
          "settings": {
            "api_key": "sk-..."
          }
        }
      ]
    },
    {
      "id": "google_translate",
      "hooks": [
        {
          "settings": {
            "project_id": "...",
            "credentials": {
              "type": "service_account",
              "private_key": "-----BEGIN PRIVATE KEY-----..."
            }
          }
        }
      ]
    }
  ]
}
```

### Expected behavior

The frontend should receive only non-sensitive integration metadata required to render the settings UI.

Secret fields should be omitted or redacted server-side before serialization. Raw credentials such as API keys and private keys should never be returned in browser-facing JSON responses.

### Environment

Docker

### Cloud Provider

None

### Platform

Browser

### Operating system

Ubuntu Linux

### Browser and version

chrome

### Docker (if applicable)

Docker version 29.2.1, build a5c7197

Client: Docker Engine - Community
 Version:           29.2.1
 API version:       1.53
 Go version:        go1.25.6
 Git commit:        a5c7197
 Built:             Mon Feb  2 17:17:42 2026
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          29.2.1
  API version:      1.53 (minimum version 1.44)
  Go version:       go1.25.6
  Git commit:       6bc6209
  Built:            Mon Feb  2 17:17:42 2026
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v2.2.1
  GitCommit:        dea7da592f5d1d2b7755e3a161be07f43fad8f75
 runc:
  Version:          1.3.4
  GitCommit:        v1.3.4-0-gd6d73eb8
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
Client: Docker Engine - Community
 Version:    29.2.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.31.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v5.0.2
    Path:     /usr/libexec/docker/cli-plugins/docker-compose
  model: Docker Model Runner (Docker Inc.)
    Version:  v1.0.10
    Path:     /usr/libexec/docker/cli-plugins/docker-model

Server:
 Containers: 7
  Running: 6
  Paused: 0
  Stopped: 1
 Images: 5
 Server Version: 29.2.1
 Storage Driver: overlayfs
  driver-type: io.containerd.snapshotter.v1
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: dea7da592f5d1d2b7755e3a161be07f43fad8f75
 runc version: v1.3.4-0-gd6d73eb8
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.14.0-37-generic
 Operating System: Ubuntu 25.04
 OSType: linux
 Architecture: x86_64
 CPUs: 24
 Total Memory: 92.08GiB
 Name: **
 ID: **
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
 Firewall Backend: iptables
Docker Compose version v5.0.2

### Additional context

This was discovered while investigating a conversation loading issue, but the secret exposure is not conversation-specific.

The issue appears to come from the integrations apps response itself, not from browser tooling. In our case, the response included live integration credentials, which required immediate credential rotation after discovery.

This should likely be treated as a security issue, and it may be worth auditing other integration/configuration endpoints for similar secret leakage.

blank0810:
The authenticated integrations endpoint `GET /api/v1/accounts/:id/integrations/apps` was serializing the full `settings` JSONB blob of every connected hook to administrators. For self-hosted instances configuring OpenAI, Google Translate, or Dialogflow, this returned raw API keys and Google service-account credentials (including `private_key`) in the browser network response. Even authorized admins should not have configured secrets re-served to their browser, where they can be observed by extensions, screen-share tools, or any script running on the page.

The codebase already had the necessary infrastructure: each integration in `config/integration/apps.yml` declares a `visible_properties` allowlist (today consumed only by `MultipleIntegrationHooks.vue` for table column rendering). This change wires the same allowlist through the backend serializer so secrets remain on the server while the dashboard UI keeps the data it actually needs.

## Closes

- Closes #14042

## How to reproduce (before this fix)

1. On a self-hosted instance, sign in as an administrator
2. Configure the OpenAI, Google Translate, or Dialogflow integration with real credentials
3. Open browser devtools Network tab and trigger `GET /api/v1/accounts/:account_id/integrations/apps`
4. Observe `payload[].hooks[].settings` containing raw `api_key` / `credentials.private_key` values

## What changed

- Adds `Integrations::Hook#filtered_settings`, returning `settings` filtered by the integration's `visible_properties` allowlist when defined, and the unmodified `settings` otherwise. Integrations that do not declare `visible_properties` (Slack, Linear, Notion, Webhook, Dashboard apps, Shopify) keep their existing serialization to avoid breaking their dashboards.
- Switches `app/views/api/v1/models/_hook.json.jbuilder` to render `filtered_settings` for admins.
- Removes `api_key` from OpenAI's `visible_properties`. OpenAI is `allow_multiple_hooks: false`, so its UI uses `SingleIntegrationHooks.vue`, which does not read `visible_properties` — there is no UI impact from this removal.
- Updates the two existing controller specs that previously asserted admins received the full `settings` blob, now asserting `api_key` is filtered out. Adds one new spec verifying that an admin response contains only the visible (non-sensitive) keys.

## Trade-offs considered

- **Allowlist (`visible_properties`) vs. denylist of known-sensitive keys**: chose the allowlist because the convention already exists in the codebase and is opt-in per integration. A denylist would silently miss any new sensitive key a future integration introduces.
- **Backfill / migration of existing data**: not needed. Stored data is unchanged; only the API response is filtered.
- **Tests for Dialogflow and Google Translate**: not added in this PR. The factory only ships an `:openai` trait, and adding factory traits for the others would expand scope. The OpenAI test exercises the same code path; the same allowlist mechanism handles every integration that declares `visible_properties`.

### DIFF
--- a/app/models/integrations/hook.rb
+++ b/app/models/integrations/hook.rb
@@ -64,6 +64,19 @@ class Integrations::Hook < ApplicationRecord
     update(status: 'disabled')
   end
 
+  # Returns settings filtered by the integration's `visible_properties` allowlist
+  # so secrets configured under `settings` (api keys, service-account credentials, etc.)
+  # are not serialized back to the dashboard. Integrations that do not declare
+  # `visible_properties` retain the previous behavior to preserve compatibility.
+  def filtered_settings
+    return settings if app.blank?
+
+    visible_properties = app.params[:visible_properties]
+    return settings if visible_properties.blank?
+
+    settings.to_h.slice(*visible_properties.map(&:to_s))
+  end
+
   def process_event(_event)
     # OpenAI integration migrated to Captain::EditorService
     # Other integrations (slack, dialogflow, etc.) handled via HookJob
--- a/app/views/api/v1/models/_hook.json.jbuilder
+++ b/app/views/api/v1/models/_hook.json.jbuilder
@@ -5,5 +5,5 @@ json.inbox resource.inbox&.slice(:id, :name)
 json.account_id resource.account_id
 json.hook_type resource.hook_type
 
-json.settings resource.settings if Current.account_user&.administrator?
+json.settings resource.filtered_settings if Current.account_user&.administrator?
 json.reference_id resource.reference_id if Current.account_user&.administrator?
--- a/config/integration/apps.yml
+++ b/config/integration/apps.yml
@@ -55,7 +55,7 @@ openai:
         'validation': '',
       },
     ]
-  visible_properties: ['api_key', 'label_suggestion']
+  visible_properties: ['label_suggestion']
 linear:
   id: linear
   logo: linear.png
--- a/spec/controllers/api/v1/accounts/integrations/apps_controller_spec.rb
+++ b/spec/controllers/api/v1/accounts/integrations/apps_controller_spec.rb
@@ -66,7 +66,7 @@ RSpec.describe 'Integration Apps API', type: :request do
         end
       end
 
-      it 'will return sensitive information for openai app for admins' do
+      it 'redacts sensitive openai hook settings even for admins' do
         openai = create(:integrations_hook, :openai, account: account)
         get api_v1_account_integrations_apps_url(account),
             headers: admin.create_new_auth_token,
@@ -75,7 +75,22 @@ RSpec.describe 'Integration Apps API', type: :request do
         expect(response).to have_http_status(:success)
 
         app = response.parsed_body['payload'].find { |int_app| int_app['id'] == openai.app.id }
-        expect(app['hooks'].first['settings']).not_to be_nil
+        settings = app['hooks'].first['settings']
+        expect(settings).not_to be_nil
+        expect(settings).not_to have_key('api_key')
+      end
+
+      it 'serializes only the visible settings keys for admins' do
+        hook = create(:integrations_hook, :openai, account: account, settings: { api_key: 'sk-secret', label_suggestion: true })
+        get api_v1_account_integrations_apps_url(account),
+            headers: admin.create_new_auth_token,
+            as: :json
+
+        expect(response).to have_http_status(:success)
+
+        app = response.parsed_body['payload'].find { |int_app| int_app['id'] == hook.app.id }
+        settings = app['hooks'].first['settings']
+        expect(settings).to eq('label_suggestion' => true)
       end
     end
   end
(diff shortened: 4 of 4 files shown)


aakashb95:
Thanks for your contribution.
This was addressed in https://github.com/chatwoot/chatwoot/pull/14147, so I will close this PR.

PR #14242, opened 2026-04-26
fix: incorrect country mapping causing Zimbabwe fallback in contact views

fortunkapishe-tech:
## What this fixes
Fixes incorrect rendering of "Zimbabwe" for contacts without a valid country.

## Root cause
Country mapping relied on `country.code`, which could lead to incorrect or missing lookups. Additionally, lookup prioritised a less reliable key first.

## Solution
- Updated country mapping to use `country.dial_code`
- Adjusted lookup order to prioritise exact matches (`countryCode`) before fallback
- Applied fix consistently across affected components

## Notes
Minimal change to avoid unintended side effects while improving correctness.


sojan-official:
Thanks for working on this. We are going to proceed with the earlier community PR for the contact-list surface (#14078), and handle the same issue in search results separately via #14098.

This PR overlaps with the same country-code/Zimbabwe fallback bug, but also changes lookup behavior with `dial_code`, which is broader than we need for this fix. I’m closing it as a duplicate.

Issue #10973, opened 2025-02-25
Bug: Files with Accents in Names Downloaded with Incorrect Extensions


psilva-sv:
### Describe the bug

I've encountered a bug where files received through Chatwoot that have accents in their filenames are being returned with incorrect extensions when users attempt to download them. This issue causes confusion and requires manual renaming to access the file properly.

### To Reproduce

When a file that contains an accent (e.g., "Número") in its name is recieved, the downloaded file gets an incorrect extension like .docx-filename= instead of the proper .docx. The same happens in other filetypes as PDFs for example.

### Expected behavior

The behavior should be the same as if the file had no accent, and the file extension should stay the same. The accented character can either have the accent removed, keeping the letter, or be replaced with an underscore.

### Environment

Kubernetes

### Cloud Provider

None

### Platform

Browser

### Browser and version

Happens in every browser

### Additional context

Prints:

![Image](https://github.com/user-attachments/assets/f57db06b-af99-4347-a91d-95eec0aa212e)

![Image](https://github.com/user-attachments/assets/ca3fcabc-2e90-4e6c-a6fd-256b752ca2be)

jmelati:
I had already opened an issue on this subject, but it was closed. More information here:
#10527

jmelati:
Here another #10602

cospin:
Hello, same issue here

orsaada:
I have the same issue, it maybe happens only on the chatwoot mobile.
I received a file _-12.pdf-filename_=

muxa-mva:
For my Chatwoot instance hosted on DigitalOcean VPS with Easypanel I was able to find a solution by creating a bind mount to override a source file. Here's how it worked:
1. The issue of attachments that have special characters in the file name creating this bug can be resolved by editing the _download_attachment_file_ function in the **app/services/whatsapp/incoming_message_whatsapp_cloud_service.rb** file.
2. I created a local copy of the file and edited it
Changed from this:
```
class Whatsapp::IncomingMessageWhatsappCloudService < Whatsapp::IncomingMessageBaseService
  private

  def processed_params
    @processed_params ||= params[:entry].try(:first).try(:[], 'changes').try(:first).try(:[], 'value')
  end

  def download_attachment_file(attachment_payload)
    url_response = HTTParty.get(inbox.channel.media_url(attachment_payload[:id]), headers: inbox.channel.api_headers)
    # This url response will be failure if the access token has expired.
[...]

PR #14245, opened 2026-04-26
Fix: preserve original filename for WhatsApp attachments with special…

matthewdekock501:
## What

Fixes an issue where files received via WhatsApp with special characters (e.g. accented filenames) are downloaded with incorrect file extensions.

## Why

The filename for downloaded attachments was being inferred from the HTTP response headers or URL, which can be unreliable when special characters are present. This resulted in malformed filenames such as:

Número.pdf → Número.pdf-filename=

## How

Instead of relying on the downloaded file’s inferred metadata, this change explicitly sets the filename using the value provided in the WhatsApp attachment payload (`attachment_payload[:filename]`), which serves as the source of truth.

### DIFF
--- a/app/services/whatsapp/incoming_message_whatsapp_cloud_service.rb
+++ b/app/services/whatsapp/incoming_message_whatsapp_cloud_service.rb
@@ -8,13 +8,26 @@ class Whatsapp::IncomingMessageWhatsappCloudService < Whatsapp::IncomingMessageB
     @processed_params ||= params[:entry].try(:first).try(:[], 'changes').try(:first).try(:[], 'value')
   end
 
-  def download_attachment_file(attachment_payload)
-    url_response = HTTParty.get(
-      inbox.channel.media_url(attachment_payload[:id]),
-      headers: inbox.channel.api_headers
-    )
-    # This url response will be failure if the access token has expired.
-    inbox.channel.authorization_error! if url_response.unauthorized?
-    Down.download(url_response.parsed_response['url'], headers: inbox.channel.api_headers) if url_response.success?
+def download_attachment_file(attachment_payload)
+  url_response = HTTParty.get(
+    inbox.channel.media_url(attachment_payload[:id]),
+    headers: inbox.channel.api_headers
+  )
+  # This url response will be failure if the access token has expired.
+  inbox.channel.authorization_error! if url_response.unauthorized?
+
+  return unless url_response.success?
+
+  file = Down.download(
+    url_response.parsed_response['url'],
+    headers: inbox.channel.api_headers
+  )
+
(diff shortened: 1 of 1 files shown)


Gasgeber:
Thanks for digging into this — we hit the exact same bug in production and arrived
at the same root cause independently. Confirming it's still present on `develop`:
`incoming_message_base_service.rb:144` continues to use `attachment_file.original_filename`.

This also looks like the underlying cause of #10973.

For context on the mechanism: when a WhatsApp document name contains non-ASCII or
special characters, Meta's media response includes an RFC 5987 `filename*=` parameter
alongside the plain `filename=`. `Down` folds the beginning of that parameter into
`original_filename`, so the stored name ends up as e.g. `robert_mu_ller_1_.pdf-filename*=`
(DB blob: `…pdf;filename*=`). On Windows the file no longer carries a `.pdf` extension
and won't open; macOS opens it anyway because it sniffs the type, which is why this is
easy to miss in testing.

Your fix is the right idea. Three things I'd suggest, based on running a variant of it
in production since early August:

**1. Apply it in the base service rather than the Cloud service.**
`download_attachment_file` is defined twice — the Cloud override in
`incoming_message_whatsapp_cloud_service.rb`, and the generic one in
`incoming_message_service_helpers.rb` used by the 360dialog path. But `filename:` is
assigned in exactly one place, `incoming_message_base_service.rb:144`. Changing it
there fixes both providers with a single edit and avoids the
`define_singleton_method(:original_filename)` indirection.

**2. Also strip the leaked parameter, not just prefer the payload name.**
`attachment_payload[:filename]` is only present for *documents*. Images, audio, video
and stickers have no filename in the payload, so they still fall through to
`original_filename` and keep the corrupted value. Stripping `;filename*=…` (and any
surrounding quotes) covers those too:

```ruby
def attachment_filename(attachment_payload, attachment_file)
  raw = attachment_payload[:filename].presence || attachment_file.original_filename.to_s
  cleaned = raw.split(/;\s*filename\*?=/i).first.to_s.gsub(/\A["']+|["']+\z/, '').strip
  cleaned.presence || attachment_file.original_filename
end
```

Worth noting for reviewers: document.filename is client-supplied, but it flows into
ActiveStorage, which sanitises filenames on write — so this doesn't introduce a path
traversal vector.

3. A spec would help this land.
We have ~10 examples covering both providers, documents vs. non-document media,
the filename*= leak, quoted names, and the empty-name fallback.

Two small things blocking CI on the current diff: the method body lost its indentation
(it now sits at column 0 inside the class), and the file is missing a trailing newline —
Rubocop will flag both.

Happy to help however is easiest: I can push these changes to your branch, or open a
follow-up PR that credits you as the original author. Just let me know which you prefer —
I'd rather not open a competing PR.

PR #14249, opened 2026-04-26
Fix: Incorrectly rendered country for contacts missing country attribute

### LINKED ISSUE #14238: Contact List View incorrectly renders "Zimbabwe" for contacts missing a country attribute but possessing a country_code
### Describe the bug

In the Contact List view (via ContactsCard.vue), a contact's location is incorrectly rendered as "Zimbabwe" when the contact record contains a country_code (e.g., "US") but does not contain a country name string in its additional_attributes.

This occurs because a misconfigured computed property (countriesMap) evaluates countriesMap[undefined] to the last element of the countries array (Zimbabwe). The detail view does not seem to suffer from this bug and renders the country correctly.

### To Reproduce

1. Create a contact using the Chatwoot API (or an external integration) and provide a country_code (e.g., "US") in the additional_attributes payload, but omit the country key.
2. Navigate to the Contacts tab to view the contact list.
3. Observe that the contact's location badge shows the Zimbabwe flag and reads "Zimbabwe" (or "City, Zimbabwe"), despite the underlying code being "US".
4. Click on the contact to open the Detail View, and observe that it correctly renders as "United States".

### Expected behavior

_No response_

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

None

### Operating system

_No response_

### Browser and version

_No response_

### Docker (if applicable)

_No response_

### Additional context

The root cause is located in app/javascript/dashboard/components-next/Contacts/ContactsCard/ContactsCard.vue.

The computed map is:

    acc[country.code] = country;  // <-- Bug is here
    acc[country.id] = country;
    return acc;
  }, {});
});

The imported countries.js array objects do not possess a code property (they use id and dial_code). Therefore, country.code evaluates to undefined on every iteration. Because Zimbabwe is the final element in the array, acc[undefined] permanently maps to the Zimbabwe object.

The execution lookup: 

const { country, countryCode, city } = attributes;

// ...
const activeCountry =
  countriesMap.value[country] || countriesMap.value[countryCode];

Because the contact payload only provided a country_code, the destructured country variable is undefined. Javascript evaluates the left side of the OR statement (countriesMap.value[undefined]), which returns the Zimbabwe object, entirely bypassing the correct countryCode lookup on the right.

Suggested Fix:

1. Remove acc[country.code] = country; from the reducer, OR change it to acc[country.dial_code] = country;.
2. Swap the short-circuit order when resolving activeCountry so it prioritizes the exact identifier match before attempting a fuzzy/undefined match:

javascript
const activeCountry = countriesMap.value[countryCode] || countriesMap.value[country];

Zandilev6:
Fixes #14238  incorrect country rendering in the contact list.

Changes made
Added guard checks before accessing countriesMap
Ensured undefined values are not used as lookup keys
Reasoning

The issue occurred because the code attempted to access countriesMap with an undefined key, leading to incorrect fallback behavior.

This fix ensures only valid country identifiers are used.

Notes

Kept the solution minimal and consistent with existing logic.

### DIFF
--- a/app/javascript/dashboard/components-next/Contacts/ContactsCard/ContactsCard.vue
+++ b/app/javascript/dashboard/components-next/Contacts/ContactsCard/ContactsCard.vue
@@ -63,8 +63,13 @@ const countryDetails = computed(() => {
 
   if (!country && !countryCode) return null;
 
-  const activeCountry =
-    countriesMap.value[country] || countriesMap.value[countryCode];
+  let activeCountry = null;
+
+  if (country && countriesMap.value[country]) {
+    activeCountry = countriesMap.value[country];
+  } else if (countryCode && countriesMap.value[countryCode]) {
+    activeCountry = countriesMap.value[countryCode];
+  }
 
   if (!activeCountry) return null;
 


sojan-official:

PR #14261, opened 2026-04-27
fix(contacts): resolve Zimbabwe fallback bug when countryCode is set …

### LINKED ISSUE #14238: Contact List View incorrectly renders "Zimbabwe" for contacts missing a country attribute but possessing a country_code

jackie20:
…but country is absent

The countriesMap reducer was keying on `country.code`, which doesn't exist in the countries constant (the property is `id`). Every iteration wrote to `acc[undefined]`, leaving Zimbabwe — the last entry — as the sole value at that key. Any contact whose `country` attribute was unset would resolve to `acc[undefined]` (Zimbabwe) instead of falling through to the correct `countryCode` lookup.

Two-part fix:
1. Remove the dead `acc[country.code]` line from the reducer.
2. Guard the `country` lookup so an unset attribute never hits the map (`(country && countriesMap[country]) || countriesMap[countryCode]`).

Adds unit specs covering the regression and the happy-path lookups.

Closes #14238

# Pull Request Template

## Description

Fixes # (issue)

## Type of change

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/app/javascript/dashboard/components-next/Contacts/ContactsCard/ContactsCard.vue
+++ b/app/javascript/dashboard/components-next/Contacts/ContactsCard/ContactsCard.vue
@@ -51,7 +51,6 @@ const isFormInvalid = computed(() => contactsFormRef.value?.isFormInvalid);
 
 const countriesMap = computed(() => {
   return countries.reduce((acc, country) => {
-    acc[country.code] = country;
     acc[country.id] = country;
     return acc;
   }, {});
@@ -64,7 +63,7 @@ const countryDetails = computed(() => {
   if (!country && !countryCode) return null;
 
   const activeCountry =
-    countriesMap.value[country] || countriesMap.value[countryCode];
+    (country && countriesMap.value[country]) || countriesMap.value[countryCode];
 
   if (!activeCountry) return null;
 
--- a/app/javascript/dashboard/components-next/Contacts/ContactsCard/specs/ContactsCard.spec.js
+++ b/app/javascript/dashboard/components-next/Contacts/ContactsCard/specs/ContactsCard.spec.js
@@ -1,0 +1,70 @@
+import countries from 'shared/constants/countries';
+
+// Reproduces the countriesMap and countryDetails logic from ContactsCard.vue
+// to verify the fix for the Zimbabwe display bug (issue #14238).
+function buildCountriesMap() {
+  return countries.reduce((acc, country) => {
+    acc[country.id] = country;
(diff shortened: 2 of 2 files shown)


sojan-official:

PR #7871, opened 2023-09-07
chore: Support multiple values for automation message content

sojan-official:
This change will allow automation rules with `message -> content` `equal` or `not equal` rules to accept an array of values. 
example rule 

```
#<AutomationRule:0x0000000107db6408
 id: 3,
 account_id: 1,
 name: "hello",
 description: "hello",
 event_name: "message_created",
 conditions:
  [{"values"=>["hi", "test", "Test", "magic"],
    "attribute_key"=>"content",
    "filter_operator"=>"equal_to",
    "custom_attribute_type"=>""}],
 actions:
  [{"action_name"=>"send_message", "action_params"=>["automation working"]}],
 created_at: Thu, 07 Sep 2023 00:19:49.873168000 UTC +00:00,
 updated_at: Thu, 07 Sep 2023 00:42:11.988807000 UTC +00:00,
 active: true>
```

PR #14269, opened 2026-04-27
fix: automation rule content filter incorrectly splits on commas (#13785)

### LINKED ISSUE #13785: Support multiple values for automation message content causes issue when we use comma in condition
**### Describe the bug**

_Support multiple values for automation message content [feature](https://github.com/Bootsschule1/B1-Chatwoot/commit/9ba5adfd60ce1399cd14a2453669554940a4d96c#diff-b0a5b64d0d54ad2b4ace6b06ca1a977de653038076c460d9f80ee89596112744) causes issue when we use comma in condition._

_This peace of code causing issue described below:_
> const generateValues = item => {
>     if (item.attribute_key === 'content') {
>       const values = item.values || '';
>       return values.split(',');
>     }
>   ...
> };

**### To Reproduce**

_1. I created a test an Automation Rule Template with Comma in condition:_

Image

_2. When I sent a request to save an Automation Rule Template the payload looks like this:_

Image

_The value is already splitted by the comma._

_3. When I changed a condition by removing a comma and sent request to save an Automation Rule Template the payload looks like this:_

Image

_Now value isn't splitted._
 

**### Expected behavior**

_I think we should add another type of condition (something like MESSAGE EQUALS with different condition type) or do something different which will allow us to hold values with commas._

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

None

### Operating system

_No response_

### Browser and version

_No response_

### Docker (if applicable)

_No response_

### Additional context

_No response_

williambsm9:
Automation rules with a `content` condition do not match messages that contain commas. The condition value is split on `,` before being sent to the API, so a rule configured for `hello, world` is stored as two separate tokens and never matches the original message.

Closes #13785

## What changed

`generateValues` in `app/javascript/dashboard/helper/filterQueryGenerator.js` had a special-case branch:

```js
if (item.attribute_key === 'content') {
  const values = item.values || '';
  return values.split(',');
}
```

This was already flagged with a `FIX ME: Make this generic option instead of using the key directly here` comment in `generatePayload`.

The fix removes the branch entirely. String values already fall through to the existing `return [item.values]` path, which produces the correct single-element array. Arrays continue to flow through `setArrayValues`. The stale `FIX ME` comment is also removed.

## Before / After

For an automation rule with content `hello, world`:

**Before**
```json
{ "values": ["hello", " world"] }
```

**After**
```json
{ "values": ["hello, world"] }
```

## Alternatives considered

- **Keep the special-case but escape commas (e.g. `\,`).** Rejected — adds UX complexity (users would need to learn an escape syntax) and there is no UI affordance for multi-value content conditions in the first place. The split was solving a problem that doesn't actually exist for this attribute.
- **Add server-side normalisation in `AutomationRules::ConditionsFilterService`.** Considered as defense-in-depth, but rejected to keep this PR scoped. The frontend is the source of the malformed payload; the backend already (correctly) expects an array. Adding server-side coercion here would conflate two concerns.

## Cross-stack check

I read `app/services/automation_rules/conditions_filter_service.rb` and `app/services/automation_rules/condition_validation_service.rb` to confirm the backend contract. The service treats `values` as an array throughout (e.g. `query_hash['values'].first` for label conditions, array operators in SQL building), so producing a correct array on the frontend is sufficient — no Rails change is needed.

## How to verify

Automated:

```bash
pnpm test app/javascript/dashboard/helper/specs/filterQueryGenerator.spec.js
```

All tests should pass, including the two new cases covering content with and without commas.

Manual:

1. Create an automation rule: condition `Message → Content → equal_to → hello, world`, action `Send Email`
2. Send an incoming message with body `hello, world`
3. Confirm the rule fires (before this change it would not, because the stored condition was `["hello", " world"]`)

## Notes

- The same helper is consumed by `ChatList`, `ContactsIndex`, segment exports and conversation getters. The fix transparently improves comma handling in all of these — no consumer code changes were needed.
- No translation strings touched.


sojan-official (commented):
Thanks for working on this.

The existing `FIX ME` comment was added because comma parsing should not be tied directly to the `content` attribute key. The intended follow-up was to introduce a generic input type/component for multiple text values, but we did not have the bandwidth to build it at the time.

Removing the split fixes content containing a literal comma, but it also breaks the multi-value behavior introduced in #7871. Existing production rules can configure `hello,refund` as:

```json
["hello", "refund"]
```

With this change, editing and saving that rule would turn it into:

```json
["hello,refund"]
```

We need to support both cases:

- Multiple independent content values.
- A single value containing a comma, such as `hello, world`.

The safer path would be a free-form multi-text/tagged input. Each tag represents one value, Enter creates another tag, and commas remain normal text inside a tag. Existing stored arrays can load directly as tags and be submitted unchanged, so this does not require a backend change or data migration.

We should also cover the round-trip behavior for existing multi-value rules and a single value containing a literal comma. Would you be open to updating the PR with this approach?

scmmishra:
Updating this PR based on the review feedback: the only reliable and unambiguous solution is a dedicated multi-text input that preserves values as arrays throughout the full edit → save cycle.

Each value is represented as a separate tag, and pressing Enter adds another value. Commas remain ordinary characters inside a tag, so `hello,world` stays one value, while `hello` and `refund` remain two independent values. Existing rules already store these values as arrays, so they load directly into tags without requiring a backend change or data migration.

The alternatives are inherently suboptimal:

- Simply removing comma splitting fixes literal commas but silently breaks existing multi-value rules when they are edited and saved.
- Continuing to use commas as delimiters cannot distinguish between two values and one value containing a comma.
- Introducing escaping or quoting syntax would expose implementation details to users, complicate the UI, and still leave existing ambiguous values unresolved.
- Normalizing this on the backend cannot recover the user’s original intent; the backend already has the correct array-based contract.
- Reusing the existing TagInput would couple automation filters to an unrelated component’s styling and semantics. The new `MultiTextInput` keeps this behavior explicit and localized.

The updated implementation also includes round-trip coverage for both existing multi-value rules and a single value containing a real comma.

Screenshot of the updated interaction:

CleanShot 2026-08-25 at 13 57 16@2x

scmmishra:
@codex please review

iamsivin (approved):
LGTM

PR #14272, opened 2026-04-27
fix(webhooks): strip trailing newlines from webhook message content

### LINKED ISSUE #13459: TipTap trailing newlines sent raw via webhook to external integrations
## Description

The TipTap/ProseMirror editor in Chatwoot v4.10.1 stores agent messages with trailing newlines (`\n\n\n`). While there are internal fixes for native channels (PR #12943), the **webhook payload** and **API responses** still include the raw content with trailing whitespace.

This affects all external integrations that consume Chatwoot messages via webhook — in our case, Evolution API for WhatsApp, where messages arrive with visible blank space below the text.

## Environment

- **Chatwoot:** v4.10.1 (self-hosted, Docker)
- **Integration:** Evolution API v2.3.7 (receives messages via Chatwoot webhook)

## Steps to Reproduce

1. Self-host Chatwoot v4.10.1
2. Set up a webhook or external integration that receives outgoing messages
3. Send a message as an agent from the Chatwoot UI
4. Inspect the `content` field in the webhook payload

**Expected:** `content` is trimmed (no trailing newlines)
**Actual:** `content` ends with `\n\n\n`

## Impact

Any external integration (WhatsApp via Evolution API, Telegram bots, custom webhooks) that forwards the message content will include the trailing whitespace, resulting in poor UX for end users.

## Suggested Fix

Trim trailing newlines from `content` before sending webhook payloads and API responses for outgoing messages. This could be done at the model level (before_save callback) or at the serializer/webhook builder level.

## Related

- #10141 — CTRL+ENTER newline bug (same root cause: TipTap editor)
- #12943 — Partial fix for native channels (not merged)
- #12878 — Backend markdown rendering (related but doesn't fix trailing newlines)
- Evolution API issue: https://github.com/EvolutionAPI/evolution-api/issues/2415

## Workaround

We are patching Evolution API's `chatwoot.service.js` to add `.replace(/\n+$/, "")` on the content before sending to WhatsApp. This needs to be reapplied after every Evolution API upgrade.

ramalau0:
The TipTap/ProseMirror editor stores agent messages with trailing paragraph nodes that produce trailing newlines (e.g. \`\n\n\n\`) in the \`content\` field. While Chatwoot's native channel delivery already handles this, webhook payloads and API responses were returning raw content with trailing whitespace — causing visible blank space below messages in every external integration that consumes Chatwoot webhooks (WhatsApp via Evolution API, Telegram bots, custom webhook consumers).

Closes #13459

## Root cause

\`Messages::WebhookContentNormalizer\` already strips CommonMark hard line breaks (\`\\\` + newline) for webhook consumers, but it did not strip trailing whitespace. All webhook and API responses flow through this normaliser, so it is the single correct place to apply the fix without touching stored data.

## What changed

Added \`.rstrip\` to \`Messages::WebhookContentNormalizer.normalize\`:

\`\`\`ruby
# before
text.gsub(/\\\r?\n/, "\n")

# after
text.gsub(/\\\r?\n/, "\n").rstrip
\`\`\`

## Trade-offs considered

| Option | Decision |
|---|---|
| \`before_save\` on \`Message\` model | Would clean stored data but is a broader change affecting all message creation paths and would require a data migration for existing records. Out of scope for this bug. |
| Trim in each channel's send path | DRY violation — many channels, each would need the same patch. |
| Fix at normaliser level (chosen) | Single location, only affects webhook/API output, zero risk to stored data or native channel delivery. |

**Known limitation:** existing messages in the database still have trailing newlines in storage. They will be delivered correctly through webhooks after this fix, but a follow-up migration could clean stored content if needed.

## How to reproduce

1. Send an agent reply from the Chatwoot UI
2. Inspect the \`content\` field of the outgoing \`message_created\` webhook payload
3. Observe trailing \`\n\n\n\` after the message text

After this fix, the \`content\` field is trimmed before delivery.


iamsivin:
@codex review?

iamsivin (approved):
@ramalau0 Thanks for the PR, LGTM.
@sony-mathew Could you please take a quick pass on this?

sony-mathew (approved):
LGTM

sony-mathew:
Thank you for your contribution @ramalau !

PR #14273, opened 2026-04-27
fix(agent-bots): destroy permissibles on AgentBot deletion and skip orphans in index

### LINKED ISSUE #13407: 500 on GET /platform/api/v1/agent_bots when platform_app_permissibles reference deleted AgentBots
### Describe the bug

What happens

GET /platform/api/v1/agent_bots can return 500 Internal Server Error when there are rows in platform_app_permissibles that point to AgentBots that no longer exist (e.g. the AgentBot was deleted but the permissible was not).

Root cause
1. The action loads all platform_app_permissibles with permissible_type: 'AgentBot'.
2. The view platform/api/v1/agent_bots/index.json.jbuilder iterates over these records and calls resource.permissible for each one.
3. For a permissible that references a deleted AgentBot (e.g. permissible_id: 41 with no matching row in agent_bots), the polymorphic belongs_to :permissible tries to load that AgentBot and raises ActiveRecord::RecordNotFound.
4. That exception is not handled during rendering, so the response becomes 500.

### To Reproduce

1. Create a Platform App and an AgentBot, and add the AgentBot to the app’s permissibles (so there is a row in platform_app_permissibles for that AgentBot).
2. Delete the AgentBot (e.g. from the account UI or via API), without removing the corresponding platform_app_permissibles row.
3. Call GET /platform/api/v1/agent_bots with a valid platform app access token.

### Expected behavior

_No response_

### Environment

Kubernetes

### Cloud Provider

None

### Platform

None

### Operating system

_No response_

### Browser and version

_No response_

### Docker (if applicable)

_No response_

### Additional context

Log:
[extract-2026-01-30T14_47_33.072Z.csv](https://github.com/user-attachments/files/24971268/extract-2026-01-30T14_47_33.072Z.csv)

ramalau0:
\`GET /platform/api/v1/agent_bots\` returns 500 when any \`AgentBot\` that was previously registered with a Platform App has since been deleted. The bug was introduced by a missing \`dependent: :destroy\` on the \`AgentBot\` model — deleting a bot left orphaned rows in \`platform_app_permissibles\`, which the index action later iterated over and crashed rendering with a \`NoMethodError\` on \`nil\`.

Closes #13407

## Root cause

The index action loads all \`platform_app_permissibles\` for the platform app and passes each \`resource.permissible\` (the associated \`AgentBot\`) to a Jbuilder partial. When the \`AgentBot\` no longer exists, \`resource.permissible\` returns \`nil\` and the partial crashes calling \`.id\`, \`.name\`, etc. on it.

Every other \`AgentBot\` association (\`agent_bot_inboxes\`, \`messages\`, \`assigned_conversations\`) had a \`dependent:\` option — \`platform_app_permissibles\` was the only one missing it. There was also an N+1 query: the index fired a separate SQL query per permissible to load each bot.

## What changed

**1. Model — prevent orphans at deletion time**
\`\`\`ruby
has_many :platform_app_permissibles, as: :permissible, dependent: :destroy
\`\`\`

**2. Controller — eager-load to eliminate N+1**
\`\`\`ruby
@resources = @platform_app.platform_app_permissibles
               .where(permissible_type: 'AgentBot')
               .includes(:permissible)
\`\`\`

**3. Jbuilder — defensive nil guard for pre-existing orphans**
\`\`\`ruby
bot = resource.permissible
next if bot.nil?
json.partial! '...', resource: bot
\`\`\`

## Trade-offs considered

| Option | Decision |
|---|---|
| Rescue \`NoMethodError\` in jbuilder | Hides the failure rather than fixing it. Rejected. |
| Only add the nil guard, skip the model fix | Leaves the data integrity gap open, future deletions continue creating orphans. Rejected. |
| Both layers (chosen) | Model fix prevents new orphans; nil guard is defence-in-depth for any orphans that survived before deployment. |
| \`dependent: :nullify\` | Doesn't apply, a nullified permissible would still cause the same nil dereference. Rejected. |

## How to reproduce

1. Create an AgentBot via the Platform API
2. Delete the AgentBot via any path (admin UI, API, or direct model call)
3. Call \`GET /platform/api/v1/agent_bots\` with a Platform App token
4. Observe 500

After this fix, the endpoint returns 200 with an empty array.

### DIFF
--- a/app/controllers/platform/api/v1/agent_bots_controller.rb
+++ b/app/controllers/platform/api/v1/agent_bots_controller.rb
@@ -3,7 +3,7 @@ class Platform::Api::V1::AgentBotsController < PlatformController
   before_action :validate_platform_app_permissible, except: [:index, :create]
 
   def index
-    @resources = @platform_app.platform_app_permissibles.where(permissible_type: 'AgentBot').all
+    @resources = @platform_app.platform_app_permissibles.where(permissible_type: 'AgentBot').includes(:permissible)
   end
 
   def show; end
--- a/app/models/agent_bot.rb
+++ b/app/models/agent_bot.rb
@@ -32,6 +32,7 @@ class AgentBot < ApplicationRecord
   has_many :agent_bot_inboxes, dependent: :destroy_async
   has_many :inboxes, through: :agent_bot_inboxes
   has_many :messages, as: :sender, dependent: :nullify
+  has_many :platform_app_permissibles, as: :permissible, dependent: :destroy
   has_many :assigned_conversations, class_name: 'Conversation',
                                     foreign_key: :assignee_agent_bot_id,
                                     dependent: :nullify,
--- a/app/views/platform/api/v1/agent_bots/index.json.jbuilder
+++ b/app/views/platform/api/v1/agent_bots/index.json.jbuilder
@@ -1,3 +1,6 @@
 json.array! @resources do |resource|
-  json.partial! 'platform/api/v1/models/agent_bot', formats: [:json], resource: resource.permissible
+  bot = resource.permissible
+  next if bot.nil?
+
+  json.partial! 'platform/api/v1/models/agent_bot', formats: [:json], resource: bot
 end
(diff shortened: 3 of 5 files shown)


sojan-official (approved):
Thanks a lot for the PR. Really appreciate you taking the time to work on this.

PR #14279, opened 2026-04-27
handle the country code

lunix2000:
# Pull Request Template

## Description
Fixed an issue in the Contact List where some contacts incorrectly showed Zimbabwe as the location.

The list view was attempting to resolve the country using a mapper lookup against country.code, but that value was not present in this data path. This caused an invalid lookup and resulted in the wrong fallback country being displayed.

Fixes # (issue)
Removed country code from [ContactsCard.vue] and [SearchRsultContactItem.vue] in the Mapper

## Testing
Ran specs for updated scenarios
Verified normalization logic locally
Checked updated files for syntax issues


sojan-official:

PR #14286, opened 2026-04-28
Fix: include private messages in webhook conversation payload

### LINKED ISSUE #14023: Private message webhook sends incorrect content

thabo1498:
## Description

When a private message (internal note) is created, the webhook fires 
correctly but sends the wrong content in the payload. Instead of the 
private message content, it sends the content of the last public message.

Root Cause:
The webhook_push_messages method in Conversations::EventDataPresenter 
uses the .chat scope which filters out private messages 
(where private: false). This causes it to fall back to the last public 
message instead of the private message that triggered the event.

Fix:
Replace .chat with .where.not(message_type: :activity) in 
webhook_push_messages only. This includes private messages in the webhook 
payload while still excluding activity messages. The UI push_messages 
method is intentionally unchanged.

Before:
Messages array content: "This is a public message" 

After:
Messages array content: "This is a secret internal note" 
Fixes #14023

## Type of change

✅ Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Added regression test to spec/listeners/webhook_listener_spec.rb
that creates a public message followed by a private message, then 
asserts the webhook payload contains the private message content.

Run: bundle exec rspec spec/listeners/webhook_listener_spec.rb
Result: 27 examples, 0 failures

## Checklist:
✅ My code follows the style guidelines
✅ I have performed a self-review
✅ I have added tests that prove my fix works
✅ New and existing unit tests pass locally


sony-mathew:
Thanks for the contribution. We already have an earlier open PR tracking this same issue: #14148.

PR #14298, opened 2026-04-28
feat(billing): attach attribution metadata to cloud subscriptions

sojan-official:
Adds billing attribution metadata to Chatwoot Cloud Stripe customer and subscription creation. This remains a draft while we confirm whether the payment analytics provider reads attribution from direct Stripe Subscription metadata.

Closes
N/A

Why
We already track acquisition on the Chatwoot website. When a user later creates a Chatwoot Cloud billing subscription from the app, we need a low-surface-area way to carry that attribution into Stripe without exposing this as a broader self-hosted/community feature.

What this change does
- Keeps checkout behind the existing cloud-only guard.
- Reads billing attribution IDs from existing browser cookies on the dashboard.
- Sends the attribution payload when the billing subscription setup runs.
- Adds attribution metadata to Stripe customers and subscriptions created by Chatwoot Cloud.
- Keeps app-facing naming generic as billing_attribution; provider-specific keys stay at the cookie and Stripe metadata boundary.

Validation
- bundle exec rspec spec/enterprise/services/enterprise/billing/create_stripe_customer_service_spec.rb spec/enterprise/jobs/enterprise/create_stripe_customer_job_spec.rb spec/enterprise/controllers/enterprise/api/v1/accounts_controller_spec.rb
- pnpm test app/javascript/dashboard/api/enterprise/specs/account.spec.js

### DIFF
--- a/app/javascript/dashboard/api/enterprise/account.js
+++ b/app/javascript/dashboard/api/enterprise/account.js
@@ -1,6 +1,5 @@
 /* global axios */
 import ApiClient from '../ApiClient';
-import { getBillingAttribution } from '../../helper/billingAttribution';
 
 class EnterpriseAccountAPI extends ApiClient {
   constructor() {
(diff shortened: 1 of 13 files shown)


sojan-official:
closing in favor of https://github.com/chatwoot/chatwoot/pull/14796

PR #14333, opened 2026-04-30
feat(voice): Assignment aware visibility and join conflict for inbound calls

muhsin-k:
### Description
Inbound voice calls now route ownership cleanly: the call widget is hidden from agents who aren't the conversation assignee, the first agent to pick up becomes the assignee, and any later join attempt by another agent is rejected with a clear "<agent> is already handling the call." alert.

Fixes https://linear.app/chatwoot/issue/PLA-98/inbound-voice-calls-assignment-aware-visibility-auto-assignment-on 

### How to test

  1. As Agent A and Agent B, open the dashboard for the same voice inbox in two browsers.
  2. Place an inbound call to the inbox with the conversation **unassigned** — both agents should see the call widget.
  3. Have Agent A click **Join**. Agent A's widget transitions to the active call; Agent B's widget disappears (conversation is now assigned to Agent A).
  4. While the call is in progress, attempt to join from a third agent (e.g., via the bubble in the conversation timeline) — the join is rejected with the toast `Agent A is already handling the call.`
  5. Resolve the conversation, then place a second call to a conversation that is already manually assigned to Agent A — only Agent A sees the widget; nobody else does.
  6. Race test: trigger two near-simultaneous join attempts (two agents click Join within a few hundred ms of each other) — exactly one wins; the other gets the conflict alert.

### DIFF
--- a/app/javascript/dashboard/components-next/message/bubbles/VoiceCall.vue
+++ b/app/javascript/dashboard/components-next/message/bubbles/VoiceCall.vue
@@ -36,18 +36,15 @@ const isOutbound = computed(() => call.value?.direction === 'outgoing');
 const isFailed = computed(() =>
   [VOICE_CALL_STATUS.NO_ANSWER, VOICE_CALL_STATUS.FAILED].includes(status.value)
 );
-// Authoritative answerer (only the Call record). Used to decide if the current
-// user is the one on the call.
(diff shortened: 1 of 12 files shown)


muhsin-k:
@codex please review

muhsin-k:
@codex please review

sony-mathew (approved):
LGTM.

PR #14344, opened 2026-05-01
feat(voice): Attach call recordings + show call duration on the bubble

muhsin-k:
When an inbound voice call ends, the conversation bubble now (1) renders an inline audio player as soon as Twilio finishes the recording and (2) shows the call duration alongside "Call ended" so the agent gets the at-a-glance summary without opening the recording.

Fixes https://linear.app/chatwoot/issue/PLA-118/feat-recordings-on-calls-should-be-attached-on-the-conversation and  https://linear.app/chatwoot/issue/PLA-119/duration-of-the-call-is-not-visible-on-the-chat-bubble

## How to test

1. Set up a Twilio voice inbox and trigger an inbound call.
2. Answer the call from an agent, talk for a few seconds, then hang up.
3. As soon as the call ends, the bubble should read **"Call ended — 0:NN"** (where NN is the call duration in seconds).
4. Wait a few seconds for Twilio to finish processing the recording (usually <30s after hangup).
5. The same bubble should now show an inline audio player below the duration. Press play; the recording should be audible.
6. Refresh the page — both the duration and the player should still be there.
7. End a second call on the same conversation — its bubble should get its own duration + player, independent of the first.


muhsin-k:
@codex please review

enterprise/app/models/call.rb
```diff
@@ -55,6 +55,7 @@ class Call < ApplicationRecord
 
   scope :active, -> { where.not(status: TERMINAL_STATUSES) }
   scope :by_conference_sid, ->(sid) { where("meta->>'conference_sid' = ?", sid) }
+  scope :by_twilio_conference_sid, ->(sid) { where("meta->>'twilio_conference_sid' = ?", sid) }
```
sony-mathew, line 58: I do not see `meta` column in calls table having a gin index. We might need to check the performance of this scoped queries.

enterprise/app/services/voice/provider/twilio/recording_attachment_service.rb
```diff
@@ -0,0 +1,82 @@
...
+      call.save!
+    end
+  end
+
+  def already_attached?
+    call.recording.attached? && call.recording_sid.to_s == recording_sid.to_s
+  end
+
+  def download_recording
+    Down.download(recording_url, http_basic_authentication: [account_sid, auth_token])
```
sony-mathew: Can we use SafeFetch for this? We want to use it in all places where we fetch from external URLs. Please check for usages of `SafeFetch.fetch`.

spec/enterprise/services/voice/provider/twilio/recording_attachment_service_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+    file.define_singleton_method(:original_filename) { 'recording.wav' }
+    file.define_singleton_method(:content_type) { 'audio/wav' }
+    file
+  end
+
+  before do
+    allow(Twilio::VoiceWebhookSetupService).to receive(:new)
+      .and_return(instance_double(Twilio::VoiceWebhookSetupService, perform: "AP#{SecureRandom.hex(8)}"))
+
+    allow(Down).to receive(:download)
```
sony-mathew: Please change the `Down` to `SafeFetch`.

sony-mathew (commented):
Added some comments.

enterprise/app/services/voice/provider/twilio/recording_attachment_service.rb
muhsin-k: Switched to `SafeFetch.fetch` in 9e7fca5357 — restricts to `audio/*` content types, drops the manual `ensure ... close!` (SafeFetch owns the tempfile lifecycle), and keeps Twilio basic auth via `http_basic_authentication`.

spec/enterprise/services/voice/provider/twilio/recording_attachment_service_spec.rb
muhsin-k: Spec stub updated to `SafeFetch.fetch` (block-yielding a `SafeFetch::Result`) in 9e7fca5357. All 5 examples pass.

enterprise/app/models/call.rb
muhsin-k: @sony-mathew Great catch. I'll add a follow-up PR with expression btree indexes on both keys.

sony-mathew (approved):
LGTM

PR #14378, opened 2026-05-05
feat(inbox): Add beta badge for TikTok and Voice channels

muhsin-k:
TikTok and Voice channels in the inbox creation flow now display a small "Beta" badge next to their title, signaling that these integrations are still being polished while keeping them available for users to try.
Fixes https://linear.app/chatwoot/issue/CW-7026/add-beta-label-for-tiktok-and-voice-inboxes


iamsivin (approved):
LGTM

PR #10843, opened 2025-02-05
feat: show email subject in conversation search results

bfontaine:
# Pull Request Template

## Description

This addresses #10842. It exposes `additional_attributes` in the conversations search endpoint, uses it in `SearchResultConversationsList` to pass `conversation.additional_attributes?.mail_subject` down to `SearchResultConversationItem`, which in turn displays it.

Fixes #10842

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

I have tested this locally by searching for conversations. See this screenshot where I searched for "noreply":
Screenshot from 2025-02-05 11-04-54

I would love to add automated tests but I’m not sure how to do that.

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #14383, opened 2026-05-06
fix(new-conversation): stop writing literal "undefined" as mail_subject

gabrieljablonski:
## Description

This PR fixes a data-corruption bug in the new conversation flow that surfaces in conversation search results as `Subject: undefined`.

### The Problem
When creating a conversation through the "New conversation" modal without typing a subject (every non-email channel never shows the subject input, and email channels can be left blank), the conversation gets persisted with `additional_attributes.mail_subject = "undefined"` (the literal string). The bogus value shows up in conversation search results as `Subject: undefined`, which started rendering after #10843.

### Root Cause
`createConversationPayload` in `app/javascript/dashboard/store/modules/contactConversations.js` appends the `mail_subject` field unconditionally:

```js
payload.append('additional_attributes[mail_subject]', mailSubject);
```

`composeConversationHelper.js` only sets `payload.mailSubject` when `subject` is truthy, so when no subject is provided, `mailSubject` is destructured as `undefined` in `createConversationPayload`. `FormData.append` coerces `undefined` to the string `"undefined"`, and the backend persists it as-is into the JSONB column.

### The Fix
Skip the append when `mailSubject` is falsy. The backend already treats a missing key the same as an empty subject (the email mailer falls back to a default subject when `mail_subject` is `nil`), so omitting the field is safe across channels.

### Key Changes
- Guarded the `additional_attributes[mail_subject]` append in `createConversationPayload`.
- Added a spec covering the case where `mailSubject` is omitted.

> Note: existing rows persisted before this fix will continue to show `Subject: undefined` in search until cleaned up. A simple SQL cleanup for non-email inboxes:
>
> ```sql
> UPDATE conversations
> SET additional_attributes = additional_attributes - 'mail_subject'
> FROM inboxes
> WHERE conversations.inbox_id = inboxes.id
>   AND inboxes.channel_type <> 'Channel::Email'
>   AND conversations.additional_attributes->>'mail_subject' = 'undefined';
> ```
>
> I'm leaving any data-cleanup migration out of this PR since it's a maintenance concern that may be handled differently per installation.

## Type of change

## How Has This Been Tested?

1. Open the "New conversation" modal on a non-email inbox (e.g., WhatsApp, SMS, API).
2. Create a conversation without filling any subject (field is not present on those, so regular flow).
3. Open the global search and search for the conversation.
4. **Before:** the result card shows `Subject: undefined`.
   **After:** the subject row is hidden.
5. Repeat on an email inbox leaving the subject blank — same result.
6. On an email inbox, type a subject and verify it still persists and renders correctly.
7. Run the new spec: `pnpm test contactConversations`.

## Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works


sony-mathew (approved):
LGTM

sony-mathew:
**Local Repro Steps**

1. Checkout `develop` or the PR base commit before the fix: `274e92e0e421fe4afd193ba1e1d7b68e6faa0c5f`.
2. Start the app locally and log in.
3. Use a non-email inbox, for example API/SMS/WhatsApp. The email blank-subject path is less reliable now because the current form requires subject for email inboxes in [ComposeNewConversationForm.vue](/Users/sonymathew/.codex/worktrees/25d0/chatwoot/app/javascript/dashboard/components-next/NewConversation/components/ComposeNewConversationForm.vue:109).
4. Open the New Conversation modal, pick a contact and that non-email inbox, enter a message, and send.
5. Inspect the request to `POST /api/v1/accounts/:account_id/conversations`: before this PR, FormData includes `additional_attributes[mail_subject]: undefined`.
6. Confirm in Rails console: `Conversation.order(created_at: :desc).first.additional_attributes['mail_subject']` returns `"undefined"`.
7. Search for that conversation globally. The result shows `Subject: undefined`.
8. On PR head `04f6cb2e324842d3a728510dbee5b5dab22d44ad`, repeat the flow: the FormData key is omitted, the DB attribute is absent, and the search card hides the subject row.

I also reproduced the core browser behavior locally:

```sh
node -e 'const fd = new FormData(); fd.append("additional_attributes[mail_subject]", undefined); console.log(fd.get("additional_attributes[mail_subject]")); console.log(typeof fd.get("additional_attributes[mail_subject]"));'
```

Output is `undefined` with type `string`.

Verified that this is no longer reproducible after the fix is in.

sony-mathew:
Thank you @gabrieljablonski for the contribution. Appreciate it!

PR #14394, opened 2026-05-07
fix: resolve SendReplyJob flaky specs

aakashb95:
`SendReplyJob` was caching reloadable service class objects in `CHANNEL_SERVICES`. In test, a request spec can trigger Rails constant reloading after `SendReplyJob` has already been loaded, leaving the job with stale class objects while later specs stub the reloaded constants. This resolves the channel service at perform time so the job follows the current Rails constant table.

How to reproduce
Run the CircleCI shard that contains send_reply_job_spec, or the minimized order-dependent reproduction:
```sh
bundle exec rspec --format progress spec/builders/v2/reports/label_summary_builder_spec.rb spec/controllers/api/v1/accounts/bulk_actions_controller_spec.rb spec/jobs/send_reply_job_spec.rb:32
```

What changed
- Store service class names in `SendReplyJob::CHANNEL_SERVICES` instead of class objects.
- Resolve the service with constantize inside perform so reloads do not leave stale cached classes.

### DIFF
--- a/spec/jobs/send_reply_job_spec.rb
+++ b/spec/jobs/send_reply_job_spec.rb
@@ -18,6 +18,17 @@ RSpec.describe SendReplyJob do
       allow(process_service).to receive(:perform)
     end
 
+    def expect_mapped_service_to_perform(message, service_class_name)
+      channel_name = message.conversation.inbox.channel.class.name
+      service_class = described_class::CHANNEL_SERVICES.fetch(channel_name)
+
+      expect(service_class.name).to eq(service_class_name)
+      expect(service_class).to receive(:new).with(message: message).and_return(process_service)
+      expect(process_service).to receive(:perform)
+
+      described_class.perform_now(message.id)
+    end
+
     it 'calls Facebook::SendOnFacebookService when its facebook message' do
       stub_request(:post, /graph.facebook.com/)
       facebook_channel = create(:channel_facebook_page)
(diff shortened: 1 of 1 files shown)


app/jobs/send_reply_job.rb
```diff
@@ -2,17 +2,17 @@ class SendReplyJob < ApplicationJob
...
-    'Channel::Telegram' => ::Telegram::SendOnTelegramService,
-    'Channel::Whatsapp' => ::Whatsapp::SendOnWhatsappService,
-    'Channel::Sms' => ::Sms::SendOnSmsService,
-    'Channel::Instagram' => ::Instagram::SendOnInstagramService,
-    'Channel::Tiktok' => ::Tiktok::SendOnTiktokService,
-    'Channel::Email' => ::Email::SendOnEmailService,
-    'Channel::WebWidget' => ::Messages::SendEmailNotificationService,
-    'Channel::Api' => ::Messages::SendEmailNotificationService
+    'Channel::TwitterProfile' => 'Twitter::SendOnTwitterService',
+    'Channel::TwilioSms' => 'Twilio::SendOnTwilioService',
```
aakashb95: i don't like this given we made class names stringified, we lose the IDE navigation, introduce possible string typos.. not sure if theres a better way

aakashb95: cc @sony-mathew

spec/jobs/send_reply_job_spec.rb
```diff
@@ -18,6 +18,17 @@
       allow(process_service).to receive(:perform)
     end
 
+    def expect_mapped_service_to_perform(message, service_class_name)
+      channel_name = message.conversation.inbox.channel.class.name
+      service_class = described_class::CHANNEL_SERVICES.fetch(channel_name)
+
+      expect(service_class.name).to eq(service_class_name)
+      expect(service_class).to receive(:new).with(message: message).and_return(process_service)
```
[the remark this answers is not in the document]
sony-mathew: Doesn't make sense at this point because we think the class referenced here is the one which will receive the call and that is what we are stubbing here. So discarding this for now.

app/jobs/send_reply_job.rb
sony-mathew: Made some other changes with this commit: [609472e](https://github.com/chatwoot/chatwoot/pull/14394/commits/609472ec0544de3e7aa4d4703b0bbdab989226eb). We do not want to touch the production code side of things. Tried to fix this in spec itself.

sony-mathew (approved):
LGTM

PR #14400, opened 2026-05-07
fix: portals#create returns 500 when custom_domain is omitted

### LINKED ISSUE #14397: Portal create endpoint returns silent 500 when `custom_domain` is omitted from request body (URI.parse(nil))
### Summary

`POST /api/v1/accounts/:account_id/portals` returns a generic 500 (`{"status":500,"error":"Internal Server Error"}`) whenever the request body does not include a `custom_domain` field. The cause is a `URI::InvalidURIError` raised by `parsed_custom_domain` when `@portal.custom_domain` is `nil`. The error is masked as a generic 500 with no useful response body, which makes the endpoint very hard to use programmatically.

The bug is present on `develop` (latest) and reproduces against `chatwoot/chatwoot:latest`.

### Affected file

[`app/controllers/api/v1/accounts/portals_controller.rb`](https://github.com/chatwoot/chatwoot/blob/develop/app/controllers/api/v1/accounts/portals_controller.rb)

### Reproduction

Against any account, with an admin `api_access_token`:

```bash
curl -X POST "https://<your-chatwoot-host>/api/v1/accounts/<account_id>/portals" \
  -H "Content-Type: application/json" \
  -H "api_access_token: <admin_user_api_token>" \
  -d '{
    "name": "Test Portal",
    "slug": "test-portal",
    "color": "#3b82f6",
    "config": {"default_locale": "en", "allowed_locales": ["en"]}
  }'
```

Response:
```json
{"status":500,"error":"Internal Server Error"}
```

The same call WITH `"custom_domain": ""` succeeds:
```bash
# Adding "custom_domain": "" makes it work
-d '{ "name": "Test Portal", "slug": "test-portal", "custom_domain": "", ... }'
# → 200 OK with portal payload
```

I verified both behaviours against the latest `chatwoot/chatwoot` Docker image.

### Root cause

```ruby
# app/controllers/api/v1/accounts/portals_controller.rb
def create
  @portal = Current.account.portals.build(portal_params.merge(live_chat_widget_params))
  @portal.custom_domain = parsed_custom_domain   # ← raises here
  @portal.save!
  process_attached_logo
end

def parsed_custom_domain
  domain = URI.parse(@portal.custom_domain)      # URI.parse(nil) → URI::InvalidURIError
  domain.is_a?(URI::HTTP) ? domain.host : @portal.custom_domain
end
```

When the request body omits `custom_domain`, `portal_params` doesn't include it, `@portal.custom_domain` is `nil` after `.build()`, and `URI.parse(nil)` raises `URI::InvalidURIError: bad URI (is not URI?): nil`. Rails translates that to 500 with no useful body.

Confirmed on the running container with a Rails console runner:
```
$ rails runner 'begin; URI.parse(nil); rescue => e; puts "#{e.class}: #{e.message}"; end'
URI::InvalidURIError: bad URI (is not URI?): nil
```

A separate `Portal.create!` (model layer) with the same params works fine, so the validation/model is not at fault — it's purely the controller's `parsed_custom_domain` helper.

### Suggested fix

```ruby
def parsed_custom_domain
  return @portal.custom_domain if @portal.custom_domain.blank?

  domain = URI.parse(@portal.custom_domain)
  domain.is_a?(URI::HTTP) ? domain.host : @portal.custom_domain
end
```

Two-line change, fully backward-compatible. Falls back to the existing else-branch behaviour when `custom_domain` is blank.

### Related (probably worth fixing in the same PR)

The `create` action also calls `process_attached_logo` unconditionally:

```ruby
def create
  ...
  @portal.save!
  process_attached_logo   # ← called even when no blob_id is in the request
end
```

The `update` action correctly guards it with `if params[:blob_id].present?`. On `develop`, `process_attached_logo` itself was [partially hardened](https://github.com/chatwoot/chatwoot/blob/develop/app/controllers/api/v1/accounts/portals_controller.rb) (`@portal.logo.attach(blob) if blob`), but moving the guard up to the call site (matching `update`) is more consistent and avoids one signed-blob-lookup per create.

### Severity

P1 for API consumers / automation tooling: every programmatic portal creation 500s unless the caller knows to pass an empty `custom_domain`, and the 500 response has no information that points to the cause.

### Versions tested

* Docker image: `chatwoot/chatwoot:latest` (pulled 2026-05-07)
* Reproduces in `develop` (verified by reading the source on this repo)

### Workaround for affected callers

Until the fix lands: send `"custom_domain": ""` in the create request body. `URI.parse("")` returns a `URI::Generic`, so the ternary falls through and the save succeeds.

---

Happy to open a PR with the suggested fix + a controller spec covering the no-`custom_domain` case if a maintainer agrees with the approach. 🙏

RenatoAscencio:
## Description

`POST /api/v1/accounts/:account_id/portals` returns a generic 500 (`{"status":500,"error":"Internal Server Error"}`) whenever the request body omits `custom_domain`. Root cause: `parsed_custom_domain` calls `URI.parse(@portal.custom_domain)` and `URI.parse(nil)` raises `URI::InvalidURIError`. Existing callers either had to know to pass `"custom_domain": ""` as a workaround or hit a 500 with no useful diagnostic.

This PR guards `parsed_custom_domain` against blank values so the existing fall-through (`else @portal.custom_domain`) applies — equivalent to passing an empty string.

It also moves the `process_attached_logo` guard from the helper into the `create` call site so `create` mirrors `update` (`process_attached_logo if params[:blob_id].present?`) and avoids an unnecessary signed-blob lookup on every create that doesn't include a logo.

Fixes #14397

## Type of change

## How Has This Been Tested?

Two new request specs in `spec/controllers/api/v1/accounts/portals_controller_spec.rb` covering the regression:

- `creates portal when custom_domain is omitted from request body` — the previously-broken case, now returns 200.
- `creates portal when custom_domain is blank` — verifies the existing workaround (`"custom_domain": ""`) still works after the change.

Manually verified against `chatwoot/chatwoot:latest` Docker image before the fix (500) and against this branch (200) using the curl repro from the issue.

```bash
curl -X POST "https://<host>/api/v1/accounts/<account_id>/portals" \
  -H "Content-Type: application/json" \
  -H "api_access_token: <token>" \
  -d '{"name":"Test Portal","slug":"test-portal","color":"#3b82f6"}'
```

Before: `{"status":500,"error":"Internal Server Error"}`
After: `200 OK` with the portal payload.

## Checklist

- [ ] I have made corresponding changes to the documentation (no doc change needed — controller behaviour, fully backward-compatible)


sojan-official (approved):
Thanks for fixing this. We’ll get this merged.

PR #14407, opened 2026-05-08
chore: verbosely log errors for leadsquare activity failure

scmmishra:


app/services/crm/leadsquared/processor_service.rb
```diff
@@ -89,11 +89,19 @@ def create_conversation_activity(conversation:, activity_type:, activity_code_ke
...
-    Rails.logger.error "Error creating #{activity_type} activity in LeadSquared: #{e.message}"
+    log_activity_error(e, activity_type, conversation)
+  end
+
+  def log_activity_error(error, activity_type, conversation, payload: nil)
+    ChatwootExceptionTracker.new(error, account: @account).capture_exception
+    context = "account_id=#{conversation.account_id}, conversation_display_id=#{conversation.display_id}"
+    if payload
+      context += ", http_status=#{error.code}, prospect_id=#{payload[:lead_id]}, " \
+                 "activity_event=#{payload[:activity_code]}, note_bytes=#{payload[:activity_note].to_s.bytesize}"
```
sony-mathew, line 102: `payload[:activity_note].to_s.bytesize` - Will this raise some error? We are already inside the rescue? Do we need try here to be safe?

sony-mathew (approved):
Approved with a minor comment.

app/services/crm/leadsquared/processor_service.rb
scmmishra: This is safe! `String#to_s` converts a possible `nil` to `""`, and `String#bytesize` is total (it never raises). So no try or removal needed.

app/services/crm/leadsquared/processor_service.rb
```diff
@@ -89,11 +89,19 @@ def create_conversation_activity(conversation:, activity_type:, activity_code_ke
...
+    end
+    Rails.logger.error("LeadSquared #{activity_type} activity failed: #{error.message} (#{context})")
```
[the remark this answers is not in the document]
scmmishra: Handled this

PR #14412, opened 2026-05-08
fix(notifications): respect conversation access when notifying agents

pranavrajs:
Agents with limited custom roles were receiving notifications (creation, assignment, mentions, new messages, SLA) for conversations they couldn't actually open. For example, an agent whose custom role only grants `conversation_unassigned_manage` was getting notified about conversations assigned to other agents.

Notifications now go through the same `ConversationPolicy#show?` check that gates the conversation view itself, so an agent only gets notified for conversations they're permitted to see. Administrators and agents without custom roles are unaffected.


app/services/messages/mention_service.rb
```diff
@@ -8,8 +8,8 @@ def perform
     return if validated_mentioned_ids.blank?
 
     Conversations::UserMentionJob.perform_later(validated_mentioned_ids, message.conversation.id, message.account.id)
-    generate_notifications_for_mentions(validated_mentioned_ids)
```
pranavrajs, line 11: When someone @mentions them on a conversation they aren't yet involved in, the original order was:

  1. Generate notification → policy runs → not assignee, not participant yet → denied → no notification
  2. Add as participant (too late)

  So the mention would be silently dropped

muhsin-k (approved):

muhsin-k (approved):
LGTM and tested with following scenarios

  Setup: Single account, single inbox with all three users as members.
  - Admin — administrator, no custom role
  - Agent B — agent, no custom role
  - Agent A — agent, custom role conversation_unassigned_manage

  **Scenario 1 — Conversation assigned to Agent B (the bug case)**

  Agent A must NOT be notified for any event on this conversation.

  - Agent A, conversation_creation — expected 0, got 0 ✅
  - Agent A, conversation_assignment — expected 0, got 0 ✅
  - Agent A, conversation_mention — expected 0, got 0 ✅
  - Agent A, assigned_conversation_new_message — expected 0, got 0 ✅
  - Agent A, sla_missed_first_response — expected 0, got 0 ✅
  - Agent B, conversation_assignment — expected 1, got 1 ✅
  - Admin, conversation_assignment — expected 1, got 1 ✅

  **Scenario 2 — Conversation unassigned**

  Agent A's role covers unassigned conversations, so the notification should fire.

  - Agent A, conversation_assignment — expected 1, got 1 ✅

  **Scenario 3 — Conversation assigned to Agent A himself**

  Agent A's role also covers conversations assigned to self.

  - Agent A, conversation_assignment — expected 1, got 1 ✅
  - Agent A, assigned_conversation_new_message — expected 1, got 1 ✅

PR #14442, opened 2026-05-13
feat: don't let onboarding write domain

scmmishra:
Stop the onboarding flow from writing the user's company website into `accounts.domain`. That column is reserved for the inbound email domain used to construct reply-to addresses (`reply+<uuid>@<domain>`), and silently overloading it from onboarding was breaking email continuity for accounts whose domain MX didn't point at Chatwoot's inbound — customer replies were going to an unreachable address.

The website value now lives in `custom_attributes.website`, which is what the rest of the app already treats as the "company website" field.


sojan-official (approved):
LGTM

PR #14457, opened 2026-05-14
feat: bidirectional contact<->conversation label sync + four automation actions

Hari-Bonda:
## Summary

Adds four contact-aware label automation actions, plus bidirectional Contact ↔ Conversation label propagation at the model layer so contact and conversation labels stay in sync regardless of which side changed. Propagation is **scoped to open conversations only** — resolved / pending / snoozed conversations stay frozen so historical labels are never retroactively rewritten. Includes a small drive-by fix for a duplicate `"All"` entry in the conversation status condition dropdown.

### Part 1 — Four automation rule actions (opt-in, agent-driven, per-label)

| Action | Scope | Use case |
|---|---|---|
| `add_label_to_contact` | contact only | Promote a conversation label onto the contact |
| `remove_label_from_contact` | contact only | Demote a contact attribute |
| `add_label_everywhere` | contact + every **open** sibling conversation | Globally tag this person across their open chats |
| `remove_label_everywhere` | contact + every **open** sibling conversation | Globally untag this person |

All four take a multi-select label list, are idempotent, and live inside automation rules so users opt in selectively per rule. Each rule's conditions are user-configurable — including the **conversation status** (Open / Resolved / Pending / Snoozed); users can target any subset of statuses per rule, the rule isn't hard-coded to one status.

The contact-only actions wrap their `contact.update_labels` call in a `without_label_propagation` helper that sets `Current.label_propagation_in_progress = true` for the duration, so the Contact propagation callback (below) does not fan the change out to conversations. This preserves the contact-only contract: only the contact's labels change. Users wanting fan-out should pick `add_label_everywhere` instead.

A no-params `sync_conversation_labels_everywhere` action is also included for "any time this conversation is updated, push its labels onto the contact + every open sibling conversation" rules without per-label parameters.

### Part 2 — Automatic Contact → conversations propagation

`Contact#propagate_label_changes_to_conversations` (in the new `ContactLabelPropagation` concern) fires on `after_update_commit` when `label_list` changes. The diff (labels added + removed) is applied to **every open conversation** of that contact. Resolved/pending/snoozed conversations stay frozen.

This handles the integration-driven flow: e.g., your billing system or scheduling plugin updates a contact's labels via the contacts API → every open conversation that contact has automatically picks up the diff.

### Part 3 — Automatic Conversation → contact + open siblings propagation

`Conversation#propagate_label_changes_to_contact_and_siblings` (in the new `ConversationLabelPropagation` concern) fires on `after_update_commit` when `label_list` changes. The diff is applied to the contact, then fanned out to every **open** sibling conversation.

The same concern's `inherit_contact_labels_on_creation` fires on `after_create_commit` so a brand-new conversation from a labelled contact automatically inherits the contact's current label set.

### Together: full bidirectional sync, scoped to open conversations

Both events the integration-driven flow needs are now handled:

- `conversation_created` → copies contact's current label set onto the new conversation (Conv concern's `after_create_commit`).
- `contact.updated` → diffs `changed_attributes['label_list']` (effectively, via `previous_changes`) and applies the delta to all of that contact's **open** conversations only — resolved / pending / snoozed stay frozen as a historical snapshot.

This solves the "cancel paid customer, all old paid-tagged conversations keep the stale label forever" problem: dropping `paid` from a contact strips it from every open conversation that contact has, while preserving historical records.

### Loop safety + side-effect suppression

`Current.label_propagation_in_progress` (new `thread_mattr_accessor` on the existing `Current` module) is set inside both propagation callbacks. It does three things:

1. **Prevents callback re-entry** — when one callback is mid-propagation, the other's entry is short-circuited.
2. **Suppresses `conversation.updated` event dispatch** for propagated label writes (`Conversation#notify_conversation_updation` checks the flag and returns early). Without this, a single contact label edit would emit `conversation.updated` for every sibling conversation and fire every user-defined `conversation_updated` automation rule (incl. non-idempotent actions like send-message, send-webhook, send-email) on every sibling — clearly unintended.
3. **Suppresses "Automation System added X" activity messages** during propagation (`ActivityMessageHandler#handle_label_change` checks the flag), so sibling conversations don't get spammed with activity rows.

Direct/user-driven writes still emit events and activity messages normally — the guard is only set inside the propagation callbacks, scoped with `ensure`.

### Small bug fix: duplicate "All" in conversation status condition dropdown

`useAutomationValues.js` was rendering the `"All"` option twice in the status condition dropdown because `statusFilterItems` already included `all` AND the computed `statusFilterOptions` appended another `{ id: 'all', name: 'All' }` on top. Removed the duplicate.

## Recipes (Part 1 examples)

**Promote a conversation label onto the contact only (not siblings)**
```
Event:     Conversation Updated
Condition: Labels / Includes / paid
Action:    Add a Label to Contact → paid
```

**Globally tag a customer as VIP across past + future open conversations**
```
Event:     Conversation Updated
Condition: Labels / Includes / vip
Action:    Add a Label Everywhere → vip
```

**Sync labels across the board on every conversation update (no per-label rules needed)**
```
Event:     Conversation Updated
Condition: Status / Equal to / Open + Resolved + Pending + Snoozed   ← any combination
Action:    Sync Conversation Labels Everywhere                       ← no-params action
```

If you want every label to sync globally without writing per-label rules, you don't need any Part 1 rules — Parts 2 and 3 handle it automatically at the model layer.

## Why all three parts

- Part 1 is the **per-label, conditional, rule-configurable surface** — for installs that want to propagate only specific labels under specific conditions / inboxes / statuses.
- Part 2 is the **integration-driven flow** — external systems writing to the contacts API (CRM, booking plugins, etc.) automatically propagate without touching the conversations API.
- Part 3 is the **agent-driven flow at the model level** — an agent labels a conversation in the dashboard and every related record picks it up. Replaces per-label "propagate X everywhere" rules with one always-on mechanism.

The companion #14455 (closed as superseded) handled contact → *new* conversation (`inherit_contact_labels` on `conversation_created`); this PR completes the rest of the matrix and includes that action as well.

## What's in the box

- **Backend (Part 1)**: 6 entries on `AutomationRule#actions_attributes`; 6 idempotent private methods on `AutomationRules::ActionService`. Contact-only methods use `without_label_propagation` helper to preserve their scope.
- **Backend (Parts 2 & 3)**: two new concerns (`ContactLabelPropagation`, `ConversationLabelPropagation`) with `after_update_commit` + `after_create_commit` hooks; re-entry + side-effect guard via `Current.label_propagation_in_progress`; targeted suppression of `Conversation#notify_conversation_updation` and `ActivityMessageHandler#handle_label_change` while the guard is set.
- **Frontend (Part 1)**: new action keys in `AUTOMATION_ACTION_TYPES` with appropriate `inputType`s; added to `AUTOMATIONS[event].actions` for `message_created`, `conversation_created`, `conversation_updated`; `getActionOptions` populates the multi-select dropdowns from the account's label set; i18n strings added in `en/automation.json`; `validations.js` updated for no-params actions.
- **Frontend (drive-by)**: `useAutomationValues.js` no longer duplicates the `"All"` status option.
- **Spec**: cases for each new action (happy path, merge, idempotency, contact-only contract); Contact callback ("open-only" scoping, no event dispatch during fan-out); Conversation callback (incl. inherit-on-create).

## Open questions for maintainers

- Parts 2 and 3 are currently always-on at the model layer. Happy to gate them behind an account setting if you'd prefer a more conservative default. Today's behavior matches what most teams probably expect (labels on a Contact represent person-attributes that should be visible on every open conversation), but existing installs upgrading will see new propagation behavior immediately.
- The new actions live on `AutomationRules::ActionService` rather than the shared `ActionService` parent — kept out of macros' allow-list intentionally. Happy to move them up if you'd prefer.
- "Open-only" was chosen as the propagation scope to avoid retroactively rewriting historical (resolved) labels. Open to changing to "all statuses" if maintainers prefer.

## Test plan

- [x] `bundle exec rspec spec/services/automation_rules/action_service_spec.rb` — covers all new actions plus `inherit_contact_labels`.
- [x] `bundle exec rspec spec/models/contact_spec.rb` — covers Part 2, including the "open-only" scoping and the "no `conversation.updated` dispatch during fan-out" guarantee.
- [x] `bundle exec rspec spec/models/conversation_spec.rb` — covers Part 3, including new-conversation inheritance.
- [x] Deployed and verified on a self-hosted 4.13.0 install: external API writes to contact `label_list`, every **open** conversation of that contact picks it up (resolved conversations untouched); agent labels a conversation, contact and every **open** sibling conversation pick it up; rule editor renders the new actions and the duplicate "All" dropdown entry is gone.


Hari-Bonda:
Thanks @chatgpt-codex-connector — good catch on the contract leak.

Fixed in [7f98677a7](https://github.com/chatwoot/chatwoot/pull/14457/commits/7f98677a7): the contact-only actions (`add_label_to_contact` / `remove_label_from_contact`) now wrap the `contact.update_labels` call in a `without_label_propagation` helper that sets `Current.label_propagation_in_progress = true` (the same guard the Contact callback respects). The Contact callback short-circuits during the action, so the labels stay scoped to the contact and the action's documented contract holds.

Added a spec case asserting that `add_label_to_contact` does NOT also write the labels to the source conversation, locking the contract in.

Users who want fan-out to every conversation should pick `add_label_everywhere` (which intentionally bypasses the helper).

Hari-Bonda:
Thanks @chatgpt-codex-connector — that's a real concern. Fixed in [407d42887](https://github.com/chatwoot/chatwoot/pull/14457/commits/407d42887):

- `Conversation#notify_conversation_updation` now returns early if `Current.label_propagation_in_progress` is set, so propagated label writes do NOT dispatch `conversation.updated`. User-configured automation rules and webhook subscribers on `conversation_updated` therefore do not fire across sibling conversations as a side effect of an internal label sync.
- `ActivityMessageHandler#handle_label_change` likewise short-circuits so we don't spam every sibling conversation with "Automation System added X" activity rows.
- Spec added: asserts no `conversation.updated` dispatch happens during the Contact → conversations fan-out.

The `Current.label_propagation_in_progress` guard now serves three purposes: prevents callback re-entry, suppresses event dispatch on propagated writes, and suppresses propagation-driven activity messages. The flag is set + cleared inside both the Contact and Conversation callbacks (with `ensure`), so the suppression is scoped strictly to the propagation chain — direct agent/user writes still emit normally.

Hari-Bonda:
Heads-up: the only remaining red check on the latest commit (`acc1ea863`) is `ci/circleci: lint`, and it's the **same pre-existing `bundle audit` failure** that affects every PR opened against `develop` right now. It's flagging 8 Rails 7.1.5.2 CVEs disclosed 2026-05-12 in the upstream `Gemfile.lock` — none of which this PR introduces or could fix from inside its scope (the patched versions are on Rails 7.2.x / 8.0.x / 8.1.x lines, not 7.1.x).

Specific job: https://circleci.com/gh/chatwoot/chatwoot/151709 → only "Bundle audit" step is red.

All other CI is green:
- `lint-backend` (Rubocop): ✅
- `lint-frontend` (eslint): ✅
- All 16 `backend-tests` shards: ✅
- `test-build` (linux/amd64, linux/arm64): ✅
- `frontend-tests`, `Validate PR title`, `log_lines_check`, `security/snyk`, MFA, Size Limit: ✅

Happy to provide a separate PR bumping the Rails constraint if maintainers want, but it felt out of scope for this label-sync work.

Hari-Bonda:
@muhsin-k this is my first PR requesting help to merge/suggestions

PR #14467, opened 2026-05-15
fix: make SAML callback session independent

scmmishra:
This PR makes SAML login independent of Rails session cookies

## Problem

The normal SAML login flow should be straightforward:

- User opens Chatwoot.
- Chatwoot creates `_chatwoot_session`.
- User starts SSO.
- Chatwoot redirects the browser to the SAML provider.
- The provider authenticates the user.
- The provider sends the browser back to Chatwoot's ACS URL.
- Chatwoot reads the SAML response, finds or creates the user, and logs them in.

The fragile step is the ACS callback. Most SSO flows return to the app through browser redirects where cookies usually pass through as expected. **ADFS commonly returns the SAML response with a cross-site POST**. With Chatwoot's session cookie using `SameSite=Lax`, browsers may not send `_chatwoot_session` on that POST.

SAML validation itself does not need the old Rails session cookie. The problem was our callback handoff after validation. DeviseTokenAuth stores the verified OmniAuth payload in Rails session, then redirects to a second callback route. If the browser does not preserve that session, Chatwoot has already received a valid SAML response but can no longer finish login.

## Solution

This PR removes the session-backed handoff for SAML only:

- The SAML callback completes login in the same request where OmniAuth validates the SAML response.
- Chatwoot reads the verified auth payload directly from `request.env['omniauth.auth']`.
- Account context and RelayState come from callback params or OmniAuth env data, not Rails session.
- Other OmniAuth providers continue using the existing DeviseTokenAuth flow.
- Mobile SAML still works when the IdP returns `RelayState=mobile`; the callback redirects to the mobile deep link with the generated SSO token.

The previous SAML override used `303 See Other` to avoid replaying the SAML POST into the second callback route. This change keeps that intent, but removes the second callback route for SAML entirely.

### SP Initiated
https://github.com/user-attachments/assets/b0735e93-3864-4cc3-b6fc-419fff4b549e

### IDP Initiated
https://github.com/user-attachments/assets/3ded0246-933c-4c85-9b7c-fa15fdc34883

## Testing

Manual validation:

- Complete a SAML login.
- In the browser network trace, find the IdP POST to `/omniauth/saml/callback?account_id=<account-id>`.
- Confirm it redirects directly to `/app/login?...sso_auth_token=...` for web login.
- For mobile, confirm `RelayState=mobile` redirects to the configured mobile deep link.
- Confirm there is no intermediate `/auth/saml/callback` request.

Testing with mocksaml.com:

- Configure Chatwoot with a public `FRONTEND_URL`.
- Set the mocksaml ACS URL to:

```text
https://<chatwoot-host>/omniauth/saml/callback?account_id=<account-id>
```

- Set the mocksaml audience/SP entity ID to the value shown in Chatwoot SAML settings, usually:

```text
https://<chatwoot-host>/saml/sp/<account-id>
```

- Use an email returned by mocksaml that exists in the SAML-enabled account.
- Start login from Chatwoot's SSO login page.
- Confirm the callback redirects directly to the app login URL with an SSO token.

### DIFF
--- a/enterprise/app/controllers/enterprise/devise_overrides/omniauth_callbacks_controller.rb
+++ b/enterprise/app/controllers/enterprise/devise_overrides/omniauth_callbacks_controller.rb
@@ -15,7 +15,7 @@ module Enterprise::DeviseOverrides::OmniauthCallbacksController
   end
 
   def omniauth_failure
-    return super unless params[:provider] == 'saml'
+    return super unless params[:strategy] == 'saml'
 
     relay_state = saml_relay_state
     error = params[:message] || 'authentication-failed'
--- a/enterprise/app/controllers/enterprise/devise_overrides/omniauth_callbacks_controller.rb
+++ b/enterprise/app/controllers/enterprise/devise_overrides/omniauth_callbacks_controller.rb
@@ -1,9 +1,4 @@
 module Enterprise::DeviseOverrides::OmniauthCallbacksController
-  def saml
-    # Call parent's omniauth_success which handles the auth
-    omniauth_success
-  end
-
   def redirect_callbacks
     return omniauth_success if params[:provider] == 'saml'
 
--- a/enterprise/app/controllers/enterprise/devise_overrides/omniauth_callbacks_controller.rb
+++ b/enterprise/app/controllers/enterprise/devise_overrides/omniauth_callbacks_controller.rb
@@ -5,22 +5,9 @@ module Enterprise::DeviseOverrides::OmniauthCallbacksController
   end
 
   def redirect_callbacks
-    # derive target redirect route from 'resource_class' param, which was set
-    # before authentication.
-    devise_mapping = get_devise_mapping
-    redirect_route = get_redirect_route(devise_mapping)
+    return omniauth_success if params[:provider] == 'saml'
 
-    # preserve omniauth info for success route. ignore 'extra' in twitter
-    # auth response to avoid CookieOverflow.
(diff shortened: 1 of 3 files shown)


scmmishra:
@codex review

muhsin-k (dismissed):

sojan-official:
Ran a deeper review using Claude and Codex. The main SAML callback path works as intended, and the PR looks good to merge for the cookieless SAML success flow.

Two issues noted:

1. `def saml` appears to be dead code

`/omniauth/saml/callback` resolves to `devise_overrides/omniauth_callbacks#redirect_callbacks` with `provider=saml`. The later `devise/omniauth_callbacks#saml` route is shadowed, so the `saml` method in `Enterprise::DeviseOverrides::OmniauthCallbacksController` does not appear to be invoked by the actual callback path.

This is cleanup-level and not a blocker.

2. SAML failure handling appears unreachable

`omniauth_failure` currently checks `params[:provider] == 'saml'`, but OmniAuth failure redirects use `/omniauth/failure?message=...&strategy=saml`. That means real SAML failures fall through to `super` instead of the custom SAML failure branch.

This looks pre-existing and does not block the cookieless success-path fix, but since this PR touches SAML callback behavior, we should either fix it narrowly here or explicitly defer it.

scmmishra:
@sojan-official addressed both issues, thanks :)

scmmishra:
@codex please review

sojan-official (approved):

PR #14469, opened 2026-05-15
feat: enable quoted reply for everyone

scmmishra:
Quoted email replies is now available to every account by default. Previously this was gated behind the `quoted_email_reply` account-level feature flag, so accounts needed it toggled on (via Super Admin) before agents saw the toggle in the reply box.

## How to test

1. Open any conversation on an email inbox.
2. Confirm the **Quote previous email** toggle is visible in the reply box (and is **not** visible on private notes or non-email channels).
3. Toggle it on, type a reply, and send — the outbound email should include the quoted prior email below your message.
4. Toggle it off and send another reply — the quoted block should not appear.
5. The toggle preference should persist per channel type (UI setting), as before.
6. Verify the toggle works on a brand new account with no feature flags flipped on (previously it would have been hidden).

## What changed

- Removed all `isFeatureEnabledonAccount(..., QUOTED_EMAIL_REPLY)` gates from `ReplyBox.vue`, so the toggle and quoted-content behavior are unconditional on email channels.
- Removed the `QUOTED_EMAIL_REPLY` constant from `dashboard/featureFlags.js`.
- Marked the flag as `deprecated: true` in `config/features.yml` (kept the entry in place to preserve FlagShihTzu bit positions on existing accounts; `deprecated: true` hides it from the Super Admin UI).
- Dropped the now-unnecessary `account.enable_features('quoted_email_reply')` setup from the message builder spec.

### DIFF
--- a/app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
+++ b/app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue
@@ -5,7 +5,6 @@ import { useAlert } from 'dashboard/composables';
 import { useUISettings } from 'dashboard/composables/useUISettings';
 import { useTrack } from 'dashboard/composables';
 import keyboardEventListenerMixins from 'shared/mixins/keyboardEventListenerMixins';
-import { FEATURE_FLAGS } from 'dashboard/featureFlags';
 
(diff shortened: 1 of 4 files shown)


scmmishra:
@codex please review

sojan-official (approved):

Issue #14474, opened 2026-05-16
Account#feature_flags will overflow at the 64th feature (bigint sign-bit ceiling)


propiter:
## Summary

`config/features.yml` currently defines **63 features** (after `advanced_assignment` from #13036), and `accounts.feature_flags` is stored as a signed `bigint`. The 64th feature added — whichever it ends up being — will land on bit 63, which is the sign bit of a signed `bigint`. Any account that enables it will hit `ActiveModel::RangeError` on every write to the row.

The risk surface is concrete: a recent inserting-in-the-middle scenario already triggered this in a downstream installation (full reproduction below). Once `develop` adds its own 64th feature, the failure path will be reachable in upstream too.

## Reproduction (real-world evidence)

Setup: a fork of `develop` with `template_builder` from #12231 cherry-picked. Across 12 accounts in production, the result was:

1. Silent feature-list misalignment for 12 accounts (features after the inserted slot were all reported one position off — `advanced_search_indexing` showing as `template_builder`, `reply_mailer_migration` as `advanced_search_indexing`, etc.)
2. For accounts where the shift set bit 63, every SuperAdmin save raised:

   ```
   ActiveModel::RangeError (18444351259918991351 is out of range
                           for ActiveModel::Type::Integer with limit 8 bytes)
     enterprise/app/controllers/enterprise/super_admin/accounts_controller.rb:24
   ```

The same outcome would occur in stock `develop` the moment a feature is appended at the end without first reclaiming a `deprecated: true` slot.

## Why a column-type migration is not the obvious answer

The intuitive fix would be `change_column :accounts, :feature_flags, :decimal, precision: 78, scale: 0`. However:

- PostgreSQL bitwise operators (`&`, `|`, `<<`, `>>`, `#`) are **only defined for integer types** (`smallint`, `integer`, `bigint`, `bit varying`) — not for `numeric`/`decimal`.
- `flag_shih_tzu` in `:bit_operator` mode (which Chatwoot uses, see `Featurable#QUERY_MODE`) issues those bitwise operators directly to the database. Migrating to `numeric` would silently break every `Account.with_<feature_name>` scope.

So the migration path requires either:

- Replacing the column with `bit varying` (arbitrary-length bit string, supports bit ops), which `flag_shih_tzu` does not natively support, or
- Adding a second `feature_flags_2` bigint column and splitting features across both, which `flag_shih_tzu` does support via multiple `has_flags` declarations, or
[...]

PR #12231, opened 2025-08-19
feat: Add whatsapp template builder

Aman-14:
# Pull Request

## Description

The PR includes a template manager for whatsapp to create templates from Chatwoot #10886

Feature Overview

  - Adds full WhatsApp message template management: REST endpoints for list/show/
  create/delete ([app/controllers/api/v1/accounts/message_templates_controller.rb](https://github.com/chatwoot/chatwoot/pull/12231/files#diff-651266cc61affd0a16328a03edc0b0f113abbd07d8029ad47af57fb6a85825b6R1)),
  new MessageTemplate model with policy and migration ([db/migrate/20250901081601_create_message_templates.rb](https://github.com/chatwoot/chatwoot/pull/12231/files#diff-baabcfb95cf3c13ffb02c8169ac0d0f5d33585a9d78c5ce033d2ab3cf1273317R1)),
  plus formatter/sync services for Meta integration ([app/services/whatsapp/
  template_formatter_service.rb](https://github.com/chatwoot/chatwoot/pull/12231/files#diff-4761b267f4631aa8d70d7e489842112e0773b8b43950c07d733645d724b30dd3R2), [app/services/whatsapp/template_sync_service.rb](https://github.com/chatwoot/chatwoot/pull/12231/files#diff-ecc2cf7105a702f5eb68e44be17d1142b4358c933796cc936fb4324095059828R1)).
  - Extends WhatsApp Cloud provider flow to handle template creation with media
  uploads by introducing a dedicated Facebook upload client and restructuring
  creation logic ([app/services/whatsapp/template_creation_service.rb](https://github.com/chatwoot/chatwoot/pull/12231/files#diff-e8788a61a62cfeaec8191feb1483df36b246c01f77a416758b15bd31c7353826R1), [app/
  services/whatsapp/facebook_upload_service.rb](https://github.com/chatwoot/chatwoot/pull/12231/files#diff-ed84ba7056d8fb5290b6163c10056771d399671fd80c55302e1dcef78be66f3aR2), [app/services/whatsapp/providers/
  whatsapp_cloud_service.rb](https://github.com/chatwoot/chatwoot/pull/12231/files#diff-1c1b4c147a74aeb2d0abe69da6788ee4410382290e52c2d8b59b70685110b7a2R201))
- This feature is gated behind a Template builder feature flag

Frontend

  - Introduces template builder UI with modular header/body/footer/buttons sections,
  routing, and Vuex store to manage templates ([app/javascript/dashboard/routes/
[...]

PR #14475, opened 2026-05-16
chore(features): document bit-position constraint and guard 63-feature ceiling

propiter:
## Why

`config/features.yml` positions map 1-to-1 to bit positions in the `accounts.feature_flags` signed bigint column, which gives only 63 usable slots (bit 63 is the sign bit). The file is currently at **exactly 63 entries**, so the next feature appended without first reclaiming a `deprecated: true` slot will trigger `ActiveModel::RangeError` on every account that enables it.

This isn't theoretical — we hit it in our environment when #12231 was rebased on `develop`: the inserted-in-the-middle entry shifted `advanced_assignment` from bit 62 to bit 63, silently misaligned feature lists for 12 accounts, and broke every SuperAdmin save with:

```
ActiveModel::RangeError (18444351259918991351 is out of range
                        for ActiveModel::Type::Integer with limit 8 bytes)
  enterprise/app/controllers/enterprise/super_admin/accounts_controller.rb:24
```

Full write-up in #14474; reproduction details in the comment on #12231.

The file header already says `# DO NOT change the order of features EVER`, but it doesn't explain *why* or what to do once the list is full — both of which become non-obvious when a contributor is racing to ship.

## What this PR does

- **Expand the header in `config/features.yml`** to document the protocol explicitly: append-only, reclaim `deprecated: true` slots when full, never delete or reorder. References the recent reclaim precedents (`c7ce26c9`, `d6827eb2`, `b25e85ed`, `bb235ea5`).

- **Add `spec/config/features_spec.rb`** that fails CI when:
  - the entry count exceeds 63 (with a message pointing at the reclaim pattern), or
  - a duplicate name slips in (e.g. a merge conflict mapping two features to the same bit).

The spec is intentionally minimal — it's a regression net, not an architectural change. The structural follow-up (lifting the 63-bit ceiling itself, likely via a second `feature_flags` column when reclaim runs out) is tracked in #14474.

## How to test

```bash
bundle exec rspec spec/config/features_spec.rb
```

Both examples should pass on current `develop`. To verify the count guard fails as intended, append a 64th entry to `config/features.yml` locally and rerun.

## Notes for reviewers

- No runtime code paths change; this is purely a doc + CI guard.
- `Featurable::FEATURE_LIST` already loads the same YAML, but I kept the spec's loading independent so a regression in `Featurable` can't mask a regression in the file itself.
- I'm happy to fold this into a larger structural PR if maintainers prefer to address the ceiling here too — let me know which direction is more useful.

### DIFF
--- a/config/features.yml
+++ b/config/features.yml
@@ -1,11 +1,31 @@
 # DO NOT change the order of features EVER
 ############################################
+# Positions in this file map 1-to-1 to bit positions in the
+# `accounts.feature_flags` bigint column (via `flag_shih_tzu`). Inserting,
+# removing, or reordering entries shifts every following bit and silently
+# reinterprets stored `feature_flags` values for every existing account.
+#
+# The column is a signed bigint (max 2^63 − 1), which gives 63 usable bits.
+# Bit 63 is the sign bit: enabling the 64th feature on any account triggers
+# `ActiveModel::RangeError` on save. A spec in `spec/config/features_spec.rb`
+# enforces this ceiling.
+#
+# Protocol for changes:
+#   - Adding a feature → APPEND to the end of this file. Never insert in
+#     the middle.
+#   - Retiring a feature → mark `deprecated: true`. Never delete the entry
+#     (deletion shifts every following bit).
+#   - Out of slots (63 reached) → RECLAIM a `deprecated: true` entry by
+#     renaming it in place. Recent precedents: commits c7ce26c9, d6827eb2,
+#     b25e85ed, bb235ea5.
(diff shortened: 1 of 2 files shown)


scmmishra:
Hey @propiter this is a widely known limitation, the `ActiveModel::RangeError` in itself is a signal enough for this. We have internal docs outlining how to reclaim unused bits (we can make them public perhaps). As such the changes in the PR are not necessary at the moment. It's good to have them, but then considering there's already signals present for the developer, the changes become redundant.

PR #14491, opened 2026-05-18
feat: Add an option on the dashboard to allow switching help center layout

pranavrajs:
Screenshot 2026-05-18 at 12 32 55 PM


iamsivin (approved):
@pranavrajs LGTM, I’ve updated the card component to use the existing `RadioCard.vue` component.

PR #14514, opened 2026-05-21
fix: apply setUser email before pre-chat

### LINKED ISSUE #12256: Pre-form still asks for email even though it's provided via `setUser` call
### Describe the bug

When user logs in to the website I am calling `window.$chatwoot.setUser()` with similar data:

```js
window.$chatwoot.setUser(123, {
  email: "email@domain.com",
  name: "Full Name",
  phone_number: "1111111111"
});
```

Still when I am starting a new conversation an email is being asked in the pre-form, while it should be already known. At least it should be a default value.

Image

### To Reproduce

1. Enable pre-form with email field
2. Before starting a new conversation call `setUser` with email
3. Start new conversation
4. Observe email still being asked

### Expected behavior

Email is not asked when it's already known. Or at least the field contains the user email as a value giving the user an opportunity to change it.

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

Browser

### Operating system

Mac OS

### Browser and version

Safari 18.6 (20621.3.11.11.3)

### Docker (if applicable)

_No response_

### Additional context

_No response_

Will-thom:
When a website calls the Chatwoot SDK setUser API with an email address before the visitor starts a conversation, the widget now applies the returned contact state immediately. This lets the pre-chat form see that the contact already has an email, so it no longer asks the visitor to enter the same email again while the contact refresh request is still pending.

Closes

- Closes #12256

How to reproduce

1. Enable the pre-chat form with the email field.
2. Call window.$chatwoot.setUser(id, { email, name }) before starting a new conversation.
3. Start a new conversation from the widget.

How to test

1. Repeat the reproduction steps above.
2. Confirm the pre-chat form does not ask for an email that was already provided via setUser.
3. Confirm visitors without an email in setUser still see the email field when it is enabled.

What changed

- Forward the stored identifier together with the user payload when the SDK sends set-user after the widget loads.
- Commit the set_user API response into the widget contact store immediately before the async contact refresh.
- Updated contact store specs and added SDK coverage for the loaded set-user payload.


iamsivin:
@Will-thom Thanks for the PR. Could you please check the Codex comments?

sony-mathew:
Please check this findings:

1. **P1 — Reset restores the previous visitor’s identity and conversation.** **Matches the existing open review thread.**  
   The corrected [load-time replay](/worktrees/pr-14514/chatwoot/app/javascript/sdk/IFrameHelper.js:181) also runs after `reset()`, which retains the cached user, identifier, and verification hash.

   Browser reproduction with identity verification enabled:
   - **Base:** reset produces an anonymous visitor; previous conversation disappears.
   - **PR:** reset re-identifies the previous contact and displays their conversation again.

   This breaks logout isolation and confirms the [existing P1 comment](https://github.com/chatwoot/chatwoot/pull/14514#discussion_r3585285141).

2. **The original numeric-ID case remains unresolved.** **Pre-existing behavior, newly verified.**  
   When contact `"123"` already exists, `setUser(123, …)` returns success but leaves the visitor anonymous without an email, on both base and head. Passing `"123"` succeeds. The [contact-matching comparison](/worktrees/pr-14514/chatwoot/app/actions/contact_identify_action.rb:78) treats the numeric and stored string identifiers as different.

PR #14521, opened 2026-05-21
feat(idb-cache): add more models to IDB cache

scmmishra:
With the push-based cache system in place (#14696), this PR expands the IndexedDB cache from 3 models to 6: canned responses, agents, and custom attribute definitions now paint instantly on warm boots and revalidate through the same pushed key map. The diff is mostly mechanical, which is the point: adding a model is now a config entry plus server-side bump hooks.

> Stacked on #14696. Review that first; this retargets to develop once it merges.

## What changed

- `cacheable_models` grows to `[Label, Inbox, Team, CannedResponse, AccountUser, CustomAttributeDefinition]`; the endpoint, channel transmit, and broadcasts pick the new models up automatically.
- The three API clients (`agents`, `attributes`, `cannedResponse`) become cache-enabled via constructor options. `DATA_VERSION` is bumped to create the new object stores (a deliberate one-time global cache reset).
- Invalidation triggers, the careful part:
  - `CannedResponse` / `CustomAttributeDefinition`: own-table bumps via `AccountCacheRevalidator`.
  - `AccountUser` / `User`: **column-filtered** bumps (`AGENT_CACHE_RELEVANT_COLUMNS`), so high-churn writes like token rotation don't invalidate the agent cache on every login.
  - Avatars: attachment changes don't dirty `users` columns, so `invalidate_avatar_cache` is called explicitly from profile update/delete and `Avatar::AvatarFromUrlJob`.
  - `CustomRole` (enterprise): role details are embedded in the agent payload, and `dependent: :nullify` skips callbacks, hence explicit bumps.
  - `Portal` to `inbox`, `TeamMember` to `team`: these associations are embedded in existing cached payloads and were silently missing from the invalidation loop.

## How to test

1. Boot the dashboard twice. The second boot should paint canned responses, agents, and custom attributes instantly (no `/canned_responses`, `/agents`, or `/custom_attribute_definitions` requests on load).
2. In a second tab or browser, edit a canned response, change an agent's name or avatar, or update a custom attribute. The first tab should reflect it within seconds.
3. Add or remove a team member, or rename a portal. Team and inbox data stay fresh in open dashboards.

## Notes for reviewers

- The column filter on `User`/`AccountUser` is load-bearing: without it, every sign-in would broadcast an invalidation and make all open dashboards refetch agents.
- Watch for the same trap in future changes: anything embedded in a cached payload needs a bump, and `update_all`/`dependent: :nullify` writes skip callbacks.

### DIFF
--- a/app/controllers/api/v1/profiles_controller.rb
+++ b/app/controllers/api/v1/profiles_controller.rb
@@ -13,11 +13,17 @@ class Api::V1::ProfilesController < Api::BaseController
     @user.assign_attributes(profile_params)
     @user.custom_attributes.merge!(custom_attributes_params)
     @user.save!
+
+    # Profile updates can change cached agent fields, including avatar-backed thumbnails.
+    @user.invalidate_avatar_cache
   end
 
   def avatar
     @user.avatar.attachment.destroy! if @user.avatar.attached?
     @user.reload
+
+    # Agent thumbnails are cached separately, and avatar attachment deletes do not dirty user columns.
+    @user.invalidate_avatar_cache
   end
 
   def auto_offline
--- a/app/javascript/dashboard/api/agents.js
+++ b/app/javascript/dashboard/api/agents.js
@@ -1,10 +1,10 @@
 /* global axios */
 
-import ApiClient from './ApiClient';
+import CacheEnabledApiClient from './CacheEnabledApiClient';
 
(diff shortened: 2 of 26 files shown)


scmmishra:
@codex please review

iamsivin:
@codex review

PR #14538, opened 2026-05-22
fix(widget): improve dark mode select options

sojan-official:
Fixes the web widget select dropdown styling in dark mode so native select options remain readable against the widget's dark UI.

Closes
None

## Screenshots

Previous state:

Previous dark mode dropdown issue

Current state:

Current dark mode dropdown styling

## How to test

1. Enable the web widget pre-chat form with a list/select field.
2. Load the widget with dark mode enabled.
3. Open the select field and verify the select control and option text remain readable in dark mode.

## What changed

- Adds light/dark color-scheme handling for widget native selects.
- Applies explicit option background and text colors so dark-mode select options do not inherit unreadable colors from the browser popup.

PR #14540, opened 2026-05-22
fix: improve widget select option contrast

sojan-official:
Fixes the native select dropdown contrast in the web widget so options remain readable when the widget is using the dark theme.

## Closes

N/A

## How to test

- Open a dark-themed web widget with a form that includes a select field.
- Open the select dropdown and verify all options are readable against the native dropdown background.
- Confirm the select input itself still matches the widget theme.

## What changed

- Sets native widget select options to a light background with dark text, independent of the widget dark theme variables.

## Validation

- `pnpm exec scss-lint app/javascript/widget/assets/scss/woot.scss`


sojan-official:
duplicate : https://github.com/chatwoot/chatwoot/pull/14538

PR #14546, opened 2026-05-22
fix(email): strip null bytes from inbound mailbox messages

vishnu-narayanan:
Inbound emails containing null bytes could fail while Chatwoot persisted the incoming message, causing IMAP sync to drop and repeatedly retry the same malformed email. This strips null bytes at the inbound mailbox message persistence boundary so the rest of the email can be saved normally.

Fixes https://linear.app/chatwoot/issue/CW-7165/argumenterror-string-contains-null-byte-argumenterror
Sentry: https://chatwoot-p3.sentry.io/issues/7405946944/

## Test
- [x] added specs
- [x] manually tested the logic on prod for the inbox(email) which was throwing the error

## What changed

- Sanitizes null bytes only while building attributes for inbound mailbox message persistence.
- Uses the sanitized source ID for both duplicate lookup and message creation.
- Adds regression coverage for the mailbox message creation path.


sony-mathew (commented):
Got the following find from codex review.

**Findings**
[P1] Sanitization happens too late for full inbound email processing. The new helper sanitizer only runs inside create_message, but both full mailbox paths touch mail-derived values before that: ReplyMailbox saves a new conversation before calling create_message (reply_mailbox.rb (line 14)), NewConversationStrategy stores raw in_reply_to and mail_subject in additional_attributes (new_conversation_strategy.rb (line 63)), and IMAP does the same before message creation (imap_mailbox.rb (line 101)). Reply/reference lookup also still queries raw null-byte headers via Message.find_by(source_id: ...) (in_reply_to_strategy.rb (line 28), references_strategy.rb (line 46)). I verified PostgreSQL still raises on these cases: null-byte source_id binds raise ArgumentError: string contains null byte, and jsonb with \u0000 raises PG::UntranslatableCharacter. So malformed emails with null bytes in Subject, In-Reply-To, or References can still abort and retry. The sanitizer should be applied before all DB queries/persistence using mail-derived strings, with integration coverage through ReplyMailbox/ImapMailbox.

app/mailboxes/mailbox_sanitizer.rb
```diff
@@ -0,0 +1,34 @@
...
+  def sanitized_content_attributes
+    sanitize_mailbox_value(
+      email: processed_mail.serialized_data,
+      cc_email: processed_mail.cc,
+      bcc_email: processed_mail.bcc
+    )
+  end
+
+  def sanitize_mailbox_value(value)
+    return value.delete(NULL_BYTE) if value.is_a?(String)
```
sony-mathew, line 28: @vishnu-narayanan Let's add a "nil" handling case also here? There are some recursions happening in here. It's better to have that case also handled.

sony-mathew: On second thought, I think this should be okay.

sony-mathew (approved):
@vishnu-narayanan I have made changes on top of your changes. Can you review those once?

app/mailboxes/mailbox_helper.rb
```diff
@@ -1,27 +1,16 @@
...
-      content_attributes: {
-        email: processed_mail.serialized_data,
-        cc_email: processed_mail.cc,
-        bcc_email: processed_mail.bcc
-      }
-    )
+    source_id = sanitize_mailbox_value(processed_mail.message_id)
+    return if @conversation.messages.find_by(source_id: source_id).present?
+
+    @message = @conversation.messages.create!(sanitized_message_attributes(source_id))
```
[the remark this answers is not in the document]
vishnu-narayanan: Will take this as a followup

vishnu-narayanan:
@sony-mathew Looks good.

PR #14570, opened 2026-05-25
feat(onboarding): allow full color channel icons

scmmishra:
`ChannelIcon` previously only rendered a monochrome glyph for each channel. It now accepts an opt-in `use-brand-icon` prop that renders the full-color brand logo (Gmail, Outlook, WhatsApp, Instagram, TikTok, Messenger, Line, Telegram) when one exists, falling back to the glyph otherwise. Existing usages are unaffected since the prop defaults off.

image


scmmishra:
@codex please review

iamsivin (approved):
LGTM

PR #14616, opened 2026-06-02
feat: ability to open article in widget via SDK

scmmishra:


scmmishra:
@codex please review

iamsivin (commented):
@scmmishra I tested this PR locally and found an issue. If any unread bubbles are visible and I run `window.$chatwoot.openArticle('slug')`, the article doesn't open correctly. I think it's reusing the same height calculation as the unread bubble wrapper, which causes the layout issue. Please see the attached screencast for reference.

Also, could you please update the PR description?

https://github.com/user-attachments/assets/837fbc2f-79af-4d3b-9bb7-7510ffb9b956

PR #14610, opened 2026-06-01
feat(conversations): remove unread count feature flag (CW-7237)

sony-mathew:
## Description

Make conversation unread counts always available at runtime by removing account feature checks from the API endpoint, unread-count listener, notifier, and ActionCable broadcast path.

Update the dashboard to fetch sidebar unread counts for the active account without checking FEATURE_FLAGS.CONVERSATION_UNREAD_COUNTS, and remove the now-unused store clear action that only supported the disabled state.

Keep the feature entry in config/features.yml to preserve flag bit order, but mark it enabled and deprecated so fresh installs default to the always-on behavior while feature-management UI hides it.

Leave existing installation default rows untouched; no migration is included, so upgraded installs may still store the old flag value but runtime behavior no longer depends on it.

Update specs around the new always-on contract and remove obsolete disabled-feature assertions.

Fixes # CW-7237

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

Update specs around the new always-on contract and remove obsolete disabled-feature assertions. Ran specs locally for the changes.

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #14625, opened 2026-06-02
fix: Un-revert unread count feature flag removal

sony-mathew:
We will merge this PR only after https://github.com/chatwoot/chatwoot/pull/14609 is merged.

This PR un-reverts #14623 by reverting its merged commit, reapplying the conversation unread count feature flag removal from #14610. Conversation unread counts are again treated as always-on while the feature remains deprecated.

## Closes
- None

## What changed
- Removes the restored `conversation_unread_counts` feature-flag gates from the unread-count API, ActionCable, notifier/listener, and dashboard sidebar paths.
- Marks `conversation_unread_counts` as enabled and deprecated again.
- Removes the disabled-feature specs restored by #14623.

## How to test
- Verify the dashboard sidebar fetches and refreshes unread counts without checking account feature flag state.
- Verify unread count API requests and unread-count broadcasts continue to work for accounts where `conversation_unread_counts` is not explicitly enabled.

### DIFF
--- a/app/controllers/api/v1/accounts/conversations/unread_counts_controller.rb
+++ b/app/controllers/api/v1/accounts/conversations/unread_counts_controller.rb
@@ -1,16 +1,6 @@
 class Api::V1::Accounts::Conversations::UnreadCountsController < Api::V1::Accounts::BaseController
-  before_action :ensure_unread_counts_enabled
-
   def index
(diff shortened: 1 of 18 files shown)


sony-mathew:
Will pickup this later.

PR #14642, opened 2026-06-03
feat(help-center): support per-locale portal title, name & header

pranavrajs:
Portals can now override their name, page title and header text per locale, with a fallback chain of locale override → default locale → base value. Overrides are stored under portal.config.locale_translations and validated with a JSON schema.

Editing is exposed as a "Localize content" action in each non-default locale's menu on the Locale page, and all public-facing portal views (classic + documentation layouts) render the localized values. The live chat widget is also loaded in the active portal locale.

Screenshot 2026-06-03 at 2 48 59 PM

https://github.com/user-attachments/assets/f5affbd2-7ad4-415a-b376-57c759fc2aaa

### DIFF
--- a/app/controllers/api/v1/accounts/portals_controller.rb
+++ b/app/controllers/api/v1/accounts/portals_controller.rb
@@ -80,8 +80,13 @@ class Api::V1::Accounts::PortalsController < Api::V1::Accounts::BaseController
       :id, :color, :custom_domain, :header_text, :homepage_link,
       :name, :page_title, :slug, :archived,
       { config: [:default_locale, :layout, { allowed_locales: [] }, { draft_locales: [] },
-                 { social_profiles: %i[facebook x instagram linkedin youtube tiktok github whatsapp] }] }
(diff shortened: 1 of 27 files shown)


iamsivin (approved):
LGTM

sony-mathew (approved):
Overall looks good to me. Just one observation and a question: I can see that everywhere we are passing the locale to display title of the portal. Is there some way like setting Portal.current.locale upstream somewhere and display_title automatically resolves. I understand, there might be some cases where we want to explicitly pass the locale for which we can optionally accept a parameter for the same.

pranavrajs:
> Overall looks good to me. Just one observation and a question: I can see that everywhere we are passing the locale to display title of the portal. Is there some way like setting Portal.current.locale upstream somewhere and display_title automatically resolves. I understand, there might be some cases where we want to explicitly pass the locale for which we can optionally accept a parameter for the same.

I tried doing this, but the solution was more complicated than this. We can consider the refactor in a separate PR.

PR #14645, opened 2026-06-04
feat: add per-inbox toggle to disable incoming calls

tds-1:
## Description

Adds a per-inbox "Allow incoming calls" toggle for voice-enabled WhatsApp and Twilio inboxes. When turned off, the setting is persisted on the channel; actually rejecting inbound calls is handled in a follow-up PR.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

## Screenshot
Screenshot 2026-06-04 at 11 44 07 AM

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #14646, opened 2026-06-04
feat: enforce per-inbox incoming-calls toggle

tds-1:
Enforces the per-inbox **Allow incoming calls** toggle added in #14645. When an admin turns the setting off, incoming calls are now actually stopped at the inbound webhook — before any conversation, Call row, or agent ring.

Stacked on top of #14645 (base branch: `feat/inbox-incoming-calls-toggle`).

Closes — follow-up to #14645

## What changed
- **WhatsApp** — `Whatsapp::IncomingCallService#create_inbound_call` returns early (logs only) when `inbox.channel.inbound_calls_enabled?` is false. Outbound calls and call-status webhooks are unaffected.
- **Twilio** — `Twilio::VoiceController#call_twiml` responds with `<Reject/>` TwiML for a fresh contact-initiated inbound leg when the toggle is off. The gate lives in `call_twiml` (not `set_inbox!`) so agent/outbound legs of an in-progress call are never dropped.

## How to test
1. On a voice-enabled WhatsApp or Twilio inbox, turn **Allow incoming calls** off.
2. Place a call to the business number from a customer.
3. WhatsApp: the call is ignored — no ring, no conversation. Twilio: the caller is declined.
4. Turn it back on → calls ring agents and create a conversation as before.


sony-mathew (commented):
Can we add specs for this case @tds-1 for both Whatsapp and Twilio?

sony-mathew (approved):
LGTM

tds-1:
@codex please review

PR #14657, opened 2026-06-05
fix(whatsapp): prevent duplicate brazil contacts in whatsapp co-existence

### LINKED ISSUE #13986: Duplicate contacts created in Brazil (9th digit issue) when using WhatsApp Coexistence with device-initiated conversations
### Describe the bug

When using WhatsApp Coexistence in Chatwoot (self-hosted), contacts are duplicated due to inconsistent phone number formats in Brazil (with and without the 9th digit).

If a conversation is initiated from the native WhatsApp device (echo message), Chatwoot creates a new contact using the number format returned by WhatsApp (without the 9th digit), even when a contact already exists with the normalized format (with the 9th digit).

This results in:

Duplicate contacts for the same user
Separate conversations for the same customer
Broken conversation history and assignment

According to the official Chatwoot documentation, the system should normalize Brazilian numbers and match both formats automatically. However, this behavior is not working in the Coexistence scenario.

### To Reproduce

To Reproduce
Use Chatwoot with WhatsApp Coexistence enabled
Ensure a contact exists with Brazilian number in E.164 format (with 9th digit), e.g.:
+55 88 9XXXXXXX
Send a message from the native WhatsApp device (outside Chatwoot)
Wait for the message to be received in Chatwoot (echo scenario)
Observe that:
A new contact is created with the same number but without the 9th digit
A new conversation is opened instead of linking to the existing contact

Image

### Expected behavior

Chatwoot should normalize Brazilian phone numbers and:

Image

Match incoming messages (with or without 9th digit) to the existing contact
Avoid creating duplicate contacts
Maintain a single conversation thread per customer

### Environment

Docker

### Cloud Provider

None

### Platform

Browser

### Operating system

 Ubuntu 24.04.3

### Browser and version

Google Chrome Versão 146.0.7680.167 (Versão oficial) 64 bits

### Docker (if applicable)

_No response_

### Additional context

The issue occurs specifically when conversations are initiated from the WhatsApp device (echo messages)
Does not seem to happen when conversations are initiated directly via Chatwoot
Screenshots attached showing:
Duplicate contacts (with and without 9th digit)
Separate conversations created for the same user

This behavior contradicts the official documentation:
[https://www.chatwoot.com/hc/user-guide/articles/1758697086-inconsistencies-for-whats_app-numbers-in-brazil-mexico-and-argentina](https://www.chatwoot.com/hc/user-guide/articles/1758697086-inconsistencies-for-whats_app-numbers-in-brazil-mexico-and-argentina?utm_source=chatgpt.com)

muhsin-k:
When using WhatsApp Coexistence, a message sent from the native WhatsApp app can create a duplicate contact and a separate conversation for a customer who already exists in Chatwoot. This happens for Brazilian numbers when WhatsApp delivers the number without the mobile 9th digit while the existing contact is stored with the normalized number.

Chatwoot previously deduplicated incoming WhatsApp identifiers mostly at the `contact_inbox.source_id` level. When no matching `contact_inbox` existed, contact creation fell back to an exact `phone_number` lookup using the raw provider number, so an existing normalized contact could be missed. This change makes contact lookup try safe phone-number candidates so the same customer resolves to the existing contact and conversation history stays intact.

### Closes

- https://linear.app/chatwoot/issue/CW-6724/duplicate-contacts-created-in-brazil-9th-digit-issue-when-using
- Closes https://github.com/chatwoot/chatwoot/issues/13986

### How to reproduce

1. Enable WhatsApp Coexistence on a self-hosted instance.
2. Ensure a contact exists with a Brazilian number including the mobile 9th digit, for example `+5511912345678`.
3. From the native WhatsApp Business app, send a message to that customer so Chatwoot receives an echo event with the number without the 9th digit, for example `+551112345678`.
4. Before this change, Chatwoot creates a duplicate contact and a separate conversation.
5. After this change, Chatwoot reuses the existing contact and the message lands in the existing conversation context.

### Things to know

- Exact phone-number matches still win before generated candidates, so existing contacts stored in the provider-delivered format are preserved.
- Brazil candidate matching is limited to legacy mobile ranges so landlines are not rewritten into mobile-looking numbers.
- The same contact candidate logic is applied to WhatsApp Cloud, WhatsApp Coexistence echo events, Twilio WhatsApp, and Enterprise WhatsApp calls.

### What changed

- Added safe contact phone-number candidate lookup to `ContactInboxWithContactBuilder`.
- Centralized candidate generation through `Whatsapp::PhoneNumberNormalizationService#phone_number_candidates`.
- Added Brazil-specific contact candidate rules and regression coverage for raw, normalized, landline, profile-name, Twilio, and Enterprise call paths.


muhsin-k:
@codex please review

tds-1 (approved):

PR #14662, opened 2026-06-06
fix(voice): reuse conversations for outbound calls

### LINKED ISSUE #14663: Twilio voice calls create new conversations despite single-conversation setting
## Problem

When a Twilio voice call is started from an existing conversation, Chatwoot can create a new conversation for the call instead of adding the voice call bubble to the existing reusable conversation.

This is surprising when the inbox is configured to keep only one conversation per contact, because incoming Twilio messages already respect that setting and reuse the contact's conversation.

## Steps to reproduce

1. Configure a Twilio voice-enabled inbox.
2. Enable the inbox setting to lock the contact to a single conversation.
3. Open an existing conversation for a contact in that inbox.
4. Click the phone call button in the conversation sidebar.
5. Complete or observe the Twilio voice call creation flow.

## Expected behavior

The outgoing voice call should be added to the reusable conversation for that contact. If the inbox is locked to a single conversation and the last conversation is resolved, the call flow should reuse and reopen that conversation before adding the voice call bubble.

## Actual behavior

The outbound Twilio call path can create a fresh conversation for the same contact instead of attaching the call to the existing/reusable conversation.

## Notes

The issue appears to be in the outbound voice call builder. It only reuses a conversation when the controller passes an explicit open conversation hint; otherwise it falls back to creating a new conversation. It does not apply the same open-or-single-conversation routing used by incoming Twilio/WhatsApp message and call handling.

salmonumbrella:
Outbound Twilio voice calls now attach the call bubble to the reusable conversation for that contact instead of always creating a fresh conversation. This makes the voice call flow respect the same open-or-single-conversation routing used by incoming messages.

## Closes

Closes #14663

## How to reproduce

1. Configure a Twilio voice-enabled inbox.
2. Enable the inbox setting to lock the contact to a single conversation.
3. Open an existing conversation for a contact in that inbox.
4. Start a Twilio voice call from the conversation sidebar.

## What changed

- Updated `Voice::OutboundCallBuilder` to resolve an existing conversation before creating a new one.
- Reused the last conversation when `lock_to_single_conversation` is enabled, reopening it before adding the outgoing voice call bubble.
- Preserved the existing behavior of creating a new conversation when the previous one is resolved and the single-conversation lock is disabled.
- Added Enterprise voice service specs for open conversation reuse, locked conversation reuse, and unlocked resolved conversation behavior.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -593,7 +593,7 @@ GEM
       sidekiq
     newrelic_rpm (9.6.0)
       base64
-    nio4r (2.7.3)
+    nio4r (2.7.5)
     nokogiri (1.19.3)
       mini_portile2 (~> 2.8.2)
       racc (~> 1.4)
@@ -682,7 +682,7 @@ GEM
     pry-rails (0.3.9)
       pry (>= 0.10.4)
     public_suffix (7.0.5)
-    puma (6.4.3)
+    puma (8.0.2)
       nio4r (~> 2.0)
     pundit (2.3.0)
       activesupport (>= 3.0.0)
--- a/enterprise/app/services/voice/outbound_call_builder.rb
+++ b/enterprise/app/services/voice/outbound_call_builder.rb
@@ -19,7 +19,7 @@ class Voice::OutboundCallBuilder
 
     ActiveRecord::Base.transaction do
       contact_inbox = ensure_contact_inbox!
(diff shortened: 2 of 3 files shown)


salmonumbrella:
@tds-1 tiny QOL 🙏

PR #14680, opened 2026-06-09
fix(portal): emit valid BCP 47 html lang attribute

pranavrajs:
Locales are stored with underscores (e.g. pt_BR), but the HTML lang attribute requires hyphens (pt-BR). Convert via a helper in the public portal layouts.


iamsivin (approved):
LGTM 

**Before**

Image

**After**

Image

PR #14687, opened 2026-06-09
feat(backend): add ability to disable inboxes

muhsin-k:
Adds an inbox-level disable toggle so admins can temporarily stop an inbox from accepting or sending messages without deleting it. Disabled inboxes remain visible and configurable, but conversations become read-only and customer-facing entry points stop creating new activity.

Fixes https://linear.app/chatwoot/issue/CW-4998/add-ability-to-disable-an-inbox-without-deleting-it

## How to test

1. Open inbox settings and disable an inbox.
2. Confirm the inbox shows a disabled badge in inbox/settings/conversation surfaces.
3. Open an existing conversation from that inbox and confirm the composer is disabled for replies and private notes.
4. Try creating a new conversation/message through the widget or public API and confirm it is blocked.
5. Re-enable the inbox and confirm normal conversation and messaging behavior resumes.
6. For a web widget inbox, confirm the launcher is hidden while the inbox is disabled.


muhsin-k:
@copilot resolve the merge conflicts in this pull request

sony-mathew (commented):
please take a look at these findings.

## Findings

### [P1] An in-flight one-off campaign continues sending after disable

The active check happens only once before campaign execution in [campaign.rb](/worktrees/e3e2/chatwoot/app/models/campaign.rb:58). The SMS, Twilio, and WhatsApp services then iterate the entire audience without checking persisted inbox state before each provider call, such as [oneoff_sms_campaign_service.rb](/worktrees/e3e2/chatwoot/app/services/sms/oneoff_sms_campaign_service.rb:19).

I reproduced this with two SMS contacts and disabled the inbox during the first provider send:

```text
sends: 2
inbox_active: false
campaign_status: completed
```

The second customer still received a send and the campaign was marked completed.

The active state needs to be rechecked at the actual provider-send boundary or before each recipient, with explicit semantics for the interrupted campaign status.

### [P1] Macros and transcript APIs remain writable for disabled inboxes

The macro execution endpoint queues work without checking any conversation inbox in [macros_controller.rb](/worktrees/e3e2/chatwoot/app/controllers/api/v1/accounts/macros_controller.rb:50). The asynchronous service subsequently executes every action in [execution_service.rb](/worktrees/e3e2/chatwoot/app/services/macros/execution_service.rb:10).

I reproduced a macro changing the priority of a disabled-inbox conversation:

```text
inbox_active: false
priority_after_macro: urgent
```

Message-builder actions fail, but the service catches those failures and continues. Later actions can still update status/assignment/labels or enqueue webhooks and transcript emails. The direct transcript API also sends without an active-inbox gate in [conversations_controller.rb](/worktrees/e3e2/chatwoot/app/controllers/api/v1/accounts/conversations_controller.rb:71), as does the shared [ActionService](/worktrees/e3e2/chatwoot/app/services/action_service.rb:82).

Add a job/service-side gate so disabling after enqueue is also safe. The API should return the established disabled-inbox error where possible.

### [P1] The required rebase would let delayed automations execute after disable

Current `develop` contains delayed automations that were added after this PR’s merge base. A pending execution scheduled while an inbox is active can become due after it is disabled. The [pending-execution job](https://github.com/chatwoot/chatwoot/blob/7142bc43a585c410a0cf0deab9344e6f00a42b6c/app/jobs/automation_rules/process_pending_execution_job.rb#L6-L15) checks account feature state and rule conditions, but not inbox state, before invoking actions.

The PR only prevents new listener-time automation execution. After rebase, existing pending rows can still mutate conversations or enqueue email/webhook actions.

The rebase should add an inbox check immediately before claiming or executing pending work and define whether disabled work is paused or skipped. Cover active → scheduled → disabled before due, plus re-enable behavior.

### [P2] Skipping continuity email leaves its one-hour deduplication key behind

The notification service sets the Redis key before scheduling the delayed job in [send_email_notification_service.rb](/worktrees/e3e2/chatwoot/app/services/messages/send_email_notification_service.rb:8). The new inactive-inbox return in [conversation_reply_email_job.rb](/worktrees/e3e2/chatwoot/app/jobs/conversation_reply_email_job.rb:4) occurs before the key is deleted.

I reproduced:

```text
key_after_skip: "1"
jobs_after_reenable: 0
```

After re-enabling, the next valid message cannot schedule continuity email until the key expires, potentially one hour later. Ensure the key is removed on disabled/account-inactive exits as well.

PR #14701, opened 2026-06-10
fix: include widget ET translation files

linbaublys:
Include ET translation files


linbaublys:
@sojan-official @sony-mathew 
is there any way to push this small fix into latest build of chatwoot?

sojan-official (approved):
Thanks for fixing this! Looks good 👍

PR #14708, opened 2026-06-11
fix(conversation): hide transcript assignee option when email missing

### LINKED ISSUE #14553: Send conversation transcript to assigned agent fails with 422 when assignee email is missing from conversation meta
### Describe the bug

The dashboard **Send conversation transcript** modal can submit a transcript request without an `email` parameter when choosing **Send the transcript to the assigned agent**.

The backend then returns:

```text
422 Unprocessable Content
{ error: "email param missing" }
```

This appears to happen because the frontend enables the assigned-agent option when `currentChat.meta.assignee` exists, but then reads `currentChat.meta.assignee.email`. In at least some conversation payloads, `meta.assignee` exists but does not include `email`.

### To Reproduce

1. Open a conversation that has an assigned agent.
2. Open the **Send conversation transcript** modal.
3. Select **Send the transcript to the assigned agent**.
4. Submit.

### Expected behavior

The transcript should be sent to the assigned agent, or the UI should not offer/enable the assigned-agent option unless an email is available.

### Actual behavior

The frontend sends a POST request without a usable `email` parameter:

```text
POST /api/v1/accounts/:account_id/conversations/:id/transcript
```

Observed Rails params from production logs:

```ruby
{
  "account_id" => "2",
  "id" => "3148",
  "conversation" => {}
}
```

The request completes with:

```text
Completed 422 Unprocessable Content
```

### Relevant code

Frontend (`EmailTranscriptModal.vue`):

```js
selectedEmailAddress() {
  const { meta } = this.currentChat;
  switch (this.selectedType) {
    case 'contact':
      return meta.sender.email;
    case 'assignee':
      return meta.assignee.email;
    case 'other_email_address':
      return this.email;
    default:
      return '';
  }
}
```

Backend (`Api::V1::Accounts::ConversationsController#transcript`):

```ruby
def transcript
  render json: { error: 'email param missing' }, status: :unprocessable_entity and return if params[:email].blank?
  # ...
end
```

### Possible fixes

One of:

1. Include `email` in `currentChat.meta.assignee` when serializing conversation meta.
2. Hide/disable **Send to assigned agent** unless `meta.assignee.email` is present.
3. Resolve the assigned agent email server-side if `params[:email]` is blank and the conversation has an assignee.

The safest frontend UX fix may be (1) or (2); a server-side fallback would also make the endpoint more robust.

### Environment

- Chatwoot self-hosted
- Chatwoot Cloudron package `com.chatwoot.cloudronapp`
- Package version: `1.52.0`
- Docker image: `cloudron/com.chatwoot.cloudronapp:202605191426440000`

augustocbx:
## Description

The **Send conversation transcript** modal offered the "Send to assigned agent" option whenever a conversation had an assignee, then submitted `meta.assignee.email` — which is absent for some assignees (e.g. agent bots). The request was posted without an `email` param and the backend returned `422 "email param missing"`. Fixes chatwoot/chatwoot#14553.

### Root cause

`EmailTranscriptModal.vue` rendered the assignee radio with `v-if="currentChat.meta.assignee"`, but `selectedEmailAddress()` reads `meta.assignee.email`. When the assignee has no email, an empty `email` is submitted and `Api::V1::Accounts::ConversationsController#transcript` rejects it with 422.

### Solution

Guard the assignee radio with `currentChat.meta.assignee.email`, mirroring the existing contact-radio guard (`currentChat.meta.sender.email`) in the same template, so the option only appears when a usable email is available.

## Type of change

## How Has This Been Tested?

* Added `app/javascript/dashboard/components/widgets/conversation/EmailTranscriptModal.spec.js`: asserts the assignee option is hidden when the assignee has no email and shown when it does; fails without the fix.
* `pnpm run test EmailTranscriptModal.spec.js` ✅  ·  `pnpm run eslint` ✅
* Manual QA: assign a conversation to an assignee whose meta lacks an email → open the transcript modal → "Send to assigned agent" no longer appears, so the 422 can no longer be triggered.

## Checklist

- [X] I have commented my code in hard-to-understand areas (where needed)
- [ ] I have made corresponding changes to the documentation (N/A — product docs are separate)
- [X] I have added tests that prove my fix is effective
- [X] Targets the `develop` branch


augustocbx:
@sojan-official 
Just checking in: is there anything else needed from my side for this PR? I’d be happy to make any requested changes.

sojan-official:
Thanks for the contribution and for following up here.

The reported AgentBot case has since been addressed in #14866 by using the assignee type to exclude AgentBot assignees from the transcript option.

Since the underlying issue is now resolved through the assignee-type check, we’re closing this PR as superseded by #14866. Thanks again for investigating this and adding the test coverage.

PR #14722, opened 2026-06-13
fix: TypeError in EmailMeta when message has no sender (Fixes #14720)

### LINKED ISSUE #14720: TypeError in EmailMeta breaks the conversation view when a message has no sender
## Describe the bug

Opening a conversation that contains an email-bubble message **without an associated `sender`** crashes the rendering of the entire conversation view:

```
TypeError: Cannot read properties of null (reading 'email')
```

All messages in the conversation disappear (broken view), not just the offending one.

## To Reproduce

1. Create a message with email content attributes but no sender (sender is an optional association). E.g. in `rails console`:
   ```ruby
   conversation.messages.create!(
     message_type: :outgoing,
     content: 'hello',
     content_attributes: { email: { from: ['agent@example.com'] } }
   )
   ```
   In the wild this happens with messages created through the API or synced from an external email system.
2. Open the conversation in the dashboard.
3. The conversation view fails to render.

## Root cause

`app/javascript/dashboard/components-next/message/bubbles/Email/EmailMeta.vue` — the `senderName` computed does `sender.value.email ?? ''` without a null guard, but `sender` is nullable by design (`belongs_to :sender, optional: true`; the `Message.vue` prop defaults to `null`; the provider typedef documents `Ref<Sender|null>`). The `?? ''` only protects the `.email` result, not the `sender.value` access itself.

## Expected behavior

The meta block should fall back to showing the raw `from` address when the message has no sender record.

Happy to submit the one-line fix (optional chaining, same pattern as `contentAttributes` in the same file) — PR incoming.

---
*Found while running Chatwoot v4.13.0 in production; verified the unguarded access still exists on current `develop`. Diagnosed and fixed with the assistance of Claude (Anthropic) as AI pair programmer.*

nico-stever:
# Pull Request Template

## Description

Opening a conversation that contains an email-bubble message without an associated `sender` crashes the rendering of the entire conversation view with:

```
TypeError: Cannot read properties of null (reading 'email')
```

**Root cause**

In `app/javascript/dashboard/components-next/message/bubbles/Email/EmailMeta.vue`, the `senderName` computed dereferences the sender without a null guard:

```js
const senderEmail = sender.value.email ?? '';
```

However, `sender` is explicitly nullable throughout the stack:

- `Message` model: `belongs_to :sender, polymorphic: true, optional: true`, and `merge_sender_attributes` only includes `sender` in the payload when it is present.
- `Message.vue` declares the prop as `sender: { type: Object, default: null }`.
- The message context typedef in `components-next/message/provider.js` documents it as `Ref<Sender|null>` with default `null`.

The `?? ''` fallback only protects against `email` being null/undefined after the property access succeeds — it does not protect against `sender.value` itself being `null`. When a message has email content attributes (so `fromEmail[0]` is truthy and the meta block renders) but no sender record — e.g. messages created via the API or synced/imported from an external email system — the computed throws during render and the whole conversation message list fails to render (blank/broken conversation, every message disappears, not just the offending one).

**Fix**

Use optional chaining, matching the pattern already used for `contentAttributes` in the same file:

```js
const senderEmail = sender.value?.email ?? '';
```

The only other `sender` access in this computed, `sender.value.name`, is unreachable when `sender` is null after this change: it requires `fromEmailAddress === senderEmail` with `senderEmail` non-empty, which implies a non-null sender (when both are empty the computed returns early). So no further guards are needed.

**Steps to reproduce**

1. Create a message with email content attributes but no sender — `sender` is an optional association, so e.g. in `rails console`:
   ```ruby
   conversation.messages.create!(
     message_type: :outgoing,
     content: 'hello',
     content_attributes: { email: { from: ['agent@example.com'] } }
   )
   ```
   (In the wild this happens with messages created through the API / imported from external email systems, where `content_attributes.email` is set but no `User`/`Contact` sender is associated.)
2. Open the conversation in the dashboard.
3. The conversation view fails to render with `TypeError: Cannot read properties of null (reading 'email')` from the `senderName` computed.

Fixes #14720

## Type of change

## How Has This Been Tested?

- Reproduced the crash on v4.13.0 and verified the same unguarded access exists on current `develop` (`EmailMeta.vue`, `senderName` computed).
- With the fix applied (running in production on our fork since 2026-06-08), conversations containing sender-less email messages render correctly: the meta block falls back to showing just the raw `from` address (the `v-else` branch), and `senderName` behaves exactly as before for messages that do have a sender.
- Note on specs: there are currently no specs for the `components-next/message/bubbles/Email` components; the component consumes the inject-based message context plus a Vuex getter (`getSelectedChatAttachments`), so a unit spec needs a small wrapper harness around `provideMessageContext`. Happy to add one in this PR if the maintainers want it.

## Checklist:

- [ ] I have made corresponding changes to the documentation (N/A)
- [ ] I have added tests that prove my fix is effective or that my feature works (see note above — can add on request)
- [ ] Any dependent changes have been merged and published in downstream modules (N/A)

---
*Developed and verified in production with the assistance of Claude (Anthropic) as AI pair programmer.*

### DIFF
--- a/app/javascript/dashboard/components-next/message/bubbles/Email/EmailMeta.vue
+++ b/app/javascript/dashboard/components-next/message/bubbles/Email/EmailMeta.vue
@@ -28,7 +28,8 @@ const ccEmail = computed(() => {
 
 const senderName = computed(() => {
   const fromEmailAddress = fromEmail.value[0] ?? '';
-  const senderEmail = sender.value.email ?? '';
+  // sender can be null (e.g. API-created or imported messages without a sender)
+  const senderEmail = sender.value?.email ?? '';
 
   if (!fromEmailAddress && !senderEmail) return null;
 


a99430331-wq (commented):
Thanks

PR #14723, opened 2026-06-13
fix: preserve contact_inboxes when EDIT_CONTACT receives a partial payload (Fixes #14721)

### LINKED ISSUE #14721: Contact loses contact_inboxes after websocket contact.updated — New Conversation modal shows no inboxes
## Describe the bug

After a contact is updated via websocket (`contact.updated`), the contact **loses its `contact_inboxes` in the dashboard store**. Symptom: opening the **New Conversation** modal for that contact shows **no inboxes to choose from** until a full page reload.

## To Reproduce

1. Open the dashboard with a contact that has at least one contactable inbox.
2. Trigger any contact update that arrives via websocket (e.g. edit the contact's name from another tab/session, or any automation touching the contact).
3. Open "New Conversation" for that contact → the inbox selector is empty.
4. Hard-reload the page → inboxes are back.

## Root cause

`app/javascript/dashboard/store/modules/contacts/mutations.js` — the `EDIT_CONTACT` mutation replaces the whole record:

```js
[types.EDIT_CONTACT]: ($state, data) => {
  $state.records[data.id] = data;
},
```

Websocket `contact.updated` payloads do not include `contact_inboxes`, so the replacement silently drops them from the store.

## Expected behavior

A partial payload should not wipe `contact_inboxes` — merge over the existing record, keeping the previous `contact_inboxes` when the payload doesn't carry them.

PR with the fix + specs for both cases incoming.

---
*Found while running Chatwoot v4.13.0 in production; verified the mutation is unchanged on current `develop`. Diagnosed and fixed with the assistance of Claude (Anthropic) as AI pair programmer.*

nico-stever:
# Description

## Problem

If a contact's page is open in the dashboard and a `contact.updated` websocket event arrives for that contact (e.g. the contact was renamed in another tab, by another agent, or via the API), the "New conversation" modal for that contact shows **no inboxes to send from**. The list only comes back after a full page reload.

## Root cause

The `EDIT_CONTACT` mutation replaces the whole store record instead of merging:

```js
// app/javascript/dashboard/store/modules/contacts/mutations.js
[types.EDIT_CONTACT]: ($state, data) => {
  $state.records[data.id] = data;
},
```

The websocket path commits the raw push payload:

- `actionCable.js`: `'contact.updated'` → `onContactUpdate` → `dispatch('contacts/updateContact', data)`
- `contacts/actions.js`: `updateContact` → `commit(types.EDIT_CONTACT, updateObj)` (no API call)

That payload is built by `Contact#push_event_data` (`app/models/contact.rb`), which does **not** include `contact_inboxes`. So the `contact_inboxes` previously fetched via the contact show API are dropped from the store, and the new-conversation flow (which reads `contact.contact_inboxes` to build the list of contactable inboxes) renders an empty inbox selector.

Note the inconsistency with the neighbouring `SET_CONTACT_ITEM` mutation, which already does a shallow merge with the existing record for the same reason.

## How to reproduce

1. Open a contact that has at least one contact inbox (visiting the contact page loads `contact_inboxes` into the store).
2. While that page stays open, trigger a `contact.updated` broadcast for it — e.g. rename the contact from another browser tab/agent session, or `PATCH /api/v1/accounts/:account_id/contacts/:id`.
3. Click "New conversation" for that contact → the inbox dropdown is empty.
4. Reload the page → inboxes are back.

## Fix

Preserve the previously known `contact_inboxes` when the incoming payload doesn't carry them. Payloads that do include `contact_inboxes` (e.g. the contact show/update API responses) still take precedence:

```js
[types.EDIT_CONTACT]: ($state, data) => {
  const existingContact = $state.records[data.id] || {};
  $state.records[data.id] = {
    ...data,
    contact_inboxes: data.contact_inboxes || existingContact.contact_inboxes,
  };
},
```

I considered making `EDIT_CONTACT` a full shallow merge like `SET_CONTACT_ITEM`, but that could resurrect stale keys that a fresh API payload intentionally no longer carries, so I kept the change minimal and targeted at the field that is actually lost and user-visible. Happy to switch to a full merge if maintainers prefer.

This has been running in production on our v4.13.0 deployment since June 2026 with no regressions.

Fixes #14721

## Type of change

## How Has This Been Tested?

- Added two unit tests to `app/javascript/dashboard/store/modules/specs/contacts/mutations.spec.js`:
  - `preserves contact_inboxes when payload does not include them`
  - `uses contact_inboxes from the payload when present`
- Existing `EDIT_CONTACT` spec passes unchanged.
- Manually verified the reproduction above before/after the fix on a production deployment.

## Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas

---
*Developed and verified in production with the assistance of Claude (Anthropic) as AI pair programmer.*

## Scope note

This is a **client-side mitigation**: it protects the window between a websocket `contact.updated` and the next REST fetch. After a contact address change, the server-side `ContactInbox` row keeps its old `source_id` (no backend sync exists today), so a later REST response that includes `contact_inboxes` will reintroduce that entry verbatim — which is fine in practice because every New-Conversation entry point already refetches contactable inboxes (`Contacts::ContactableInboxesService` recomputes from the current addresses). A backend sync of `source_id` on address change would be the deeper fix; intentionally out of scope here.

### DIFF
--- a/app/javascript/dashboard/store/modules/contacts/mutations.js
+++ b/app/javascript/dashboard/store/modules/contacts/mutations.js
@@ -60,20 +60,38 @@ export const mutations = {
   [types.EDIT_CONTACT]: ($state, data) => {
     // Websocket `contact.updated` payloads don't include `contact_inboxes`;
     // preserve them from the existing record so the new conversation modal
-    // doesn't lose the list of contactable inboxes. Contactable inboxes are
-    // derived from the contact's current email/phone_number, so only carry
-    // them over when those identifying addresses are unchanged — otherwise
-    // drop them rather than risk composing to a stale address.
+    // doesn't lose the list of contactable inboxes. Channels that derive
+    // their source_id from an identifying address (email / phone number)
+    // are dropped when that address changes — their entries would point at
+    // the stale address; other channels (API, web widget, …) are kept.
+    const EMAIL_CHANNELS = ['Channel::Email'];
+    const PHONE_CHANNELS = [
+      'Channel::Whatsapp',
+      'Channel::Sms',
+      'Channel::TwilioSms',
+    ];
     const existingContact = $state.records[data.id] || {};
-    const addressesUnchanged =
-      data.email === existingContact.email &&
-      data.phone_number === existingContact.phone_number;
-    $state.records[data.id] = {
-      ...data,
-      contact_inboxes:
-        data.contact_inboxes ||
-        (addressesUnchanged ? existingContact.contact_inboxes : undefined),
-    };
+    let contactInboxes = data.contact_inboxes;
+    if (!contactInboxes && existingContact.contact_inboxes) {
+      const staleChannels = [];
+      if (existingContact.email && data.email !== existingContact.email) {
+        staleChannels.push(...EMAIL_CHANNELS);
+      }
+      if (
+        existingContact.phone_number &&
+        data.phone_number !== existingContact.phone_number
+      ) {
+        staleChannels.push(...PHONE_CHANNELS);
+      }
+      contactInboxes = existingContact.contact_inboxes.filter(
+        contactInbox => !staleChannels.includes(contactInbox.inbox?.channel_type)
+      );
+    }
+    const record = { ...data };
+    if (contactInboxes !== undefined) {
+      record.contact_inboxes = contactInboxes;
+    }
+    $state.records[data.id] = record;
   },
 
   [types.DELETE_CONTACT]: ($state, id) => {
--- a/app/javascript/dashboard/store/modules/contacts/mutations.js
+++ b/app/javascript/dashboard/store/modules/contacts/mutations.js
@@ -73,11 +73,18 @@ export const mutations = {
     const existingContact = $state.records[data.id] || {};
     let contactInboxes = data.contact_inboxes;
     if (!contactInboxes && existingContact.contact_inboxes) {
+      // An address only counts as changed when the payload carries the field
+      // and a previous value is known — payloads that omit it say nothing.
       const staleChannels = [];
-      if (existingContact.email && data.email !== existingContact.email) {
+      if (
+        data.email !== undefined &&
+        existingContact.email &&
+        data.email !== existingContact.email
+      ) {
         staleChannels.push(...EMAIL_CHANNELS);
       }
       if (
+        data.phone_number !== undefined &&
         existingContact.phone_number &&
         data.phone_number !== existingContact.phone_number
       ) {
--- a/app/javascript/dashboard/store/modules/contacts/mutations.js
+++ b/app/javascript/dashboard/store/modules/contacts/mutations.js
@@ -1,6 +1,16 @@
 import types from '../../mutation-types';
 import * as Sentry from '@sentry/vue';
 
+// Channels whose contact_inbox source_id derives from an identifying address
+// of the contact (see Contacts::ContactableInboxesService and
+// ContactInboxBuilder): their entries go stale when that address changes.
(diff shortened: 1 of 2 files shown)


nico-stever:
Good catch — agreed. `contact_inboxes` entries derive their `source_id` from the contact's current email/phone_number (`Contacts::ContactableInboxesService`), so blindly carrying them over across an address change could leave the compose modal pointing at a stale address.

Updated in fc6ba36: the mutation now preserves `contact_inboxes` only when the payload omits them **and** the identifying addresses (`email`/`phone_number`) are unchanged. When an address changes, the stale entries are dropped — same outcome as the current behavior for that case, so no regression there — rather than risking a send to an outdated address. Added a spec covering the address-change case.

nico-stever:
Fair point again — and digging into it surfaced that matching `source_id` strings was the wrong axis altogether: it duplicates backend serialization rules in the frontend, and it missed Twilio WhatsApp, whose source_id is `whatsapp:+E164` (`ContactInboxBuilder#twilio_source_id`), so a stale entry would have survived a phone change there.

Reworked in c04cd48 to filter by `inbox.channel_type` instead, which both serializer shapes carry:

- email changed → drop `Channel::Email` entries
- phone changed → drop `Channel::Whatsapp` / `Channel::Sms` / `Channel::TwilioSms` entries
- everything else (API, web widget, …) is kept, so the compose flow keeps working without a refetch for the channels that are still valid

Also tightened per the earlier notes: an address only counts as changed when a previous value is known (avoids false drops for records hydrated only via `fetchContactableInbox`), and the record no longer gains a phantom `contact_inboxes: undefined` key when neither side has one. Specs updated, including the Twilio WhatsApp case.

On the remaining edge (the dropped channel stays absent until a refetch): that matches today's behavior for that case, and recomputing requires an async fetch that doesn't belong in a mutation — happy to follow up with a refetch-on-address-change in the websocket action handler if the maintainers think it's worth it.

nico-stever:
Agreed — `undefined !== existing` was reading an omitted field as a change. Updated in adc14de: an address only counts as changed when the payload actually carries the field (`data.email !== undefined`, same for `phone_number`) **and** a previous value is known. A name-only partial update now preserves all entries untouched; spec added for that case.

nico-stever:
Follow-up in 7a102db (no behavior change): channel constants hoisted to module scope, prettier width fix, clarified the comment, and two more edge-case specs (cleared email, explicit empty `contact_inboxes` array). Also added a **Scope note** to the description: this is a client-side mitigation for the websocket→next-fetch window; the server-side `ContactInbox.source_id` keeps the old address (compose entry points already refetch contactable inboxes, which recomputes from current addresses) — a backend sync would be the deeper fix and is intentionally out of scope.

nico-stever:
Note: this review ran against adc14de — 7a102db (pushed shortly before) added a **Scope note** to the description covering exactly this case. To recap the reasoning:

- **Replacing the source_id with the new address client-side isn't viable**: these entries mirror server-side `ContactInbox` rows. Fabricating one in the store would point the compose flow at a record that doesn't exist server-side.
- **Recomputing requires an async fetch**, which can't live in a Vuex mutation; and dispatching it from the websocket handler for *every* `contact.updated` with an address change would fire for any contact in the account, not just the one being viewed.
- The dropped-until-refetch window **matches upstream's current behavior for this case** — today the entire list is wiped on *any* update, address change or not; this PR strictly narrows the loss.

If the maintainers want the remaining gap closed in-repo, the natural place is the compose flow itself: dispatch `contacts/fetchContactableInbox` when the modal opens for an active contact, mirroring what `EditContact.vue` / `ContactInfo.vue` already do on mount. Happy to add that here or as a follow-up PR — whichever scope you prefer.

PR #14757, opened 2026-06-16
fix(companies): persist active sidebar tab on company detail

salmonumbrella:
On a company's detail page, the sidebar tab you last viewed (History / Notes / Contacts) is now remembered when you navigate away and come back. 

### How to reproduce 
1. Open a company detail page.
2. Switch the right sidebar to the **Contacts** tab.
3. Click into one of the contacts (navigates to the contact's page).
4. Press the browser **Back** button.
5. ❌ The company page reopens on the **History** tab instead of **Contacts**.

### What changed
- The active sidebar tab is persisted to the route query (`?tab=contacts`) via `router.replace`, so switching tabs updates the URL without adding extra back-history entries.
- On mount, the view initializes the active tab from the query (validated against the known tabs, defaulting to `history`) instead of always forcing `history`.
- If the restored tab is **Notes**, its data is fetched on load.
Single-file change; `TabBar` already reacts to `initial-active-tab`, so no other components needed updates.

### DIFF
--- a/app/javascript/dashboard/routes/dashboard/companies/pages/CompanyDetailView.vue
+++ b/app/javascript/dashboard/routes/dashboard/companies/pages/CompanyDetailView.vue
@@ -68,6 +68,11 @@ const SIDEBAR_TABS_OPTIONS = [
   { key: 'CONTACTS', value: 'contacts' },
 ];
 
+const getTabFromQuery = () =>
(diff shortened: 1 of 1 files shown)


salmonumbrella:
cc @sojan-official @sony-mathew — would appreciate a review when you get a chance 🙏

PR #14768, opened 2026-06-17
feat(tiktok): add cloud warning

muhsin-k:
Adds Cloud-only warning on the TikTok connect page while North American account connections may be temporarily unavailable.

Fixes https://linear.app/chatwoot/issue/CW-7322/cloud-show-tiktok-north-america-warning-on-connect-page


db/migrate/20260617000000_enable_tiktok_for_paid_accounts.rb
```diff
@@ -0,0 +1,9 @@
+class EnableTiktokForPaidAccounts < ActiveRecord::Migration[7.0]
```
sojan-official: we don't need to commit this as a migration in code . just run directly from production console .

muhsin-k: Fixed via https://github.com/chatwoot/chatwoot/pull/14768/commits/75ffb849fb7c82fb748d639f7ce68c3aada21044

sony-mathew (approved):
LGTM

PR #14774, opened 2026-06-17
fix: Surface inline-marked email images as regular attachments when their CID is not referenced in the HTML body

### LINKED ISSUE #14633: Inbound email: inline image with Content-ID but not referenced in HTML body is silently dropped (Apple Mail <attachment>)
### Describe the bug

This is reproducible with Apple Mail when attachments are set to display as icons via the `DisableInlineAttachmentViewing` default: in that mode Apple Mail emits its proprietary `<attachment ...>` placeholder tag in the HTML instead of ``, while still marking the image part as `Content-Disposition: inline` with a `Content-Id`. With the default setting (flag off), the image is embedded as `` and renders correctly.

### To Reproduce

1. On macOS, set Apple Mail to display/send attachments as icons:
   `defaults write com.apple.mail DisableInlineAttachmentViewing -bool yes`
   Then fully quit Mail (Cmd+Q) and reopen it.
2. Compose an HTML email and attach an image.
3. Send it to an Email channel inbox.
4. The image is uploaded to storage (confirmed present in the S3 bucket) but does not appear anywhere in the conversation.

Confirmation of the trigger: removing the flag with
   `defaults delete com.apple.mail DisableInlineAttachmentViewing`
(then quit/reopen Mail) makes Apple Mail embed the image as ``, and it then renders correctly in Chatwoot.

Resulting failing MIME structure (image marked inline, with Content-Id, but not referenced by cid: in the HTML):

Content-Type: multipart/alternative; boundary="A"
--A
Content-Type: text/plain
Hello
--A
Content-Type: multipart/related; type="text/html"; boundary="B"
--B
Content-Type: text/html
<html><body>Hello<attachment title="photo.jpeg" type="image/jpeg"></attachment></body></html>
--B
Content-Type: image/jpeg; name="photo.jpeg"
Content-Disposition: inline; filename="photo.jpeg"
Content-Id: <ABC123@example>
<base64...>
--B--
--A--

### Expected behavior

An image marked inline but NOT referenced by its CID in the HTML body should fall back to being shown as a regular attachment, instead of being silently dropped.

Root cause: in app/mailboxes/mailbox_inline_attachment_helper.rb, `inline_attachment?` returns true for any part that is `attachment.inline?`, without checking whether its CID is referenced in the HTML (the `body_references_cid?` check is only applied to the non-inline branch). Such parts go to `upload_inline_image`, which gsub-replaces `cid:<id>` in the HTML; when no matching `cid:` is present (Apple's `<attachment>` tag), nothing is replaced, and the part is also excluded from `process_regular_attachments`. The blob is stored but never surfaced.

Suggested fix: for inline-marked image parts, only treat as inline when the CID is actually referenced in the body; otherwise process as a regular attachment.

### Environment

Docker

### Cloud Provider

None

### Platform

Browser

### Operating system

macOS (Apple Mail 16)

### Browser and version

_No response_

### Docker (if applicable)

Docker Version 29.5.1

### Additional context

Chatwoot version v4.14.1 (CE)

mvanhorn:
## Summary
When an inbound email (e.g. Apple Mail with DisableInlineAttachmentViewing set) marks an image part as Content-Disposition: inline with a Content-Id but does NOT reference that CID via cid: in the HTML body (instead emitting Apple's proprietary <attachment> tag), the image is silently dropped.

Fixes #14633


scmmishra (approved):

scmmishra:
@codex please review

PR #14777, opened 2026-06-17
fix: resolve double-encoded JSON attributes in Message model (#14660)

### LINKED ISSUE #14660: Message#content_attributes is double-encoded on disk (store coder: JSON over a json column) — SQL JSON operators silently return NULL
### Describe the bug

`Message#content_attributes` is declared with `store ... coder: JSON`, but the underlying column is a native PostgreSQL `json` column. This causes **double JSON encoding**: the serialized-attribute layer runs `JSON.dump(hash)` and then the column's `Json` type encodes that string again. The value persisted on disk is a **JSON string scalar**, not a JSON object.

Reads through the model/API work (the same coder parses it back), so the bug is invisible from the app. But **any raw SQL using JSON operators (`->`, `->>`, `#>`, `@>`) on `content_attributes` silently returns `NULL` / matches nothing**, because the operators run against a string scalar instead of an object.

Source — `app/models/message.rb` (unchanged on current `develop`); the column is annotated `content_attributes :json`:

​```ruby
store :content_attributes,
      accessors: [..., :is_unsupported, :data],
      coder: JSON
​```

This looks like a leftover from when `content_attributes` was a `text` column; after the migration to `json`, `coder: JSON` was never removed.

**Impact:** any code path querying `content_attributes` via SQL JSON operators is silently broken. Example in OSS — the automation condition checking whether a message was created by an automation rule (`messages.where("content_attributes ->> 'automation_rule_id' IS NOT NULL")`) always returns empty, because `->> 'automation_rule_id'` is always `NULL`. The same `store coder: JSON` pattern is also applied to `external_source_ids` (a `jsonb` column) and likely has the same issue.

### To Reproduce

On a Rails console:

​ruby
m = Message.last
m.update!(content_attributes: { foo: 'bar' })

# 1) Model read works (coder decodes on the way out):
m.reload.content_attributes
# => {"foo"=>"bar"}

# 2) Raw SQL with a JSON operator does NOT match (expected: true):
Message.where("content_attributes ->> 'foo' = 'bar'").exists?
# => false

# 3) Because the value is stored as a JSON *string*, not an object (expected: "object"):
ActiveRecord::Base.connection.select_value(
  "SELECT json_typeof(content_attributes) FROM messages WHERE id = #{m.id}")
# => "string"
​

### Expected behavior

`json_typeof(content_attributes)` should be `object`, and JSON operators should work — e.g. `content_attributes ->> 'foo'` should return `'bar'`. Instead the object is stored escaped as a JSON string and JSON operators return `NULL`.

### Environment

Linux VM

### Cloud Provider

### Platform

Browser

### Operating system

Self-hosted

### Browser and version

_No response_

### Docker (if applicable)

_No response_

### Additional context

Reproduced at scale on a self-hosted instance. `json_typeof(content_attributes)` across the `messages` table — **every populated row is a `string`, none is an `object`**:

​```
count      json_typeof
8316234    (null)        # never-written / NULL
 928406    string        # written through the model — all double-encoded
      0    object        # expected type for populated rows — none exist
​```

Contrast on a single populated message (note the double-escaped raw value, and `->>` returning nil):

​```
== message id=648, key=submitted_values ==
model read   : {"submitted_values"=>{"csat_survey_response"=>{"feedback_message"=>"", "rating"=>5}}}
raw json text: "\"{\\\"submitted_values\\\":{\\\"csat_survey_response\\\":{\\\"feedback_message\\\":\\\"\\\",\\\"rating\\\":5}}}\""
json_typeof  : string
SQL ->> submitted_values : nil
​```

**Root cause:** `store` layers a `serialize` / `Coders::JSON` (dump on write, load on read) on top of an attribute whose column type (`json`) already (de)serializes JSON. The two layers stack, producing a string-encoded object on disk.

kevinjosh10:
# Pull Request Template

## Description

Resolves the issue where `Message#content_attributes` and `Message#external_source_ids` were being double-encoded as JSON strings on disk, which caused SQL JSON operators to silently fail and return NULL.

- Removed `coder: JSON` from the `store` definitions in `app/models/message.rb`. The database columns are already `json`/`jsonb` types, so ActiveRecord natively manages the hash serialization without a coder.
- Added a data migration `FixDoubleEncodedJsonInMessages` using native Postgres `#>>` operations to properly extract and re-cast the existing double-encoded strings into raw JSON objects in the database.

Fixes #14660

## Type of change

## How Has This Been Tested?

Verified that existing `Message` tests pass locally and manually verified that the Postgres adapter correctly wraps JSON when written. The data migration unwraps the double encoding effectively.
Tested in the Rails console via:
1. `bundle exec rails db:migrate` to unwrap any double-encoded strings in the `messages` table.
2. Created a message with `content_attributes` in the console.
3. Verified via `Message.where("content_attributes ->> 'test' = 'value'").exists?` that raw SQL operators can now correctly parse the JSON nodes since they are stored as JSON scalars instead of string scalars.

## Checklist:

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -105,14 +105,26 @@ class Message < ApplicationRecord
   # [:email] : Used by conversation_continuity incoming email messages
   # [:in_reply_to] : Used to reply to a particular tweet in threads
   # [:deleted] : Used to denote whether the message was deleted by the agent
+  # Pass-through coder to allow `store` to provide HashWithIndifferentAccess and dirty tracking
+  # without performing JSON stringification, since the underlying column is already `json`/`jsonb`.
+  class PassThroughCoder
+    def self.dump(obj)
+      obj
+    end
+
+    def self.load(obj)
+      obj || {}
+    end
+  end
+
   # [:external_created_at] : Can specify if the message was created at a different timestamp externally
   # [:external_error : Can specify if the message creation failed due to an error at external API
   # [:data] : Used for structured content types such as voice_call
-  store_accessor :content_attributes, :submitted_email, :items, :submitted_values, :email, :in_reply_to, :deleted,
+  store :content_attributes, accessors: [:submitted_email, :items, :submitted_values, :email, :in_reply_to, :deleted,
                                          :external_created_at, :story_sender, :story_id, :external_error,
-                                         :translations, :in_reply_to_external_id, :is_unsupported, :data
+                                         :translations, :in_reply_to_external_id, :is_unsupported, :data], coder: PassThroughCoder
 
-  store_accessor :external_source_ids, :slack, prefix: :external_source_id
+  store :external_source_ids, accessors: [:slack], prefix: :external_source_id, coder: PassThroughCoder
 
   scope :created_since, ->(datetime) { where('created_at > ?', datetime) }
   scope :chat, -> { where.not(message_type: :activity).where(private: false) }
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -110,9 +110,9 @@ class Message < ApplicationRecord
   # [:data] : Used for structured content types such as voice_call
   store :content_attributes, accessors: [:submitted_email, :items, :submitted_values, :email, :in_reply_to, :deleted,
                                          :external_created_at, :story_sender, :story_id, :external_error,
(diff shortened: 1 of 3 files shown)


kevinjosh10:
all github checks are done and the bugs been fixed, can you please verify it once again and merge it to the main

PR #14791, opened 2026-06-19
feat(instagram): support post/reel comment management

smarttecuberaba:
## Describe your changes

The Instagram channel (Instagram API with Instagram Login) currently handles
**direct messages only** (`SUPPORTED_EVENTS = [:message, :read]`). This PR adds
ingestion of **post/reel comments** and **public replies** via the Comments API.

When someone comments on a connected professional account's post or reel, the
comment becomes a conversation in the agent inbox; an agent (or automation) can
reply, and the reply is posted **publicly** on the comment thread.

### What changed
- Subscribe to the `comments` webhook field on the IG account
- Route comment change events by `field` (avoids colliding with the IG
  test-event path and with direct messages)
- `Instagram::CommentText` + `Messages::Instagram::CommentBuilder` create the
  comment conversation/message, grouped per media
- `Instagram::SendOnInstagramService` replies via `POST /{comment-id}/replies`
- Robust webhook signature verification for comment entries
- Add `instagram_business_manage_comments` to `REQUIRED_SCOPES`
- Read `content_attributes` / `additional_attributes` in Ruby (these `json`
  columns make the Postgres `->>` operator unreliable)
- `rake instagram:resubscribe_comments` to migrate already-connected channels

### Tested
Validated end-to-end on a self-hosted instance: an Instagram comment on a
connected professional account's post created a conversation in Chatwoot, and an
agent reply was posted back as a **public comment reply** on Instagram.

Specs added: `comment_text_spec`, `send_on_instagram_service_comment_spec`.

### Note
Production use requires `instagram_business_manage_comments` approved in App Review.

## Type of change

### DIFF
--- a/app/builders/messages/instagram/comment_builder.rb
+++ b/app/builders/messages/instagram/comment_builder.rb
@@ -1,0 +1,94 @@
+class Messages::Instagram::CommentBuilder
+  def initialize(comment, inbox)
+    @comment = comment.with_indifferent_access
+    @inbox = inbox
+  end
+
+  def perform
+    return if @inbox.channel.reauthorization_required?
+    return if message_already_exists?
+
+    ActiveRecord::Base.transaction do
+      conversation.messages.create!(message_params)
+    end
+  rescue StandardError => e
+    ChatwootExceptionTracker.new(e, account: @inbox.account).capture_exception
+  end
+
+  private
+
+  def contact
+    @contact ||= contact_inbox&.contact
+  end
+
+  def contact_inbox
+    @contact_inbox ||= @inbox.contact_inboxes.find_by(source_id: commenter_id)
+  end
+
+  def conversation
+    @conversation ||= find_open_conversation || build_conversation
+  end
+
+  # Group all comments from the same media + contact into a single conversation,
+  # keeping comment threads separate from direct-message conversations.
+  # additional_attributes is a `json` column, so we match in Ruby instead of
+  # relying on the Postgres `->>` operator (unreliable here).
+  def find_open_conversation
+    @inbox.conversations
+          .where(contact_id: contact.id)
+          .where.not(status: :resolved)
+          .order(created_at: :desc)
+          .find { |conv| conv.additional_attributes['instagram_media_id'] == media_id }
+  end
+
+  def build_conversation
+    Conversation.create!(
+      account_id: @inbox.account_id,
+      inbox_id: @inbox.id,
+      contact_id: contact.id,
+      contact_inbox_id: contact_inbox.id,
+      additional_attributes: {
+        type: 'instagram_comment',
+        instagram_media_id: media_id
+      }
+    )
+  end
+
+  def message_params
+    {
+      account_id: @inbox.account_id,
+      inbox_id: @inbox.id,
+      message_type: :incoming,
+      status: :sent,
+      source_id: comment_id,
+      content: comment_text,
+      sender: contact,
+      content_attributes: {
+        type: 'instagram_comment',
+        instagram_comment_id: comment_id,
+        instagram_media_id: media_id,
+        instagram_parent_comment_id: @comment[:parent_id]
+      }
+    }
+  end
+
+  def message_already_exists?
+    Message.find_by(source_id: comment_id).present?
+  end
+
+  def commenter_id
+    @comment.dig(:from, :id).to_s
+  end
+
+  def comment_id
+    @comment[:id]
+  end
+
+  def media_id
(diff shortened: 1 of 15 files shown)


salmonumbrella:
This is useful thanks @smarttecuberaba cc: @muhsin-k

smarttecuberaba:
Thanks @salmonumbrella, glad it's useful!

@muhsin-k happy to walk through the approach or adjust anything to match
Chatwoot's conventions. The feature was validated end-to-end on a self-hosted
instance — an Instagram comment was ingested into Chatwoot and an agent reply
was posted back as a public comment reply. Happy to address any feedback.

nestordavalos:
Supporting Instagram post/reel comments inside Chatwoot is a very useful direction.

I think the overall approach makes sense: Instagram comments should be handled differently from Instagram DMs. A comment reply should not go through the normal DM flow; it should be published as a public reply to the specific Instagram comment using the Comments API.

One UX concern I have is around using the regular reply box for this flow. If an Instagram comment is represented inside a conversation, the agent may assume that typing in the standard reply box sends a normal private message. For comments, that can be confusing because the expected behavior is public and tied to a specific comment, not to the whole conversation.

A safer design could be:

1. Treat the Instagram comment message as the reply target.
2. Add a context-menu action on that specific message, for example: **“Reply to comment”**.
3. When selected, open a dedicated composer/panel labeled clearly as **“Public reply to comment”**.
4. Show a warning/helper text such as: **“This will be published below the comment on Instagram.”**
5. Keep **“Send DM”** as a separate secondary action, if supported, so agents clearly understand the difference between public comment replies and private messages.

This would also help technically, because the reply action would carry the exact `instagram_comment_id` being replied to, instead of relying on the conversation context or the latest incoming comment. That becomes important when multiple comments from the same user exist in the same post/reel conversation.

In short, I think comments should not be answered from the generic conversation reply box by default. The reply should be explicitly triggered from the Instagram comment message itself, so the target comment is clear and the agent understands that the response will be public.

image

smarttecuberaba:
Thanks @nestordavalos, great feedback — I agree the direction is right and that a
comment reply shouldn't be treated like a DM.

On the UX: an explicit "Reply to comment" action on the specific comment message
(clearly labeled as a public reply) is a solid improvement, and it removes any
ambiguity about which comment is being answered.

For context on the current implementation: the backend already targets a specific
comment (it resolves the incoming comment's instagram_comment_id and replies via
the Comments API), picking the latest comment that existed when the reply was
composed to avoid mismatches. So the common case is handled — but I agree an
explicit per-comment action is clearer and more robust for threads with multiple
comments from the same user.

Happy to add the dedicated composer/action either in this PR or as a focused
follow-up — whichever the maintainers prefer.

@muhsin-k any preference on scope here: include the UI in this PR, or land the
backend first and iterate on the UX?

salmonumbrella:
@nestordavalos +1 on the UX. @muhsin-k this is amazing, can could possibly work for TikTok, Facebook.

smarttecuberaba:
Friendly bump on this one 🙂

@muhsin-k the PR is green and has been stable for a few weeks. Happy to adjust
scope if that helps — for example, adding the explicit "Reply to comment" action
suggested above, or splitting the UI work into a follow-up PR.

Just let me know what direction you'd prefer and I'll adapt it.

smarttecuberaba:
@nestordavalos thanks for the feedback, and I agree with the concern. Replying from the standard composer is ambiguous — the agent has no clear signal that their message will be posted publicly on the post/reel rather than sent as a DM.

I'd suggest splitting it:

- **This PR** stays as the backend groundwork: webhook subscription, comment ingestion, conversation grouping per media, and the public reply path through the Comments API. It's self-contained and covered by specs.
- **A follow-up PR** adds the explicit "Reply to comment" action on the message, plus the visual affordance marking the thread as public.

That keeps this one reviewable and gives the UX room to be designed properly instead of bolted on. If maintainers would rather see it all in a single PR, I'm happy to fold it in — just let me know which is easier to review.

---

Status note for maintainers: this has been open since Jun 19. I re-checked today and it's still `mergeable: clean` against current `develop`, with no conflicts despite roughly 330 commits of drift since. CI was green on both GitHub Actions and CircleCI.

One thing worth flagging: the Codex cards above are **outdated** — they refer to earlier commits and were all addressed in later pushes, with the exception of one P2 about honoring `in_reply_to` when picking the target comment, which I judged an edge case outside the normal flow and left deliberately.

Would appreciate a review pass whenever there's capacity. Happy to rebase, split, or adjust scope if that makes it easier to land.

PR #14794, opened 2026-06-19
fix: force account picker on microsoft email inbox OAuth

vishnu-narayanan:
## Description

The Microsoft authorize URL had no `prompt` parameter. With Microsoft's default behavior, a browser carrying an active Microsoft session is silently signed in to that account. If the same account is already attached to an inbox in the same Chatwoot account, the callback treats the OAuth response as a re-auth and routes the user to the existing inbox settings page instead of letting them create the new inbox they intended.

Adding `prompt=select_account` interrupts SSO and always presents the Microsoft account picker, so the user explicitly chooses the mailbox they want to connect.

This is distinct from `prompt=consent`, which was removed in [#13962](https://github.com/chatwoot/chatwoot/pull/13962) to fix the admin consent loop. `select_account` only interrupts SSO and does not retrigger the consent dialog.

Closes INF-76

## Type of change

## How Has This Been Tested?

`bundle exec rspec spec/controllers/api/v1/accounts/microsoft/authorization_controller_spec.rb` (3 examples, 0 failures). The regression test introduced in [#13962](https://github.com/chatwoot/chatwoot/pull/13962) is updated to assert `prompt=select_account` instead of asserting absence of the parameter.

## Checklist:


app/controllers/api/v1/accounts/microsoft/authorizations_controller.rb
```diff
@@ -6,7 +6,10 @@ def create
       {
         redirect_uri: "#{base_url}/microsoft/callback",
         scope: scope,
-        state: state
+        state: state,
+        # Force the Microsoft account picker so an already-signed-in account does not
+        # silently authorize and re-bind to an existing inbox in the new-inbox flow.
+        prompt: 'select_account'
```
scmmishra, line 12: Will this break the consent flow? Which originally caused the admin consent loop?

scmmishra: Irrelvant comment, addressed in the PR description. Closing

scmmishra (approved):

PR #14579, opened 2026-05-27
feat: voice-call UX fixes

tds-1:
## Linear ticket
https://linear.app/chatwoot/issue/CW-7187/voice-calls-followup-tasks

## Description

Improvements to the WhatsApp voice-calling experience plus a cheaper, more accurate audio-transcription model.
- First-time callers now get a real name. An inbound WhatsApp call creates the contact from the caller's WhatsApp profile name instead of the bare phone number.
- Clear, consistent call attribution. Call bubbles show a unified "Handled by {agent}" 
- Cleaner call widget. The dismiss (✕) button is shown only for incoming calls
- WhatsApp calling for manual inboxes. voice_calling_supported? now covers any whatsapp_cloud inbox
- Transcription: whisper-1 → gpt-4o-mini-transcribe.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #14815, opened 2026-06-22
fix: normalize WhatsApp .oga audio extension to .ogg for transcription

### LINKED ISSUE #14689: Audio transcription broken for WhatsApp voice notes since v4.14 — gpt-4o-mini-transcribe rejects .oga extension (HTTP 400)
### Summary

Since the transcription model was switched from `whisper-1` to `gpt-4o-mini-transcribe` in #14579, **WhatsApp voice notes no longer transcribe**. WhatsApp audio attachments are stored with the `.oga` extension, which `whisper-1` accepted but the `gpt-4o-*-transcribe` models reject with HTTP 400.

Because `Messages::AudioTranscriptionJob` uses `discard_on Faraday::BadRequestError` (#13487), the job is silently dropped — `transcribed_text` stays empty with no user-visible error.

### Affected version

v4.14.x (regression from #14579, `enterprise/app/services/messages/audio_transcription_service.rb`).

### Root cause

`fetch_audio_file` names the temp upload after the blob's original filename, preserving `.oga`. `extension_from_content_type` only fills in an extension when the blob has **none**, so `.oga` files pass through unchanged. The OpenAI `gpt-4o-mini-transcribe` / `gpt-4o-transcribe` endpoints reject `oga` as a format.

### Reproduction

Take any WhatsApp voice note (`audio/opus`, `.oga`) and call the transcription endpoint:

```
# same bytes, only the extension differs:
file=clip.oga  model=gpt-4o-mini-transcribe  -> HTTP 400 {"code":"unsupported_value","message":"Unsupported file format oga"}
file=clip.ogg  model=gpt-4o-mini-transcribe  -> HTTP 200 {"text":"...transcript..."}
```

`whisper-1` accepts both, which is why this worked before #14579.

### Suggested fix

Normalize the upload extension for Opus/WhatsApp audio in `fetch_audio_file` — e.g. always remap `audio/opus` / `.oga` → `.ogg` (or apply `extension_from_content_type` unconditionally rather than only when the extension is blank).

### Note

A second, deployment-side gotcha: the new model must be enabled in the OpenAI **project**'s model permissions, or every request returns `403 "Project ... does not have access to model gpt-4o-mini-transcribe"` (silent, since whisper-1 was previously allowed org-wide).

gobeyondpty:
## Description

WhatsApp voice notes are stored as ActiveStorage blobs with the `.oga` extension. Since the transcription model was switched from `whisper-1` to `gpt-4o-mini-transcribe` in #14579, these uploads fail: the `gpt-4o-*-transcribe` endpoints reject `.oga` with `HTTP 400 {"code":"unsupported_value","message":"Unsupported file format oga"}`, whereas `whisper-1` accepted it.

Because `Messages::AudioTranscriptionJob` uses `discard_on Faraday::BadRequestError` (#13487), the job is silently dropped and `transcribed_text` stays empty with no user-visible error.

The same bytes named `.ogg` transcribe successfully — only the extension the API sees matters.

This normalizes the upload extension in `fetch_audio_file`: `.oga` is remapped to `.ogg` via a small `EXTENSION_REMAP` table before the temp file is created. The audio bytes are unchanged. This keeps the existing "fill extension from content type when blank" behaviour intact and only remaps known-unsupported extensions.

Fixes #14689

## Type of change

## How Has This Been Tested?

Added specs to `spec/enterprise/services/messages/audio_transcription_service_spec.rb`:

- `.oga` filename is remapped to `.ogg`
- a supported extension (`.mp3`) is preserved unchanged
- the existing "adds extension from content type when filename has no extension" case still passes

Verified in production (Chatwoot v4.15.1) that WhatsApp `.oga` voice notes which previously 400'd now transcribe successfully once the extension is normalized to `.ogg`.

## Checklist:

### DIFF
--- a/enterprise/app/services/llm/speech_to_text_service.rb
+++ b/enterprise/app/services/llm/speech_to_text_service.rb
@@ -10,6 +10,12 @@ class Llm::SpeechToTextService < Llm::LegacyBaseOpenAiService
   # skips transcription.
   BYTE_LIMIT = 25_000_000
 
+  # Extensions the transcription endpoint rejects, mapped to an accepted name for
+  # the same container. WhatsApp voice notes arrive as `.oga` and come back as
+  # "Unsupported file format oga" even though the bytes are valid Ogg/Opus audio;
+  # `.ogg` names the same data in a way the endpoint accepts.
+  EXTENSION_REMAP = { 'oga' => 'ogg' }.freeze
+
   attr_reader :blob, :account, :transcription_model
 
   # Transcription runs on Captain's OpenAI credentials and consumes its response credits.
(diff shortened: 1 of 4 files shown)


gobeyondpty:
Rebased this PR on current `develop` and moved the fix to where the code now lives.

Since the original patch, the temporary-file handling moved out of `Messages::AudioTranscriptionService` into `Llm::SpeechToTextService`, which is shared by voice-note attachments and voice-call recordings. The `.oga` rejection is still reproducible there, so the branch now carries a single commit against that service, plus coverage for a `.oga` blob and for an extension the endpoint already accepts.

https://claude.ai/code/session_01BCAg5BTURQSx4HpTv88xWS

PR #14818, opened 2026-06-23
fix: recover from stale LeadSquared lead ids on activity sync

scmmishra:
LeadSquared sync now recovers automatically when a contact's cached lead has been deleted or merged on the LeadSquared side. Previously the stale lead id was never cleared, so every new conversation or contact update for that contact failed with "Lead not found" (`MXInvalidEntityReferenceException`) indefinitely.

## What changed
- Activity sync: on a "Lead not found" error while posting a conversation/transcript activity, clear the cached `leadsquared_id`, re-resolve the contact to a fresh lead, and retry the activity once (guarded against loops and duplicate leads).
- Contact sync: on the same error while updating an existing lead, clear the cached id and create a fresh lead instead.
- Fix `get_lead_id` to actually return early for unidentifiable contacts (the guard previously fell through).

## How to reproduce
1. For a LeadSquared-enabled account, point a contact's cached lead id at a lead that no longer exists in LeadSquared.
2. Update the contact, or create/resolve a conversation for it.
3. Before: the sync fails repeatedly with "Lead not found" and never self-corrects. After: the stale id is cleared, a fresh lead is resolved/created, and subsequent syncs reuse the healed id.


tds-1:
@codex please review

tds-1:
@scmmishra 
Same staleness bug exists on the contact path. 
`create_or_update_lead` calls `update_lead(lead_data, lead_id)` with the cached leadsquared_id; if that lead was deleted/merged it fails with the same MXInvalidEntityReferenceException but only gets logged, never cleared — so contact.updated stays broken indefinitely. Can we reuse the clear-and-retry healing here too?

scmmishra:
@codex please review

scmmishra:
@tds-1 this is good to go

tds-1 (approved):

PR #14852, opened 2026-06-25
chore: Track cloud signup conversions

sojan-official:
## Summary
- enqueue `cloud_signup` conversion tracking after website attribution is stored on a new cloud account
- keep authenticated add-workspace requests out of attribution/conversion tracking
- reuse the existing marketing conversion tracking service and config already available on `develop`

## Notes
- This PR only wires the signup path.
- Billing/plan activation conversion tracking is intentionally left out for a separate rollout.
- The conversion service still no-ops outside Chatwoot Cloud and when stored attribution has no supported click identifier.

### DIFF
--- a/enterprise/app/services/internal/accounts/marketing_attribution_service.rb
+++ b/enterprise/app/services/internal/accounts/marketing_attribution_service.rb
@@ -50,6 +50,7 @@ class Internal::Accounts::MarketingAttributionService
         'stored_at' => Time.current.iso8601
       }.compact
     )
+    enqueue_signup_conversion
   end
 
   private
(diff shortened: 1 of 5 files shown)


enterprise/app/services/internal/accounts/marketing_attribution_service.rb
```diff
@@ -79,4 +80,8 @@ def validate_payload(payload)
   def internal_attributes_service
     @internal_attributes_service ||= Internal::Accounts::InternalAttributesService.new(account)
   end
+
+  def enqueue_signup_conversion
+    Internal::Accounts::MarketingConversionTrackingJob.perform_later(account.id, 'cloud_signup', account.created_at.iso8601)
```
pranavrajs: why do we need iso8061

pranavrajs (approved):

PR #14855, opened 2026-06-25
fix(widget): prevent duplicate email-collect messages

hahuma:
## Pull Request Description

The email-collect web-widget template can run twice for the same conversation and create a duplicate pair of messages (the "Give the team a way to reach you." line + the email-input box appear twice in the widget).

`MessageTemplates::HookExecutionService#should_send_email_collect?` guards only with `email_collect_was_sent?` (does an `input_email` message already exist?), checked outside any lock. When two hook executions race for the same conversation — e.g. two incoming widget messages in quick succession — both pass the guard before either commits, so each creates the pair. Because the two messages are created in one transaction, the within-run after_create_commit re-trigger is already safe; the gap is the lack of serialization across separate executions.

This:
- Serializes creation in `MessageTemplates::Template::EmailCollect#perform` with `conversation.with_lock` and re-checks inside the lock, so concurrent executions can't both create the pair. (`reload` first because the conversation passed in from the message hook can carry an unpersisted `display_id` change, which Rails refuses to row-lock.)
- Adds `return false unless message.incoming?` to `should_send_email_collect?`, so the hook only reacts to the contact's incoming message and never to the template messages it creates — consistent with `should_send_out_of_office_message?`.

## Type of change

## How Has This Been Tested?
- `spec/services/message_templates/template/email_collect_spec.rb` — added a test that invoking the template twice for the same conversation creates only one pair.
- Existing `hook_execution_service` specs (OSS + enterprise) pass.

## Checklist


iamsivin (approved):
Thanks for the PR. LGTM 🚀
@sony-mathew Could you please take a quick pass?

sony-mathew (approved):
lgtm

PR #14857, opened 2026-06-25
fix(help-center): drop leading @ from pt_BR social link placeholder (vue-i18n crash)

hahuma:
## Pull Request Description

The Help Center portal settings "Social links" section (documentation layout) crashes for **pt_BR** users the moment a network is added. The `pt_BR` value of `HELP_CENTER.PORTAL_SETTINGS.LAYOUT_CONTENT.SOCIAL_LINKS.PLACEHOLDER` is `"@usuario"`. vue-i18n treats a leading `@` as the linked-message marker, so compiling that message throws `SyntaxError: Invalid linked format` when the input field renders (i.e. right after a platform is added), which unmounts the whole settings section.

English is unaffected (`PLACEHOLDER: "handle"`); pt_BR is the only locale whose placeholder starts with `@`. This drops the leading `@` (`"usuario"`) — the network prefix (e.g. `facebook.com/`) is already shown next to the field, so the `@` wasn't needed.

## Type of change

## How to reproduce
With the dashboard in Portuguese: Help Center → Settings → "Documentation" layout → "Add social link" → click a network → the section disappears (console: `Invalid linked format`).

## Checklist


devi-r:
@hahuma Thanks for catching this. 
This is the same `Invalid linked format crash` we fixed in [#10206](https://github.com/chatwoot/chatwoot/pull/10206) and [#10334](https://github.com/chatwoot/chatwoot/pull/10334), and the convention we settled on there is literal interpolation rather than removing the character, so I've pushed `{'@'}usuario` instead of `usuario`. That keeps the handle prefix you'd expect in the placeholder, and it's also the form that survives a translation sync.

**Small note for future translation changes**: we manage localization through [Crowdin](https://translate.chatwoot.com) , so edits made directly to the non-English locale files get overwritten on the next automated sync. You may remember that happening to part of [#11172](https://github.com/chatwoot/chatwoot/pull/11172). A crash like this one is worth fixing in the repo so it ships immediately, but the string also needs correcting in Crowdin or the sync will undo it, and I'm doing that side now. If you'd like to contribute more pt-BR updates, please follow the [translation guide](https://developers.chatwoot.com/contributing-guide/translation-guidelines) and submit them through Crowdin going forward rather than opening direct locale-file PRs.

devi-r (approved):

PR #14869, opened 2026-06-26
feat: expose internal Signals attribution API

sojan-official:
Adds a Chatwoot Cloud-only internal platform API endpoint for Signals to sync account attribution data without direct database access.

Closes [GRO-114](https://linear.app/chatwoot/issue/GRO-114)

## Why

Signals needs a narrow, read-only way to consume attribution and conversion metadata across Chatwoot Cloud accounts. The existing platform app API is scoped through permissible records, which does not fit cross-account internal reporting.

## What changed

* Adds `GET /platform/api/v1/internal/accounts` behind `CHATWOOT_CLOUD_SIGNALS_API_TOKEN`.
* Registers `CHATWOOT_CLOUD_SIGNALS_API_TOKEN` in the existing cloud internal installation config.
* Keeps the route and controller behavior scoped to Chatwoot Cloud.
* Supports incremental account sync using Chatwoot-style epoch `since` and `until` parameters against `accounts.created_at`.
* Returns account metadata, plan metadata from custom attributes, limits, and stored marketing attribution.

## Validation

* `bundle exec rails routes | rg "platform/api/v1/internal/accounts|internal_accounts"`
* `bundle exec rubocop config/routes.rb enterprise/app/controllers/platform/api/v1/internal/accounts_controller.rb`
* Ruby syntax checks for the new controller and Jbuilder view


sojan-official:
Closing this PR because the Signals account attribution export is now handled through a protected Metabase API instead of a Chatwoot code endpoint. Keeping LEA-37 open for tracking/follow-up.

Issue #6446, opened 2023-02-13
Getting 500 server error despite successful send to Chatwoot


sumnerevans:
**Describe the bug**

When under heavy load, Chatwoot returns 500 errors despite successfully saving the message.

**To Reproduce**

We are using the Matrix <-> Chatwoot bot: https://github.com/beeper/chatwoot, and when we send many messages at once (checking in with customers), the bot sends many requests to Chatwoot.

These messages get persisted, but I get a 500 error instead of a successful response. Because of this, the Matrix <-> Chatwoot bot thinks that it needs to resend, and we get duplicate messages (see screenshots).

**Expected behavior**

I expect that if the message succeeds, then a successful response is given.

If rate limiting happens, I expect to get a failure, and that the message is not persisted.

**Screenshots**

<img width="476" alt="CleanShot 2023-02-09 at 20 01 05@2x" src="https://user-images.githubusercontent.com/16734772/218583305-c363a4a9-57d5-4c8a-83ab-4f95aa0b12a6.png">

**Browser logs**

N/A

**Server logs**

Logs for a request that failed despite successfully sending the message: [Explore-logs-2023-02-13 21 50 01.txt](https://github.com/chatwoot/chatwoot/files/10726595/Explore-logs-2023-02-13.21.50.01.txt)

**Environment**

Describe whether you are using Chatwoot Cloud (app.chatwoot.com) or a self-hosted installation of Chatwoot. If you are using a self-hosted installation of Chatwoot, describe the type of deployment (Docker/Linux VM installation/Heroku/Kubernetes/Other).

- [ ] app.chatwoot.com (Chatwoot Cloud)
- [x] Self-hosted
- - [ ] Linux VM
- - [ ] Docker
- - [x] Kubernetes
- - [ ] Heroku
- - [ ] Other (Please specify)

**Cloud Provider** (If applicable)
- [x] AWS
- [ ] GCP
- [ ] Azure
- [ ] DigitalOcean
- [ ] Others

pranavrajs:
``` 
Rack::Timeout::RequestTimeoutException (Request ran for longer than 15000ms 
```

This is what I see on the logs. @sumnerevans Do you have any additional insights on why it is taking so much time? Is it waiting for the request to be accepted (adding more servers would help here), or is there something else? Any insights on this would be helpful in debugging this. (e.g., Any APM traces would be helpful).

ceolinrenato:
[...]

PR #14904, opened 2026-07-01
fix: keyboard shortcuts fire twice after switching conversations (keydown listener leak)

oAllanNascimento:
## Description

`keyboardEventListenerMixins` adds a `document` `keydown` handler on `mounted` and is meant to remove it on `unmounted`. The removal path reads `this.$el.dataset.keydownHandlerIndex`, but that index is written to the element returned by `getElementToBind()`.

Under Vue 3 this is broken for **multi-root (fragment) components**, where `this.$el` is a comment anchor node with no `dataset`. `ReplyBox` is exactly such a component — it renders `<ReplyBoxBanner>` and the `.reply-box` `<div>` as sibling roots — and it overrides `getElementToBind()` to return the `.reply-box` div. So the handler is **added** on mount but **never removed** on unmount.

Because `ReplyBox` is mounted under `v-if="currentChat.id"`, **every conversation switch leaks another `keydown` listener** on `document`. With N leaked listeners, a single <kbd>Enter</kbd> fires `onSendReply` N times → the same message is sent 2×, 3×, … This is especially visible on API channels (e.g. WhatsApp), where each duplicate becomes a real outgoing message with its own `source_id`.

`FullEditor` also overrides `getElementToBind()` and is affected by the same latent leak.

### Root cause timeline
- The `$el.dataset` cleanup scheme predates Vue 3 (single-root components, where `$el` **was** the root element).
- The Vue 3 migration introduced fragments, changing `$el` semantics for multi-root components.
- `ReplyBox` became multi-root when the bot-handoff `<ReplyBoxBanner>` was added as a sibling root, exposing the latent leak.

## Fix

Keep the exact handler reference on the instance (`this.keydownHandlerRef`) and remove **that** handler on `unmounted`, instead of relying on `$el.dataset`. This drops the fragile module-level `taggedHandlers` index scheme and works for both single-root and multi-root consumers of the mixin (`ReplyBox`, `FullEditor`).

## How to reproduce / verify

1. Open the dashboard with DevTools console.
2. `getEventListeners(document).keydown.length` → note the count.
3. Open/close or switch between a few conversations.
4. Re-run `getEventListeners(document).keydown.length` → **before this PR the count grows by one per `ReplyBox` mount; after, it stays stable.**
5. Functionally: after switching conversations several times, pressing <kbd>Enter</kbd> once in the reply box sends the message only once (previously it could send 2×+).

## Type of change

## Notes

Related: #6446 (message idempotency keys) would additionally guard the symptom server-side; this PR fixes the client-side root cause.


iamsivin (approved):
@oAllanNascimento Thanks for the PR🚀

PR #14906, opened 2026-07-01
fix: assign outbound voice call conversation to the calling agent

scmmishra:
Outbound voice calls were being auto-assigned to the wrong agent. When an agent placed an outbound call, the conversation was created without an assignee, so inboxes with auto-assignment enabled would round-robin it to a different agent instead of keeping it with the person who actually made the call. This made it hard to tell which agent was on an active call.

## What changed

- Set the calling agent as the conversation's assignee when creating an outbound voice call conversation.
- This prevents the generic auto-assignment handler from treating the conversation as unassigned and reassigning it.
- Added specs covering the assignment, including a regression case with inbox auto-assignment enabled.

**Note:** this applies to newly placed calls; it does not retroactively fix conversations that were already mis-assigned.


tds-1:
@codex please review

tds-1:
@codex please review

a99430331-wq:
I was scammed and I need help in creating a full report fom Jan 2025 ugant

a99430331-wq:
I was scammed and I need help in creating a full report fom Jan 2025 urgant pls

tds-1:
@codex please review

tds-1 (approved):

PR #14933, opened 2026-07-06
fix: Sanitize query canceled API errors

sony-mathew:
# Pull Request Template

## Description

Message creation API failures caused by PostgreSQL query cancellations now return a generic retryable error instead of exposing raw database internals such as `PG::QueryCanceled`, tuple identifiers, or relation names to customers. Existing validation failures continue to return their specific validation messages.

Closes [CW-7538](https://linear.app/chatwoot/issue/CW-7538/do-not-expose-pgquerycanceled-details-in-messages-api-errors)

## Type of change

## How Has This Been Tested?

Reproduced the messages API path by raising `ActiveRecord::QueryCanceled` from `Messages::MessageBuilder` and verified the response contains the customer-safe localized error without `PG::QueryCanceled` details.

Validation run locally:

- `bundle exec rspec spec/controllers/api/v1/accounts/conversations/messages_controller_spec.rb`
- `bundle exec rubocop app/controllers/concerns/request_exception_handler.rb spec/controllers/api/v1/accounts/conversations/messages_controller_spec.rb`
- `git diff --check`

## Checklist:

### DIFF
--- a/app/controllers/concerns/request_exception_handler.rb
+++ b/app/controllers/concerns/request_exception_handler.rb
@@ -1,6 +1,12 @@
 module RequestExceptionHandler
   extend ActiveSupport::Concern
 
+  QUERY_CANCELED_ERROR_MESSAGE_PATTERNS = [
+    'ActiveRecord::QueryCanceled',
+    'PG::QueryCanceled',
+    'canceling statement due to statement timeout'
+  ].freeze
+
   included do
     rescue_from ActiveRecord::RecordInvalid, with: :render_record_invalid
   end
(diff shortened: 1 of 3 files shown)


tds-1 (approved):

app/controllers/concerns/request_exception_handler.rb
```diff
@@ -1,6 +1,12 @@
 module RequestExceptionHandler
   extend ActiveSupport::Concern
 
+  QUERY_CANCELED_ERROR_MESSAGE_PATTERNS = [
```
pranavrajs, line 4: @sony-mathew Should we check we are not sending unfiltered errors directly to the UI?

This check would definitely help, but I think the core issue is that there may be a catch-all error handler somewhere that is passing the raw error message through to the UI.

sony-mathew: Sometimes we do send that unfiltered errors like in the case of an attribute error or parameter missing or some things like that. This is the right concern to handle all errors and error mesages. Currently we have a list of errors which we are handling in different ways. As we come across more, we can add more here on how that particular error need to be handled.

Screenshot 2026-07-06 at 10 45 01 PM

Let me know your thoughts on this.

sony-mathew: This will handle this error across all controllers where we use `render_could_not_create_error` or `handle_with_exception` being called scenarios.

PR #14938, opened 2026-07-06
feat: support whatsapp input select row descriptions

### LINKED ISSUE #14911: Support descriptions for WhatsApp input_select list rows

WhatsApp interactive list rows support an optional description/subtitle for each row. Chatwoot currently exposes `input_select` items with `title` and `value`, but rejects an item-level `description` key during message validation.

This prevents integrations from rendering compact WhatsApp list titles with longer explanatory subtitles. For example, an education/tenant bot may need list rows like:

- title: `Option 1`
- description: `Nursing - Distance Learning | 2026/1`

Without description support, the integration must either truncate meaningful labels to WhatsApp row title limits or move all details into the message body, which is a worse UX for longer menus.

Current behavior when sending `content_type: input_select` with item descriptions:

```json
{
  "content_type": "input_select",
  "content_attributes": {
    "items": [
      {
        "title": "Option 1",
        "value": "option_1",
        "description": "Nursing - Distance Learning | 2026/1"
      }
    ]
  }
}
```

Chatwoot rejects it because `description` is not in the allowed select item keys.

### Describe the solution you would like

Allow `description` as an optional key for `input_select` items and forward it when building WhatsApp list rows.

A rough implementation would be:

1. Add `description` to `ContentAttributeValidator::ALLOWED_SELECT_ITEM_KEYS`.
2. Update `Whatsapp::Providers::BaseService#create_rows` to include `description` when present:

```ruby
row = { "id" => item["value"], "title" => item["title"] }
description = item["description"] || item[:description]
row["description"] = description if description.present?
```

3. Add/update WhatsApp provider specs to assert list rows can include `description`.

This should be backward compatible because the field is optional, and existing `input_select` payloads continue to work unchanged.

### Describe alternatives you have considered

The main workaround is to put all row details in the message body and keep list row titles short, but this makes the interactive list less useful and less readable.

Another alternative is for external middleware to bypass Chatwoot and call WhatsApp directly, but that loses the benefit of using Chatwoot as the source of truth for channel delivery.

### Additional context

This applies specifically to WhatsApp list messages. WhatsApp reply buttons do not support row descriptions, so this enhancement is most useful when Chatwoot renders `input_select` as a list message.

Whatsapp screenshot of the output we want Chatwoot to do, which it currently can't:

Image

izcoser:
## Description

Fixes #14911

Adds optional `description` support for `input_select` items and forwards it when WhatsApp providers build interactive list rows. This lets integrations use compact row titles with longer explanatory subtitles in WhatsApp list messages while keeping existing `input_select` payloads backward compatible.

I kept this as an optional passthrough and did not add WhatsApp-specific length validation in the generic content attribute validator, matching the existing behavior for row titles.

## Type of change

## How Has This Been Tested?

Ran the focused lint and specs in the Chatwoot devcontainer image against current `origin/develop` after rebasing:

```bash
bundle exec rubocop app/models/concerns/content_attribute_validator.rb app/services/whatsapp/providers/base_service.rb spec/controllers/api/v1/accounts/conversations/messages_controller_spec.rb spec/services/whatsapp/providers/whatsapp_cloud_service_spec.rb spec/services/whatsapp/providers/whatsapp360_dialog_service_spec.rb
bundle exec rspec spec/services/whatsapp/providers/whatsapp_cloud_service_spec.rb spec/services/whatsapp/providers/whatsapp360_dialog_service_spec.rb spec/controllers/api/v1/accounts/conversations/messages_controller_spec.rb
```

Result: RuboCop reported no offenses; RSpec finished with 49 examples, 0 failures. Also ran `bundle exec rake swagger:build` after updating the OpenAPI source.

## Checklist:


izcoser:
Addressed the review-bot catch: described input_select items now use the WhatsApp list payload even when there are 1-3 options, since reply buttons cannot carry row descriptions. Added focused specs for both WhatsApp Cloud and 360dialog providers. Focused RSpec now passes with 49 examples, 0 failures.

izcoser:
Hey, could anyone take a look at it?

salvuccicarlos:
+1 — we hit exactly this limitation in production, with a different use case, so
here is a second real-world example in case it helps move the review along.

We run a WhatsApp bot for the residents of the buildings our company manages.
The list rows are apartments, and the row has to identify both the unit and the
building. Building names in our portfolio go up to 52 characters, so 24 is not
close to enough.

The workaround we ship today is the one described in the issue: number the
options and repeat the full text in the message body.

It works, but it duplicates every label and makes the reader match numbers
between the body and the list. With `description` this becomes one clean row per
unit and the body goes back to being a single question.

Two notes on the implementation:

- Forcing the list payload when items carry a description (instead of reply
  buttons for 1–3 options) is the right call for us too. Reply buttons cannot
  carry a subtitle, and silently dropping the description depending on how many
  options the menu happens to have would be surprising.
- Keeping the key optional means nothing changes for existing integrations —
  ours included, for the menus where 24 characters are plenty.

Happy to test this against our WhatsApp Cloud API inbox on a self-hosted 4.16.1
if that is useful.

izcoser:
> +1 — we hit exactly this limitation in production, with a different use case, so
> 
> here is a second real-world example in case it helps move the review along.
> 
> 
> 
> We run a WhatsApp bot for the residents of the buildings our company manages.
> 
> The list rows are apartments, and the row has to identify both the unit and the
> 
> building. Building names in our portfolio go up to 52 characters, so 24 is not
> 
> close to enough.
> 
> 
> 
> The workaround we ship today is the one described in the issue: number the
> 
> options and repeat the full text in the message body.
> 
> 
> 
> It works, but it duplicates every label and makes the reader match numbers
> 
> between the body and the list. With `description` this becomes one clean row per
> 
> unit and the body goes back to being a single question.
> 
> 
> 
> Two notes on the implementation:
> 
> 
> 
> - Forcing the list payload when items carry a description (instead of reply
> 
>   buttons for 1–3 options) is the right call for us too. Reply buttons cannot
> 
>   carry a subtitle, and silently dropping the description depending on how many
> 
>   options the menu happens to have would be surprising.
> 
> - Keeping the key optional means nothing changes for existing integrations —
> 
>   ours included, for the menus where 24 characters are plenty.
> 
> 
> 
> Happy to test this against our WhatsApp Cloud API inbox on a self-hosted 4.16.1
> 
> if that is useful.

Hi. Yeah, nice to hear. We have been using this PR in prod for a while already.

muhsin-k (approved):
LGTM

PR #14943, opened 2026-07-07
chore(inbox): temporarily disable Instagram and WhatsApp embedded signup inbox creation

muhsin-k:
This temporarily turns off inbox creation for channels that rely on Meta's embedded signup — the WhatsApp Cloud signup popup, Instagram (OAuth-only), and WhatsApp Call. WhatsApp Cloud inbox creation now falls back to the manual API-key form, while the Instagram and WhatsApp Call tiles appear disabled. Both channels are also hidden from the new-account onboarding flow. Existing inboxes are not affected.


sony-mathew (approved):
LGTM

PR #14944, opened 2026-07-07
fix: verify contact inbox HMAC server-side

vishnu-narayanan:
# Pull Request Template

## Description
This verifies `ContactInbox#hmac_verified` server-side when a contact is attached to an API inbox instead of trusting the client supplied `hmac_verified` flag.

## Type of change

## How Has This Been Tested?

- `bundle exec rspec spec/controllers/api/v1/accounts/contacts/contact_inboxes_controller_spec.rb`

## Checklist:

Fixes [https://linear.app/chatwoot/issue/CW-6937](https://linear.app/chatwoot/issue/CW-6937)
Fixes [https://linear.app/chatwoot/issue/CW-7464](https://linear.app/chatwoot/issue/CW-7464)
Fixes [https://linear.app/chatwoot/issue/CW-6921](https://linear.app/chatwoot/issue/CW-6921)


vishnu-narayanan:
Validation
------

Tested against the real account contact-inbox API path locally:

- Created a contact with an identifier.
- Called the contact-inbox creation endpoint with `hmac_verified=true` but without a valid HMAC.
- Verified DB state: the contact inbox was created, but `hmac_verified=false`.
- Called the same endpoint with the correct `identifier` and `identifier_hash`.
- Verified DB state: the contact inbox was created with `hmac_verified=true`.
- Verified the HMAC is computed server-side from the inbox HMAC token and contact identifier, instead of trusting the client supplied `hmac_verified` flag.

Regression coverage:
- `bundle exec rspec spec/controllers/api/v1/accounts/contacts/contact_inboxes_controller_spec.rb`
- Result: 6 examples, 0 failures.

sony-mathew (commented):
Changes look good. One question: does the SDK provide identifier_hash in all routes where we call `hmac_verified?`.

Got one finding from codex review:

## Findings

### [P1] Existing spoofed HMAC grants are never revoked

The new check only controls the value passed during future creation. [ContactInboxBuilder](/chatwoot/app/builders/contact_inbox_builder.rb:61) uses `first_or_create!`, so existing `hmac_verified=true` rows remain true. There is also no migration or rollout cleanup.

That persisted boolean is directly trusted by unauthenticated public API authorization in [conversations_controller.rb](/chatwoot/app/controllers/public/api/v1/inboxes/conversations_controller.rb:5) and [inboxes_controller.rb](/chatwoot/app/controllers/public/api/v1/inboxes_controller.rb:24), and by cross-session broadcasting in [action_cable_listener.rb](//chatwoot/app/listeners/action_cable_listener.rb:216).

Therefore, an attacker who set the flag through the vulnerable API before an upgrade retains the expanded conversation access after this patch.

I reproduced the persistence directly:

```ruby
{ same_record: true, hmac_verified_after_safe_rebuild: true }
```

This needs a checked-in rollout cleanup that revokes previously granted state. Because the boolean has no provenance, the conservative option is resetting verified rows and requiring legitimate clients to re-verify. The cleanup should have regression coverage proving a pre-upgrade spoofed row no longer authorizes public access.

## Follow-ups

Blocking:

- Add rollout remediation for existing `hmac_verified=true` records.
- Test the persisted pre-upgrade attack state through the downstream public API.

PR #14948, opened 2026-07-07
fix: captain hours saved metric

scmmishra:
Reworks the Captain assistant "hours saved" metric so it produces a believable number. It previously multiplied assistant reply count by customer reply-wait time, which inflated the figure into the millions of hours. It now estimates saved time as replies × a fixed ~2 min per-reply effort assumption. On the overview card the value renders in days once it passes 100 hours, and the label/hint copy was updated to match.

## How to test
1. Open the Captain assistant overview page.
2. Check the "Time saved" card shows a sensible value (hours, or days past 100h).
3. Hover the hint and confirm it reflects the per-reply estimate.

### DIFF
--- a/app/javascript/dashboard/routes/dashboard/captain/assistants/overview/Index.vue
+++ b/app/javascript/dashboard/routes/dashboard/captain/assistants/overview/Index.vue
@@ -49,6 +49,11 @@ const resolveTrendGood = (trendValue, direction) => {
 // :point, and a plain number for :absolute counts like conversation depth.
 const TREND_SUFFIX = { percent: '%', point: ' pts', absolute: '' };
 
+// Hours-saved is reported in hours, but large values read better as days. Past
+// 100h we switch the unit so the card stays legible.
+const formatDuration = hours =>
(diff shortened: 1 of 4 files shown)


aakashb95 (approved):

app/javascript/dashboard/routes/dashboard/captain/assistants/overview/Index.vue
```diff
@@ -49,6 +49,11 @@ const resolveTrendGood = (trendValue, direction) => {
 // :point, and a plain number for :absolute counts like conversation depth.
 const TREND_SUFFIX = { percent: '%', point: ' pts', absolute: '' };
 
+// Hours-saved is reported in hours, but large values read better as days. Past
+// 100h we switch the unit so the card stays legible.
+const formatDuration = hours =>
```
aakashb95, line 54: nit: i think more than 24hrs should be in days right, 60hrs would look weird?

aakashb95: correction: it will not, 60h saved looks better than 2.5 days saved. LGTM

PR #14950, opened 2026-07-08
fix(db): fail loudly when production DB credentials are missing

A831ARD0:
## Summary
`config/database.yml` falls back to the well-known weak credential `chatwoot_prod`/`chatwoot_prod` for the production database username and password when `POSTGRES_USERNAME`/`POSTGRES_PASSWORD` are unset, instead of failing fast. This changes both to a required `ENV.fetch` so a misconfigured deployment fails loudly at boot rather than silently connecting with a known default credential.

## What changed
- `production.username` and `production.password` now use `ENV.fetch('POSTGRES_USERNAME')` / `ENV.fetch('POSTGRES_PASSWORD')` with no default value.

## How to test
1. Unset `POSTGRES_USERNAME`/`POSTGRES_PASSWORD` in a production-like environment.
2. Boot the app — it should raise `KeyError` immediately instead of connecting with the default `chatwoot_prod` credential.
3. With the env vars set normally, the app boots and connects as before.


A831ARD0:
Heads up: the `ci/circleci: lint` check's **Bundle audit** step is currently failing on this PR, but it's unrelated to this change — it's a pre-existing dependency vulnerability already present on `develop`:

- `datadog` 2.19.0 — CVE-2026-50276 (High, DoS via baggage header parsing) — fix: `>= 2.32.0`
- `loofah` 2.23.1 — GHSA-9wjq-cp2p-hrgf (Medium, XSS via SVG) — fix: `>= 2.25.2`
- `rails-html-sanitizer` 1.6.1 — GHSA-cj75-f6xr-r4g7 — fix: `>= 1.7.1`

Opening a separate PR to bump these three gems.

A831ARD0:
Good catch on the DATABASE_URL/Heroku deployment path — fixed: the strict fetch is now skipped whenever `DATABASE_URL` is present, so `app.json`'s `heroku-postgresql` provisioning keeps working as-is.

A831ARD0:
Thanks @aakashb95 for pushing the cleaner validation approach with `DATABASE_URL` handling built in! Picked it up and found one more spot that needed a placeholder: the "Run Size Limit Check" workflow also runs `RAILS_ENV=production bundle exec rake assets:precompile` (via `.env.example`, which leaves `POSTGRES_PASSWORD` blank), so it hit the same build-time-only issue as `docker/Dockerfile`. Fixed by passing the same kind of placeholder creds there too — verified locally that:\n- missing creds in production still raises\n- `DATABASE_URL`-only deployments (Heroku) still work\n- both asset-precompile-only CI paths (Docker build and size-limit check) now pass placeholder creds and succeed

aakashb95:
@A831ARD0 thanks for raising this PR.
We won’t be merging it because it would turn the existing credential fallback into a startup failure and could break older self-hosted installations that rely on it.

PR #10568, opened 2024-12-10
feat(v4): Compose a new conversation from a phone number.

iamsivin:
# Pull Request Template

## Description

This PR adds the ability to create new contacts from phone numbers in the compose conversation form.

Fixes https://linear.app/chatwoot/issue/CW-3781/ability-to-create-new-conversation-contacts-from-phone-numbers-in

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

**Loom video**
https://www.loom.com/share/d9e16714c66d4095acccedac1f75e876?sid=10f4db7b-886f-4e01-a9c6-82513a8438e2

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #14962, opened 2026-07-08
fix: Handle phone numbers gracefully in the new conversation To: field

salmonumbrella:
Typing or pasting a phone number into the new conversation "To:" field now works the way agents expect. Numbers in any common format — `901-788-0795`, `(901) 788-0795`, `+1 901-788-0795` — are recognized as phone numbers, the create-contact suggestion shows the number in international format so the inferred country code is visible, and the contact is created with a clean E.164 phone number. Numbers typed without a country prefix use the agent's country, inferred from the browser timezone via the same helper the contact form's phone input already uses.

## How to reproduce

1. Open the compose dialog and type `901-788-0795` (a number not in your contacts) in the To: field → nothing happens; the input is silently validated as an email (`ContactSelector` classifies input as a phone number only when it starts with `+`).
2. Type `+1 901-788-0795` and choose the create suggestion → contact creation fails silently: the raw formatted string is sent as `phone_number`, and the Contact model requires strict E.164.
3. Type the number of an existing contact and press Enter → a duplicate contact is created, because the Enter-key matcher only compares menu items by email.
4. Search for an existing contact by typing their number with dashes → no results, since stored numbers are E.164 and the search is a literal `ILIKE`.

## What changed

- `tagInputHelper.js`: phone-like input detection (`detectInputType`/`isPhoneLikeInput`), default-country-aware validation/formatting (via the existing `getActiveCountryCode()` timezone helper), new `normalizePhoneNumber` (E.164), and a type-aware `findMatchingMenuItem` so Enter selects an existing contact by number instead of duplicating.
- `ContactSelector.vue`: uses `detectInputType` instead of `startsWith('+')`.
- `TagInput.vue`: passes the input type to `findMatchingMenuItem`; a tel-mode search result that is an email-only contact (matched by a numeric name or identifier) is now selected as an email instead of being refused by phone validation.
- `composeConversationHelper.js`: `createNewContact` sends E.164; contact search queries that look like phone numbers are normalized to digits so they match stored E.164 numbers. (Trade-off: a numeric identifier containing separators, e.g. `123-456`, is searched as digits in this field; pure-digit identifiers are unaffected.)

Frontend-only; no new dependencies (libphonenumber-js and the timezone helpers were already in use). Extends the phone compose flow introduced in #10568.


salmonumbrella:
@sojan-official @sony-mathew Create outbound call to a number who is not in contacts

iamsivin:
@codex review?

PR #14968, opened 2026-07-09
chore: improve account suspended message

muhsin-k:
Updates the suspended account screen copy to use clearer policy and safety-oriented language, while giving users a direct support path if they believe the suspension is a mistake.


sony-mathew (approved):
LGTM

PR #14970, opened 2026-07-09
feat: add captain sessions model [CW-7485]

scmmishra:
This adds a `captain_sessions` table to log every Captain run, starting with Assistant Responses and Copilot Responses. Each session records the assistant, model, credits consumed, the FAQs/documents/scenario that contributed to the response, and the full run context — giving customers visibility into how a response was generated and giving us durable stats on credit, FAQ, and document usage (which today only exist as ephemeral trace metadata and an aggregate account counter).

## What changed

- New `Captain::Session` model with a `session_type` enum (`assistant`, `copilot`). The subject (`Conversation` / `CopilotThread`) and result (`Message` / `CopilotMessage`) classes are inferred from the session type, so the table stores plain `subject_id` / `result_id` ids. `result_id` is nullable so failed runs that still consumed credits can be logged.
- Composite indexes on `[session_type, subject_id]`, `[session_type, result_id]`, and `[account_id, session_type, created_at]` for lookup and usage-stats queries.
- Factory and model specs.

This PR is schema + model only; the writer/instrumentation that records sessions from the assistant and copilot flows will follow.


scmmishra:
@codex please review

db/migrate/20260709091147_create_captain_sessions.rb
```diff
@@ -0,0 +1,24 @@
+class CreateCaptainSessions < ActiveRecord::Migration[7.1]
+  def change
+    create_table :captain_sessions do |t|
```
sony-mathew: Let's rename this to something generic related to LLMs rather than captain? Maybe `llm_sessions`? What do you think?

db/migrate/20260709091147_create_captain_sessions.rb
```diff
@@ -0,0 +1,24 @@
...
+      t.bigint :subject_id, null: false
+      t.bigint :result_id
+      t.references :account, null: false, index: true
+      t.references :assistant, null: false, index: true
+      t.references :user, index: true
+      t.string :llm_model
+      t.float :credits_consumed
+      t.jsonb :faq_ids, default: []
+      t.jsonb :document_ids, default: []
+      t.bigint :scenario_id
```
sony-mathew: This should also be an jsonb and an array right? We can refer to multiple scenarios in the same run, in the loop that we have.

db/migrate/20260709091147_create_captain_sessions.rb
```diff
@@ -0,0 +1,24 @@
+class CreateCaptainSessions < ActiveRecord::Migration[7.1]
+  def change
+    create_table :captain_sessions do |t|
+      t.integer :session_type, null: false
+      t.bigint :subject_id, null: false
```
sony-mathew: Remind me, what is subject here? Should this be polymorphic?

db/migrate/20260709091147_create_captain_sessions.rb
```diff
@@ -0,0 +1,24 @@
...
+      t.jsonb :faq_ids, default: []
+      t.jsonb :document_ids, default: []
+      t.bigint :scenario_id
+      t.jsonb :run_context, default: {}
+
+      t.timestamps
+    end
+
+    add_index :captain_sessions, [:session_type, :subject_id]
+    add_index :captain_sessions, [:session_type, :result_id]
```
sony-mathew: Should we add account scope of this index also?

db/migrate/20260709091147_create_captain_sessions.rb
```diff
@@ -0,0 +1,24 @@
...
+      t.float :credits_consumed
+      t.jsonb :faq_ids, default: []
+      t.jsonb :document_ids, default: []
+      t.bigint :scenario_id
+      t.jsonb :run_context, default: {}
+
+      t.timestamps
+    end
+
+    add_index :captain_sessions, [:session_type, :subject_id]
```
sony-mathew: should we add account scope of this also?

sony-mathew (commented):
Hey take a look at these findings:

**Findings**
[P1] `captain_sessions` can retain account-scoped LLM context after account/assistant deletion. The migration adds `account_id`, `assistant_id`, and `run_context` on `captain_sessions`, but the new model is not associated from `Account` or `Captain::Assistant` for dependent cleanup. Existing Enterprise cleanup lists Captain assistants/responses/documents/tools and copilot threads, but not sessions; `Captain::Assistant` also has dependent cleanup for documents/responses/inboxes/copilot threads/scenarios, but not sessions. The follow-up PR attached to `CW-7485` stores current-turn conversation history in `run_context`, so this becomes a privacy/data-retention issue once capture lands.
Refs: [migration](/chatwoot/db/migrate/20260709091147_create_captain_sessions.rb:7), [account concern](/chatwoot/enterprise/app/models/enterprise/concerns/account.rb:12), [assistant model](/chatwoot/enterprise/app/models/captain/assistant.rb:26)

[P2] The `subject`/`result` helpers dereference global records instead of preserving the session’s account boundary. `Conversation.find_by`, `Message.find_by`, `CopilotThread.find_by`, and `CopilotMessage.find_by` should scope through `account`/`account_id` so future APIs or UI cannot render another account’s record if a bad writer, console repair, or import creates a mismatched `subject_id` or `result_id`.
Ref: [session model](/chatwoot/enterprise/app/models/captain/session.rb:42)

**Risk Audit**
Security/privacy: needs work because `run_context` is intended to hold LLM context.  
Schema/modeling: new `captain_sessions` table; indexes cover subject/result lookup and account/session/time lookup.  
API/backcompat: no API surface changed in this PR.  
Performance: no hot-path writer yet in this PR; future stats by `faq_ids`/`document_ids` may need a query-specific index or normalized usage table.  
Dependencies/external services: no new dependency or external service.  
Feature flags/plans: Enterprise-only model, no new flag.

**Areas Impacted**
Enterprise Captain assistant/copilot session logging, future Captain stats/audit UI, account deletion/data retention, and future session readers.

**Alternatives And Follow-Ups**
There is already an unresolved inline thread about whether `captain_sessions` should be more generic, like `llm_sessions`; I did not duplicate that as a new finding. If the product direction is “all LLM runs,” renaming before merge is cheaper than migrating later.

db/migrate/20260709091147_create_captain_sessions.rb
scmmishra: Yeah, we can do that, it's invisible to user, so it doesn't matter. I would still prefer Captain since that is the umbrella term we use

db/migrate/20260709091147_create_captain_sessions.rb
scmmishra: Subject can either be a Copilot thread or Conversation, based on the session type. The polymorphism is implied through the code in the model that fetches the subject and result

db/migrate/20260709091147_create_captain_sessions.rb
scmmishra: Yeah, makes sense, will add

db/migrate/20260709091147_create_captain_sessions.rb
scmmishra: Yeah, makes sense, will add

db/migrate/20260709091147_create_captain_sessions.rb
scmmishra: Let me think this through, I originally though the final Scenario will be added here. A turn ideally will have just one scenario.

db/migrate/20260709091147_create_captain_sessions.rb
scmmishra: @sony-mathew wdyt?

db/migrate/20260709091147_create_captain_sessions.rb
scmmishra: At the moment there's no way to capture all the subagents in the loop. but gonna make it a jsonb array for future proofing

db/migrate/20260709091147_create_captain_sessions.rb
scmmishra: Fixed it to use rails native polymorphism

db/migrate/20260709091147_create_agent_sessions.rb
```diff
@@ -0,0 +1,24 @@
+class CreateCaptainSessions < ActiveRecord::Migration[7.1]
+  def change
+    create_table :captain_sessions do |t|
+      t.integer :session_type, null: false
+      t.bigint :subject_id, null: false
+      t.bigint :result_id
+      t.references :account, null: false, index: true
```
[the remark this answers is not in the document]
scmmishra: Stale comment

enterprise/app/models/captain/agent_session.rb
```diff
@@ -0,0 +1,64 @@
...
+  self.table_name = 'captain_sessions'
+
+  SUBJECT_TYPES = { 'assistant' => 'Conversation', 'copilot' => 'CopilotThread' }.freeze
+  RESULT_TYPES = { 'assistant' => 'Message', 'copilot' => 'CopilotMessage' }.freeze
+
+  belongs_to :account
+  belongs_to :assistant, class_name: 'Captain::Assistant'
+  belongs_to :user, optional: true
+  belongs_to :subject, ->(session) { where(account_id: session.account_id) }, polymorphic: true
+  belongs_to :result, ->(session) { where(account_id: session.account_id) }, polymorphic: true, optional: true
```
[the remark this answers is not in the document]
scmmishra: Solved

enterprise/app/models/captain/assistant.rb
```diff
@@ -37,6 +37,7 @@ class Captain::Assistant < ApplicationRecord
   has_many :messages, as: :sender, dependent: :nullify
   has_many :copilot_threads, dependent: :destroy_async
   has_many :scenarios, class_name: 'Captain::Scenario', dependent: :destroy_async
+  has_many :sessions, class_name: 'Captain::Session', dependent: :destroy_async
```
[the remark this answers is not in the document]
scmmishra: Solved

enterprise/app/models/captain/session.rb
```diff
@@ -0,0 +1,84 @@
...
+  belongs_to :result, ->(session) { where(account_id: session.account_id) }, polymorphic: true, optional: true
+
+  enum :session_type, { assistant: 0, copilot: 1 }, prefix: :session
+
+  before_validation :ensure_account
+
+  validate :subject_type_matches_session_type
+  validate :result_type_matches_session_type, if: -> { result_type.present? }
+  validate :subject_belongs_to_account
+  validate :result_belongs_to_account, if: -> { result.present? }
```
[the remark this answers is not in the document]
scmmishra: Solved

db/migrate/20260709091147_create_captain_sessions.rb
sony-mathew: I would say it would still be confusing later down the road. something like `agent_sessions` makes more sense to me.

sony-mathew (approved):
LGTM

enterprise/app/models/captain/agent_session.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def subject_belongs_to_account
+    return if subject.nil? || subject.account_id == account_id
+
+    errors.add(:subject, 'must belong to the session account')
+  end
+
+  def result_belongs_to_account
+    target_class = result_type.safe_constantize
+    actual_account_id = target_class && target_class.unscoped.where(id: result_id).pick(:account_id)
```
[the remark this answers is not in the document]
scmmishra: It's fine, we're gonna access this via ActiveRecord only

PR #14978, opened 2026-07-10
feat(captain): group conversation FAQ signals (2/4)

aakashb95:
Captain can now turn resolved conversations into FAQ suggestions that people can review. When a human support agent gives a reusable answer, Captain saves the question and answer as an observation. Captain groups matching observations into one suggestion instead of creating a pending FAQ for every conversation. This PR does not approve or publish FAQs.

PR [#14977](https://github.com/chatwoot/chatwoot/pull/14977) adds the data model and should be reviewed first. The controller and UI PRs will add the review flow. The three PRs should merge together.

## Closes

[CW-7495](https://linear.app/chatwoot/issue/CW-7495/backend-llm-changes-to-make-conversation-faqs-as-signalssuggestions). This is PR 2 of 3. The issue is complete after the full stack lands.

## What changed

1. Runs FAQ generation in the low priority queue after a conversation is resolved.
2. Reads only customer messages and answers written by human support agents.
3. Uses the assistant's product details, instructions, response rules, and guardrails to reject spam and unrelated conversations.
4. Stores each reusable question and answer as an observation.
5. Uses exact text similarity search within the conversation language to find likely matches.
6. Asks the LLM whether both FAQs ask the same question and give the same answer.
7. Does not create a new suggestion when an approved FAQ already covers the observation.
8. Adds matching observations to an existing open suggestion and updates its source count.
9. Does not suggest the same FAQ again after someone dismisses it.
10. Creates a new open suggestion only when no approved FAQ or existing suggestion covers the observation.

## How to test

1. Resolve a conversation where a human support agent gives a reusable answer. Confirm that Captain creates one open suggestion with one source.
2. Resolve another conversation with the same question and answer. Confirm that Captain adds a source to the existing suggestion instead of creating another suggestion.
3. Resolve a conversation that is already covered by an approved FAQ. Confirm that Captain creates no new suggestion.
4. Dismiss a suggestion, then resolve another conversation with the same question and answer. Confirm that Captain does not suggest the FAQ again.
5. Resolve a spam or unrelated conversation. Confirm that Captain creates no suggestion.

PR #14979, opened 2026-07-10
feat(captain): add FAQ suggestion review API (3/4)

aakashb95:
Agents can review recurring FAQ suggestions when they can access at least one supporting conversation. The detail view returns only source conversations the agent can access. Administrators can review every suggestion and can edit, approve, or dismiss it. Approval creates one approved Captain FAQ and removes the stored source observations.

This is the third PR in the CW-7495 stack. It is built on [#14978](https://github.com/chatwoot/chatwoot/pull/14978), which adds the FAQ suggestion models and generation flow.

## Closes

Closes [CW-7495](https://linear.app/chatwoot/issue/CW-7495/backend-llm-changes-to-make-conversation-faqs-as-signalssuggestions).

## What changed

1. Added a paginated suggestion list with assistant, status, and search filters.
2. Limited agents to suggestions that have at least one source conversation they can access.
3. Limited the detail response to the 50 most recent source conversations the current user can access.
4. Allowed administrators to edit, approve, and dismiss open suggestions.
5. Added approval that creates one approved Captain FAQ, closes the suggestion, and removes its source observations.
6. Rejected approval when the suggestion language does not match the account language.
7. Added row locking so an edit or dismissal cannot overwrite an approval.
8. Prevented FAQ generation from attaching a new observation after a suggestion has closed.

## How to test

1. Sign in as an agent who has access to one inbox but not another.
2. Confirm the agent sees only suggestions with at least one source conversation from an accessible inbox.
3. Open a suggestion and confirm the source list does not contain conversations from restricted inboxes.
4. Sign in as an administrator and confirm all account suggestions are available.
5. Edit an open suggestion and approve it. Confirm one approved FAQ is created and the suggestion no longer has source observations.
6. Try to approve a suggestion in a different language from the account language. Confirm the request is rejected.
7. Dismiss another open suggestion and confirm it leaves the open review queue.


enterprise/app/services/captain/faq_suggestion_approval_service.rb
```diff
@@ -0,0 +1,27 @@
...
+  def perform
+    suggestion.with_lock do
+      raise ActiveRecord::RecordNotFound unless suggestion.open?
+
+      suggestion.update!(attributes) if attributes.present?
+
+      response = suggestion.assistant.responses.create!(
+        question: suggestion.question,
+        answer: suggestion.answer,
+        status: :approved
```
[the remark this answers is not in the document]
aakashb95: fixed in [4a73de1](https://github.com/chatwoot/chatwoot/pull/14978/commits/4a73de15818dd1741b98067ba5e18ab6f8af6292)

enterprise/app/controllers/api/v1/accounts/captain/faq_suggestions_controller.rb
```diff
@@ -0,0 +1,64 @@
+class Api::V1::Accounts::Captain::FaqSuggestionsController < Api::V1::Accounts::BaseController
+  before_action :current_account
+  before_action -> { check_authorization(Captain::Assistant) }
```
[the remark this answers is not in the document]
aakashb95: fixed in[64d2751](https://github.com/chatwoot/chatwoot/pull/14979/commits/64d2751c0601ce4f3e83f6fb366c629dfc147dce)

enterprise/app/controllers/api/v1/accounts/captain/faq_suggestions_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+                               .where(conversation_id: accessible_conversations.select(:id))
+                               .includes(:conversation)
+                               .order(created_at: :desc)
+                               .limit(SOURCE_PREVIEW_LIMIT)
+  end
+
+  def update
+    raise ActiveRecord::RecordNotFound unless @suggestion.open?
+
+    @suggestion.update!(suggestion_params)
```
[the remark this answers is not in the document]
aakashb95: fixed in [f6d97c5](https://github.com/chatwoot/chatwoot/pull/14979/commits/f6d97c56b4e48a48aed2dd7ae58c58af525390c1)

enterprise/app/views/api/v1/models/captain/_faq_suggestion.json.jbuilder
```diff
@@ -0,0 +1,12 @@
+json.id resource.id
+json.account_id resource.account_id
+json.question resource.question
+json.answer resource.answer
```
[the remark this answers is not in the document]
aakashb95: A non-admin agent can see a suggestion only when they can access at least one supporting conversation, which is intentional because FAQ are account level knowledge. There is no PII anyway since the prompt takes care of that.

enterprise/app/controllers/api/v1/accounts/captain/faq_suggestions_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    if permitted_params[:search].present?
+      search_term = "%#{permitted_params[:search]}%"
+      base_query = base_query.where('question ILIKE :search OR answer ILIKE :search', search: search_term)
+    end
+
+    base_query
+  end
+
+  def set_suggestions
+    @suggestions = Current.account.captain_faq_suggestions.includes(:assistant).ordered
```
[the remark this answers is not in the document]
aakashb95: resolved in a later comment

sony-mathew (commented):
Agentic review findings:

## Findings

1. **[P2] Revalidate in-flight matches after an edit.**  
   [faq_suggestions_controller.rb:29](/chatwoot/enterprise/app/controllers/api/v1/accounts/captain/faq_suggestions_controller.rb:29) locks the suggestion while editing, but a generation job may already have classified a candidate against the old question/answer. When it later reaches [conversation_faq_service.rb:88](/chatwoot/enterprise/app/services/captain/llm/conversation_faq_service.rb:88), `with_lock` reloads the edited row and only verifies that it remains open. It then attaches evidence generated for the old FAQ, inflates `source_count`, and can leave approval backed by mismatched sources. The new base-branch mutex serializes generation jobs, but not API edits, so it does not close this race. This is also the remaining valid [[unresolved GitHub thread](https://github.com/chatwoot/chatwoot/pull/14979#discussion_r3622291695)](https://github.com/chatwoot/chatwoot/pull/14979#discussion_r3622291695).

2. **[P3] The restricted detail-response contract lacks request coverage.**  
   [faq_suggestions_controller_spec.rb:64](/chatwoot/spec/enterprise/controllers/api/v1/accounts/captain/faq_suggestions_controller_spec.rb:64) only exercises `show` as an administrator. The PR’s main privacy invariant—an agent with access to one source must not receive observations from restricted conversations—is implemented correctly, but not tested. Add a mixed-access agent example asserting that only the accessible observation is serialized.

enterprise/app/controllers/api/v1/accounts/captain/faq_suggestions_controller.rb
```diff
@@ -0,0 +1,79 @@
+class Api::V1::Accounts::Captain::FaqSuggestionsController < Api::V1::Accounts::BaseController
+  before_action :current_account
+  before_action -> { check_authorization(Captain::Assistant) }
+  before_action :set_suggestions
```
sony-mathew: This should be set only for index? We don't need to set suggestions for show, update and delete actions right? So let's gate this only for those actions or gate this to done for only index action.

enterprise/app/controllers/api/v1/accounts/captain/faq_suggestions_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  private
+
+  def apply_filters(base_query)
+    base_query = base_query.where(assistant_id: permitted_params[:assistant_id]) if permitted_params[:assistant_id].present?
+    base_query = base_query.where(status: permitted_params[:status]) if permitted_params[:status].present?
+
+    if permitted_params[:search].present?
+      search_term = "%#{permitted_params[:search]}%"
+      base_query = base_query.where('question ILIKE :search OR answer ILIKE :search', search: search_term)
```
sony-mathew, line 55: It's not a good thing to do like queries in pg DB right? But for now, let it be. But add a comment #TODO to move this to ES.

enterprise/app/services/captain/faq_suggestion_approval_service.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      suggestion.update!(attributes) if attributes.present?
+      validate_language!
+
+      response = suggestion.assistant.responses.create!(
+        question: suggestion.question,
+        answer: suggestion.answer,
+        status: :approved
+      )
+      suggestion.observations.delete_all
```
sony-mathew: Why do we need to delete all observations? Will it be a problem if we don't delete it?

enterprise/app/services/captain/faq_suggestion_approval_service.rb
```diff
@@ -0,0 +1,39 @@
...
+      suggestion.approved!
+      response
+    end
+  end
+
+  private
+
+  attr_reader :suggestion, :attributes
+
+  def validate_language!
```
sony-mathew: this can be moved to suggestion model itself right?

sony-mathew (commented):
Added some comments.

enterprise/app/controllers/api/v1/accounts/captain/faq_suggestions_controller.rb
aakashb95: This should have been named better, something like `set_accessible_suggestions`, since we have admins who can approve and agents who can only see and some restricted agents that may not have access to conversations, we need to prepare the query appropriately before listing the suggestions. We can take a call if FAQ suggestions should only be approved by admins or agents or both.

enterprise/app/services/captain/faq_suggestion_approval_service.rb
aakashb95: Hmmm we should keep it for audits i guess? just in case we ever want to show based on which observation the FAQ was suggested and approved.. it won't be a problem if we delete because the suggested FAQ has been approved by this point and is part of the knowledge base now

enterprise/app/services/captain/faq_suggestion_approval_service.rb
aakashb95: this helped spotting another bug, currently if a suggested FAQ is in a different language than the account's locale, the approval will not go through. need to remove this check altogether.

enterprise/app/policies/captain/faq_suggestion_policy.rb
```diff
@@ -0,0 +1,21 @@
...
+  def show?
+    true
+  end
+
+  def update?
+    true
+  end
+
+  def approve?
+    true
```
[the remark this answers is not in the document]
aakashb95: this is a product call, since agents are closer to actual customers than admins

aakashb95:
fixed all comments @sony-mathew

enterprise/app/controllers/api/v1/accounts/captain/faq_suggestions_controller.rb
aakashb95: Decided to allow both

sony-mathew (approved):
lgtm

PR #14981, opened 2026-07-10
fix: avoid full scan in IMAP email dedup on large inboxes

### LINKED ISSUE #14682: IMAP inbound stops on large inboxes: email_already_present? full-scans messages (ORDER BY created_at LIMIT 1) → connection drops
### Summary

On self-hosted Chatwoot **v4.14.1**, inbound IMAP email silently stopped on the email inboxes that have a **large message history**, while small inboxes on the same mail server kept working fine. Outbound (SMTP) was unaffected.

Sidekiq logged this every fetch cycle for the affected inboxes:

```
[ActiveJob] [Inboxes::FetchImapEmailsJob] Error for email channel - <id> : closed stream
```

…and the inbox ingested **nothing** (zero progress), so the un-fetched backlog kept growing. It correlates with the **inbox's total message history size**, not with how much new mail arrives.

### Root cause

`Imap::BaseFetchEmailService#email_already_present?`:

```ruby
def email_already_present?(channel, message_id)
  channel.inbox.messages.find_by(source_id: message_id).present?
end
```

`find_by` on `channel.inbox.messages` carries the model's default `ORDER BY created_at ASC` plus `LIMIT 1`. On a large `messages` table, the Postgres planner satisfies `ORDER BY created_at LIMIT 1` by walking `index_messages_on_created_at` and filtering on `inbox_id`/`source_id`. For a `source_id` that isn't in the DB yet (i.e. a **new** message) it walks the whole table before giving up:

```
EXPLAIN (ANALYZE, BUFFERS) for the generated query
  SELECT "messages".* FROM "messages"
  WHERE "messages"."inbox_id" = <N> AND "messages"."source_id" = '<new-message-id>'
  ORDER BY "messages"."created_at" ASC LIMIT 1

Limit  (actual time=2105.692..2105.692 rows=0 loops=1)
  ->  Index Scan using index_messages_on_created_at on messages
        Filter: ((inbox_id = <N>) AND (source_id = '<new-message-id>'::text))
        Rows Removed by Filter: 1551501
        Execution Time: 2105.709 ms
```

~**2.1 s per message**. `fetch_mail_for_channel` calls `email_already_present?` once per message in the `SINCE` window, so a single fetch run for a busy inbox takes **75 s+**. During that loop the IMAP socket is idle (no commands sent), the mail server eventually closes it, and the next `imap.fetch` raises:

```
IOError: closed stream
  openssl/buffering.rb:351  OpenSSL::SSL::SSLSocket#syswrite
  net-imap-0.4.24 .../net/imap.rb  put_string -> send_command -> fetch_internal
```

The job aborts before processing any message, so the inbox makes **zero progress** and can never catch up.

### Why this started after upgrading from 4.7.x

The slow query itself is identical in 4.7 and 4.14. What changed is the header-fetch batching in `fetch_message_ids_with_sequence`:

- **4.7.0:** `seq_nums.each_slice(10)` — an `imap.fetch` of headers every 10 messages keeps the IMAP connection active, so it never idles long enough to be dropped. The run was slow but **completed and ingested**.
- **4.14.1:** `seq_nums.each_slice(MAX_MESSAGES_PER_SYNC)` with `MAX_MESSAGES_PER_SYNC = 500` — one big header fetch, then a long (~75 s) `email_already_present?` loop with **no IMAP activity** in between, so the connection idles out and the server drops it.

So 4.14 turned a "slow but working" fetch into "connection dropped, zero progress" on large inboxes.

### Workaround (verified)

A composite index makes the planner use it and removes the sort:

```sql
CREATE INDEX CONCURRENTLY idx_messages_inbox_source_created ON messages (inbox_id, source_id, created_at);
ANALYZE messages;
```

Same query after: `Index Scan using idx_messages_inbox_source_created`, **`Execution Time: 0.085 ms`** (down from 2105 ms). Fetch runs now finish in <1 s and the backlog ingests normally.

Note: a plain `(inbox_id, source_id)` index was **not** picked by the planner — given a poor row estimate it preferred sorting via the `created_at` index. The 3-column index works because it already supplies `created_at` order, so no sort is needed.

### Suggested fix

1. `email_already_present?` is a presence check; it shouldn't carry an `ORDER BY` or materialize the row. Using `exists?` drops the ordering and lets the planner use the existing `index_messages_on_source_id`:
   ```ruby
   channel.inbox.messages.where(source_id: message_id).exists?
   ```
2. Optionally ship a migration adding an index on `messages (inbox_id, source_id)` (or `(source_id)`), since source_id dedup is a hot path during IMAP sync.
3. Optionally reconsider running the entire dedup loop while holding the IMAP connection idle — the large `each_slice` makes the idle window much worse and is what makes the connection drop.

### Environment

- Chatwoot **v4.14.1**, self-hosted (Docker), PostgreSQL
- `net-imap` **0.4.24**
- Channel: **Email (IMAP, basic auth / SSL 993)**
- Affected: email inboxes with a large message history (in our case one inbox with ~436k messages, on a ~1.5M-row `messages` table). Small inboxes on the same server were unaffected.

vishnu-narayanan:
## Description

`Imap::BaseFetchEmailService#email_already_present?` used `find_by(source_id:)`, which inherits `Message`'s `default_scope { order(created_at: :asc) }`, adding an `ORDER BY created_at ASC LIMIT 1` to what is only a presence check.

On inboxes with a large message history, that `ORDER BY` lets Postgres satisfy the sort by walking `index_messages_on_created_at` instead of the selective `index_messages_on_source_id`. For a not-yet-seen `source_id` (every new email) it can scan the whole table before returning, taking seconds per message. The dedup loop runs with no IMAP activity in between, so the idle socket is dropped by the mail server and the fetch job aborts with `closed stream`. The inbox then stops ingesting mail entirely, while smaller inboxes on the same server keep working.

`exists?` issues `SELECT 1 ... LIMIT 1` with no `ORDER BY`, so the planner uses `index_messages_on_source_id` regardless of table size. No schema change is required. The fix lives in the shared base class, so it covers both the IMAP and Microsoft fetch paths.

Fixes #14682

## Type of change

## How Has This Been Tested?

The change is behavior-preserving: `find_by(source_id:).present?` and `exists?(source_id:)` return the same boolean, so it is covered by the existing `spec/services/imap` and `spec/jobs/inboxes` fetch specs, which pass locally. The query-plan improvement (dropping the `ORDER BY` so the planner uses `index_messages_on_source_id`) was verified with `EXPLAIN` on a large `messages` table.

## Checklist:


sony-mathew (approved):
LGTM

PR #14987, opened 2026-07-10
fix: dedupe API messages by source_id on bridge retries

deagleon:
## Summary
- Deduplicate incoming API messages when `source_id` already exists (Evolution/Baileys webhook retries were creating spam in Chatwoot while WhatsApp showed the message once).
- Harden native WhatsApp webhook locking to atomic Redis `SET NX` to avoid concurrent duplicate inserts.

## Test plan
- [ ] On prod, confirm conversation #2411 inbox is `Channel::Api` and inspect the 4 spam rows' `source_id` values
- [ ] Confirm the bridge maps WhatsApp `key.id` → Chatwoot `source_id`
- [ ] Replay/retry the same incoming webhook twice and verify only one message is created
- [ ] Smoke-test a normal new incoming message still creates correctly

### DIFF
--- a/app/builders/messages/message_builder.rb
+++ b/app/builders/messages/message_builder.rb
@@ -21,6 +21,10 @@ class Messages::MessageBuilder
   end
 
   def perform
+    # Bridges (Evolution/Baileys) may retry the same webhook; return existing message when source_id matches.
+    existing = existing_message_with_source_id
+    return existing if existing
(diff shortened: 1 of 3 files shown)


deagleon:
Opened against upstream by mistake; this change is for our fork deploy (deagleon/chatwoot-original).

PR #15003, opened 2026-07-13
fix(captain): honor mandatory handoff guidelines

aakashb95:
Ensures Captain follows explicit mandatory-transfer rules from active Response Guidelines and Guardrails instead of allowing the generic consent-first fallback to override those rules.

## What changed

- Made explicit transfer requirements take precedence over generic consent-first handoff defaults only when their condition matches.
- Added explicit Response Guideline and Guardrail transfer rules to the human-handoff protocol.
- Added focused prompt regression coverage.

## How to reproduce

Configure a Response Guideline or Guardrail that requires immediate transfer for a specific condition, then send a request matching that condition. Captain should invoke the human-handoff path without asking the user to consent again. Unmatched requests continue to use the existing consent-first fallback.

The assistant prompt renderer, agent prompt context, and focused regression specs pass locally.


sony-mathew (approved):
LGTM

PR #15011, opened 2026-07-14
fix(twilio): register messaging webhook for voice inboxes and add webhook health tab

tds-1:
When a Twilio number is set up through the Voice flow, Chatwoot registers the voice webhooks but never the messaging one. Since a voice channel is really an SMS channel with voice enabled, the number ends up able to receive calls but not SMS, with no way to fix it from the dashboard. This registers the messaging webhook on the voice path, and adds a Health tab to Twilio inboxes that shows whether each webhook Chatwoot expects is actually set on the number, with a button to re-register them.

## How to reproduce

1. Create a Voice inbox with a Twilio number.
2. In the Twilio Console, open the number: the voice webhook points at Chatwoot, but "A message comes in" is untouched.
3. Send an SMS to the number, nothing arrives in Chatwoot.

## How to test

1. Create a Voice inbox with a Twilio number, the messaging webhook is now set automatically and inbound SMS lands in the inbox.
2. Open the inbox, go to Settings > Account Health. Each expected webhook (messaging, plus voice, voice status and the TwiML app when voice is on) shows as configured.
3. Point the messaging webhook somewhere else in Twilio and reload the tab, it shows a mismatch with the reason and a hint. Click Register webhooks and it goes back to green.
4. Point the inbox at a number that is not in the connected Twilio account, the tab surfaces Twilio's own message instead of a blank card.

## Screenshots

A number whose voice webhooks drifted to an old tunnel — each webhook is flagged with why Chatwoot cannot receive on it:

health-misconfigured

After clicking Register webhooks, all four are back on the current installation URL and the repair button drops away:

health-healthy

When Twilio itself rejects the lookup, its message is shown rather than an empty state:

health-error

## What changed

- The voice inbox creation path now registers the messaging webhook alongside the voice ones.
- `Twilio::HealthService` reports the account, the sender and every webhook Chatwoot depends on, with a reason per webhook (not set, URL mismatch, wrong HTTP method, overridden by a trunk / TwiML app / the number's own webhook, missing TwiML app).
- The health and register-webhook endpoints moved from a WhatsApp-only concern to a shared `InboxHealthManagement` concern.
- `Twilio::WebhookSetupService` authenticates through `Channel::TwilioSms#client`, so inboxes on API-key credentials can repair their webhooks too.
- Inbound MMS media downloads use `api_key_secret` when the channel stores one, so voice inboxes keep their attachments.
- The voice toggle saves on flip instead of waiting for a separate Update click.


iamsivin:
@codex review?

sony-mathew (approved):
looks good to me. @iamsivin can you take a pass at the frontend changes?

Please take a look at this finding @tds-1 :

- **[P2] Repair reports success when the configured Twilio number no longer exists.**  
  For a phone-number-backed inbox, [`register_webhook`](/worktrees/fc50/chatwoot/app/controllers/api/v1/accounts/concerns/inbox_health_management.rb:51) always returns HTTP 200 when the setup service completes. However, [`update_phone_number`](/worktrees/fc50/chatwoot/app/services/twilio/webhook_setup_service.rb:26) only logs and returns when Twilio returns no matching number. The voice equivalent also silently returns in [`configure_number_webhooks!`](/worktrees/fc50/chatwoot/enterprise/app/services/twilio/voice_webhook_setup_service.rb:34).

  An exact-head request probe with `incoming_phone_numbers.list` returning `[]` produced:

  ```json
  {"status":200,"body":{"message":"Webhook registered successfully"}}
  ```

  Nothing was repaired, and the asynchronous setup job would also treat the no-op as success rather than retrying. This is attributable to the PR because the base branch rejected Twilio on this endpoint; the PR newly routes Twilio repairs through this pre-existing no-op path.

  **Suggested fix:** raise a clear error when the number lookup is empty, align the voice setup path, and add request/job coverage asserting a non-success result.

PR #15018, opened 2026-07-14
perf(captain): add cosine FAQ search index

aakashb95:
This draft is future work and must not be merged as part of the first FAQ suggestion rollout.

Production currently runs pgvector 0.4.1. Filtered iterative vector scans require pgvector 0.8.0 or newer. The first rollout will use an exact cosine search instead.

Keep this draft for the pgvector upgrade discussion. After production and supported self hosted installations can run pgvector 0.8.0 or newer, this change can replace the unused L2 index with a cosine index.

## Closes

Part of [CW-7495](https://linear.app/chatwoot/issue/CW-7495/backend-llm-changes-to-make-conversation-faqs-as-signalssuggestions).

## How to reproduce

1. Upgrade pgvector to 0.8.0 or newer.
2. Use an assistant with a large FAQ library.
3. Search within that assistant's approved FAQs.
4. Confirm that the search returns five matching rows from the assistant scope.
5. Confirm that PostgreSQL uses the cosine vector index.

## What changed

The migration replaces the old L2 IVFFlat index with a cosine IVFFlat index. It creates and removes both indexes concurrently. FAQ search enables relaxed iterative scans so account, assistant, and status filters do not hide valid matches. The migration stops when pgvector is older than 0.8.0.


aakashb95:
Closing this because we don't have the required pgvector version, we can revisit this once we have the scale

PR #15021, opened 2026-07-15
test: specs for delayed automation rules

tds-1:
## Description

Specs for time based automation changes

## Type of change

## Checklist:


sony-mathew:
## Findings

1. **[P1] The “exactly once” test misses the retry path that duplicates side effects.**  
   [process_pending_execution_job_spec.rb:101](/chatwoot/spec/jobs/automation_rules/process_pending_execution_job_spec.rb:101) invokes the worker twice sequentially after the first execution is fully persisted. It never covers: action succeeds → `status: executed` persistence fails → row remains `processing` → stale reclaim repeats the action. I reproduced that path with `send_message`: the first attempt created one message, the stale retry created a second, then marked the row executed. This confirms the existing Ito QA report. The parent feature needs an explicit delivery guarantee/idempotency design and a regression spec.

2. **[P2] Pause/resume coverage omits stale processing rows.**  
   [automation_rule_pending_execution_spec.rb:294](/chatwoot/spec/models/automation_rule_pending_execution_spec.rb:294) covers only `pending` rows. The parent implementation also reclaims stale `processing` rows, but `reschedule_paused` leaves their old `due_at` untouched; after a long pause they are reclaimed and immediately skipped as `expired`. This is already an unresolved finding on parent PR #15022 and should be encoded here.

3. **[P2] The factory creates invalid message-backed executions.**  
   [automation_rule_pending_executions.rb:7](/chatwoot/spec/factories/automation_rule_pending_executions.rb:7) always derives a status episode key with `message=nil`. My probe with an outgoing message produced `status:*`, while production computed `reply_chase:0`; `episode_current?` was therefore `false`. Derive the key using the supplied message.

sony-mathew (commented):
Other than codex comments, rest looks good. Might need to change some of these specs based on changes in the other PR.

PR #15024, opened 2026-07-15
fix: freeze SLA misses after resolution

sony-mathew:
## Description

Resolved conversations now preserve historical SLA misses without allowing their displayed duration to keep growing. Applied SLAs record a stable completion timestamp that is shared through REST and realtime payloads, and the dashboard freezes FRT, NRT, and RT misses at that point.

Legacy completed SLAs without a reliable timestamp remain visible as a static missed state. Terminal SLAs remain frozen when a conversation is reopened; a reopen before finalization continues the same SLA without resetting its deadlines.

### Closes

[CW-7597](https://linear.app/chatwoot/issue/CW-7597/freeze-sla-miss-durations-after-conversation-resolution)

## Type of change

## How to reproduce

1. Apply an SLA with a resolution-time threshold to a conversation.
2. Let the threshold breach, then resolve the conversation.
3. Observe that the recorded miss duration continues increasing every minute even though the conversation is resolved.

## What changed

- Added nullable `applied_slas.completed_at` and exposed it as `sla_completed_at` in conversation, report, and websocket payloads.
- Captured completion before broadcasting resolution and preserved it for terminal applied SLAs.
- Frozen recorded FRT, NRT, and RT durations in classic and next-generation conversation labels, including a static fallback for legacy rows.
- Added a dry-run-first, resumable Rails runner for account-scoped or explicitly global historical repair without enqueuing jobs or touching `updated_at`.

Account-scoped production rollout starts with:

```sh
ACCOUNT_ID=168154 bundle exec rails runner script/backfill_applied_sla_completed_at.rb
ACCOUNT_ID=168154 APPLY=true bundle exec rails runner script/backfill_applied_sla_completed_at.rb
```

## How Has This Been Tested?

- Verified resolution stamping, nonterminal reopen clearing, and terminal reopen preservation.
- Verified dry-run, apply, account/global scope, resume, skip, idempotency, and timestamp-preserving backfill behavior.
- Verified all three miss types freeze and existing conversation-card behavior remains intact.
- 71 focused RSpec examples and 37 focused Vitest examples pass.
- RuboCop, ESLint, and diff checks pass.

## Checklist:


app/javascript/dashboard/components-next/Conversation/ConversationCard/SLACardLabel.vue
```diff
@@ -39,7 +47,24 @@ const updateSlaStatus = () => {
   });
 };
 
+const clearTimer = () => {
```
iamsivin: The timer lifecycle logic is duplicated across all three `SLACardLabel.vue` variants. Might be worth extracting into a shared composable in a follow-up.

app/javascript/dashboard/components-next/Conversation/Sla/SLACardLabel.vue
```diff
@@ -39,7 +54,24 @@ const updateSlaStatus = () => {
   });
 };
 
+const clearTimer = () => {
```
iamsivin: https://github.com/chatwoot/chatwoot/pull/15024/changes#r3630732215

app/javascript/dashboard/components/widgets/conversation/components/SLACardLabel.vue
```diff
@@ -65,7 +80,24 @@ const updateSlaStatus = () => {
   });
 };
 
+const clearTimer = () => {
```
iamsivin: https://github.com/chatwoot/chatwoot/pull/15024/changes#r3630732215

iamsivin (commented):
@sony-mathew Added couple of comments. Please take a look and let me know if they're valid.


1. Existing resolved + missed conversations will lose their overdue duration until `completed_at` is backfilled. Since `isSLACompleted` now returns `true` for resolved conversations, the card shows only **"Missed"** without the time. We should run the backfill as part of the deployment, not after.
app/javascript/dashboard/helper/slaHelper.js:106

2. Just to confirm, is it intentional that reopening a conversation whose SLA is already `hit`/`missed` keeps the SLA frozen? The early return skips clearing `completed_at`, so the timer never resumes.
enterprise/app/models/enterprise/conversation.rb:43

3. `update!` now runs before the other resolve callbacks. If it raises, later callbacks like activity creation and notifications won't run. Just wanted to check if that's expected.
enterprise/app/models/enterprise/conversation.rb:45

app/javascript/dashboard/helper/slaHelper.js
```diff
@@ -84,47 +103,51 @@ export const evaluateSLAStatus = ({ appliedSla, chat, slaEvents = [] }) => {
...
-  if (sla.slaFrtDueAt && shouldCheckFirstResponse) {
-    const threshold = sla.slaFrtDueAt - currentTime;
-    slaStatuses.push({
-      type: 'FRT',
-      threshold,
-      icon: threshold <= 0 ? 'flame' : 'alarm',
-      isSlaMissed: threshold <= 0,
-    });
-  }
+  if (!isCompleted) {
```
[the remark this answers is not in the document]
sony-mathew: Fixed in https://github.com/chatwoot/chatwoot/commit/16e91d07b477a80ace0eebc4ad4b2d4ee926ce52. Completed nonterminal SLAs now evaluate an unrecorded overdue FRT against sla_completed_at, while completion before the FRT deadline remains unmissed. Added focused regression coverage for both cases.

app/javascript/dashboard/components-next/Conversation/ConversationCard/SLACardLabel.vue
sony-mathew: Fixed in https://github.com/chatwoot/chatwoot/commit/9c2d094eeb4f688a635acf51d78fb84e51a6140f. Extracted SLA evaluation, timer setup and cleanup, and refresh handling into useSlaStatus, then switched all three SLACardLabel variants to the shared composable. Added focused lifecycle coverage.

app/javascript/dashboard/components-next/Conversation/Sla/SLACardLabel.vue
sony-mathew: Fixed together with the shared lifecycle extraction in https://github.com/chatwoot/chatwoot/commit/9c2d094eeb4f688a635acf51d78fb84e51a6140f. This SLACardLabel variant now uses useSlaStatus.

app/javascript/dashboard/components/widgets/conversation/components/SLACardLabel.vue
sony-mathew: Fixed together with the shared lifecycle extraction in https://github.com/chatwoot/chatwoot/commit/9c2d094eeb4f688a635acf51d78fb84e51a6140f. This SLACardLabel variant now uses useSlaStatus.

app/javascript/dashboard/components/widgets/conversation/components/SLACardLabel.vue
```diff
@@ -65,7 +80,24 @@ const updateSlaStatus = () => {
...
+};
+
 const createTimer = () => {
+  clearTimer();
+  if (
+    !shouldRefreshSLAStatus({
+      appliedSla: appliedSLA.value,
+      chat: props.chat,
+    })
+  ) {
```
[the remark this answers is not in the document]
sony-mathew: Fixed in https://github.com/chatwoot/chatwoot/commit/9ddeb567f6c3a4c4c63cf79e8d3456a07df9d9f7. The shared composable now watches the reactive conversation and SLA fields used by the calculation, so resolving stops the timer and reopening the same conversation object restarts it immediately. Added a fake-timer regression for the in-place resolve and reopen path.

app/javascript/dashboard/helper/slaHelper.js
```diff
@@ -47,6 +47,22 @@ const toUnixTimestamp = value => {
     : Math.floor(parsedTimestamp / 1000);
 };
 
+const isTerminalSLAStatus = status => ['hit', 'missed'].includes(status);
+
+const isSLACompleted = (sla, conversation) => {
+  return Boolean(
+    sla.slaCompletedAt ||
+      isTerminalSLAStatus(sla.slaStatus) ||
+      conversation.status === 'resolved'
```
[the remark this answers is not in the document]
sony-mathew: Fixed in https://github.com/chatwoot/chatwoot/commit/ecdd25ef611fffe43324d179d13402eebc30a76a. Reopened nonterminal SLAs no longer treat a stale sla_completed_at value as completion or use it to freeze existing miss durations, so the live timer resumes immediately. The composable regression now includes the stale completion value from the production Vuex path, and focused helper coverage verifies active_with_misses continues from current time.

enterprise/app/models/applied_sla.rb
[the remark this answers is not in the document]
sony-mathew: Thanks for the detailed report. We are not planning to run the historical backfill as part of this rollout, and live dashboard convergence is not a requirement for optional maintenance repair. The repair intentionally uses bulk SQL without callbacks or jobs and without touching updated_at, so widening the after_update_commit guard would not affect that path. A connected client may require a reload after such maintenance; no code change is planned for this finding.

sony-mathew:
@iamsivin Following up on items 1 and 2 from https://github.com/chatwoot/chatwoot/pull/15024#pullrequestreview-4755000898:

1. We are not planning to run the historical completed_at backfill as part of this rollout. Legacy rows without a reliable completion timestamp will retain the static Missed fallback, while new resolutions record the completion time.
2. Yes, keeping terminal hit or missed SLAs frozen after a reopen is intentional. Only a nonterminal SLA resumes live evaluation on reopen.

We will evaluate the callback failure semantics from item 3 separately before responding to that point.

sony-mathew:
@iamsivin Good point. The observation is correct: `update_applied_sla_completion` runs after the conversation is committed but before the remaining status events, activity creation, count invalidation, and `conversation.updated` broadcast. If `applied_sla.update!` raises, those later callbacks will not run.

We considered changing this, but the current ordering is intentional. `completed_at` is nullable and the update does not change any fields involved in the `AppliedSla` validation, so a validation failure is not expected for an existing valid record. The base resolved-status handler also performs a database write before the same callbacks.

Most importantly, the completion timestamp must be persisted before `conversation.updated` is dispatched so the client receives a consistent SLA snapshot. We prefer an unexpected database or invariant failure to surface rather than rescuing it and broadcasting incomplete SLA state.

If we want to remove the partial-after-commit failure possibility entirely, the correct follow-up would be to move SLA completion into the same transaction as the conversation status update, rather than swallowing the exception here. No change is planned for this PR.

iamsivin (approved):
LGTM

enterprise/app/models/enterprise/conversation.rb
```diff
@@ -33,6 +33,18 @@ def with_captain_activity_context(reason:, reason_type:)
...
   private
 
+  def handle_resolved_status_change
+    super
+    update_applied_sla_completion
+  end
+
+  def update_applied_sla_completion
+    return unless saved_change_to_status?
+    return if applied_sla.blank? || applied_sla.hit? || applied_sla.missed?
```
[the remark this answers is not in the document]
sony-mathew: Fixed in https://github.com/chatwoot/chatwoot/commit/8fee877b38a8cba01cc5d239fc3156de09cf57b4. Resolution now stamps a missing completed_at even when SLA evaluation has already moved the row to hit or missed, while reopen and re-resolution continue to preserve an existing terminal timestamp. Added a deterministic regression that terminalizes the SLA inside the completion callback; all 14 focused Enterprise conversation examples and focused RuboCop pass.

sony-mathew:
Current behaviour:
* FRT, NRT, and RT miss durations freeze when the conversation is resolved.
* Active unresolved SLA timers continue updating.
* Reopening a nonterminal SLA resumes its timer without resetting deadlines.
* Reopening a terminal SLA preserves its frozen result.
* The resolution/evaluation race is handled, so terminal SLAs still receive a completion timestamp.
* REST, reporting, and realtime payloads expose sla_completed_at.
* Legacy records without a completion timestamp display a static “Missed” state, since historical backfill will not be run.

PR #15035, opened 2026-07-15
fix(whatsapp): restore embedded signup reauthorization banner

pranavrajs:
Show the reauthorization banner whenever an embedded signup inbox has a reauthorization error instead of suppressing it when the manual migration flow is enabled. Reword the manual migration banner and dialog as a recommendation to reconnect with your own Meta app, and switch their styling from amber warning to blue info.


sony-mathew (approved):
LGTM

PR #15039, opened 2026-07-16
refactor: align conversation direct uploads with standard account auth

scmmishra:
Conversation attachment uploads now go through the same authentication that every other account-scoped API endpoint uses. Agents continue to attach files exactly as before, and the upload request is now tied to the agent's dashboard session instead of a separately serialized access token.

Because the upload request is now authenticated, the dashboard proves the agent's session directly instead of passing `currentUser.access_token`. This keeps uploads working alongside the profile access-token changes in #14973, including on accounts where that token is serialized as empty.

## What changed

- `Api::V1::Accounts::Conversations::DirectUploadsController` now runs the standard account auth stack: API access token when the `api_access_token` header is present, dashboard session (devise-token-auth) otherwise, with agent-bot tokens rejected. Previously it inherited `ActiveStorage::DirectUploadsController` directly and did not run any authentication.
- `EnsureCurrentAccountHelper#ensure_current_account` now returns `401` when a request has neither an authenticated user nor a bot resource, instead of continuing. This closes the same gap for any controller that relies on the helper.
- The dashboard direct-upload paths (`useFileUpload.js` and the legacy `fileUploadMixin.js`) now attach the agent's session headers to the upload request via a new `directUploadsHelper.js`, instead of sending `currentUser.access_token`.

## How to test

1. As a logged-in agent, open a conversation and attach a file. Upload should succeed as before, on installs with direct uploads enabled.
2. Confirm attachments still work for an agent on an account whose profile access token is not serialized (e.g. a Cloud plan without `api_and_webhooks`).
3. Send a `POST` to `/api/v1/accounts/:account_id/conversations/:conversation_id/direct_uploads` with no credentials, an empty `api_access_token`, or an invalid token, and confirm it returns `401`.
4. Confirm a valid agent of the account (via API token or session) gets `200`, while an agent of a different account gets `401`.

### DIFF
--- a/app/controllers/api/v1/accounts/conversations/direct_uploads_controller.rb
+++ b/app/controllers/api/v1/accounts/conversations/direct_uploads_controller.rb
@@ -1,5 +1,13 @@
 class Api::V1::Accounts::Conversations::DirectUploadsController < ActiveStorage::DirectUploadsController
+  include DeviseTokenAuth::Concerns::SetUserByToken
+  include RequestExceptionHandler
+  include AccessTokenAuthHelper
   include EnsureCurrentAccountHelper
+
+  around_action :handle_with_exception
+  before_action :authenticate_access_token!, if: :authenticate_by_access_token?
+  before_action :validate_bot_access_token!, if: :authenticate_by_access_token?
+  before_action :authenticate_user!, unless: :authenticate_by_access_token?
   before_action :current_account
   before_action :conversation
 
@@ -10,6 +18,10 @@ class Api::V1::Accounts::Conversations::DirectUploadsController < ActiveStorage:
   end
 
   private
+
+  def authenticate_by_access_token?
+    request.headers[:api_access_token].present? || request.headers[:HTTP_API_ACCESS_TOKEN].present?
+  end
 
   def conversation
     @conversation ||= Current.account.conversations.find_by(display_id: params[:conversation_id])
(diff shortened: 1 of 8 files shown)


scmmishra:
@codex please review

sony-mathew (approved):
LGTM.

Please take a look at this finding.

[P2] API-token uploads still require a CSRF token — [direct_uploads_controller.rb](/chatwoot/app/controllers/api/v1/accounts/conversations/direct_uploads_controller.rb:1)

The controller inherits `ActiveStorage::DirectUploadsController`, whose base controller enables `protect_from_forgery with: :exception`. Unlike normal Chatwoot API controllers, this controller does not skip CSRF verification.

With forgery protection enabled, I reproduced a valid `api_access_token` request returning `422 ActionController::InvalidAuthenticityToken` before the new authentication callbacks ran. The committed request spec passes because the test environment disables forgery protection.

Dashboard uploads remain functional because the Active Storage browser client automatically adds `X-CSRF-Token`, but token-only API clients cannot use the endpoint as described. Conditionally skip `verify_authenticity_token` for `authenticate_by_access_token?` requests and add a production-like regression spec.

Non-blocking follow-up: add a focused test for `directUploadsHelper.js`; current frontend tests do not invoke the XHR delegate or assert the five session headers.

scmmishra:
@sony-mathew fixed

PR #15047, opened 2026-07-17
fix(whatsapp): skip template sync for suspended accounts

muhsin-k:
Suspended accounts no longer participate in scheduled WhatsApp template syncs. This avoids unnecessary external API calls while keeping the existing refresh behavior unchanged for active accounts.


sony-mathew (approved):
LGTM

PR #15048, opened 2026-07-17
test(onboarding): fix inbox channel dialog spec

muhsin-k:
Restores the Inbox Channels dialog’s Facebook-gating test coverage by isolating account feature dependencies that are unrelated to these cases. This keeps the onboarding checks focused on whether Facebook is configured and avoids initializing router and account-store state.


sony-mathew (approved):
LGTM

PR #15060, opened 2026-07-18
fix(attachment): resolve inline audio URL via configured Active Storage route

### LINKED ISSUE #15006: Attachment#inline_audio_url hardcodes rails_storage_redirect_url, ignoring resolve_model_to_route = :rails_storage_proxy
### Describe the bug

`Attachment#inline_audio_url` explicitly calls `rails_storage_redirect_url`, bypassing the `config.active_storage.resolve_model_to_route` setting entirely. When `resolve_model_to_route = :rails_storage_proxy` is configured (required for S3 setups to avoid CORS/ORB failures), audio URLs still point directly to S3 instead of going through the Rails proxy. Chrome blocks the cross-origin range request with `ERR_BLOCKED_BY_ORB` and audio playback fails.

**Environment**: Self-hosted (Docker Compose)

### To Reproduce

1. Set `config.active_storage.resolve_model_to_route = :rails_storage_proxy` (required when using S3 to avoid CORS issues)
2. Configure `ACTIVE_STORAGE_SERVICE=amazon`
3. Receive an audio message from any channel
4. Inspect the audio player URL in DevTools. it points directly to S3, not to the Rails proxy
5. Chrome blocks the range request: `ERR_BLOCKED_BY_ORB`, audio fails to play

### Expected behavior

Audio URLs respect `resolve_model_to_route` and go through the Rails proxy, just like `file_url` does.

### Environment

### Cloud Provider

DigitalOcean

### Platform

Browser

### Operating system

Ubuntu 22.04

### Browser and version

Chrome 136

### Docker (if applicable)

_No response_

### Additional context

The inconsistency is in `app/models/attachment.rb`. `file_url` uses `url_for(file)`, which respects the configured route. `inline_audio_url` does not:

```ruby
# respects config ✓
def file_url
  url_for(file)
end

# hardcoded, ignores config ✗
def inline_audio_url
  rails_storage_redirect_url(file, disposition: 'inline')
end
```
Fix is one line:
```ruby
def inline_audio_url
  return '' unless file.attached?
  url_for(file)
end
```

Also related: `normalize_opus_content_type` in `config/initializers/active_storage.rb` updates the database blob record from `audio/opus` to `audio/ogg` before sending to Meta/WhatsApp, but does not update the S3 object metadata. The Meta API downloads the file directly from S3, receives `Content-Type: audio/opus`, and rejects it with error **131053** ("Unsupported Voice mime type"). The S3 object needs an in-place copy with metadata_directive: 'REPLACE' to fix the Content-Type at rest.

augustocbx:
## Description
`Attachment#inline_audio_url` hardcoded `rails_storage_redirect_url`, so audio playback ignored `config.active_storage.resolve_model_to_route`. On S3 setups configured with `:rails_storage_proxy` (required to avoid CORS/ORB failures) audio URLs still pointed directly at S3, and Chrome blocked the cross-origin range request with `ERR_BLOCKED_BY_ORB`, so playback failed. This switches the method to `url_for(file)` — the same helper `file_url` already uses — so the configured Active Storage route is honoured. Fixes #15006.

### Root cause
`app/models/attachment.rb:118-122` called `Rails.application.routes.url_helpers.rails_storage_redirect_url(file, disposition: 'inline')`, which always builds a redirect URL regardless of `resolve_model_to_route`. `file_url` (`:57`) uses `url_for(file)` and respects the setting — audio was the only inconsistency.

### Solution
Use `url_for(file)` in `inline_audio_url`, mirroring `file_url`.
- Default installs (`:rails_storage_redirect`) are unchanged — they keep redirect URLs, preserving the audio-on-redirect behavior introduced in #14440 (which keeps large recordings off the proxy range-limiter).
- Only installs that explicitly set `:rails_storage_proxy` now serve audio through the proxy — exactly what those admins configured, and consistent with all other media via `file_url`.
- Inline playback is preserved by the `content_types_allowed_inline` allowlist in `config/initializers/active_storage.rb`, so dropping the explicit `disposition: 'inline'` has no effect on playback.
- The `ActiveStorageProxyRangeLimit` mitigation from #14440 is untouched and still applies to any proxy traffic.

Note: the issue also mentions opus S3 object-metadata handling (Meta error 131053). That is a separate, larger change and is intentionally out of scope here.

## Type of change

## How Has This Been Tested?
- Added `spec/models/attachment_spec.rb` example asserting the audio `data_url` honours `resolve_model_to_route` when the proxy route is configured; it fails without the fix (URL was `.../blobs/redirect/...`) and passes after.
- `bundle exec rspec spec/models/attachment_spec.rb` ✅ (27 examples, 0 failures)
- `bundle exec rubocop --parallel` ✅
- Manual QA: N/A — reproducing requires an S3 backend with `resolve_model_to_route = :rails_storage_proxy`; behavior is covered by the spec.

## Checklist
- [x] I have commented my code in hard-to-understand areas (where needed)
- [ ] I have made corresponding changes to the documentation (N/A — product docs are separate)
- [x] Targets the `develop` branch


sony-mathew (approved):
lgtm

sony-mathew:
Thank you for the contribution @augustocbx ! Appreciate it!

PR #15069, opened 2026-07-19
fix: use direct media_id upload for WhatsApp Cloud API attachments (fixes #13540)

### LINKED ISSUE #13540: Media sending via WhatsApp Cloud API fails intermittently with error 131053 (HTTP 429) due to Meta's ASN-based rate limiting on fwdproxy — Chatwoot should support direct media upload via media_id
## Describe the bug

When sending media attachments (images, PDFs, audio) through WhatsApp Cloud API integration, messages fail intermittently with error `131053: Media upload error`. Text messages are delivered correctly 100% of the time. The issue started around **February 6, 2026** and is progressively getting worse, affecting thousands of self-hosted users.

**The root cause is architectural.** Chatwoot currently sends media by providing a URL for Meta's servers to download. Meta's internal download service (fwdproxy) applies rate limiting based on the hosting provider's **ASN (Autonomous System Number)**, not per individual account. When many Chatwoot/Evolution API instances share the same hosting provider (e.g., Hostinger — ASN 47583), the aggregate download volume exceeds Meta's threshold and **all users on that provider are blocked intermittently**.

The exact error from Meta's webhook:

```
Downloading media from weblink failed with http code 429,
status message Too Many Requests,
ratelimit reason: Request ratelimit by fwdproxy.
Details: [GlobalCountingRatelimiter] Request hit ratelimit policy
— destination matcher asn_list: 47583
```

**This is not a temporary issue.** As Chatwoot adoption grows (especially in markets like Brazil where WhatsApp dominates), more providers will hit the same limit. The sustainable fix is to change media upload from URL-based to direct upload via `media_id`, which completely bypasses Meta's fwdproxy.

### Current flow (URL-based — causes the problem)

```
1. User attaches file in Chatwoot
2. Chatwoot uploads file to storage (local/S3/MinIO) ✅
3. Chatwoot sends file URL to WhatsApp API ✅
4. Meta's fwdproxy tries to download file from user's server ❌ RATE LIMITED BY ASN
5. WhatsApp returns error 131053
```

### Proposed flow (direct upload via media_id — no fwdproxy involved)

```
1. User attaches file in Chatwoot
2. Chatwoot uploads file to storage ✅
3. Chatwoot uploads file directly to WhatsApp Media API → receives media_id ✅
4. Chatwoot sends message referencing media_id ✅
5. WhatsApp delivers media to recipient ✅
```

This is how tools like n8n's WhatsApp node handle media — and they are not affected by this issue.

### Evidence

- [[Discussion #13517](https://github.com/orgs/chatwoot/discussions/13517)](https://github.com/orgs/chatwoot/discussions/13517) — Multiple users reporting the same issue
- Hostinger officially confirmed Meta is rate-limiting their ASN
- Meta developer support confirmed the 429 is from fwdproxy's GlobalCountingRatelimiter targeting destination ASN
- Users who migrated to providers with different ASNs report the issue disappears completely
- The WhatsApp Cloud API officially supports both methods: [[URL-based](https://developers.facebook.com/docs/whatsapp/cloud-api/guides/send-messages#media-messages)](https://developers.facebook.com/docs/whatsapp/cloud-api/guides/send-messages#media-messages) and [[Direct upload](https://developers.facebook.com/docs/whatsapp/cloud-api/reference/media#upload-media)](https://developers.facebook.com/docs/whatsapp/cloud-api/reference/media#upload-media)

## To Reproduce

1. Deploy Chatwoot (any recent version) on a hosting provider with high Chatwoot/Evolution API density (e.g., Hostinger — ASN 47583)
2. Configure WhatsApp Cloud API inbox
3. Use any storage backend (local, S3-compatible, MinIO) — the issue occurs with all of them
4. Send a media attachment (image, PDF, or audio) to a WhatsApp contact
5. Observe that text messages are always delivered, but media fails intermittently with:

```
content_attributes: {
  "external_error" => "131053: Media upload error"
}
```

6. Check server logs — the file is uploaded to storage successfully:

```
S3 Storage (36.9ms) Uploaded file to key: 488mr7jp78rvfayn4m2suyagxa4l
```

7. But the Active Storage serves it via redirect:

```
Processing by ActiveStorage::Blobs::RedirectController#show as PDF
Redirected to https://storage.example.com/chatwoot/488mr7jp...?X-Amz-Expires=300&X-Amz-Signature=...
Completed 302 Found in 19ms
```

8. Meta's fwdproxy attempts to download from the redirect URL, hits the ASN rate limit, and returns 429

**Important:** The failure is intermittent because it depends on the aggregate traffic volume from all Chatwoot instances on the same ASN at that moment. It fails more during business hours (high traffic) and works more often at night (low traffic).

**Environment:**
- Chatwoot version: Multiple versions affected (tested on 4.10.1 and others)
- Channel: WhatsApp Cloud API (official)
- Storage: Tested with S3-compatible (MinIO) and local disk — same result
- Docker version: Docker Engine 24.x / 28.x
- OS: Ubuntu 24.04

## Expected behavior

Media attachments should be delivered reliably to WhatsApp contacts regardless of the hosting provider's ASN or aggregate traffic from other Chatwoot instances.

**Proposed implementation:**

1. **Primary:** Chatwoot should upload media files directly to the WhatsApp Cloud API media endpoint (`POST /{phone_number_id}/media`), receive a `media_id`, and send the message using `"document": {"id": "media_id"}` instead of `"document": {"link": "https://..."}`. This completely eliminates the dependency on Meta's fwdproxy and makes media sending immune to ASN-based rate limiting.

2. **Alternative:** If a full migration is not immediately feasible, add a configurable option:

```
WHATSAPP_MEDIA_UPLOAD_STRATEGY=direct  # "direct" for media_id upload, "link" for current behavior
```

This approach would also fix the related issue where Active Storage's 302 redirects cause `131053` errors due to WhatsApp not following redirects properly (returning `Unsupported mime type text/html` instead of the actual file).

### Environment

Docker

### Cloud Provider

### Platform

Browser

### Operating system

Ubuntu 22.04

### Browser and version

Chrome Latest

### Docker (if applicable)

_No response_

### Additional context

_No response_

AdarshJ173:
## Problem

Chatwoot currently sends WhatsApp media attachments via URL (`link: attachment.download_url`), requiring Meta's `fwdproxy` to download the file from the hosting server. `fwdproxy` applies rate limits **per destination ASN**, not per WABA account — so when many Chatwoot/Evolution API instances share a hosting provider (e.g. Hostinger — ASN 47583), the aggregate download volume triggers the global rate limiter and **all users on that ASN receive intermittent `131053: Media upload error`**.

This has been affecting thousands of self-hosted users since at least February 2026, and worsens as Chatwoot adoption grows. The exact Meta error from webhooks:

```
Downloading media from weblink failed with http code 429
ratelimit reason: Request ratelimit by fwdproxy
[GlobalCountingRatelimiter] Request hit ratelimit policy — destination matcher asn_list: 47583
```

Additionally, ActiveStorage's `RedirectController` serves pre-signed S3 URLs via `302 Found`. Meta's fwdproxy **does not follow redirects**, causing a secondary failure path (`Unsupported mime type text/html`) even when ASN limits aren't hit.

Closes #13540

---

## Solution

Upload media **directly to the WhatsApp Cloud API media endpoint** (`POST /{phone_number_id}/media`) before sending the message, then reference the returned `media_id` in the payload with `{ id: media_id }` instead of `{ link: url }`. This completely bypasses fwdproxy — Meta serves the media from their own CDN after upload, with no download from the Chatwoot server.

This is how tools like n8n's WhatsApp node handle media (and why they are not affected by this issue).

---

### New: `app/services/whatsapp/media_upload_service.rb`
Single-responsibility service: takes a channel + attachment, POSTs the file to the WhatsApp media API via multipart upload, and returns the `media_id`. Raises on failure with full error details logged.

### Modified: `app/services/whatsapp/providers/whatsapp_cloud_service.rb`
- `build_attachment_content` now routes through `upload_and_build_media_id_content` when using the `direct` strategy
- Falls back gracefully to `link:` if the upload raises (network error, rate limit on upload endpoint, etc.) — no silent failure
- Adds `direct_upload_strategy?` helper controlled by `WHATSAPP_MEDIA_UPLOAD_STRATEGY` env var

### New: `spec/services/whatsapp/media_upload_service_spec.rb`
Covers: success path, auth header, error 400, error 429, and custom `WHATSAPP_CLOUD_BASE_URL`.

### New: `spec/services/whatsapp/providers/whatsapp_cloud_service_media_upload_spec.rb`
Covers: `direct` strategy uses `media_id`, `link` strategy uses URL, upload failure falls back to link with warning log.

---

## Configuration

| Env var | Default | Description |
|---|---|---|
| `WHATSAPP_MEDIA_UPLOAD_STRATEGY` | `direct` | Set to `link` to revert to URL-based sending |
| `WHATSAPP_CLOUD_BASE_URL` | `https://graph.facebook.com` | Existing var, respected by the upload service |

No database migrations required. No breaking changes for operators — existing deployments get the fix automatically on upgrade. Operators who want to revert temporarily can set `WHATSAPP_MEDIA_UPLOAD_STRATEGY=link`.

---

## Testing

```bash
bundle exec rspec spec/services/whatsapp/media_upload_service_spec.rb
bundle exec rspec spec/services/whatsapp/providers/whatsapp_cloud_service_media_upload_spec.rb
```

---

## References
- Issue: #13540
- Meta docs — Upload Media: https://developers.facebook.com/docs/whatsapp/cloud-api/reference/media#upload-media
- Meta docs — Send Media Messages: https://developers.facebook.com/docs/whatsapp/cloud-api/guides/send-messages#media-messages

### DIFF
--- a/app/services/whatsapp/media_upload_service.rb
+++ b/app/services/whatsapp/media_upload_service.rb
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+
+# Uploads a file attachment directly to the WhatsApp Cloud API media endpoint
+# and returns the resulting media_id.
+#
+# Using media_id instead of a link URL completely bypasses Meta's fwdproxy
+# download path, which is rate-limited per destination ASN. This prevents
+# intermittent 131053 errors for self-hosted instances sharing a hosting
+# provider with other Chatwoot/Evolution API deployments.
+#
+# See: https://developers.facebook.com/docs/whatsapp/cloud-api/reference/media#upload-media
+class Whatsapp::MediaUploadService
+  WHATSAPP_SUPPORTED_TYPES = %w[
+    audio/aac audio/mp4 audio/mpeg audio/amr audio/ogg audio/opus
+    application/pdf
+    image/jpeg image/png image/webp
+    video/mp4 video/3gpp
+    application/vnd.ms-powerpoint application/msword
+    application/vnd.openxmlformats-officedocument.presentationml.presentation
+    application/vnd.openxmlformats-officedocument.wordprocessingml.document
+    application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
+    application/vnd.ms-excel text/plain
+  ].freeze
+
+  def initialize(channel:, attachment:)
+    @channel = channel
+    @attachment = attachment
+  end
+
+  # Returns the media_id string on success, raises on failure.
+  def upload
+    file_content = @attachment.file.download
+    mime_type = @attachment.file.content_type.presence || 'application/octet-stream'
+    filename = @attachment.file.filename.to_s
+
+    response = HTTParty.post(
+      upload_url,
+      headers: auth_header,
+      body: build_multipart_body(file_content, mime_type, filename),
+      multipart: true
+    )
+
+    raise upload_error(response) unless response.success?
+
+    media_id = response.parsed_response['id']
+    raise '[WhatsApp Media Upload] API returned success but no media_id in response' if media_id.blank?
+
+    media_id
+  rescue StandardError => e
+    Rails.logger.error("[WhatsApp Media Upload] Failed for attachment #{@attachment.id}: #{e.message}")
+    raise
+  end
+
+  private
+
+  def upload_url
+    base = ENV.fetch('WHATSAPP_CLOUD_BASE_URL', 'https://graph.facebook.com')
+    phone_number_id = @channel.provider_config['phone_number_id']
+    "#{base}/v18.0/#{phone_number_id}/media"
+  end
+
+  def auth_header
+    { 'Authorization' => "Bearer #{@channel.provider_config['api_key']}" }
+  end
+
+  def build_multipart_body(file_content, mime_type, filename)
+    {
+      messaging_product: 'whatsapp',
+      type: mime_type,
+      file: UploadIO.new(StringIO.new(file_content), mime_type, filename)
+    }
+  end
+
+  def upload_error(response)
+    error_msg = response.parsed_response.is_a?(Hash) ? response.parsed_response.dig('error', 'message') : response.body
+    "[WhatsApp Media Upload] HTTP #{response.code}: #{error_msg}"
+  end
+end
--- a/app/services/whatsapp/providers/whatsapp_cloud_service.rb
+++ b/app/services/whatsapp/providers/whatsapp_cloud_service.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class Whatsapp::Providers::WhatsappCloudService < Whatsapp::Providers::BaseService
   def send_message(phone_number, message)
     @message = message
@@ -16,7 +18,7 @@ class Whatsapp::Providers::WhatsappCloudService < Whatsapp::Providers::BaseServi
 
     request_body = {
       messaging_product: 'whatsapp',
-      recipient_type: 'individual', # Only individual messages supported (not group messages)
+      recipient_type: 'individual',
       to: phone_number,
       type: 'template',
       template: template_body
@@ -32,7 +34,6 @@ class Whatsapp::Providers::WhatsappCloudService < Whatsapp::Providers::BaseServi
   end
 
   def sync_templates
-    # ensuring that channels with wrong provider config wouldn't keep trying to sync templates
     whatsapp_channel.mark_message_templates_updated
     templates = fetch_whatsapp_templates("#{business_account_path}/message_templates?access_token=#{whatsapp_channel.provider_config['api_key']}")
     whatsapp_channel.update(message_templates: templates, message_templates_last_updated: Time.now.utc) if templates.present?
@@ -47,7 +48,6 @@ class Whatsapp::Providers::WhatsappCloudService < Whatsapp::Providers::BaseServi
     end
 
     next_url = next_url(response)
-
     return response['data'] + fetch_whatsapp_templates(next_url) if next_url.present?
 
     response['data']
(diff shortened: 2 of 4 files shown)


nestordavalos:
Hi @AdarshJ173 and everyone, thank you for working on this important issue.

The direct-upload direction is valid and aligns with the Meta Media API: upload the attachment first, receive a `media_id`, and send the message using that ID instead of a remote `link`.

However, the current implementation has a few blockers:

1. The multipart upload is not sending the attachment bytes correctly. With HTTParty 0.24, the current `UploadIO` backed by `StringIO` is not recognized as a file part because it does not expose a filesystem path. The request serializes the Ruby object rather than the file content, the upload fails, and every message falls back to the existing URL flow.

2. The upload endpoint is hardcoded to an old Graph API version. It should follow the application configuration so media upload and message sending remain compatible over time.

3. The upload path loads the complete attachment into memory before sending it to Meta. The direct-upload implementation should handle ActiveStorage blobs in a storage-backend-compatible way and avoid unnecessary memory pressure for larger files.

4. The declared supported MIME type list is unused. The direct-upload path needs validation for media type and size according to Meta requirements, with explicit fallback behavior.

5. The direct-upload flow needs observability. Operators should be able to distinguish a successful `media_id` upload from a URL fallback and see the API error that caused it.

In our implementation, Cloud API attachments already use a Rails proxy URL rather than an ActiveStorage redirect. This addresses storage reachability and redirect-related failures, but it still relies on Meta fetching a remote URL. Therefore, it does not address `fwdproxy` rate limiting when that is the actual cause of error 131053.

The correct direction is to preserve the existing URL path as a controlled fallback while making direct media upload the reliable primary path after multipart handling, API-version consistency, streaming, validation, and monitoring are addressed.

tds-1:
Closing this as #15702, which builds on this branch, has been merged. Thanks for the work here @AdarshJ173 — it set the direction for the fix.

PR #15071, opened 2026-07-19
fix: preserve original filenames for Telegram inbound attachments

### LINKED ISSUE #15070: Bug: Telegram inbound attachments lose their original filename (renamed to Telegram's internal file name)
### Describe the bug

When a Telegram user uploads a file (document/audio/video) to a Telegram inbox, the attachment stored in Chatwoot is **renamed** instead of keeping the original filename the user uploaded.

For example, a user sends `Screenshot 2021-09-27 at 2.01.14 PM.png`, but in the Chatwoot conversation it shows up with a generic name like `file_5.png`.

**Root cause**

In `Telegram::IncomingMessageService#attach_files`, the filename is taken from the downloaded file's `original_filename`, which the `Down` gem derives from the **Telegram file-server download URL** — not from the user's upload:

https://github.com/chatwoot/chatwoot/blob/develop/app/services/telegram/incoming_message_service.rb#L144-L166

```ruby
attachment_file = Down.download(
  inbox.channel.get_telegram_file_path(file[:file_id])
)

@message.attachments.new(
  account_id: @message.account_id,
  file_type: file_content_type,
  file: {
    io: attachment_file,
    filename: attachment_file.original_filename,   # derived from Telegram's internal file path
    content_type: attachment_file.content_type
  }
)
```

`get_telegram_file_path` builds the URL from Telegram's `getFile` response, whose `file_path` is Telegram's internal storage path (e.g. `documents/file_5.png`) — not the name chosen by the user:

https://github.com/chatwoot/chatwoot/blob/develop/app/models/channel/telegram.rb#L55-L60

Meanwhile the original filename **is** available in the webhook payload as `file_name` on the `document` / `audio` / `video` object, but `attach_files` only reads `file[:file_id]` and never `file[:file_name]`. So the user-provided name is discarded.

There is currently no test asserting the resulting filename (the attachment specs only check `file_type`), which is why this has gone unnoticed.

### To Reproduce

1. Connect a Telegram inbox to Chatwoot.
2. From a Telegram client, send a document with a recognizable name, e.g. `Quarterly-Report-2025.pdf`.
3. Open the conversation in Chatwoot and look at the received attachment.
4. The filename is a generic Telegram name (e.g. `file_5.pdf`) instead of `Quarterly-Report-2025.pdf`.

### Expected behavior

The attachment should retain the original filename provided by the sender (`Quarterly-Report-2025.pdf`), matching the behavior of other channels.

**Suggested fix**

Prefer the payload's `file_name` and fall back to the download's name:

```ruby
filename: file[:file_name].presence || attachment_file.original_filename,
```

`file_name` is present for `document`, `audio`, and `video` messages; photos/stickers/voice notes don't include one, so the fallback preserves current behavior there.

### Environment

Docker

### Additional context

Confirmed present on the current `develop` branch (v4.16.0). The relevant code path (`attach_files`) is unchanged from earlier self-hosted versions where this was first observed.

AfiqAqil:
## Description

Telegram inbound attachments were stored under a generic, renamed filename (e.g. `file_5.pdf`) instead of the name the sender actually uploaded (e.g. `Quarterly-Report-2025.pdf`).

**Root cause:** In `Telegram::IncomingMessageService#attach_files`, the attachment filename was taken from the `Down`-downloaded file's `original_filename`, which `Down` derives from the Telegram file-server download URL. That URL is built from Telegram's `getFile` response, whose `file_path` is Telegram's *internal* storage path — not the sender's chosen name. The real name is present in the webhook payload as `file_name` on the `document` / `audio` / `video` object, but it was never read.

Because `Attachment#set_extension` derives the stored `extension` from the filename, this also corrupted the recorded extension for documents whose Telegram path lacked a proper one.

**Fix:** Prefer the payload's `file_name`, falling back to the downloaded name when it's absent (photos, stickers, voice notes — which Telegram sends without a `file_name`). This mirrors how the Line channel (`message['fileName']`) and WhatsApp Cloud channel already handle inbound attachment filenames.

```ruby
filename: file[:file_name].presence || attachment_file.original_filename,
```

Fixes #15070

## Type of change

## How Has This Been Tested?

Added/updated specs in `spec/services/telegram/incoming_message_service_spec.rb`:

- Extended the existing document-message example to assert the stored attachment retains the original payload filename (`Screenshot 2021-09-27 at 2.01.14 PM.png`) rather than Telegram's internal download name.
- Added a fallback example (a `photo` payload, which carries no `file_name`) asserting the attachment is still created and the filename falls back to the downloaded name — guarding the `.presence || …` branch and confirming no regression for media types without a `file_name`.

> Note: my local machine does not have the Ruby/Redis toolchain to run the suite, so I have not run RSpec/RuboCop locally — I'm relying on CI to validate. Suggested command for reviewers: `bundle exec rspec spec/services/telegram/incoming_message_service_spec.rb`.

## Checklist:

- [ ] New and existing unit tests pass locally with my changes (unable to run locally — see note above; relying on CI)

### DIFF
--- a/app/services/telegram/incoming_message_service.rb
+++ b/app/services/telegram/incoming_message_service.rb
@@ -159,7 +159,9 @@ class Telegram::IncomingMessageService
       file_type: file_content_type,
       file: {
         io: attachment_file,
-        filename: attachment_file.original_filename,
+        # Telegram's download URL uses an internal storage path, so Down derives a generic name from it.
+        # Prefer the original filename from the payload (present for document/audio/video) when available.
+        filename: file[:file_name].presence || attachment_file.original_filename,
         content_type: attachment_file.content_type
       }
     )
--- a/app/services/telegram/incoming_message_service.rb
+++ b/app/services/telegram/incoming_message_service.rb
@@ -150,9 +150,7 @@ class Telegram::IncomingMessageService
       return
     end
 
-    attachment_file = Down.download(
-      inbox.channel.get_telegram_file_path(file[:file_id])
-    )
+    attachment_file = Down.download(file_download_path)
 
     @message.attachments.new(
       account_id: @message.account_id,
--- a/spec/services/telegram/incoming_message_service_spec.rb
+++ b/spec/services/telegram/incoming_message_service_spec.rb
@@ -299,7 +299,29 @@ describe Telegram::IncomingMessageService do
         described_class.new(inbox: telegram_channel.inbox, params: params).perform
         expect(telegram_channel.inbox.conversations.count).not_to eq(0)
         expect(Contact.all.first.name).to eq('Sojan Jose')
-        expect(telegram_channel.inbox.messages.first.attachments.first.file_type).to eq('file')
+        attachment = telegram_channel.inbox.messages.first.attachments.first
+        expect(attachment.file_type).to eq('file')
+        # Retain the original filename from the payload instead of Telegram's internal download name
+        expect(attachment.file.filename.to_s).to eq('Screenshot 2021-09-27 at 2.01.14 PM.png')
+      end
+    end
+
+    context 'when attachment params have no file_name' do
+      it 'falls back to the downloaded file name' do
+        allow(telegram_channel.inbox.channel).to receive(:get_telegram_file_path).and_return('https://chatwoot-assets.local/sample.png')
+        params = {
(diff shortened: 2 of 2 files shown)


sojan-official (approved):

sojan-official:
Thanks for fixing this, @AfiqAqil! 🙌

PR #15073, opened 2026-07-19
fix: Enforce authorization in direct uploads controller

### LINKED ISSUE #15072: Unauthenticated ActiveStorage direct-upload (missing authentication) on the conversation direct_uploads endpoint
### Describe the bug

reported on 13 June 2026 - https://github.com/chatwoot/chatwoot/security/advisories/GHSA-5h82-j98m-7r5c

Affected version: commit 2663a84

### Summary

The conversation direct-upload controller subclasses `ActiveStorage::DirectUploadsController` directly instead of the authenticated API base, so `authenticate_user!` is never in its ancestry, and the only remaining gate short-circuits for unauthenticated callers. An unauthenticated network attacker can therefore create ActiveStorage blobs in any tenant account and obtain signed PUT upload URLs to write arbitrary bytes to the app's storage backend. Confirmed against the controller/helper source with `current_user => nil`: the request reached the ActiveStorage upload sink and returned a signed upload URL.

### Details

Sink: `app/controllers/api/v1/accounts/conversations/direct_uploads_controller.rb:9` calls `super` -> `ActiveStorage::DirectUploadsController#create` (Rails 7.1.5.2), which creates an `ActiveStorage::Blob` row and returns a signed, single-use PUT upload URL. Route: `POST /api/v1/accounts/:account_id/conversations/:conversation_id/direct_uploads` (`config/routes.rb:153`).

Root cause: the controller subclasses `ActiveStorage::DirectUploadsController`, not `Api::V1::Accounts::BaseController`. `authenticate_user!` is registered only in `Api::BaseController` (`app/controllers/api/base_controller.rb`), which is never in this controller's ancestry. The sole gate is `EnsureCurrentAccountHelper#ensure_current_account`, whose membership check is wrapped in `if current_user` (`ensure_current_account_helper.rb:13-15`). For an unauthenticated caller `current_user` is nil, so the check is skipped and the account is returned.

Flow: `current_account` does `Account.find(account_id)` (any active account); because `current_user` is nil, `account_accessible_for_user?` is never called, so no membership is enforced. `conversation` does `Current.account.conversations.find_by(display_id: conversation_id)` where `display_id` is a small per-account sequential integer, trivially guessable. `create` early-returns only if conversation/account is nil; otherwise `super` creates the blob and returns the signed URL. CSRF is globally skipped in ApplicationController.

The boundary crossed is an unauthenticated network attacker reaching the authenticated-only storage subsystem of any tenant. The sibling widget endpoint (`Api::V1::Widget::DirectUploadsController`) correctly gates on `WebsiteTokenHelper`; the account endpoint is the outlier that dropped auth.

### PoC

```
POST /api/v1/accounts/1/conversations/1/direct_uploads        (no api_access_token, no Devise token)
{ "blob": { "filename":"x.bin","byte_size":10,"checksum":"...","content_type":"application/octet-stream" } }
-> 200 with a signed direct_upload.url (PUT) into the app storage backend
```

Validated by installing the repo's Rails (`actionpack 7.1.5.2`), loading the verbatim repo files (`ensure_current_account_helper.rb` + the ral controller logic), stubbing only external models, forcing `current_user => nil`, and driving the ActionController callback machinery via Rack::Test:

```
HTTP status: 200
Body: SINK REACHED: blob created + signed upload URL returned
Current.account set? true
```

The membership check was wired to raise if invoked; it never fired, proving the `if current_user` guard short-circuits all authorization for unauthenticated callers and the request reaches the ActiveStorage upload sink.

### Impact

Unauthenticated creation of storage blobs in any account and issuance of signed PUT URLs that allow arbitrary bytes to be written to the configured storage backend (local disk / S3 / GCS). Enables storage abuse and resource exhaustion, staging of attacker-controlled files on the application's storage domain, and orphan-blob DB pollution, all pre-auth and cross-tenant. Impact rises where a deployment serves the storage origin as trusted content.

### Remediation

Add authentication to the controller, e.g. include the authenticated API base or explicitly enforce auth and conversation authorization:

```ruby
class Api::V1::Accounts::Conversations::DirectUploadsController < ActiveStorage::DirectUploadsController
  include AccessTokenAuthHelper
  before_action :authenticate_user!          # or authenticate_access_token! for token clients
  # keep current_account + conversation, and require the conversation be accessible to current_user
  # via ConversationPolicy#show? / PermissionFilterService
end
```

so an authenticated agent can only upload to conversations in inboxes/teams they can access.

### Expected behavior

_No response_

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

None

### Operating system

_No response_

### Browser and version

_No response_

### Docker (if applicable)

_No response_

### Additional context

_No response_

Mister2005:
Users were able to attach files to any conversations if they have a valid conversation_id due to a missing pundit authorization check. This commit includes Pundit::Authorization and explicitly authorizes the conversation before proceeding. It also implements pundit_user to provide the proper authorization context for the controller. Fixes #15072

### DIFF
--- a/app/controllers/api/v1/accounts/conversations/direct_uploads_controller.rb
+++ b/app/controllers/api/v1/accounts/conversations/direct_uploads_controller.rb
@@ -3,6 +3,7 @@ class Api::V1::Accounts::Conversations::DirectUploadsController < ActiveStorage:
   include RequestExceptionHandler
   include AccessTokenAuthHelper
   include EnsureCurrentAccountHelper
+  include Pundit::Authorization
 
   skip_before_action :verify_authenticity_token, if: :authenticate_by_access_token?
 
@@ -22,6 +23,14 @@ class Api::V1::Accounts::Conversations::DirectUploadsController < ActiveStorage:
 
   private
 
+  def pundit_user
+    {
+      user: Current.user,
+      account: Current.account,
+      account_user: Current.account_user
+    }
+  end
+
   def authenticate_by_access_token?
     request.headers[:api_access_token].present? || request.headers[:HTTP_API_ACCESS_TOKEN].present?
   end
@@ -33,6 +42,7 @@ class Api::V1::Accounts::Conversations::DirectUploadsController < ActiveStorage:
   end
 
   def conversation
-    @conversation ||= Current.account.conversations.find_by(display_id: params[:conversation_id])
+    @conversation ||= Current.account.conversations.find_by!(display_id: params[:conversation_id])
+    authorize @conversation, :show?
   end
 end
--- a/app/controllers/api/v1/accounts/conversations/direct_uploads_controller.rb
+++ b/app/controllers/api/v1/accounts/conversations/direct_uploads_controller.rb
@@ -11,6 +11,7 @@ class Api::V1::Accounts::Conversations::DirectUploadsController < ActiveStorage:
   before_action :authenticate_access_token!, if: :authenticate_by_access_token?
   before_action :validate_bot_access_token!, if: :authenticate_by_access_token?
   before_action :authenticate_user!, unless: :authenticate_by_access_token?
+  before_action :set_current_user, unless: :authenticate_by_access_token?
   before_action :current_account
   before_action :validate_token_api_access, if: :authenticate_by_access_token?
   before_action :conversation
@@ -22,6 +23,11 @@ class Api::V1::Accounts::Conversations::DirectUploadsController < ActiveStorage:
   end
 
   private
+
+  def set_current_user
+    @user ||= current_user
+    Current.user = @user
+  end
 
   def pundit_user
     {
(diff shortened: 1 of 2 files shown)


scmmishra:
Thanks for working on this. The unauthenticated direct-upload issue reported in [#15072](https://github.com/chatwoot/chatwoot/issues/15072) was already addressed by [#15039](https://github.com/chatwoot/chatwoot/pull/15039), which added authentication and account-membership checks to this endpoint.

This PR adds conversation-level authorization, but direct uploads only create an Active Storage blob, they do not attach it to the specified conversation. The later message creation endpoint already authorizes access to the conversation. Since the blob is not bound to the conversation ID used during upload, this change does not provide meaningful additional protection for the reported vulnerability.

Closing this as superseded by [#15039](https://github.com/chatwoot/chatwoot/pull/15039). Thanks again for the contribution!

PR #15081, opened 2026-07-20
fix: rate limit agent management APIs

sony-mathew:
# Pull Request Template

## Description

Limits agent create and delete requests to 30 per minute per authenticated user or API token, with an IP fallback when no user identity is available. This reduces the rate at which a compromised administrator can cycle agent invitations while keeping allowances independent across users.

Related to [CW-7637](https://linear.app/chatwoot/issue/CW-7637/prevent-agent-invitation-email-abuse-after-july-20-incident).

## Type of change

## How Has This Been Tested?

Verified that each endpoint allows requests 1 through 30 and returns HTTP 429 for request 31. Also verified user, API-token, and IP-fallback keys, including independent allowances for different users.

## Checklist:

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -223,7 +223,7 @@ class Rack::Attack
 
   ## Prevent abuse of agent delete API (per account)
   throttle('/api/v1/accounts/:account_id/agents/:id DELETE',
-           limit: ENV.fetch('RATE_LIMIT_AGENT_DELETE', '100').to_i, period: 1.day) do |req|
+           limit: ENV.fetch('RATE_LIMIT_AGENT_DELETE', '50').to_i, period: 1.day) do |req|
     next unless req.delete?
 
     match_data = %r{\A/api/v1/accounts/(?<account_id>\d+)/agents/(?<id>\d+)/?\z}.match(req.path_without_extensions)
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -36,6 +36,13 @@ class Rack::Attack
     def path_without_extensions
       path[/^[^.]+/]
     end
+
+    def user_or_ip_identifier
+      user_uid = get_header('HTTP_UID')
+      api_access_token = get_header('HTTP_API_ACCESS_TOKEN') || get_header('api_access_token')
+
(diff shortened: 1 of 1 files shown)


vishnu-narayanan:
Pushed 5322878 with a few adjustments, happy to discuss or revert any of it:

- **Throttle key is now the account_id from the path** instead of uid/api_access_token. Rack::Attack runs before authentication, so those headers are client-controlled at throttle time: with a valid api_access_token, a caller could send a different uid on every request and the counter would never accumulate. The path-based key cannot be rotated without switching target accounts, and it matches the existing account-level reports throttle in the same file. Tradeoff: the budget is shared across all admins of an account, which felt acceptable for an action this rare, but flag it if you see a bulk-onboarding flow this would hurt.
- **The create regex now also matches `agents/bulk_create`**, since one request there fans out to an invitation email per agent in the payload and it was outside the previous pattern.
- **Lowered the create default to 10/min** (delete stays at 30). During the Julu incident the invite volume ran at roughly 48/min, so 30/min per identity would have allowed much the same storm to continue. Both remain ENV-overridable.

The unused `user_or_ip_identifier` helper came out with the key change. Was there a case you had in mind where per-user keying matters more than per-account? If so we could layer both.

vishnu-narayanan:
Follow-up in fc37b1b: switched the create limit from 10/min to **100/day per account**. The abuse we saw ran deliberately slow (one invite every ~6 seconds for two days), so any per-minute rate only stretches it out, while a daily budget caps the total email volume an account can generate. There is no practical case for onboarding more than 100 agents in a day, and it stays overridable via RATE_LIMIT_AGENT_CREATE if one shows up. Delete remains 30/min since it does not send email.

vishnu-narayanan:
One more in d1de40b: delete is now also **100/day per account** (was 30/min). Same reasoning as create, deletion is what frees seats for the re-invite loop, and the abuse deleted at a drip rate per-minute limits never catch. We sized both limits from production data: the busiest legitimate day of agent creation in the last 90 days was 55, the largest team is 206 agents, and only one account exceeds 100 agents total. 100/day clears all of that with headroom, and both remain ENV-overridable.

vishnu-narayanan (approved):

PR #15083, opened 2026-07-20
fix: automation conditions silently fail for NULL/empty fields (priority=none, unassigned, etc.)

### LINKED ISSUE #12797: Issues with automation involving unfilled fields
### Describe the bug

 The automation does not work when we monitor fields that do not have initial values, such as the priority field, email, or even labels. It is not possible, for example, to create automations for leads where the label list does not contain a specific label. The automation does not run.

### To Reproduce

1. Self-hosted environment
2. Create an automation
3. Select the "message created" event
4. Add a condition and select the message type "incoming"
5. Add a condition and select "priority", "equal to", and finally "none"
6. In actions, select add a "label" and add any label
7. Receive a message from any lead without priority

Note: This happens with any field that is empty. It seems that the condition does not work.

### Expected behavior

The expected behavior in this example would be to add the label to a lead who contacted without having any priority defined.

### Environment

### Cloud Provider

AWS

### Platform

None

### Operating system

Linux

### Browser and version

_No response_

### Docker (if applicable)

_No response_

### Additional context

I'm using WhatsApp Channel

AdarshJ173:
## Problem

Closes #12797

Automation rules with conditions on fields that have no value (e.g. `priority = none`, `assignee = unassigned`) **never trigger**, even when the conversation perfectly matches.

### Root cause

When the filter operator is `equal_to` or `not_equal_to`, the service calls `filter_values(query_hash)` which maps user-facing values to DB integers/enums. For `priority = none`, this resolves to `[nil]` because `Conversation.priorities['none'.to_sym]` returns `nil` (no-priority is stored as `NULL` in the DB).

That `nil` array gets bound into the SQL as:
```sql
conversations.priority IN (NULL)
```

**`IN (NULL)` is always `FALSE` in SQL** — `NULL = NULL` evaluates to `UNKNOWN`, not `TRUE`. The correct form is `IS NULL`.

This silently drops every automation that targets unfilled fields, with no error logged to the user.

### Affected fields

| Field | DB NULL value | User-facing label |
|---|---|---|
| `priority` | `nil` | "None" |
| `assignee_id` | `nil` | Unassigned |
| `team_id` | `nil` | No team |
| Any standard nullable column | `nil` | Empty/unset |

## Fix

Added two private helpers to `AutomationRules::ConditionsFilterService`:

- **`null_value_condition?(query_hash)`** — returns `true` when `filter_values` resolves to an all-nil array, indicating the user is targeting a NULL DB column.
- **`build_null_condition_string(table_name, attribute_key, query_hash, query_operator)`** — emits `IS NULL` for `equal_to` and `IS NOT NULL` for `not_equal_to`.

Both `conversation_query_string` and `contact_query_string` now check `null_value_condition?` before delegating to `filter_operation`, routing to `build_null_condition_string` when needed.

The existing `filter_operation` / `IN()` path is **completely unchanged** for all non-null values — no regression risk on the happy path.

## Tests added

`spec/services/automation_rules/conditions_filter_service_null_spec.rb`

- `priority = none` matches conversation with `NULL` priority ✅
- `priority = none` does **not** match conversation with a real priority ✅
- `priority != none` matches conversation with any priority set ✅
- `priority != none` does **not** match conversation with `NULL` priority ✅
- Non-null path (`priority = medium`) still works correctly ✅
- Compound `AND` condition: `priority = none AND status = open` ✅

## How to test manually

1. Create a Chatwoot automation: event `message_created`, condition `priority = none`, action `add label: needs-triage`
2. Start a new conversation via WhatsApp/any channel without setting priority
3. **Before fix**: no label is added
4. **After fix**: `needs-triage` label is applied correctly

### DIFF
--- a/app/services/automation_rules/conditions_filter_service.rb
+++ b/app/services/automation_rules/conditions_filter_service.rb
@@ -113,6 +113,7 @@ class AutomationRules::ConditionsFilterService < FilterService
     query_operator = query_hash['query_operator']
 
     attribute_key = 'processed_message_content' if attribute_key == 'content'
+    attribute_key = 'private' if attribute_key == 'private_note'
 
     filter_operator_value = filter_operation(query_hash, current_index)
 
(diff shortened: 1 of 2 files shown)


aakashb95:
Thanks for working on this. Before we merge, could you address two points?
1. The 3.x UI saves Priority "None" as "nil", not "none". The 3.x filter service also does not convert names such as "medium" to the stored priority value. I prepared the correction in commit [60151ddea0](https://github.com/chatwoot/chatwoot/commit/60151ddea0d2be2c54ce39f8b1f6315815cb5e69). Please cherry-pick it.
2. If the label changes remain in this PR, please reuse the existing tag_filter_query or use an equivalent NOT EXISTS query. The current tags.name != check can incorrectly match a conversation that has both the excluded label and another label.

PR #15084, opened 2026-07-20
feat(conversation): add label search to right-click context menu

NickThePigeon:
When you right-click a conversation in the list and open **Assign label**, the submenu now has a search box at the top. Type to fuzzy-filter your labels instead of scrolling through the full list. This makes labelling fast on accounts that have many labels.

The other label pickers (the conversation sidebar and bulk actions) already have search — this brings the right-click context menu in line with them.

> Supersedes #15080 (that PR could not be reopened after a force-update of the branch).

### How to test
1. Create/have an account with a good number of labels (10+ makes the difference obvious).
2. In the conversation list, right-click any conversation card.
3. Hover **Assign label** to open the submenu.
4. Type in the new search box at the top — the label list fuzzy-filters as you type.
5. Assigned labels sort to the top so they're easy to find; clicking a result assigns/removes it as before; a "No labels found" message shows when nothing matches.

### What changed
- Added a search input to the label submenu in `contextMenu/Index.vue`, filtering with `picoSearch` (the same `@scmmishra/pico-search` helper the sidebar `LabelDropdown` already uses).
- Sort assigned labels first, then alphabetical within each group.
- `keydown`/`click` are stopped on the input so typing doesn't trigger conversation-list keyboard shortcuts.
- Fixed `ContextMenu.vue` closing on `focusout`: it now stays open when focus moves to an element inside it (the search input), and only closes when focus leaves the menu entirely. Previously clicking the search box dismissed the whole menu.
- Styling notes: the sticky search bar uses `-top-1` so, when the list is scrolled, the opaque bar also covers the submenu's `p-1` top padding and label rows don't peek above it. The input uses `reset-base` to opt out of the global `input[type]` styling (`field-base h-10`), which would otherwise force a 40px height and a 1rem bottom margin.
- Added `SEARCH_LABELS` and `NO_LABELS_FOUND` strings to `en.json`.


iamsivin:
@codex please review?

iamsivin (approved):
@NickThePigeon Thanks for the great work on this PR! I made a few small tweaks, and everything looks good to me. Approved

PR #15100, opened 2026-07-21
feat: monitor whatsapp phone number health

muhsin-k:
Adds scheduled WhatsApp Cloud API phone-number health synchronization and expands the Account Health page with phone, capacity, business-account, webhook, coexistence, and recovery details. Authorization failures now guide administrators to the appropriate Configuration flow without exposing technical Meta error codes.

Fixes [CW-7621](https://linear.app/chatwoot/issue/CW-7621/store-whatsapp-phone-number-health-status) 

**Preview**

CleanShot 2026-07-22 at 10 55 01@2x

CleanShot 2026-07-22 at 10 55 08@2x

CleanShot 2026-07-22 at 11 10 05@2x

### How to test

1. Open **Settings → Inboxes → a WhatsApp Cloud API inbox → Account Health**.
2. Confirm the page shows separate Phone number, Health and capacity, Business account, and Webhook configuration sections.
3. Confirm the configured webhook URL can be copied and the expected URL appears only when it differs.
4. For a coexistence number, confirm **Coexistence · Active** appears; confirm it is hidden for standard Cloud API numbers.
5. With an invalid Embedded Signup token, confirm the page asks to refresh the WhatsApp connection and **Go to Configuration** opens the Configuration tab.
6. With an invalid manually configured token, confirm the page asks the administrator to verify or replace the access token.
7. Confirm authorization states do not display Meta error codes or the manual-migration recommendation.

### What changed

- Persists the latest successful phone health snapshot, check time, and most recent error while retaining the last successful data after a failed refresh.
- Refreshes stale active Cloud API channels every six hours through low-priority jobs.
- Fetches phone, WABA, business portfolio, webhook, and coexistence details.
- Uses Meta's current `whatsapp_business_manager_messaging_limit` field while preserving the existing UI response key.
- Classifies authorization failures for setup-specific recovery guidance and logs new risky quality/status transitions.
- Adds focused service, scheduler, job, trigger, and API coverage.

Internal alerts, throttling, automatic inbox disablement, and customer notifications remain outside this PR and are tracked separately in CW-7622.


muhsin-k:
@codex please review

sony-mathew (commented):
Please take a look at these findings.

## Findings

1. **[P2] Prevent older health requests from overwriting newer results**

   [health_service.rb:55](/chatwoot/app/services/whatsapp/health_service.rb:55) fetches externally and then unconditionally updates the channel at lines 174–194. Overlapping requests can finish out of order, allowing an older snapshot or error to replace newer state while receiving the newest `checked_at`.

   I reproduced this: a newer request persisted `RED`, then an older blocked request completed and changed the final stored quality to `GREEN`.

   Add a request-generation/freshness guard covering both success and error persistence. Holding a database lock across Meta requests would be less desirable.

2. **[P2] Deduplicate scheduled health jobs before children execute**

   [health_sync_scheduler_job.rb:4](/chatwoot/app/jobs/channels/whatsapp/health_sync_scheduler_job.rb:4) selects stale rows without claiming them. Because [TriggerScheduledItemsJob](/chatwoot/app/jobs/trigger_scheduled_items_job.rb:23) runs every five minutes and health jobs use the low queue, subsequent scheduler executions can enqueue the same channels while earlier jobs are waiting.

   I reproduced nine child jobs by invoking the scheduler three times for three stale channels. In production this can repeat up to 25 channels per scheduler run, with two Meta requests per job, while also creating the write race above.

   Use a scheduler overlap lock or a per-channel enqueue lease/uniqueness mechanism. Updating `checked_at` at enqueue time would conflate “queued” with “successfully checked” and could hide lost jobs.

3. **[P2] Bind frontend responses to their initiating inbox**

   [Settings.vue:576](/chatwoot/app/javascript/dashboard/routes/dashboard/settings/inbox/Settings.vue:576) writes every completed request into shared `healthData`, `healthError`, and loading state. The inbox watcher can start another request before the previous one completes.

   Switching from inbox A to B can therefore let A’s late response overwrite B’s health view, including phone, WABA, portfolio, and webhook details.

   Capture the requested inbox ID or a generation token and ignore responses/finalizers that are no longer current.

app/jobs/trigger_scheduled_items_job.rb
```diff
@@ -19,6 +19,9 @@ def perform
 
     # Job to sync whatsapp templates
     Channels::Whatsapp::TemplatesSyncSchedulerJob.perform_later
+
+    # Job to sync whatsapp phone number health
+    Channels::Whatsapp::HealthSyncSchedulerJob.perform_later
```
sony-mathew: Do we need to do this every 5 minutes? Can we move this to 30 minutes or hourly please? Anyway we are doing 6 hour check in the scheduler -> so this cadence of 5 minutes seems very aggressive and un-necessary.

app/services/whatsapp/health_service.rb
```diff
@@ -18,30 +72,30 @@ def validate_channel!
...
+        fields: fields,
         access_token: @access_token
       }
     )
 
     handle_response(response)
   rescue StandardError => e
     Rails.logger.error "[WHATSAPP HEALTH] Error fetching health data: #{e.message}"
-    raise e
+    raise
```
sony-mathew, line 92: Why not "raise e"?

app/services/whatsapp/health_service.rb
```diff
@@ -92,4 +170,46 @@ def build_expected_webhook_url
 
     "#{frontend_url}/webhooks/whatsapp/#{@channel.phone_number}"
   end
+
+  def persist_health_status(health_status, checked_at)
+    # Health polling should not touch the inbox or run channel credential validation.
+    # rubocop:disable Rails/SkipsModelValidations
```
sony-mathew, line 178: Why are we skipping rubocop validations here? Can you please check if we can do this better? 

And why are we updating the columns using update_coumns and why note `model.update`? What is preventing us from doing this?

sony-mathew: There are ways we can skip validation for subset of columns right?

sony-mathew: Let me know what you think.

sony-mathew (commented):
Added some minor comments.

app/services/whatsapp/health_service.rb
muhsin-k: Addressed in e66338260c. I kept update_columns intentionally and clarified the code comments. Channel::Whatsapp#validate_provider_config performs a provider credential check, so update or update! would make another external request and could prevent us from recording the failure. Using save!(validate: false) would still run update callbacks, including Channelable and Enterprise audit behavior. These polling fields need a narrow callback-free write, so the RuboCop exception remains scoped to the two persistence calls.

app/services/whatsapp/health_service.rb
muhsin-k: I kept the bare raise intentionally. Inside a rescue block it re-raises the current exception with its existing context; passing the rescued exception explicitly is redundant here. This is also the idiomatic Ruby form.

app/jobs/trigger_scheduled_items_job.rb
muhsin-k: Addressed in e66338260c. The health scheduler is now triggered by Internal::TriggerHourlyScheduledItemsJob instead of the five-minute scheduler. Specs cover both the hourly enqueue and the absence of health scheduling from TriggerScheduledItemsJob.

app/services/whatsapp/health_service.rb
```diff
@@ -18,30 +72,30 @@ def validate_channel!
...
+    raise
   end
 
-  def health_fields
+  def phone_health_fields
     %w[
       id
       quality_rating
-      messaging_limit_tier
+      whatsapp_business_manager_messaging_limit
```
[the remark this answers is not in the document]
muhsin-k: Addressed in 63bc8fe003. Whatsapp::HealthService now enforces v24.0 as the minimum only for Account Health requests, while preserving any newer configured WHATSAPP_API_VERSION such as v25.0. The global WhatsApp API version and other WhatsApp flows remain unchanged. Added regression coverage for both the v24.0 floor and newer-version passthrough.

sony-mathew:
@muhsin-k got this findings. Please check if they are valid or not!

## Findings

1. **[P2, blocking] The scheduler cannot guarantee the required six-hour cadence above 150 channels.**

   The scheduler runs once per hour and selects at most 25 channels: [health_sync_scheduler_job.rb:10](/chatwoot/app/jobs/channels/whatsapp/health_sync_scheduler_job.rb:10), [schedule.yml:19](/chatwoot/config/schedule.yml:19), [limits.rb:3](/chatwoot/lib/limits.rb:3).

   Therefore, one installation can process only `25 × 6 = 150` channels per six-hour window. At 151 eligible channels, at least one channel necessarily exceeds the promised cadence; queue latency lowers the practical threshold further. The batch size or cadence must scale with the eligible population, or all stale channels should be claimed/enqueued with separate rate limiting.

2. **[P2, blocking] An older health check can overwrite a newer persisted snapshot.**

   `checked_at` is assigned only after both Graph requests complete, then success and error states are written unconditionally: [health_service.rb:60](/chatwoot/app/services/whatsapp/health_service.rb:60), [health_service.rb:180](/chatwoot/app/services/whatsapp/health_service.rb:180).

   I reproduced this with two service instances against the isolated test database:

   - Older request started and was delayed.
   - Newer request persisted `GREEN`.
   - Older request completed afterward and replaced it with `YELLOW`, carrying a later `checked_at`.
   - Probe result: `stale_overwrite: true`.

   This is introduced by the PR because `develop` has no persisted health-sync path. Use an attempt generation or start timestamp captured before external calls, and conditionally persist success/error only when that attempt has not been superseded. Do not hold a database lock across Meta requests.

3. **[P3, non-blocking] Messaging-capacity changes are logged as new risky transitions.**

   [health_service.rb:217](/chatwoot/app/services/whatsapp/health_service.rb:217) includes `messaging_limit_tier` in the risk signature. If quality remains `YELLOW` while only the tier changes, another risky-transition warning is emitted, contrary to CW-7621’s quality/status transition scope. The tier can remain in log context but should not determine whether a new risky transition occurred.

The unresolved Ito frontend stale-request thread is real but not a PR regression: the same unguarded `fetchHealthData` assignment and callers exist at the base commit. It should be handled separately.

sony-mathew (approved):
lgtm

PR #15101, opened 2026-07-21
fix(whatsapp): guide inbox recovery for invalid tokens

muhsin-k:
After #15100 added WhatsApp phone-number health monitoring, this PR adds only the authorization-recovery behavior. OAuth authorization failures detected by Account Health or template synchronization now enter Chatwoot’s existing reauthorization flow. Embedded Signup and Coexistence inboxes expose the reconfiguration action, while manually configured Cloud API inboxes receive access-token guidance. Administrators are notified once per authorization-state transition.

### What changed

- Connect Account Health authorization failures to `authorization_error!`.
- Detect the same authorization failure during WhatsApp template synchronization.
- Keep the reconfiguration action available for affected Embedded Signup and Coexistence inboxes, independent of the rollout flag.
- Hide the unrelated manual-migration action while reauthorization is required.
- Send source-specific recovery email guidance linked directly to inbox Configuration.
- Make reauthorization notifications transition-only.

### Things to know

- Rebased directly onto the merged #15100 commit; this PR now contains one focused commit.
- Account Health layout, data retrieval, persistence, polling, and API response behavior remain in #15100.
- The transition-only notification behavior applies to every model using the existing `Reauthorizable` concern.
- The template pagination helper introduced by #15100 is preserved.

### How to test

1. Trigger a Meta authorization failure from Account Health or template synchronization twice.
2. Confirm the inbox enters reauthorization after the existing threshold and administrators receive only one email while it remains in that state.
3. For Embedded Signup or Coexistence, confirm **Reconfigure inbox** is available and the manual-migration action is hidden.
4. For a manually configured Cloud API inbox, confirm the email asks the administrator to update the access token.
5. Confirm the email action opens the inbox Configuration page.

### Validation

- Focused Rails coverage: 150 examples, 0 failures.
- WhatsApp provider coverage after conflict resolution: 23 examples, 0 failures.
- Configuration UI coverage: 2 tests passed.
- Targeted RuboCop and ESLint: no offenses or errors.
- Git merge simulation against current `develop`: clean.

Related: https://github.com/chatwoot/chatwoot/pull/15100


sony-mathew (commented):
Please check the following findings.

## Findings

1. **[P1] An expired template-management token can disable a healthy inbox.**  
   On Chatwoot Cloud, template sync prefers the separate business-management token. The new hook at [whatsapp_cloud_service.rb](/worktrees/6f55/chatwoot/app/services/whatsapp/providers/whatsapp_cloud_service.rb:49) treats its code-190 failure as an inbox credential failure. After two failures, incoming webhooks are suppressed at [whatsapp_events_job.rb](/worktrees/6f55/chatwoot/app/jobs/webhooks/whatsapp_events_job.rb:130). Reconfiguring the inbox does not repair this because same-WABA recovery retains the failed token at [reauthorization_service.rb](/worktrees/6f55/chatwoot/app/services/whatsapp/reauthorization_service.rb:32). Template-token failures should instead direct admins to replace that token and must not globally disconnect the inbox.

2. **[P2] Transition-only notifications are not atomic.**  
   [reauthorizable.rb](/worktrees/6f55/chatwoot/app/models/concerns/reauthorizable.rb:40) performs a Redis read followed by an unconditional write. Concurrent health and template workers can both observe the old state and send mail/events. A deterministic two-worker probe produced `transition_side_effects=2`. Use an atomic claim such as `SET ... NX` and add concurrent coverage.

3. **[P2] Cloud recovery CTA can call an API that rejects it.**  
   [Settings.vue](/worktrees/6f55/chatwoot/app/javascript/dashboard/routes/dashboard/settings/inbox/Settings.vue:400) now shows recovery whenever the Redis state is set, independent of the rollout flag, but [authorizations_controller.rb](/worktrees/6f55/chatwoot/app/controllers/api/v1/accounts/whatsapp/authorizations_controller.rb:22) still rejects Cloud accounts without `whatsapp_embedded_signup_inbox_creation`. The PR’s stated flag-independent recovery path therefore fails. The backend should exempt valid required-state reauthorization without ungating new inbox creation.

4. **[P2] Healthy self-hosted inboxes lose their supported reconfigure action.**  
   [ConfigurationPage.vue](/worktrees/6f55/chatwoot/app/javascript/dashboard/routes/dashboard/settings/inbox/settingsPage/ConfigurationPage.vue:65) no longer checks the self-hosted `whatsapp_reconfigure` flag. The backend still authorizes that path, so healthy self-hosted inboxes with the flag enabled lose their button. The former Cloud/self-host flag split should be preserved.

The last two findings were already present in unresolved review threads and remain valid on the resolved head.

PR #15105, opened 2026-07-21
feat: improve captain overview loading and reuse stats for summary [CW-7610]

scmmishra:
The Captain overview page now loads more efficiently and handles slow stats gracefully. The AI summary reuses the stats already fetched for the metric cards instead of recomputing them on the server, halving the expensive reporting queries per page view, and is now written in the account's configured language. While stats are loading, metric cards show a skeleton state instead of empty values, transient fetch failures recover with a silent retry, and switching the date range or assistant mid-load cancels the superseded request so slow responses can never overwrite the ones you asked for.

## How to test

1. Open Captain → an assistant → Overview.
2. While the page loads, the six metric cards should show pulsing skeletons (not `—` or stale numbers); once loaded, the AI summary appears without an extra delay for stats recomputation.
3. Switch the range (e.g. This month → Last 90 days): cards return to the skeleton state immediately, and the network tab shows the previous `stats` request being cancelled.
4. Quickly flip between ranges a few times. The metric numbers always match the last selected range, and the AI summary never shows a previous range's text next to the latest metrics.
5. Simulate a flaky network (devtools request blocking on the first `stats` call): the page should recover via a single silent retry with no error toast; if both attempts fail, cards fall back to `—`.
6. Set the account language to something other than English (Settings → Account Settings → Site Language) and reload the overview: the AI summary should be written in that language, with the highlighted figures still bolded.

## What changed

- `summary` endpoint accepts the overview stats from the client (whitelisted via strong params) and no longer runs `Captain::AssistantStatsBuilder`; it only resolves the date window for the period label. `WelcomeCard` now waits for stats and passes them along, so the summary also never fires against missing data.
- The summary prompt instructs the model to write in the account's language (`account.locale_english_name`), greeting included.
- `MetricCard` gained a `loading` skeleton state; drilldown clicks are disabled while fetching.
- `fetchStats` in the overview page clears stale data up front, retries once silently on failure, and uses a fetch token plus `AbortController` (also aborted on unmount) to discard and cancel superseded requests.
- `WelcomeCard` guards its summary fetch with the same token pattern, so a summary response left in flight across a range, stats, or assistant change is discarded instead of rendered.

### DIFF
--- a/lib/integrations/openai/openai_prompts/captain_overview_summary.liquid
+++ b/lib/integrations/openai/openai_prompts/captain_overview_summary.liquid
@@ -1,6 +1,7 @@
 You are writing a short, warm summary of how an AI support assistant named "{{ assistant_name }}" performed over a reporting period, for {{ first_name }}, the person who manages it.
 
 Voice and format:
+- Write the entire summary in {{ language }}, including the opening greeting.
 - Address {{ first_name }} directly and open with "Hey {{ first_name }},". Be conversational, never robotic.
 - Always call the assistant by its name, {{ assistant_name }}. Never call it "Captain", "the assistant", or "your assistant".
 - This is a static, read-only poster on an analytics dashboard, not a chat. The reader cannot reply or ask you for anything. Never ask a question, invite a reply, offer further help, or say things like "let me know" or "I can dive in".
--- a/app/javascript/dashboard/api/captain/assistant.js
+++ b/app/javascript/dashboard/api/captain/assistant.js
@@ -26,10 +26,13 @@ class CaptainAssistant extends ApiClient {
     });
   }
 
-  getStats({ assistantId, range }) {
-    return axios.get(`${this.url}/${assistantId}/stats`, {
+  getStats({ assistantId, range, signal }) {
+    const requestConfig = {
       params: { range, timezone_offset: getTimezoneOffset() },
-    });
+    };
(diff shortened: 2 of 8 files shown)


lib/integrations/openai/openai_prompts/captain_overview_summary.liquid
```diff
@@ -1,6 +1,7 @@
 You are writing a short, warm summary of how an AI support assistant named "{{ assistant_name }}" performed over a reporting period, for {{ first_name }}, the person who manages it.
 
 Voice and format:
+- Write the entire summary in {{ language }}, including the opening greeting.
 - Address {{ first_name }} directly and open with "Hey {{ first_name }},". Be conversational, never robotic.
```
aakashb95, line 5: nit: Instead of Hey we can ask it to use the greetings of that language. I switched to Spanish and it still said Hey {name}, followed by spanish text.

aakashb95 (approved):
LGTM, added a small comment

lib/integrations/openai/openai_prompts/captain_overview_summary.liquid
scmmishra: > nit: Instead of Hey we can ask it to use the greetings of that language. I switched to Spanish and it still said Hey {name}, followed by spanish text. 

Missed this, I'll create a separate pr fixing it

PR #15106, opened 2026-07-21
feat(whatsapp): unify embedded signup feature gating

muhsin-k:
WhatsApp embedded signup now uses `whatsapp_embedded_signup_inbox_creation` as the single Chatwoot Cloud rollout gate for inbox creation, proactive reconfiguration, and disconnected inbox reauthorization. The authorization endpoint enforces the same gate, so the UI and backend remain consistent.

Self-hosted installations keep their existing behavior.

## Things to know

- This reuses the existing feature flag; there is no migration or schema change.
- The feature is shown as “WhatsApp Embedded Signup Flow” in feature management.
- `whatsapp_reconfigure` remains visible and honored for self-hosted proactive reconfiguration to preserve existing accounts. It can be deprecated after the self-hosted dependency is removed or migrated.

## How to test

1. On Chatwoot Cloud, enable `whatsapp_embedded_signup_inbox_creation` for an account.
2. Confirm that new WhatsApp inbox creation, proactive reconfiguration, and disconnected inbox reauthorization are available.
3. Disable the flag and confirm those entry points are hidden and authorization requests are rejected.
4. On self-hosted, confirm proactive reconfiguration remains controlled by the existing `whatsapp_reconfigure` account setting.

### DIFF
--- a/app/controllers/api/v1/accounts/whatsapp/authorizations_controller.rb
+++ b/app/controllers/api/v1/accounts/whatsapp/authorizations_controller.rb
@@ -50,7 +50,12 @@ class Api::V1::Accounts::Whatsapp::AuthorizationsController < Api::V1::Accounts:
   end
 
   def can_reconfigure_channel?
-    @inbox.channel.provider == 'whatsapp_cloud'
+    channel = @inbox.channel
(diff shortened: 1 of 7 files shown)


tds-1 (approved):
LGTM

PR #15113, opened 2026-07-21
fix(whatsapp): prevent invalid automation sends outside reply window

muhsin-k:
WhatsApp automations now fail locally when they attempt to send a free-form message after the 24-hour customer service window has closed. This avoids sending an invalid template request to Meta and gives users a clear, actionable error instead of “Template not found or invalid template name.”

Template messages continue to be sent whenever template parameters are present. Free-form messages continue to be sent normally while the conversation is replyable.

Fixes https://linear.app/chatwoot/issue/PLA-183/prevent-whatsapp-automations-outside-the-24-hour-window-from-producing

### How to reproduce

1. Create a WhatsApp automation that sends a message without template parameters.
2. Trigger it on a conversation whose 24-hour customer service window is closed.
3. Observe that the message previously reached the template send path and failed with a misleading provider error.

### How to test

1. Trigger an automation with template parameters and confirm it sends as a template message.
2. Trigger an automation without template parameters inside the 24-hour window and confirm it sends as a free-form message.
3. Trigger an automation without template parameters outside the 24-hour window and confirm it fails locally with a clear error and makes no request to Meta.

### Things to know

This changes only the invalid closed-window, no-template path. Existing template and in-window message behavior remains unchanged.

### DIFF
--- a/app/services/whatsapp/send_on_whatsapp_service.rb
+++ b/app/services/whatsapp/send_on_whatsapp_service.rb
@@ -6,12 +6,10 @@ class Whatsapp::SendOnWhatsappService < Base::SendOnChannelService
   end
 
   def perform_reply
-    should_send_template_message = template_params.present? || !message.conversation.can_reply?
-    if should_send_template_message
-      send_template_message
-    else
-      send_session_message
-    end
+    return send_template_message if template_params.present?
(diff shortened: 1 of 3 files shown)


muhsin-k:
@codex please review

sony-mathew (approved):
lgtm

PR #15122, opened 2026-07-22
perf: reuse resolved count for reopen rate

scmmishra:
This improves the Captain overview by loading reporting metrics and FAQ stats from separate endpoints. Range changes now refresh only the metrics, while reopen-rate calculation reuses the resolved conversation count to avoid redundant database queries.

## What changed

- Split Captain overview metrics and FAQ stats into separate APIs.
- Fetch FAQ stats independently from range-based metrics.
- Reuse resolved conversation totals when calculating reopen rate.
- Skip the reopen query when there are no resolved conversations.

### DIFF
--- a/app/javascript/dashboard/api/captain/assistant.js
+++ b/app/javascript/dashboard/api/captain/assistant.js
@@ -26,13 +26,20 @@ class CaptainAssistant extends ApiClient {
     });
   }
 
-  getStats({ assistantId, range, signal }) {
+  getMetrics({ assistantId, range, signal }) {
     const requestConfig = {
       params: { range, timezone_offset: getTimezoneOffset() },
     };
     if (signal) requestConfig.signal = signal;
 
-    return axios.get(`${this.url}/${assistantId}/stats`, requestConfig);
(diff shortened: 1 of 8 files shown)


sony-mathew (approved):
One minor comment:

- **[P2] Preserve the existing `/stats` contract during rollout.** [config/routes.rb:68](/chatwoot/config/routes.rb:68) removes `GET .../stats` immediately. An already-loaded dashboard bundle will continue calling that endpoint after deployment; both attempts will 404, clearing the metrics, knowledge card, and AI summary until the user refreshes. Direct API consumers also break.

  Keep `/stats` for a deprecation window, returning the legacy payload—`metrics` merged with `knowledge: faq_stats`—while new clients use the split endpoints. Preserve matching policy authorization as well.

scmmishra:
> * **[P2] Preserve the existing `/stats` contract during rollout.**

@sony-mathew this is acceptable, while it's technically a breaking change, the blast radius for this is very small. Should be fine to deploy, a quick reload will fix it for anyone

PR #15135, opened 2026-07-23
fix(widget): validate required checkboxes as accepted

### LINKED ISSUE #15128: Required checkbox accepted when unchecked again
### Describe the bug

When using a custom attribute of the type checkbox in a pre-chat form that is set to required, it is possible in the pre-chat UI to check and then uncheck the checkbox and the form will still be accepted.

I think that this is a bug, because a required checkbox should be set to true to be valid.

This can be due to FormKit handling 'required' and 'accepted' differently.
See [this issue in FormKit](https://github.com/formkit/formkit/issues/169) that describes how 'required' just needs the checkbox to be clicked at least once, but not necessary have a value equal to true.
See [documentation from FormKit](https://formkit.com/essentials/validation#accepted) for the validation rule 'accepted'.

I suspect that the issue can be resolved quickly, by adding the 'accepted' rule for the checkbox.

### To Reproduce

1. Create custom attribute of type checkbox.
2. Activate it in the pre-chat form.
3. Set the attribute to 'required'.
4. Open the UI.
5. Click the checkbox to check it.
6. Click the checkbox again to uncheck it.
7. Proceed with the form and see that it was possible even though the checkbox was unchecked.

### Expected behavior

The form is only validated successfully if the required checkbox is checked.

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

Browser

### Operating system

_No response_

### Browser and version

_No response_

### Docker (if applicable)

_No response_

### Additional context

_No response_

Nameless-Monster-Nerd:
Required checkbox fields in the pre-chat form now remain invalid unless they are checked. Other required field types keep their existing validation behavior, and the existing localized required message is reused.

## Closes

Closes https://github.com/chatwoot/chatwoot/issues/15128

## How to reproduce

1. Add a required checkbox custom attribute to the pre-chat form.
2. Check and then uncheck it.
3. Submit the form; submission is now blocked until the checkbox is checked.

## What changed

- Use the FormKit accepted rule for required checkbox fields.
- Map accepted validation failures to the existing pre-chat required message.


maxfriedrichsit:
Author of the underlying issue here.
This fix looks exactly like what we thought would be necessary to solve the problem.

sojan-official:
QA verified ✅

- Required checkbox blocks submission when left unchecked.
- Checking and then unchecking it still shows `Required` and prevents submission.
- Checking it allows the pre-chat form to submit; the local conversation was created with the checkbox custom attribute stored as `true`.
- Widget suite passed: 35 files / 226 tests. Focused pre-chat specs and lint also passed.

### Verification screenshots

**Unchecked — submission blocked with the required validation message**

Paper Layer widget blocks submission when required checkbox is unchecked

**Checked — required validation clears**

Paper Layer widget accepts the required checkbox when checked

sojan-official (approved):
Thanks for fixing this! 🙌

PR #15144, opened 2026-07-23
fix(whatsapp): keep Flow nfm_reply agent-only and restore i18n

pabloluna3596afk:
## Summary
- Restore missing `conversations.messages.whatsapp.flow_response` locale keys so agents see Formulario completado instead of Translation missing.
- Stop auto-sending Flow answers back to the customer (FlowConfirmationService removed); nfm_reply stays as an incoming agent-only message with whatsapp_flow_response payload.
- Documented as B-NEW-12 in docs/BUGS.md.

## Test plan
- [ ] Send a WhatsApp Flow template and complete it as the contact
- [ ] Confirm inbox shows one incoming bubble with readable field values (no Translation missing)
- [ ] Confirm the contact does **not** receive a confirmation summary message
- [ ] Confirm free-form reply is available after the Flow (24h window via incoming)

Closes B-NEW-12


pabloluna3596afk:
Opened against upstream by mistake; PaluHub fork hotfix belongs on pabloluna3596afk/chatwoot.

PR #15147, opened 2026-07-23
fix(perf): lazy load super admin dashboard stats

vishnu-narayanan:
## Description

The super admin landing page runs all its stat queries synchronously before rendering anything. On large installs the exact `COUNT(*)` on conversations and the 30-day chart group-by scan the whole table and exceed the request timeout, so the first page load fails and the console is unreachable.

This change:

- renders the page shell immediately; the dashboard component fetches stats as JSON from the same action, shows a loading skeleton, and falls back to N/A if the request fails
- caches the stats for 30 minutes
- uses the Postgres planner estimate (`pg_class.reltuples`) for the conversations count instead of an exact `COUNT(*)`
- adds an index on `conversations.created_at` (created concurrently) so the chart query is an index range scan instead of a full table scan

Fixes https://linear.app/chatwoot/issue/INF-62

## Type of change

## How Has This Been Tested?

- Browser test against a local instance: the shell and sidebar render immediately and stats fill in asynchronously. Forcing the stats request to return 500 shows N/A per metric and the page stays usable.
- Query verification against a production-sized dataset: the chart query was a 71s sequential scan without the index and becomes an index scan with it; the exact count exceeded 100s while the planner estimate returns in 250ms, within 0.1% of the exact value.

## Checklist:


sony-mathew (approved):
lgtm

sony-mathew:
@vishnu-narayanan please check if we need to do the migration outside of the deployment pipeline.

vishnu-narayanan:
@sony-mathew done

PR #15149, opened 2026-07-23
fix(reports): keep all agents as pivot summary rows

pabloluna3596afk:
## Summary
- Pivot agent/inbox/team/label tables no longer collapse to the few dimensions seen in a truncated conversation scan — rows are seeded from the same universe as the flat (non-pivot) summary, with zeros for empty segments.
- Conversation accumulation uses batched full-range scans instead of `reorder(:id).limit(2000)`, which biased busy accounts toward earliest assignees and dropped the rest.
- Adds `tmp/smoke_pivot_agent_rows.rb` to assert pivot/subset agent count ≥ flat for the demo panel.

Closes production regression from PR #29 (B-NEW-13).

## How to test
1. Seed: `docker cp tmp/seed_report_panels_demo.rb chatwoot-chatwoot-rails-1:/tmp/ && docker exec chatwoot-chatwoot-rails-1 bundle exec rails runner /tmp/seed_report_panels_demo.rb`
2. Smoke: `docker cp tmp/smoke_pivot_agent_rows.rb … && rails runner /tmp/smoke_pivot_agent_rows.rb` → `OK`
3. UI: open «Demo Agentes × Ventas» — agent row count with Columnas=Producto matches flat; agents without a product show `0` under that segment.
4. Prod after deploy: reopen the broken agent summary panel with pivot — all agents from the non-pivot view should remain; no panel config change needed.

## What changed
- `Reports::PanelRunnerService#accumulate_pivot_buckets` + `pivot_baseline_dimension_ids`
- Flat (non-pivot) summary path untouched


pabloluna3596afk:
Opened against upstream by mistake — fix lives on fork pabloluna3596afk/chatwoot (PaluHub).

PR #15153, opened 2026-07-24
feat: allow suspended accounts to access billing

scmmishra:
Administrators of suspended accounts can now reach the billing page to settle payment and restore their account, instead of being fully locked out on the suspended screen. The suspended screen shows a subtle **Manage billing** link below the contact support button (admins on Cloud only). Agents remain restricted to the suspended screen.

## How to test

1. On Cloud, suspend an account from Super Admin.
2. Log in as an administrator of that account — you land on the suspended screen with a "Manage billing" link below contact support.
3. Click it — the billing settings page opens; subscription, Stripe portal, and top-ups all work.
4. Log in as an agent — no billing link, and navigating to `/settings/billing` manually redirects back to the suspended screen.

## What changed

- Router guard allows `billing_settings_index` for administrators when the account is not active.
- Suspended screen gets a billing link gated by admin role and Cloud installation.

CleanShot 2026-07-24 at 12 25 56@2x


sony-mathew (approved):
lgtm

PR #15155, opened 2026-07-24
fix(imap): skip SSL hostname verify for Ski Dubai mailbox

Samersadek1:
## Summary
- Commits `config/initializers/imap_ssl_verify_none.rb` so Chatwoot IMAP fetch works against `mail.mysecondhomedubai.com` (cert hostname mismatch).
- Survives Railway redeploys (worker previously lost the ephemeral SSH patch because `railway.worker.toml` locks startCommand).

## Test plan
- [ ] After deploy, worker has no `certificate verify failed (hostname mismatch)` on Ski Dubai channel
- [ ] New mail to snowdogs@ appears in Ski Dubai inbox
- [ ] Remove this initializer after Gmail cutover


Samersadek1:
Opened against upstream by mistake; fix is for Samersadek1/chatwoot fork hosting only.

PR #15161, opened 2026-07-24
fix(billing): gate suspended account access by category

muhsin-k:
Administrators of suspended Cloud accounts now see and can access billing when the account was suspended for non-payment or predates suspension metadata. Accounts explicitly suspended for spam or another categorized reason remain restricted to the suspension screen.

The suspension screen also provides category-specific guidance: spam suspensions explain the detected activity, non-payment suspensions direct administrators to billing, other suspensions direct users to support, and legacy suspensions offer billing or support as the next step.

This aligns the billing recovery flow introduced in #15153 with the suspension categories introduced in #15158 without removing billing access from legacy suspended accounts. The authenticated account payload exposes only the current suspension category; suspension reasons and history remain internal.

Related: https://github.com/chatwoot/chatwoot/pull/15153
Related: https://github.com/chatwoot/chatwoot/pull/15158

### Things to know

Automatic account restoration after a successful payment is intentionally outside this PR and will be handled separately. The non-payment message therefore does not promise immediate restoration.

Legacy suspended accounts have no category, so they retain the billing access introduced in #15153.

### How to test

1. On Chatwoot Cloud, suspend an account with the non-payment category and log in as an administrator.
2. Confirm the suspension screen explains the outstanding payment, shows Manage billing, and opens billing settings.
3. Use a legacy suspended account without suspension history and confirm it shows the default message, Manage billing, and can open billing settings.
4. Suspend an account for spam and confirm the spam-specific message appears without Manage billing.
5. Suspend an account with other and confirm the support-focused message appears without Manage billing.
6. For spam and other suspensions, navigate directly to the billing settings URL and confirm it redirects to the suspension screen.
7. Confirm agents cannot access billing for any suspended account.


sony-mathew (approved):
lgtm. Please take a look at these findings:

- **P2 — blocking, matches existing thread:** [`check_suspended_billing_access`](/worktrees/e958/chatwoot/enterprise/app/controllers/enterprise/api/v1/accounts_controller.rb:6) omits `topup_checkout`. A suspended `spam` or `other` account administrator can directly call the endpoint; exact-head probes returned HTTP 200 for both categories. The production service creates and pays a Stripe invoice before granting credits. This behavior existed on the base, but leaves the PR’s restriction incomplete. Include `topup_checkout` in the callback and add request coverage for restricted, non-payment, and legacy suspensions. [Existing review thread](https://github.com/chatwoot/chatwoot/pull/15161#discussion_r3688356680).

- **P3 — non-blocking, matches existing thread:** [`routeHelpers.spec.js`](/worktrees/e958/chatwoot/app/javascript/dashboard/helper/specs/routeHelpers.spec.js:78) tests only a suspended agent. It does not protect the new administrator matrix for legacy, `non_payment`, `spam`, and `other`. The component tests covering messages and billing-link visibility were also added and then removed. Restore table-driven router tests and focused suspended-screen tests.

PR #15162, opened 2026-07-24
fix(billing): restore accounts after successful payment

muhsin-k:
Cloud accounts suspended for non-payment are now restored automatically after Stripe confirms that a subscription invoice was paid and the related subscription is active. Accounts suspended for spam or another reason, along with legacy suspensions without a category, remain unchanged.

Restoration runs at the verified Stripe webhook boundary and re-checks the latest account state under a database lock. This makes repeated webhook delivery safe and prevents a concurrent status or suspension-category change from restoring the wrong account.

Related: https://github.com/chatwoot/chatwoot/pull/15161
Related: https://github.com/chatwoot/chatwoot/pull/15158

### Things to know

The production Stripe webhook endpoint must subscribe to `invoice.paid`. One-off invoices and other paid invoices without a related subscription are ignored. A paid invoice also does not restore the account unless Stripe reports the subscription as active.

The handler supports both the legacy top-level invoice subscription field and Stripe's current nested invoice parent structure.

### How to test

1. Suspend a Cloud account with the non-payment category and ensure it is linked to a Stripe customer with an active subscription.
2. Pay an outstanding subscription invoice and deliver the resulting `invoice.paid` webhook.
3. Confirm the account becomes active and users can access it again.
4. Repeat delivery of the same webhook and confirm the account remains active without an error.
5. Repeat with accounts suspended for spam, other, and a legacy suspension without metadata; confirm they remain suspended.
6. Deliver a paid one-off invoice or an invoice whose subscription is not active; confirm the account remains suspended.


enterprise/app/services/enterprise/billing/handle_stripe_event_service.rb
```diff
@@ -168,7 +187,7 @@ def billing_period_renewed?
   end
 
   def account
-    @account ||= Account.where("custom_attributes->>'stripe_customer_id' = ?", subscription.customer).first
+    @account ||= Account.where("custom_attributes->>'stripe_customer_id' = ?", @event.data.object.customer).first
```
tds-1, line 190: is this change required?

tds-1 (approved):

sony-mathew (approved):
Looks good to me.

There is one minor finding, please see if we need to make these changes or do a followup:

## Findings

- **P2 — New: `invoice.paid` does not prove the non-payment debt was settled.** The handler restores the account whenever the related subscription is `active`, without inspecting how or what was paid ([service](/worktrees/b1e4/chatwoot/enterprise/app/services/enterprise/billing/handle_stripe_event_service.rb:28)). A focused probe confirmed that a paid, zero-amount subscription invoice restores a non-payment-suspended account.

  This is reachable in Stripe: `invoice.paid` also covers invoices marked paid out-of-band, and Stripe can automatically mark too-small invoices paid while rolling the amount into the customer balance. Furthermore, an `active` subscription does not guarantee that all outstanding invoices are paid. [Stripe event semantics](https://docs.stripe.com/api/events/retrieve?event_types-invoice.paid=), [automatic paid invoices](https://docs.stripe.com/billing/customer/balance?locale=en-GB), [subscription status semantics](https://docs.stripe.com/billing/subscriptions/webhooks?locale=en-GB).

  **Impact:** an account suspended for non-payment can regain full access after an unrelated, zero-value, out-of-band, or otherwise non-settling invoice event while debt remains.

  **Required fix:** define the restoration invariant explicitly. At minimum, distinguish a successful collection from generic `invoice.paid`—for example, use `invoice.payment_succeeded`, reject zero/out-of-band payments, and verify there are no relevant open invoices. The most precise design would associate the suspension with its Stripe invoice/subscription and restore only when that debt clears. Add repository specs covering zero-value, out-of-band, and “another invoice remains open” cases.

PR #15184, opened 2026-07-27
feat(compose): allow multiple To recipients on email inboxes

### LINKED ISSUE #15181: Compose new conversation: Email inboxes allow only a single "To" recipient

### Current Behavior

When starting a **new conversation** on an Email inbox from the compose window, the `To:` field accepts exactly one recipient.

`ContactSelector.vue` renders its `TagInput` with `mode="single"`, so once one contact is picked the input is replaced by a static chip and no further address can be entered:

https://github.com/chatwoot/chatwoot/blob/develop/app/javascript/dashboard/components-next/NewConversation/components/ContactSelector.vue#L136-L153

`Cc` and `Bcc` in the same window are multi-value (`EmailOptions.vue` uses `TagInput` in its default multiple mode), so the first email of a thread can only ever be addressed to one person, while everyone else has to be pushed into Cc.

This is inconsistent with the **reply** box on an existing conversation, where `ReplyEmailHead.vue` already exposes a multi-address `To` field, and with the backend, which has supported multiple To addresses for a while: `Messages::MessageBuilder#process_emails` reads `params[:to_emails]`, stores them in `content_attributes[:to_emails]`, and `ConversationReplyMailer#to_emails` delivers to that list.

https://github.com/chatwoot/chatwoot/blob/develop/app/builders/messages/message_builder.rb#L79-L92

So the capability exists end to end — the compose window simply never sends `to_emails`.

### Expected Behavior

On an Email inbox, the compose window should let an agent address the first message to more than one recipient, the same way Cc already works and the same way the reply box already works.

The contact selected first stays the conversation contact (the conversation model is unchanged, one contact per conversation). Any additional addresses are sent along as extra `To` recipients on that first outgoing email.

### Steps To Reproduce

1. Open the dashboard and click the compose (new conversation) button.
2. Pick a contact in `To:` and select an Email inbox in `Via:`.
3. Try to add a second recipient in `To:`.
4. The field is now a static chip — there is no way to add a second address. The only option is to put the other people in `Cc`.

### Environment

- Chatwoot version: develop
- Channel: Email

### Additional context

Related, already-closed issue about To addresses on the reply side: #8925.

I am working on a fix for this and will open a PR shortly.

vaibzde:
Agents can now address a new email thread to more than one person. Until now the compose window's `To:` field took exactly one contact, so anyone else on the thread had to be pushed into `Cc` — which reads differently to the recipients and is not what the agent meant. Pick the contact as before, choose the email inbox, and then keep adding recipients right next to it, with the same contact suggestions that `Cc` already offers.

The conversation still belongs to one contact. The extra addresses ride along as additional `To` recipients on that first outgoing email, exactly as the reply box already does on an existing conversation.

## Closes

Closes #15181

## How to test

1. Open the compose (new conversation) window.
2. Pick a contact in `To:` and select an **Email** inbox in `Via:`.
3. A second input appears next to the contact chip. Type two or more characters — matching contacts are suggested, and an address that matches nothing can still be typed in full.
4. Add a subject and a message, and send.
5. Open the created conversation. The first outgoing email lists every address in its `To` header, and the contact you picked first is the conversation contact.
6. Switch `Via:` to a non-email inbox: the extra recipient input is not offered, since no other channel can deliver one message to several addresses.

## What changed

- `ContactSelector.vue` keeps the single-contact chip that drives inbox selection, and renders a multi-value `TagInput` beside it when the target inbox is email. Suggestions come from the contact search already wired into the compose window.
- `prepareNewMessagePayload` sends `message[to_emails]`. The primary contact's address is prepended and the list de-duplicated, because `content_attributes[:to_emails]` replaces rather than extends the contact address in `ConversationReplyMailer#to_emails`.
- `contactConversations.js` forwards `to_emails` on the multipart create payload.

No backend change: `Messages::MessageBuilder#process_emails` already reads `to_emails`, validates the addresses, and stores them.

### DIFF
--- a/app/javascript/dashboard/components-next/NewConversation/ComposeConversation.vue
+++ b/app/javascript/dashboard/components-next/NewConversation/ComposeConversation.vue
@@ -49,6 +49,7 @@ const isSearching = ref(false);
 const formState = reactive({
   message: '',
   subject: '',
+  toEmails: '',
   ccEmails: '',
   bccEmails: '',
   attachedFiles: [],
@@ -57,6 +58,7 @@ const formState = reactive({
 const clearFormState = () => {
   Object.assign(formState, {
     subject: '',
+    toEmails: '',
     ccEmails: '',
     bccEmails: '',
     attachedFiles: [],
--- a/app/javascript/dashboard/components-next/NewConversation/components/ComposeNewConversationForm.vue
+++ b/app/javascript/dashboard/components-next/NewConversation/components/ComposeNewConversationForm.vue
@@ -60,6 +60,7 @@ const copilot = useCopilotReply();
 
 const showContactsDropdown = ref(false);
 const showInboxesDropdown = ref(false);
+const showToEmailsDropdown = ref(false);
 const showCcEmailsDropdown = ref(false);
 const showBccEmailsDropdown = ref(false);
 
@@ -68,6 +69,7 @@ const isCreating = computed(() => props.contactConversationsUiFlags.isCreating);
 const state = props.formState || {
   message: '',
   subject: '',
+  toEmails: '',
   ccEmails: '',
   bccEmails: '',
   attachedFiles: [],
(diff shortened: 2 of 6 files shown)


vaibzde:
Rebased-free bump on this one, just checking it's still on the radar. Happy to adjust the UX if multiple To recipients isn't the direction you want, or split the CC/BCC piece out separately if that's easier to review.

PR #15185, opened 2026-07-27
feat(reply-box): add a pinnable formatting toolbar

### LINKED ISSUE #15182: Reply box: formatting toolbar is only reachable through a text selection

### Current Behavior

The conversation reply box uses the same ProseMirror editor as the compose window, but the formatting menubar is hidden there.

`ReplyBox.vue` renders `WootMessageEditor` with the `popover-prosemirror-menu` class:

https://github.com/chatwoot/chatwoot/blob/develop/app/javascript/dashboard/components/widgets/conversation/ReplyBox.vue#L1335-L1341

and `Editor.vue` hides the menubar for that class unless the editor also has `has-selection`:

```scss
.popover-prosemirror-menu {
  .ProseMirror-menubar {
    display: none; // Hide by default
  }

  &.has-selection {
    .ProseMirror-menubar { display: flex; ... }
  }
}
```

So the toolbar only exists as a bubble menu over an active text selection. In practice that means:

- There is no visible affordance telling an agent that the reply box supports formatting at all. The compose window shows a toolbar; the reply box shows nothing.
- Commands that apply to an empty position cannot be reached. You cannot click "bullet list" or "numbered list" on a blank line to *start* a list — there is nothing to select yet. The same applies to toggling bold or italic before typing.
- The only remaining paths are markdown input rules (typing `- ` or `**`) or keyboard shortcuts, neither of which is discoverable.

The compose window (`components-next/Editor/Editor.vue`) does not set that class, so the very same editor shows a persistent toolbar there. The two entry points to writing an email behave differently.

### Expected Behavior

The reply box should offer a persistent formatting toolbar that an agent can turn on, so list, bold, italic, link and code commands are reachable from a blank line and are discoverable without knowing markdown.

The selection bubble menu is genuinely nice and should stay as the default; the ask is for an explicit toggle in the reply footer (next to emoji / attach / signature) that pins the toolbar open, with the choice remembered per agent.

### Steps To Reproduce

1. Open any conversation and focus the reply box.
2. Look for a formatting toolbar — there is none.
3. On an empty line, try to start a bulleted list. There is no button, and no text to select to summon the bubble menu.
4. Open the compose (new conversation) window on the same account: the toolbar is visible there.

### Additional context — tables

A closely related gap, raised together with this by the same users: email replies cannot contain tables, which is a common need when replying with a small set of figures or a status breakdown.

Table support is genuinely absent rather than merely hidden, and unlike the toolbar it cannot be fixed inside this repository alone:

- `FORMATTING` in `app/javascript/dashboard/constants/editor.js` has no `table` node or menu entry for any channel.
- `buildMessageSchema` in `@chatwoot/prosemirror-schema` builds no table nodes; only `fullSchema` (used by the Help Center article editor) pulls in `tableNodes` from `prosemirror-tables`.
- `messageParser.js` in the same package explicitly disables the markdown `table` rule, and `messageSerializer.js` has no table serialization.

So enabling tables in the message editor needs a change in [chatwoot/prosemirror-schema](https://github.com/chatwoot/prosemirror-schema) (schema builder, message parser, message serializer) and a released version bump before this repository can opt any channel into it.

I am happy to work on that upstream change too if maintainers are open to it — I wanted to confirm the direction before starting, since it touches the shared schema used by every channel.

### Environment

- Chatwoot version: develop
- Channel: all channels that allow formatting; most visible on Email

I am working on the toolbar part and will open a PR shortly.

vaibzde:
The reply box now has a formatting toolbar an agent can pin open. Until now formatting there was a bubble that only appeared over selected text, which meant an agent could not start a bullet list on an empty line, could not turn on bold before typing, and had no visible sign that the reply box supports formatting at all. The compose window, running the very same editor, has always shown a toolbar — so the two places you write an email behaved differently.

A new button in the reply footer pins the toolbar above the composer. The choice is stored in the agent's UI settings, so it stays put across conversations and sessions. The selection bubble remains the default for anyone who prefers it.

## Closes

Closes #15182

## How to test

1. Open any conversation and look at the reply footer. There is a new **Aa** button next to emoji and attach.
2. Click it. The formatting toolbar appears above the composer and the button shows as pressed.
3. On an empty line, click the bulleted list button. The list starts — this was not reachable before, since there was no text to select.
4. Move to a different conversation and reload the page. The toolbar is still pinned.
5. Click the button again to unpin. Select some text: the bubble menu still appears as it always did.
6. Open a conversation on an SMS or X inbox. The button is absent, because those channels have no formatting options.
7. Switch to **Private Note**. The toolbar follows the private-note formatting set, which includes strikethrough.

## What changed

- `ReplyBottomPanel.vue` gains the toggle, mirroring the existing quoted-reply button including `aria-pressed`. It is hidden when the channel's formatting menu is empty or the editor is disabled.
- `ReplyBox.vue` stores the preference under `formatting_toolbar_enabled` in UI settings.
- `Editor.vue` previously decided between bubble and menubar by reading its own root element's `classList` inside a `computed`. That has no reactive dependency, so the value froze after mount and would never have seen a runtime toggle. It is now driven by a `popoverMenu` prop, and the editor applies the class itself, so the CSS and the JS read from one source.

### On tables

The issue also raises tables in email replies. That one cannot be fixed from this repository: `buildMessageSchema` in `@chatwoot/prosemirror-schema` builds no table nodes, `messageParser.js` disables the markdown `table` rule, and `messageSerializer.js` has no table serialization — only `fullSchema`, used by the Help Center editor, pulls in `prosemirror-tables`. It needs a change and a release in that package first. Happy to take it on if maintainers want it; I left it out here to keep this reviewable.

### DIFF
--- a/app/javascript/dashboard/components/widgets/WootWriter/Editor.vue
+++ b/app/javascript/dashboard/components/widgets/WootWriter/Editor.vue
@@ -95,6 +95,9 @@ const props = defineProps({
   conversationId: { type: Number, default: null },
   medium: { type: String, default: '' },
   focusOnMount: { type: Boolean, default: true },
+  // Render the formatting menu as a bubble over the current selection instead
+  // of a menubar pinned above the editor.
+  popoverMenu: { type: Boolean, default: false },
 });
 
 const emit = defineEmits([
@@ -208,10 +211,7 @@ const editorRoot = useTemplateRef('editorRoot');
 const imageUpload = useTemplateRef('imageUpload');
 const editor = useTemplateRef('editor');
 
-const isEditorMenuPopover = computed(
-  () =>
-    editorRoot.value?.classList.contains('popover-prosemirror-menu') ?? false
-);
+const isEditorMenuPopover = computed(() => props.popoverMenu);
 
 const handleCopilotAction = actionKey => {
   if (actionKey === 'improve_selection' && editorView?.state) {
@@ -875,6 +875,7 @@ useEmitter(BUS_EVENTS.INSERT_INTO_RICH_EDITOR, insertContentIntoEditor);
     class="relative w-full"
     :class="{
       'opacity-50 cursor-not-allowed pointer-events-none': disabled,
+      'popover-prosemirror-menu': popoverMenu,
     }"
   >
     <TagAgents
--- a/app/javascript/dashboard/components/widgets/WootWriter/ReplyBottomPanel.vue
+++ b/app/javascript/dashboard/components/widgets/WootWriter/ReplyBottomPanel.vue
@@ -121,6 +121,14 @@ export default {
       type: Boolean,
       default: false,
     },
+    showFormattingToolbarToggle: {
+      type: Boolean,
+      default: false,
+    },
+    formattingToolbarEnabled: {
+      type: Boolean,
+      default: false,
+    },
     isEditorDisabled: {
       type: Boolean,
       default: false,
@@ -131,6 +139,7 @@ export default {
     'selectWhatsappTemplate',
     'selectContentTemplate',
     'toggleQuotedReply',
+    'toggleFormattingToolbar',
   ],
   setup(props) {
     const { setSignatureFlagForInbox, fetchSignatureFlagFromUISettings } =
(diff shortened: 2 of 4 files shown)


iamsivin (changes requested):
@vaibzde Thanks for the PR! Could you please resolve the conflicts?

vaibzde:
@iamsivin sorry for the delay — resolved now. Rebased onto develop and merged in your quoted-content rework from #15746. That PR removed the old quoted-reply toggle button entirely, so I dropped my toggle's leftover references to it rather than trying to keep both around. Everything else in this PR (the pinnable formatting toolbar) is unchanged.

PR #15186, opened 2026-07-27
feat(reply-box): suggest contacts in To, Cc and Bcc

### LINKED ISSUE #15183: Reply box: To/Cc/Bcc are plain text fields with no contact suggestions

### Current Behavior

Inside an existing email conversation, the reply box lets an agent edit `To`, `Cc` and `Bcc`. All three are plain free-text inputs — every address has to be typed out in full, comma separated, from memory.

`ReplyEmailHead.vue` renders three `woot-input` fields whose only logic is a comma-split email format check:

https://github.com/chatwoot/chatwoot/blob/develop/app/javascript/dashboard/components/widgets/conversation/ReplyEmailHead.vue#L96-L152

There is no contact lookup, so:

- Adding a known contact means typing the whole address by hand. A typo silently sends the reply to the wrong place, or the format check rejects the line and the agent has to hunt for which of several addresses is malformed.
- Nothing connects these fields to the contacts already in the account, even though the contact search API is right there and the compose window already uses it.

The compose window solved this already. `EmailOptions.vue` and `ContactSelector.vue` use the shared `TagInput` component with `menu-items` fed from contact search, so addresses arrive as chips picked from a dropdown:

https://github.com/chatwoot/chatwoot/blob/develop/app/javascript/dashboard/components-next/NewConversation/components/EmailOptions.vue#L92-L105

So an agent gets autocomplete when starting a thread and loses it the moment they reply to one.

### Expected Behavior

`To`, `Cc` and `Bcc` in the reply box should behave like the equivalent fields in the compose window: type two or more characters, see matching contacts, pick one, and get a removable chip. Typing a full address that matches no contact should still be accepted, the way `allow-create` already works in compose.

### Steps To Reproduce

1. Open an email conversation and focus the reply box.
2. Click `Cc` and start typing the name or partial address of a contact that exists in the account.
3. No suggestions appear. The full address has to be typed manually, and further addresses appended after a comma.
4. Open the compose (new conversation) window and type in `Cc` there: suggestions appear.

### Environment

- Chatwoot version: develop
- Channel: Email

I am working on a fix for this and will open a PR shortly.

vaibzde:
`To`, `Cc` and `Bcc` in the reply box now suggest contacts as you type. They were plain text boxes: every address had to be typed out in full from memory and separated by commas, so a typo either sent the reply to the wrong person or failed validation with no hint about which of five addresses was wrong. The compose window already solved this — replying now behaves the same way, with chips picked from a dropdown.

## Closes

Closes #15183

## How to test

1. Open an email conversation. `To` and `Cc` show the inherited recipients as removable chips.
2. Click `Cc` and type two or more characters of a contact's name or address. Matching contacts appear; pick one and it becomes a chip.
3. Type a full address that belongs to no contact and press enter or comma. It is still accepted.
4. Type something that is not a valid address. It is refused rather than silently sent.
5. Remove a chip with its ×, send the reply, and confirm the outgoing email's headers match the chips.
6. Open a conversation whose previous email had `Bcc` recipients: the `Bcc` row is now open on load instead of hidden behind **Add bcc**.
7. Remove every `To` chip. The field stays on screen so the recipients can be rebuilt.

## What changed

- `ReplyEmailHead.vue` is rewritten on `<script setup>` and uses the shared `TagInput` with `createContactSearcher`, the same pairing the compose window uses. The `v-model` contract with `ReplyBox.vue` is unchanged — still three comma-joined strings — so nothing else had to move.
- Scoped SCSS replaced with Tailwind utilities.

Two behaviour fixes fall out of the rewrite:

- `Bcc` stayed hidden when the reply already carried inherited `Bcc` addresses, because visibility tracked only whether **Add bcc** had been clicked. It now also opens when there is something to show.
- `To` rendered only when it already had a value, so clearing it once removed the field for the rest of the session. It now always renders.

`TagInput` validates each address as it is added, so `validEmailsByComma` and its spec have no callers left and are removed.


vaibzde:
Still around on this one if there's interest — happy to adjust the contact-suggestion matching logic or scope it down if the current approach isn't quite right.

PR #15192, opened 2026-07-27
feat: add manage-notification-preferences footer to agent notification emails

vishnu-narayanan:
## Description

Agent notification emails (new conversation, assignment, mention, new message, SLA misses) currently carry no indication of why the recipient received them or how to turn them off. When recipients cannot easily manage these emails, some mark them as spam, which hurts sender reputation and overall deliverability.

This adds a short footer line to agent notification emails only:

> You're receiving this email because email notifications are enabled for your account. **Manage notification preferences**.

The link points the recipient to their dashboard profile notification settings page (`/app/accounts/:account_id/profile/settings`) so they can disable notifications instead of marking the email as spam. This is a navigational link only, not a one-click unsubscribe (a real unsubscribe flow is a separate future change).

Scoping: the mailer layout (`app/views/layouts/mailer/base.liquid`) is shared by all mailers via `ApplicationMailer`. To keep the footer on agent notification emails only, `AgentNotifications::ConversationNotificationsMailer` exposes a `notification_settings_url` liquid local (built from the existing `app_account_url` route helper), and the layout renders the footer line only when that local is present. Transactional and other emails do not set it, so they are unaffected. The enterprise SLA notification methods prepend into the same mailer class, so they inherit the footer automatically.

Part of https://linear.app/chatwoot/issue/CW-7752

## Type of change

## How Has This Been Tested?

- `bundle exec rspec spec/mailers/agent_notifications/conversation_notifications_mailer_spec.rb spec/mailers/confirmation_instructions_spec.rb` — 26 examples, 0 failures. Asserts the footer link and settings URL are present in an agent notification email and absent from a non-notification (confirmation) email.
- `bundle exec rspec spec/enterprise/mailers/enterprise/agent_notifications/conversation_notifications_mailer_spec.rb` — 6 examples, 0 failures (SLA notification emails).

## Checklist:


sony-mathew (approved):
Approving this. But there are two caveats for this as we allow people to store these liquid templates in DB now. Both of the findings point to that. Currently we do not use this, so it should be good. (And we can't control what kind of layouts people store in DB and can't append random html). So this change is good to go keeping in mind that the following findings are true.


- **P2 — New, blocking:** Saved database-backed mail layouts bypass the new footer. [`base.liquid`](/worktrees/4383/chatwoot/app/views/layouts/mailer/base.liquid:132) only changes the filesystem fallback, while [`DbResolverService`](/worktrees/4383/chatwoot/app/services/email_templates/db_resolver_service.rb:58) prefers existing installation/account `base` layouts. I reproduced an account with `branded_email_templates` and a saved layout:

  ```json
  {"custom_layout_rendered":true,"preferences_footer_rendered":false,"settings_path_rendered":false}
  ```

  Those agent notifications therefore still provide no explanation or preferences link, undermining the PR’s incident-mitigation goal. Use a mailer-owned, non-overridable wrapper, or explicitly migrate/require the footer placeholder in stored layouts, with account- and installation-layout regression coverage.

- **P3 — Matches an open thread, non-blocking:** The new copy in [`base.liquid`](/worktrees/4383/chatwoot/app/views/layouts/mailer/base.liquid:135) is hardcoded and cannot participate in [`ApplicationMailer#switch_locale`](/worktrees/4383/chatwoot/app/mailers/application_mailer.rb:77). The existing thread’s P1 severity is overstated because these agent-notification templates are already largely English-only, but new source strings should still enter `en.yml` for future Crowdin translation.

PR #15194, opened 2026-07-27
fix(email): resolve reply recipients from the message being sent

### LINKED ISSUE #15193: Email inbox: replies are delivered to the recipients of the conversation's latest outgoing message, not the message being sent
On an Email inbox, the To/Cc/Bcc of an outgoing email are not read from the message that is being delivered. They are read from whatever the conversation's **latest** outgoing message happens to be at the moment the delivery job runs.

Because replies are delivered asynchronously (`SendReplyJob.perform_later`, and `wait: 2.seconds` when the message has attachments), any outgoing message created in that window silently replaces the recipient list of the mail that is already on its way out.

The case that hurts most is looping a new person into an ongoing email thread. The agent adds `vendor@example.com` to Cc, sends the follow-up, and then immediately writes a private note like "vendor looped in". A private note is an outgoing message too, and on an Email inbox it is stored with empty `to_emails` / `cc_emails` / `bcc_emails`. By the time `SendReplyJob` runs, that private note is the latest outgoing message, so the follow-up goes out with no Cc at all and with `To` falling back to the conversation contact. The newly added person never receives the mail, and therefore never gets the thread.

### Where it comes from

`ConversationReplyMailer` looks the addresses up on the conversation instead of on the message it was handed:

https://github.com/chatwoot/chatwoot/blob/develop/app/mailers/conversation_reply_mailer.rb#L176-L192

```ruby
def cc_bcc_emails
  content_attributes = @conversation.messages.outgoing.last&.content_attributes
  ...
end

def to_emails_from_content_attributes
  content_attributes = @conversation.messages.outgoing.last&.content_attributes
  ...
end
```

`email_reply(message)` already stores the message being delivered in `@message`, and the class already has a `current_message` helper (`@message || @conversation.messages.outgoing.last`) that the `from` / `reply_to` builders use. The recipient lookups are the only place that still reaches for `messages.outgoing.last` directly.

Private notes are the most reproducible trigger, but the same window is open for any two replies sent back to back, and it widens whenever the Sidekiq queue is backed up.

### Steps to reproduce

1. Open a conversation on an Email inbox.
2. Add a new address to **Cc** and send a reply.
3. Immediately add a private note to the same conversation.
4. Inspect the delivered mail: `Cc` is empty and `To` is the conversation contact, not the addresses that were entered on the reply.

### Expected

The delivered email carries the To/Cc/Bcc that were entered on that reply. Messages created afterwards, private notes included, do not change the recipients of a mail that has already been queued.

vaibzde:
When an agent loops a new address into an ongoing email thread and then adds a private note, the reply currently goes out with no Cc, and with To falling back to the conversation contact. The newly added person never receives the mail. This change makes an outgoing email keep the To/Cc/Bcc that were entered on that reply, no matter what is added to the conversation afterwards.

Closes #15193

## How to reproduce

1. Open a conversation on an Email inbox.
2. Add a new address to **Cc** and send a reply.
3. Immediately add a private note to the same conversation.
4. Inspect the delivered mail: `Cc` is empty and `To` is the conversation contact instead of the addresses entered on the reply.

## What changed

`ConversationReplyMailer#cc_bcc_emails` and `#to_emails_from_content_attributes` read the addresses from `@conversation.messages.outgoing.last` rather than from the message they were handed. Replies are delivered asynchronously (`SendReplyJob.perform_later`, and `wait: 2.seconds` when the message has attachments), so any outgoing message created in that window replaces the recipients of a mail that is already queued. A private note is the easiest way to hit it: private notes are outgoing messages, and on an Email inbox `Messages::MessageBuilder#process_emails` stores them with empty `to_emails` / `cc_emails` / `bcc_emails`.

Both lookups now go through the existing `current_message` helper (`@message || @conversation.messages.outgoing.last`), which the `from` and `reply_to` builders already use. `email_reply` sets `@message`, so it resolves the recipients of the message being delivered; `reply_with_summary` and `reply_without_summary` leave `@message` nil and keep their current behaviour.

Added a spec covering the private-note case in `spec/mailers/conversation_reply_mailer_spec.rb`.

## Note

I could not run the Ruby test suite locally — there is no Ruby toolchain in the environment I worked in, so the new spec has not been executed on my side. The change and the spec were verified by reading the code paths (`Message#send_reply`, `Email::SendOnEmailService`, `Messages::MessageBuilder#process_emails`, `ConversationReplyMailerHelper#prepare_mail`). Deferring to CI for the actual run.

### DIFF
--- a/app/mailers/conversation_reply_mailer.rb
+++ b/app/mailers/conversation_reply_mailer.rb
@@ -174,7 +174,7 @@ class ConversationReplyMailer < ApplicationMailer
   end
 
   def cc_bcc_emails
-    content_attributes = @conversation.messages.outgoing.last&.content_attributes
+    content_attributes = current_message&.content_attributes
 
     return [] unless content_attributes
     return [] unless content_attributes[:cc_emails] || content_attributes[:bcc_emails]
@@ -183,7 +183,7 @@ class ConversationReplyMailer < ApplicationMailer
   end
 
   def to_emails_from_content_attributes
-    content_attributes = @conversation.messages.outgoing.last&.content_attributes
+    content_attributes = current_message&.content_attributes
 
     return [] unless content_attributes
     return [] unless content_attributes[:to_emails]
--- a/spec/mailers/conversation_reply_mailer_spec.rb
+++ b/spec/mailers/conversation_reply_mailer_spec.rb
@@ -342,6 +342,25 @@ RSpec.describe ConversationReplyMailer do
         expect(mail.message_id).to eq("conversation/#{conversation.uuid}/messages/#{message.id}@#{conversation.account.domain}")
       end
 
+      context 'when a newer outgoing message exists in the conversation' do
+        let!(:message) do
+          create(:message, conversation: conversation, account: account, message_type: 'outgoing', content: 'Looping in the vendor',
+                           content_attributes: { to_emails: ['customer@example.com'], cc_emails: ['vendor@example.com'],
+                                                 bcc_emails: ['audit@example.com'] })
+        end
+
+        it 'sends to the recipients of the message being delivered' do
+          # a private note added right after the reply carries empty recipient lists
+          create(:message, conversation: conversation, account: account, message_type: 'outgoing', private: true,
+                           content: 'Vendor has been looped in',
+                           content_attributes: { to_emails: [], cc_emails: [], bcc_emails: [] })
+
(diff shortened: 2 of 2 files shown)


sojan-official:
Thanks for picking this up and fixing it! Really appreciate the contribution 🙌

sojan-official (approved):

PR #15199, opened 2026-07-27
fix(whatsapp): collect text header parameters

muhsin-k:
WhatsApp templates with variables in both a text header and body currently show inputs only for the body. Agents therefore cannot provide the header value in the template composer, even though the API and backend support the corresponding `processed_params.header` payload.

The composer now displays text-header variables separately, previews their substituted values, and sends them alongside body parameters. Templates using media headers remain unchanged.

Related: https://github.com/chatwoot/utils/pull/65
Fixes https://linear.app/chatwoot/issue/CW-6667/whatsapp-template-messages-with-required-header-named-params-fail-with

### Things to know

This PR consumes the released `@chatwoot/utils@0.0.57`, which adds text-header parameter construction and completeness validation.

### How to reproduce

1. Open a WhatsApp template containing text header `Welcome {{1}}` and body variables `{{1}}` and `{{2}}`.
2. Observe that the current composer displays only two body inputs and omits the header input.

### How to test

1. Open the same template in the conversation composer.
2. Confirm one header input and two body inputs are displayed.
3. Fill the values and confirm both the header and body previews update.
4. Send the template and confirm `processed_params` contains `header.1`, `body.1`, and `body.2`.


muhsin-k:
@codex please review

sony-mathew (approved):
lgtm. please take a look at this finding:

## Findings

### [P2] Valid named header parameters can be silently discarded

[template_processor_service.rb:68](/Users/sonymathew/.codex/worktrees/pr-15199/chatwoot/app/services/whatsapp/template_processor_service.rb:68) still treats `media_type` and `media_name` as reserved keys for every header:

```ruby
elsif key != 'media_type' && key != 'media_name'
```

The new frontend accepts arbitrary named text-header keys, so a valid template such as:

```text
HEADER: Welcome {{media_type}}
parameter_format: NAMED
```

renders an input, passes completeness validation, and submits `header.media_type`. The backend then drops the entire header component. My direct service probe produced only the body component, meaning Meta receives a request missing its required header parameter.

Named variables may contain lowercase letters and underscores, and named header sends require the matching `parameter_name`, as shown in this [[AWS WhatsApp/Meta integration example](https://aws.amazon.com/blogs/messaging-and-targeting/send-whatsapp-business-messages-with-aws-end-user-messaging-social/)](https://aws.amazon.com/blogs/messaging-and-targeting/send-whatsapp-business-messages-with-aws-end-user-messaging-social/).

The smallest fix is to branch on the actual template header format:

- For media headers, interpret `media_url`, `media_type`, and `media_name` as metadata.
- For `TEXT` headers, treat every key as a text parameter and pass it through `build_text_parameter`.

Add backend coverage for ordinary positional/named text headers and named variables called `media_type` and `media_name`.

PR #15206, opened 2026-07-28
fix: default Captain overview to last 7 days

scmmishra:
The Captain agents overview now defaults to the last 7 days instead of this month, so the page opens on a more recent and actionable window.

## What changed

- Overview page, range selector, and welcome card default to `7`.
- Backend `Captain::AssistantStatsWindow::DEFAULT_RANGE` changed from `30` to `7`, so requests without a `range` param (or with invalid values) also resolve to the last 7 days.

## How to test

- Open Captain → Overview: the range selector should show "Last 7 days" by default and metrics should reflect that window. Other ranges continue to work as before.


iamsivin (approved):
LGTM

PR #15207, opened 2026-07-28
fix(security): keep inbox access filtering on the participating scope

tds-1:
An agent who is removed from an inbox could still see that inbox's conversations under the **Participating** filter. Removing an agent from an inbox does not delete the conversation participant records they already had, and the participating filter was ignoring inbox access entirely — so those conversations stayed visible indefinitely. The filter now respects inbox access like every other conversation filter.

## Linear Ticket
- https://linear.app/chatwoot/issue/CW-6923

## How to reproduce

1. Add an agent to two inboxes, A and B.
2. As that agent, become a participant on a conversation in inbox B (open it, or get added as a participant).
3. Remove the agent from inbox B in Settings → Inboxes → Collaborators.
4. Log in as the agent and open Conversations → Participating.
5. The inbox B conversation is still listed, and is openable.

## What changed

`ConversationFinder#filter_by_conversation_type` **replaced** `@conversations` with `current_user.participating_conversations` for the `participating` type, discarding the inbox/permission-filtered scope built up by `Conversations::PermissionFilterService` immediately before it. It now narrows the existing scope by participating ids instead, so permission filtering survives.

### DIFF
--- a/app/finders/conversation_finder.rb
+++ b/app/finders/conversation_finder.rb
@@ -142,7 +142,8 @@ class ConversationFinder
       conversation_ids = current_account.mentions.where(user: current_user).pluck(:conversation_id)
       @conversations = @conversations.where(id: conversation_ids)
     when 'participating'
-      @conversations = current_user.participating_conversations.where(account_id: current_account.id)
+      participating_ids = current_user.participating_conversations.where(account_id: current_account.id).select(:id)
+      @conversations = @conversations.where(id: participating_ids)
     when 'unattended'
       @conversations = @conversations.unattended
     end
--- a/app/finders/conversation_finder.rb
+++ b/app/finders/conversation_finder.rb
@@ -142,8 +142,8 @@ class ConversationFinder
       conversation_ids = current_account.mentions.where(user: current_user).pluck(:conversation_id)
       @conversations = @conversations.where(id: conversation_ids)
     when 'participating'
-      participating_ids = current_user.participating_conversations.where(account_id: current_account.id).select(:id)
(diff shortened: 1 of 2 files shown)


vishnu-narayanan (approved):
@tds-1 Nice fix, intersecting instead of replacing the scope is the right shape. A few things:

**1. Add a meta-count assertion for `conversation_type=participating`.** The `/meta` path shares this scope and nothing pins it, so it's the most likely spot for a silent future regression. Would be good to cover.

**2. The "openable" claim in the PR body.** On develop, `ConversationPolicy#show?` seems to already deny a removed agent without inbox or team access, so direct-open looks blocked already, the real remaining hole was the list/preview leak. Was the openability observed against an older build, or is there a path I'm missing? Worth getting right before it's copied into the advisory closeout.

**3. Admin path on very large accounts.** Since the permission filter doesn't restrict admins, the intersected shape self-joins `conversations` twice. I ran EXPLAIN on a large account: agents get a big win (seq-scan ~33s to index nested loop ~11ms), but the admin count goes ~33s to ~63s because of the double scan. Would it be worth using the leaner `ConversationParticipant.where(account_id:, user_id:).select(:conversation_id)` subquery, the way the enterprise permission filter already does? That avoids the self-join. (The deeper misestimate is CW-7604 territory, not this PR's job, just flagging the admin regression so it's a deliberate call.)

**4. Stale `ConversationParticipant` rows.** Filtering at read time is the safer default and I agree with it, just want to confirm leaving the rows in place was intentional rather than deferred, since cleanup on `InboxMember` destroy would change re-add semantics.

PR #15226, opened 2026-07-28
fix: add (conversation_id, created_at) index to messages

### LINKED ISSUE #15225: Messages pagination / conversation list 500s on statement_timeout: missing (conversation_id, created_at) index
### Describe the bug

Message endpoints that paginate or inspect a conversation's messages intermittently return **HTTP 500** (`ActiveRecord::QueryCanceled` / `PG::QueryCanceled: canceling statement due to statement timeout`) on instances with a large `messages` table — even when the conversation itself is small (~1k messages).

Two hot paths are affected:

- `GET /api/v1/accounts/:account_id/conversations/:id/messages?before=<id>` — cursor pagination (`MessageFinder#messages_before`).
- `GET /api/v1/accounts/:account_id/contacts/:id/conversations` — the per-conversation last message in `app/views/api/v1/conversations/partials/_conversation.json.jbuilder` (`conversation.messages.where(account_id: …).last`), rendered once per conversation. This one surfaces as `ActionView::Template::Error (PG::QueryCanceled: …)`.

**Root cause (missing index, not conversation size).** Both queries filter by `conversation_id` and order by `created_at`, but no index covers `(conversation_id, created_at)`. The existing `index_messages_on_conversation_account_type_created` can't serve the ordering because `account_id` and `message_type` sit between `conversation_id` and `created_at`. So the planner falls back to an **`Index Scan Backward` on the global `index_messages_on_created_at`**, filtering by `conversation_id`. For **old/inactive** conversations the matching rows live at the far (oldest) end of that global index, so Postgres walks a huge fraction of the table before collecting the 20 (or 1) rows it needs — crossing `statement_timeout` (default `14s`) and returning 500. This is why the failing cursor "moves" between runs: it depends on table-wide load and cache, not on any corrupt row.

`EXPLAIN` on a production instance (~358k rows in `messages`; sample conversation has only 1,668 messages):

```
Limit  (cost=0.42..954.50 rows=20)
  ->  Index Scan Backward using index_messages_on_created_at on messages
        Filter: ((id < 208128) AND (conversation_id = 10278))
```
→ ~14,013 ms in ActiveRecord → statement timeout → 500.

After adding `(conversation_id, created_at)`:

```
Limit  (actual time=495.568..503.417 rows=20)
  ->  Index Scan Backward using index_messages_on_conversation_id_and_created_at
 Execution Time: 503.484 ms
```
and the per-conversation `.last` drops from ~14,000 ms to 0.044 ms.

Note: #14933 (v4.16.0) sanitizes this error's message but does not prevent the timeout; the underlying index is still missing on `develop`.

### To Reproduce

1. On a self-hosted instance whose `messages` table is large (hundreds of thousands of rows) and which has some **old/inactive** conversations.
2. Paginate an old conversation's messages backwards with the `before` cursor: `GET /api/v1/accounts/:id/conversations/:cid/messages?before=<old_message_id>`, or open such a contact's conversation list in the agent dashboard.
3. Observe intermittent HTTP 500 with `PG::QueryCanceled: canceling statement due to statement timeout` (~14s) in the Rails log; neighbouring cursors return 200.

### Expected behavior

Message pagination and the conversation list should be served by an index on `(conversation_id, created_at)` and complete in milliseconds regardless of the conversation's age, instead of scanning the global `created_at` index and hitting `statement_timeout`.

### Environment

Docker (self-hosted)

### Version

v4.15.1 (also reproducible on current `develop` — the `(conversation_id, created_at)` index is still absent in `db/schema.rb`).

qsebas:
## Description

Message endpoints that paginate/inspect a conversation's messages can return **HTTP 500** (`ActiveRecord::QueryCanceled` / `PG::QueryCanceled: canceling statement due to statement timeout`) on instances with a large `messages` table, even for conversations that are themselves small (~1k messages).

**Root cause is a missing index, not conversation size.** Two hot paths filter by `conversation_id` and order by `created_at`:

- `MessageFinder#messages_before` → `... WHERE conversation_id = ? AND id < ? ORDER BY created_at DESC LIMIT 20` (the `GET /api/v1/accounts/:id/conversations/:id/messages?before=` cursor).
- `app/views/api/v1/conversations/partials/_conversation.json.jbuilder` → `conversation.messages.where(account_id: …).last` (and `.non_activity_messages.first`), i.e. `... WHERE conversation_id = ? [AND account_id = ?] ORDER BY created_at DESC LIMIT 1`, rendered once **per conversation** in `Contacts::ConversationsController#index`.

No index covers `(conversation_id, created_at)`. The existing `index_messages_on_conversation_account_type_created` can't serve the ordering because `account_id` and `message_type` sit between `conversation_id` and `created_at`. So the planner falls back to an **`Index Scan Backward` on the global `index_messages_on_created_at`**, filtering by `conversation_id`. For **old/inactive** conversations the matching rows live at the far (oldest) end of that global index, so Postgres walks a huge fraction of the table before collecting the 20 (or 1) rows — crossing `statement_timeout` (default `14s`) and returning 500.

This complements #14933 (which sanitizes the error message but does not prevent the timeout). This PR removes the timeout at the source by adding a dedicated composite index; both the API pagination and the agent-facing conversation list benefit, with no query/logic changes.

Fixes #15225

## Type of change

## How Has This Been Tested?

Validated locally against a clean PostgreSQL 16 database loaded from `develop`'s `db:schema:load`:

- `bundle exec rails db:migrate` → applied `AddConversationCreatedIndexToMessages`, creating `CREATE INDEX index_messages_on_conversation_id_and_created_at ON public.messages USING btree (conversation_id, created_at)`.
- `bundle exec rails db:rollback STEP=1` → `remove_index` ran, index dropped cleanly (fully reversible).
- `bundle exec rails db:schema:dump` confirms the `db/schema.rb` entry and its placement in this PR match the canonical dump.
- `bundle exec rubocop db/migrate/20260728000000_add_conversation_created_index_to_messages.rb` → `1 file inspected, no offenses detected`.

Effectiveness measured on a production instance (~358k rows in `messages`; sample conversation has only 1,668 messages), with `EXPLAIN ANALYZE` before and after creating the index concurrently.

**`messages_before` — before:**

```
Limit  (cost=0.42..954.50 rows=20)
  ->  Index Scan Backward using index_messages_on_created_at on messages
        Filter: ((id < 208128) AND (conversation_id = 10278))
```
→ 14,013 ms in ActiveRecord → statement timeout → 500.

**`messages_before` — after:**

```
Limit  (actual time=495.568..503.417 rows=20)
  ->  Index Scan Backward using index_messages_on_conversation_id_and_created_at
        Index Cond: (conversation_id = 10278)
        Filter: (id < 208128)
 Execution Time: 503.484 ms
```

**`.last` (jbuilder) — before:** same global-index scan → ~14,000 ms → 500.
**`.last` (jbuilder) — after:**

```
Limit  (actual time=0.019..0.020 rows=1)
  ->  Index Scan Backward using index_messages_on_conversation_id_and_created_at
        Index Cond: (conversation_id = 10278)
        Filter: (account_id = 2)
 Execution Time: 0.044 ms
```

The index was also created concurrently on that production instance (`indisvalid = t`), confirming the statement is valid and non-blocking at scale.

## Checklist:

### DIFF
--- a/db/migrate/20260728000000_add_conversation_created_index_to_messages.rb
+++ b/db/migrate/20260728000000_add_conversation_created_index_to_messages.rb
@@ -10,13 +10,21 @@ class AddConversationCreatedIndexToMessages < ActiveRecord::Migration[7.1]
   # existing index_messages_on_conversation_account_type_created can't serve the
   # ORDER BY because account_id / message_type sit between the two columns.
   def up
+    # CREATE INDEX CONCURRENTLY on a large table can take minutes; disable the
+    # session statement_timeout so the build is not cancelled mid-run.
+    execute 'SET statement_timeout TO 0'
     add_index :messages, [:conversation_id, :created_at],
               name: 'index_messages_on_conversation_id_and_created_at',
               algorithm: :concurrently, if_not_exists: true
+  ensure
+    execute 'RESET statement_timeout'
   end
 
   def down
+    execute 'SET statement_timeout TO 0'
     remove_index :messages, name: 'index_messages_on_conversation_id_and_created_at',
                             algorithm: :concurrently, if_exists: true
+  ensure
+    execute 'RESET statement_timeout'
   end
 end
--- a/db/migrate/20260728000000_add_conversation_created_index_to_messages.rb
+++ b/db/migrate/20260728000000_add_conversation_created_index_to_messages.rb
@@ -1,0 +1,22 @@
+class AddConversationCreatedIndexToMessages < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  # Message pagination (MessageFinder#messages_before) and the per-conversation
+  # last-message lookup in api/v1/conversations/partials/_conversation.json.jbuilder
+  # both filter by conversation_id and order by created_at, but no index covers
+  # that shape. For old/inactive conversations Postgres falls back to an
+  # Index Scan Backward on the global index_messages_on_created_at, which on a
+  # large messages table crosses statement_timeout (default 14s) and 500s. The
+  # existing index_messages_on_conversation_account_type_created can't serve the
+  # ORDER BY because account_id / message_type sit between the two columns.
+  def up
+    add_index :messages, [:conversation_id, :created_at],
+              name: 'index_messages_on_conversation_id_and_created_at',
+              algorithm: :concurrently, if_not_exists: true
+  end
+
+  def down
+    remove_index :messages, name: 'index_messages_on_conversation_id_and_created_at',
+                            algorithm: :concurrently, if_exists: true
+  end
+end
--- a/db/migrate/20260728000000_add_conversation_created_index_to_messages.rb
+++ b/db/migrate/20260728000000_add_conversation_created_index_to_messages.rb
@@ -37,7 +37,7 @@ class AddConversationCreatedIndexToMessages < ActiveRecord::Migration[7.1]
   INDEX_NAME = 'index_messages_on_conversation_id_and_created_at'.freeze
 
   def drop_invalid_index_if_present
-    result = execute(<<~SQL)
+    result = execute(<<~SQL.squish)
       SELECT indisvalid
       FROM pg_class
       JOIN pg_index ON pg_index.indexrelid = pg_class.oid
--- a/db/migrate/20260728000000_add_conversation_created_index_to_messages.rb
+++ b/db/migrate/20260728000000_add_conversation_created_index_to_messages.rb
@@ -10,6 +10,10 @@ class AddConversationCreatedIndexToMessages < ActiveRecord::Migration[7.1]
   # existing index_messages_on_conversation_account_type_created can't serve the
   # ORDER BY because account_id / message_type sit between the two columns.
   def up
+    # A previous failed CONCURRENTLY build may have left an invalid index stub.
+    # Drop it so the build below does not silently skip over a broken index.
+    drop_invalid_index_if_present
+
     # CREATE INDEX CONCURRENTLY on a large table can take minutes; disable the
     # session statement_timeout so the build is not cancelled mid-run.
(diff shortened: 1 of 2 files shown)


qsebas:
Good catch — addressed in the latest commit.

`CREATE INDEX CONCURRENTLY` runs outside the migration transaction (hence `disable_ddl_transaction!`) but the session-level `statement_timeout` from `database.yml` (default 14s) still applies and will cancel the build on a large table.

Fixed by zeroing `statement_timeout` at the start of both `up` and `down`, with an `ensure` block to `RESET` it afterward so the timeout is always restored even if the index build fails mid-run:

```ruby
execute 'SET statement_timeout TO 0'
add_index ..., algorithm: :concurrently, if_not_exists: true
ensure
  execute 'RESET statement_timeout'
```

qsebas:
Good catch — addressed in the latest commit.

`if_not_exists: true` silences a same-named index regardless of `indisvalid`, so a stub left by a failed concurrent build would be skipped silently and the migration recorded as applied while queries keep timing out.

Fixed by querying `pg_index` before the `CREATE` and dropping any invalid copy first:

```ruby
def drop_invalid_index_if_present
  result = execute(<<~SQL)
    SELECT indisvalid
    FROM pg_class
    JOIN pg_index ON pg_index.indexrelid = pg_class.oid
    WHERE pg_class.relname = '#{INDEX_NAME}'
  SQL
  return if result.none? || result.first['indisvalid'] == 't'

  execute 'SET statement_timeout TO 0'
  remove_index :messages, name: INDEX_NAME, algorithm: :concurrently
ensure
  execute 'RESET statement_timeout'
end
```

Three paths now:
- Index doesn't exist → normal build.
- Index exists and is valid → `if_not_exists` skips safely (idempotent re-run).
- Index exists and is invalid → drop it first, then rebuild.

PR #15229, opened 2026-07-29
fix: harden article author updates

sony-mathew:
## Description

Article edits now ignore an `author_id` that does not belong to the current account, retain the existing author, and still apply other valid article changes. Article creation continues to reject cross-account authors with a generic validation error and without creating a record.

The authenticated article serializers still omit authors without a current-account membership so existing forged or stale records cannot expose agent profile fields. The guard now checks `current_account_user` directly to make that intent explicit.

## Closes

Follow-up to [CW-7665](https://linear.app/chatwoot/issue/CW-7665) and [#15191](https://github.com/chatwoot/chatwoot/pull/15191).

## Type of change

## How Has This Been Tested?

- Creating an article with a cross-account author returns `422` and does not create an article.
- Updating an article with a cross-account author retains the previous author while applying other valid attributes.
- Articles whose previous author is no longer an account member can still be edited without exposing that author's agent profile.
- Existing OSS and Enterprise article request coverage passes locally.

## Checklist:


vishnu-narayanan (approved):
lgtm

PR #15233, opened 2026-07-29
feat: add Captain conversation outcome model [CW-7792]

scmmishra:
Adds the `Captain::ConversationOutcome` model and its table: one row per (account, assistant, conversation) that folds Captain lifecycle events into flat facts for the upcoming value metrics report. This is the second PR in the outcomes stack, on top of the lifecycle event layer (#15213); the tracker and listener that populate it come next.

The table deliberately stores only timestamps, counts, and the handoff reason. Every classification and duration (coverage, autonomous vs assisted, durable resolution, resolution time) is derived at query time in the stats layer, so definitions can change later without backfills. The row is created on the first qualifying customer message, making `created_at` the demand-start anchor. Commit messages carry the field-by-field rationale for what was left out.

| Field | Meaning |
| --- | --- |
| `account_id`, `assistant_id`, `conversation_id`, `inbox_id` | Reporting dimensions; unique on the first three |
| `first_captain_reply_at` / `last_captain_reply_at` | First and latest public Captain reply |
| `captain_reply_count` | Public Captain replies in the conversation |
| `first_human_reply_at` | First public human agent reply, used to classify assisted resolutions |
| `handoff_at` | When Captain handed the conversation to a human |
| `handoff_reason_category` | Why it handed off (customer_request, missing_knowledge, unsupported_request, policy_restriction, tool_failure, pending_clarification, usage_limit) |
| `resolved_at` | When the conversation was resolved |
| `last_reopened_at` / `reopen_count` | Reopen facts backing reopen rate and durable-resolution checks |
| `csat_rating` / `csat_received_at` | CSAT for Captain-involved conversations |


enterprise/app/models/conversation_outcome.rb
```diff
@@ -0,0 +1,59 @@
...
+    pending_clarification
+    usage_limit
+  ].freeze
+
+  self.table_name = 'captain_conversation_outcomes'
+
+  belongs_to :account
+  belongs_to :assistant, class_name: 'Captain::Assistant'
+  belongs_to :conversation, class_name: '::Conversation'
+  belongs_to :inbox
```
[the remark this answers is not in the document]
scmmishra: Will be handled later

db/migrate/20260731140853_create_captain_conversation_outcomes.rb
```diff
@@ -0,0 +1,40 @@
+class CreateCaptainConversationOutcomes < ActiveRecord::Migration[7.1]
+  def change
+    create_table :captain_conversation_outcomes do |t|
```
sony-mathew: Shall we reuse this for conversations in general not tie it to captain? So shall we rename this to "conversation_outcomes"?

scmmishra: Should I move it outside Captain folder too? I would suggest keeping it Captain focused, since the data is not used for anything else, and would prevent us from overloading the model for other things outside captain

sony-mathew: Yes please. Let's move it out of captain folder. We could bring in some goodness from these models to other areas in reporting. Let's make it generic from the start itself. Easy to do this now vs later.

scmmishra: It's done

sony-mathew (approved):
lgtm

scmmishra:
@codex please review

scmmishra:
@codex please review

PR #15237, opened 2026-07-29
feat: add Captain outcome stats builder

scmmishra:
Adds `Captain::AssistantOutcomeStatsBuilder`, the read side of the outcomes stack (#15235): the Captain value-report metrics computed from `captain_conversation_outcomes`, exposed at `GET /captain/assistants/:id/outcome_metrics` with the same `range`/`timezone_offset` contract and current/previous/trend packing as the existing overview endpoint.

The cohort for a window is the demand that entered it (rows created inside the window), so every rate shares one denominator and conversations never migrate between windows as they progress. Every classification is a query-time expression over the row's fact columns, and both windows are computed in a single scan via FILTER aggregation. The existing `AssistantStatsBuilder` overview is untouched: it remains the source for the current cards and the V1 path, and since outcome rows only exist from deployment onward, windows reaching before that date undercount demand until the data ages in.

| Metric | Definition |
| --- | --- |
| `eligible_conversations` | Outcome rows created in the window (demand in Captain-connected inboxes) |
| `coverage_rate` | Involved / eligible, where involved = replied publicly or handed off for a non-usage-limit reason |
| `autonomous_resolutions` / `autonomous_resolution_rate` | Resolved with Captain replies, no handoff, no human reply before resolution; rate is over eligible demand |
| `assisted_resolutions` | Resolved with Captain participation and a human in the loop |
| `durable_resolution_rate` | Of autonomous resolutions older than 7 days, the share not reopened since resolving |
| `csat_score` / `human_only_csat_score` | Average CSAT on involved conversations vs conversations no assistant touched |
| `median_first_response_seconds` | Median of first Captain reply minus demand start |
| `median_resolution_seconds` | Median of resolution minus demand start for involved conversations |
| `handoff_reasons` | Current-window handoff count per reason category |


scmmishra:
Closing this for a newer PR: https://github.com/chatwoot/chatwoot/pull/15425

PR #15263, opened 2026-07-30
fix: consolidate customer message bursts before lookup

aakashb95:
Captain now treats consecutive customer messages as one request before it asks a question, searches the knowledge base, or suggests a solution. It uses the full message burst to identify the customer's goal and current state. It checks conflicting details before relying on them.

## How to test

1. Send several customer messages in a row that describe one issue. Add a short correction in a later message.
2. Confirm Captain uses all the messages before it asks a question, searches the knowledge base, or suggests a solution.
3. Send another message with a detail that conflicts with information Captain already checked. Confirm Captain checks the detail again before relying on it.


sony-mathew (approved):
Looks good to me. But there is one finding which makes sense:

## Findings

- **[P2] Direct answers to Captain’s clarification questions are classified as unconfirmed context.**

  [core_rules.liquid:8](/worktrees/f18b/chatwoot/enterprise/lib/captain/prompts/snippets/core_rules.liquid:8) says short fragments and values are “not … confirmed answers.” This applies to every user turn after an assistant response—including the expected flow where [assistant.liquid:86](/worktrees/f18b/chatwoot/enterprise/lib/captain/prompts/assistant.liquid:86) asks for a missing identifier and the customer replies only `12345`.

  The runner preserves the assistant question in history and extracts that value as the current user message at [agent_runner_service.rb:67](/worktrees/f18b/chatwoot/enterprise/app/services/captain/assistant/agent_runner_service.rb:67). Captain may therefore repeat the question, avoid the intended lookup, or hand off instead of progressing. The shared snippet affects both primary and scenario agents.

  This is the same still-unresolved [GitHub review finding](https://github.com/chatwoot/chatwoot/pull/15263#discussion_r3689677863), which I independently confirmed. The smallest fix is to remove “or confirmed answers” and explicitly preserve direct replies to the immediately preceding clarification.

aakashb95:
> [P2] Direct answers to Captain’s clarification questions are classified as unconfirmed context.
[core_rules.liquid:8](https://github.com/worktrees/f18b/chatwoot/enterprise/lib/captain/prompts/snippets/core_rules.liquid:8) says short fragments and values are “not … confirmed answers.” This applies to every user turn after an assistant response—including the expected flow where [assistant.liquid:86](https://github.com/worktrees/f18b/chatwoot/enterprise/lib/captain/prompts/assistant.liquid:86) asks for a missing identifier and the customer replies only 12345.
The runner preserves the assistant question in history and extracts that value as the current user message at [agent_runner_service.rb:67](https://github.com/worktrees/f18b/chatwoot/enterprise/app/services/captain/assistant/agent_runner_service.rb:67). Captain may therefore repeat the question, avoid the intended lookup, or hand off instead of progressing. The shared snippet affects both primary and scenario agents.
This is the same still-unresolved https://github.com/chatwoot/chatwoot/pull/15263#discussion_r3689677863, which I independently confirmed. The smallest fix is to remove “or confirmed answers” and explicitly preserve direct replies to the immediately preceding clarification.


will revisit and merge this

aakashb95:
@codex review

Issue #13446, opened 2026-02-04
Shortcuts conflicting with diacritical marks


DanielAGW:
### Describe the bug

Hi. Chatwoot shortcuts are conflicting with diacritic marks in Polish (and possibly other languages).

For example, Add attachment shortcuts conflicts with Polish accented "a" letter :( We are unable to easily type `ą`. It is the same for other shortcuts with Alt - N, E, O, S. We are unable to type `ń`, `ę`, `ó`, `ś`.

I strongly believe that this issue - https://github.com/chatwoot/chatwoot/issues/10060 - should be reopened. Shortcuts are supposed to increase usability. Making shortcuts based on Alt, which is used for accented characters, significantly worsens the user experience, I'm afraid. If shortcuts are not customizable, they can be more harmful than valuable, and I'm afraid this is exactly the case here. 

### To Reproduce

1. As a support agent, go to a conversation
2. Try to type a Polish word with accented characters, for example `moją` (mine)
3. The actual behavior is that after typing `moj`, upload is triggered

### Expected behavior

I should see `moją` in the test area, not `moj` and upload modal.

### Environment

app.chatwoot.com

### Cloud Provider

None

### Platform

None

### Operating system

Mac OS (latest)

### Browser and version

Chrome (latest)

adrianbanasiak:
I can confirm this issue with self-hosted version `v4.10.1` and chrome 144. It doesnt happen on Brave or Safari 🤔

Konrad-FSi:
It is very annoying and for polish language makes application almost useless

adrianbanasiak:
Following GIT diff allowed me to fix issue on my self-hosted version. No longer shortcut ALT+L triggers re-focus and movement of cursor.

```
diff --git a/app/javascript/dashboard/components/widgets/WootWriter/Editor.vue b/app/javascript/dashboard/components/widgets/WootWriter/Editor.vue
index 956e04dda..79b70b44d 100644
--- a/app/javascript/dashboard/components/widgets/WootWriter/Editor.vue
+++ b/app/javascript/dashboard/components/widgets/WootWriter/Editor.vue
@@ -572,7 +572,7 @@ useKeyboardEvents({
   },
   'Alt+KeyL': {
     action: focusEditorInputField,
-    allowOnFocusedInput: true,
+    allowOnFocusedInput: false,
   },
 });
```

boryn:
Hi!

[...]

Issue #10060, opened 2024-09-02
Ability to disable keyboard shorcuts


mani-zur:
### Is your feature or enhancement related to a problem? Please describe.

In the Polish language, there are many additional characters like 'ł' (Alt + L), 'ś' (Alt + S), 'ą' (Alt + A). These letters overlap with keyboard shortcuts, and I would like to have the option to disable them. Currently, to write in Polish, I have to use an external editor.

### Describe the solution you'd like

I would like to have an option in the user settings that allows disabling keyboard shortcuts entirely.

stonecharioteer:
This issue is being closed due to lack of activity for over a year. [For more details, visit this discussion.](https://github.com/orgs/chatwoot/discussions/12810)

sojan-official:
ref: https://discord.com/channels/647412545203994635/1487162415337439232 

@iamsivin  could you take a look ?

PR #15271, opened 2026-07-30
fix(shortcuts): ignore AltGr keystrokes in global keyboard handlers

przemyslvw:
Windows reports AltGr as Ctrl+Alt, and tinykeys deliberately treats an AltGraph modifier state as satisfying both Control and Alt on Win32. That made AltGr+<letter> match the $mod+Alt+<letter> keybindings, so on a Polish layout typing "a" with ogonek (AltGr+A) opened the attachment picker ($mod+Alt+KeyA) and "e" with ogonek (AltGr+E) resolved the conversation and moved to the next one ($mod+Alt+KeyE).

Add an isAltGraphEvent helper and drop such events in the global keydown handlers before any shortcut matching runs. Only the AltGraph modifier state is checked, so a purposely pressed Ctrl+Alt+A keeps working on layouts without AltGr (e.g. US QWERTY), where that state is never set.

Plain Alt+<letter> shortcuts were never affected: tinykeys rejects them when Control is also held, which AltGr implies. The same class of collision exists on other AltGr layouts (Czech, Croatian, Hungarian).

# Pull Request Template

## Description

Fixes # (issue)

## Type of change

## How Has This Been Tested?

## Checklist:


przemyslvw:
Thanks — the second comment is correct, and it convinced me to take the heuristic
out rather than narrow it further. I've reverted the guard to the simple form and
documented why. Reasoning below, since it also answers the first comment.

**Both review comments rest on one premise I could not verify:** that Firefox on
Windows reports `getModifierState('AltGraph') === true` for a deliberately pressed
Ctrl+Alt chord. I have no Firefox-on-Windows environment to test, and Gecko appears
to set the `AltGraph` state based on whether the key actually *produced* a
character — in which case `Ctrl+Alt+A` on a US layout would not set it and the
concern would not arise. If you have a source for that behaviour I'd genuinely like
to see it; it decides how much of this discussion is real.

**Taking the premise as given, the second comment is right and it kills the
heuristic.** Comparing the produced character against the Latin letter in `code`
only works on Latin layouts. On Cyrillic or Greek, `event.key` never equals the
`code` letter, so the check reports "this is text" for every keystroke and drops
`$mod+Alt+KeyE` / `$mod+Alt+KeyA`. I verified that logic:

| layout | event | verdict of the heuristic | wanted |
|---|---|---|---|
| Polish | `key: 'ą'`, `code: 'KeyA'`, AltGraph | text → drop | drop |
| US | `key: 'a'`, `code: 'KeyA'`, AltGraph | chord → pass | pass |
| Cyrillic | `key: 'у'`, `code: 'KeyE'`, AltGraph | text → drop | **pass** |
| Greek | `key: 'ε'`, `code: 'KeyE'`, AltGraph | text → drop | **pass** |

**The deeper point: this cannot be decided from a single keydown.** Windows
collapses AltGr into left Ctrl + right Alt, so AltGr+E and Ctrl+Alt+E on a
Cyrillic layout produce an identical event — same `key`, same `code`, same
modifier state. No event-level heuristic can separate them. Doing it correctly
needs `navigator.keyboard.getLayoutMap()`, to compare the produced character
against what the key yields unmodified on the active layout — and that API exists
in Chromium but not in Firefox, i.e. it is missing exactly where the concern is
raised.

I also tried skipping the guard on Apple platforms (the fix for the first
comment). It preserves `Alt+<letter>` on macOS, but leaves macOS users unable to
type accented characters at all — and the original report in #13446 came from
macOS. So that variant trades the reporter's bug for a different one.

**What the PR does now:** checks the `AltGraph` state and nothing else, with the
tradeoff stated in the code and pinned by a test, so it reads as a decision rather
than an oversight. Where a browser reports `AltGraph` for a real chord, that chord
stops firing. I've prioritised typing on the grounds that a user who cannot write
their own language is worse off than one who loses a keybinding — but that is a
product call, and if you weigh it the other way I'll happily change it.

**What I think the actual fix is:** stop binding `$mod+Alt+<letter>`. On Windows
that chord is indistinguishable from AltGr by construction, so any keybinding
built on it will collide with some layout's accented characters — Polish today,
Czech, Croatian and Hungarian equally. Two bindings are affected
(`$mod+Alt+KeyA`, `$mod+Alt+KeyE`). Moving them to a chord that does not overlap
AltGr would remove the whole class of bug without any heuristic, and would also
address #10060. That is a decision on the shortcut scheme, so I've left it to you
rather than making it in this PR — happy to do that work instead if you prefer.

scmmishra (approved):

PR #15272, opened 2026-07-30
fix: sanitize CSV exports against spreadsheet formula injection

kah-ja:
## Description

The `csv-safe` gem was added in #6444 so that CSV exports cannot be turned into spreadsheet formulas, and the v2 report exports render through `CSVSafe`. Several export paths were never migrated and still write their rows with the plain `CSV` writer, so a field starting with `=`, `+`, `-`, `@`, `%`, `|`, tab, CR or LF is emitted verbatim and is evaluated as a formula when the file is opened in Excel, LibreOffice or Google Sheets.

Two of the affected exports carry values that end users, not operators, control:

* `app/views/api/v1/accounts/csat_survey_responses/download.csv.erb` renders `csat_response.feedback_message`. `Public::Api::V1::CsatSurveyController#update` permits that field without authentication for any caller holding the conversation UUID, which Chatwoot sends to the customer in the CSAT email.
* `app/jobs/account/contacts_export_job.rb` writes `contact.send(header)` raw. `name` is in the default column set and has no format validation, and `identifier` / `custom_attributes` are admin selectable export columns that the public widget contact endpoint accepts.

The file is opened by the administrator who downloaded the report, so anything that evaluates does so in their spreadsheet client.

This PR switches the remaining writers to `CSVSafe`:

* `app/views/api/v1/accounts/csat_survey_responses/download.csv.erb`
* `app/jobs/account/contacts_export_job.rb`
* `enterprise/app/views/api/v1/accounts/applied_slas/download.csv.erb`
* `app/views/api/v2/accounts/reports/conversation_traffic.erb` (the timezone line only, the data rows already used `CSVSafe`)
* `app/controllers/api/v1/accounts/data_imports_controller.rb` (import error and skip log download)
* `lib/tasks/download_report.rake`

`Account::DataImportJob#generate_csv_data` is deliberately left on the plain writer: its output is the failed records file that admins are expected to correct and re-upload, and prefixing would corrupt values on that round trip, for example `+4915112345678` coming back as `'+4915112345678`. Happy to switch it too if you prefer full consistency.

Reported privately as `GHSA-fwcw-wx2v-q2x2` (opened 2026-05-20, still in triage). This PR is the fix for it.

### Behaviour change

Fields beginning with a formula character now get a leading apostrophe, so an E.164 phone number in the CSAT export reads `'+919999999999`. Spreadsheet clients display the value without the apostrophe, a plain text reader sees it. This is the same trade-off the already migrated report exports make.

## Type of change

## How Has This Been Tested?

Three specs were added:

* `spec/controllers/api/v1/accounts/csat_survey_responses_controller_spec.rb`: a CSAT response whose feedback message is a HYPERLINK formula comes back from the download prefixed.
* `spec/jobs/account/contacts_export_job_spec.rb`: a contact whose name is a formula is exported prefixed.
* `spec/enterprise/controllers/api/v1/accounts/applied_slas_controller_spec.rb`: the SLA policy name column is exported prefixed.

Additional checks I ran:

* RuboCop 1.75.6 with this repository's config reports no offenses on the changed Ruby files.
* The behaviour of `csv-safe` 3.3.1 (the version in `Gemfile.lock`) was verified on Ruby 3.4.4 for the three entry points used here: `CSVSafe.generate_line`, `CSVSafe.generate` with `csv << row`, and `CSVSafe.open`. All three prefix dangerous fields, and `Numeric` values pass through untouched, so the rating and id columns are unaffected.

I do not have a full local dev environment for this repository, so the complete RSpec suite was not run locally, only the checks listed above.

## Checklist:

### DIFF
--- a/app/controllers/api/v1/accounts/data_imports_controller.rb
+++ b/app/controllers/api/v1/accounts/data_imports_controller.rb
@@ -180,7 +180,7 @@ class Api::V1::Accounts::DataImportsController < Api::V1::Accounts::BaseControll
   end
 
   def logs_csv(logs)
-    CSV.generate(headers: true) do |csv|
+    CSVSafe.generate(headers: true) do |csv|
       csv << %w[created_at kind source_object_type source_object_id error_code message details]
 
       logs.order(:created_at).find_each do |log|
--- a/app/controllers/api/v1/accounts/data_imports_controller.rb
+++ b/app/controllers/api/v1/accounts/data_imports_controller.rb
@@ -158,7 +158,7 @@ class Api::V1::Accounts::DataImportsController < Api::V1::Accounts::BaseControll
   end
 
   def logs_csv(logs)
-    CSV.generate(headers: true) do |csv|
+    CSVSafe.generate(headers: true) do |csv|
       csv << %w[created_at kind source_object_type source_object_id error_code message details]
 
       logs.order(:created_at).find_each do |log|
--- a/app/jobs/account/contacts_export_job.rb
+++ b/app/jobs/account/contacts_export_job.rb
@@ -20,7 +20,7 @@ class Account::ContactsExportJob < ApplicationJob
     contacts_to_export = contacts.to_a
     preload_contact_labels(contacts_to_export) if headers.include?(LABELS_COLUMN)
 
-    csv_data = CSV.generate do |csv|
+    csv_data = CSVSafe.generate do |csv|
       csv << headers
       contacts_to_export.each do |contact|
         csv << headers.map { |header| value_for_header(contact, header) }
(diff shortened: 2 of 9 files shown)


kah-ja:
Any updates?

kah-ja:
ping

The failing `Run Chatwoot CE spec / backend-tests (16, 5)` is a pre-existing failure in `spec/lib/chatwoot_app_spec.rb` (`ChatwootApp.self_hosted_paid?` → `NoMethodError` in the EE module injection, `chatwoot_hub.rb`), unrelated to this PR. This diff only switches the remaining CSV export writers to `CSVSafe`; it doesn't touch any file in that stack trace.

PR #15276, opened 2026-07-31
feat(campaigns): track whatsapp campaign delivery analytics

muhsin-k:
WhatsApp campaigns now store delivery data for every intended recipient so delivery performance can be inspected after a campaign runs. The backend records the actual campaign start and completion times, captures sent, delivered, read, failed, and skipped outcomes, and exposes metrics and recipient-level results for the analytics UI.

The analytics APIs and recipient tracking are available for one-off WhatsApp campaigns on Enterprise installations when the `whatsapp_campaign` account feature is enabled. The shared campaign lifecycle records and exposes actual start and completion timestamps for all one-off campaigns.

Related: https://linear.app/chatwoot/issue/CW-4627/ee-basic-analytics-for-whatsapp-campaigns

### What changed

- Create a campaign recipient record for every contact selected by the campaign audience.
- Store rendered message content, provider message ID, delivery timestamps, and provider failure details.
- Update recipient status from WhatsApp delivered, read, and failed webhooks without allowing status downgrades.
- Store `started_at` when campaign processing begins and `completed_at` when it finishes.
- Add metrics and paginated, status-filtered recipient analytics endpoints.

### Things to know

- This PR contains the backend implementation only. The analytics UI is in #15369 and is based on this branch.
- Recipient analytics currently cover one-off WhatsApp campaigns only; `started_at` and `completed_at` are shared by all one-off campaign types.
- Access is controlled by the existing `whatsapp_campaign` account feature. Plan entitlement configuration is outside this PR.
- The migration creates `campaign_recipients` and adds `started_at` and `completed_at` to campaigns.
- Retry, automatic reattempt, and campaign-specific reply attribution are not included.

### How to test

1. Enable `whatsapp_campaign` for an Enterprise account.
2. Create a one-off WhatsApp campaign with a label-based audience that includes valid contacts and at least one contact without a phone number.
3. Trigger the campaign and confirm it records `started_at`, `completed_at`, and one campaign recipient per audience contact.
4. Confirm successful sends store the provider message ID and skipped or failed recipients store the corresponding reason.
5. Process delivered, read, and failed WhatsApp status webhooks and confirm recipient states and timestamps update.
6. Request the campaign metrics endpoint and confirm the audience and delivery counts.
7. Request the contacts endpoint with status and page parameters and confirm filtering and pagination.


sony-mathew (commented):
Please take a look at the following findings:

- **P2 — New: webhook updates are not tenant/inbox scoped.** [incoming_message_base_service.rb](/worktrees/7944/chatwoot/enterprise/app/services/enterprise/whatsapp/incoming_message_base_service.rb:5) looks up `CampaignRecipient` globally by `source_id`; the delayed [update_recipient_status_job.rb](/worktrees/7944/chatwoot/enterprise/app/jobs/campaigns/update_recipient_status_job.rb:5) does the same. I reproduced a webhook processed for inbox/account A changing account B’s recipient from `sent` to `delivered`. This is a cross-tenant integrity boundary violation, particularly because manual WhatsApp Cloud channels can accept unsigned callbacks when no channel secret is configured. Scope both paths to the resolved inbox/account and include that identifier in the delayed job.

- **P2 — Matches the current open thread: only the first batched status is processed.** [incoming_message_base_service.rb](/worktrees/7944/chatwoot/enterprise/app/services/enterprise/whatsapp/incoming_message_base_service.rb:5) and its parent both select `.first`. My two-status provider-shaped probe ended with `["delivered", "sent"]`; the second recipient remained stale. Meta models `statuses` as an array, and notes that status notifications may arrive out of order. [Official Meta/Postman documentation](https://www.postman.com/meta/whatsapp-business-platform/request/rgtfq23/message-status-update-notifications). This confirms the [existing open review thread](https://github.com/chatwoot/chatwoot/pull/15276#discussion_r3760427855).

- **P2 — Still present despite the resolved early-webhook thread: reconciliation is one-shot.** When a status arrives before `mark_sent!` persists `source_id`, the webhook schedules exactly one lookup after two seconds. If that job runs too early, [update_recipient_status_job.rb](/worktrees/7944/chatwoot/enterprise/app/jobs/campaigns/update_recipient_status_job.rb:4) silently returns and never retries. I ran that order explicitly: the job missed the queued recipient, `source_id` was then persisted, and the final status remained `sent`. Use bounded retries/backoff or persist pending statuses until attribution becomes possible.

- **P3 — Existing issue-comment concern: malformed `errors` crashes campaign tracking.** [campaign_recipient.rb](/worktrees/7944/chatwoot/enterprise/app/models/campaign_recipient.rb:72) assumes `errors.first` is a hash. With `errors: ['plain text error']`, it raises `TypeError: no implicit conversion of Symbol into Integer` and leaves the recipient `sent`. This violates Meta’s documented structured-error shape, so I consider it non-blocking hardening, but normalizing external webhook input would prevent retry noise and stale data.

sony-mathew (approved):

PR #15278, opened 2026-07-31
fix: quote email sender display names per RFC 5322

scmmishra:
Outgoing emails failed permanently on inboxes whose sender name contains a comma (for example `Care Team, Billing`). The comma is an address separator in email syntax, so the unquoted From header could not be parsed into a valid mailbox and the SMTP envelope sender fell back to the whole raw string, which providers rejected (Gmail: `555 5.5.2 Syntax error, cannot decode response`). Sender display names are now assembled with `Mail::Address`, which quotes them whenever needed, so any inbox or agent name produces a valid, deliverable From header.

## Before and after

Broken case, inbox name contains a comma (`Care Team, Billing <care@example.com>`):

| | From header | SMTP envelope |
|---|---|---|
| Before | Entire string absorbed into the encoded display text, no structured address remains | `MAIL FROM:<Care Team, Billing <care@example.com>>` rejected with `555 5.5.2` |
| After | `"Care Team, Billing" <care@example.com>` | `MAIL FROM:<care@example.com>` accepted |

Already working case, plain inbox name (`Care Team <care@example.com>`):

| | From header | SMTP envelope |
|---|---|---|
| Before | `Care Team <care@example.com>` | `MAIL FROM:<care@example.com>` |
| After | `Care Team <care@example.com>`, byte identical | `MAIL FROM:<care@example.com>` |

Recipients see the same sender name in every case. Quotes and MIME encoding are wire syntax that mail clients strip on display.

## References

- [RFC 5322, section 3.4 (Address Specification)](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4): defines the `mailbox` grammar where commas separate addresses in a list, which is why an unquoted comma splits the display name.
- [RFC 5322, section 3.2.4 (Quoted Strings)](https://datatracker.ietf.org/doc/html/rfc5322#section-3.2.4): the quoted-string form that makes special characters safe inside a display name.
- [RFC 2047 (MIME encoded words)](https://datatracker.ietf.org/doc/html/rfc2047): the `=?UTF-8?...?=` encoding used for non-ASCII display names.


scmmishra:
@codex please review

PR #15284, opened 2026-08-01
fix: accept inbound Twilio SMS from short codes

### LINKED ISSUE #15283: fix: accept inbound Twilio SMS from short codes
### Describe the bug

Inbound Twilio SMS callbacks can use a five- or six-digit short code in the `From` field. The callback is accepted, but asynchronous message processing attempts to save that sender as a contact phone number. Contact validation requires E.164, so the contact, conversation, and message are not created.

### Steps to reproduce

1. Configure a Twilio SMS inbox.
2. Submit an inbound callback with synthetic parameters such as `From=68849`, `To=+15551234567`, and a non-empty body.
3. Run the queued Twilio event job.

### Expected behavior

The short code remains the contact-inbox source identifier and display name, the contact phone number remains empty, and the inbound message is created.

### Actual behavior

Contact creation fails because the short code is not a valid E.164 phone number.

### Proposed fix

Only populate the contact phone-number field when the inbound SMS sender is valid E.164. Preserve all other sender identifiers for contact-inbox routing and display.

salmonumbrella:
Twilio can deliver inbound SMS from 5-6 digit short codes, but Chatwoot currently requires every Twilio SMS source to be E.164. The callback job then fails when creating both the contact and contact inbox, after the callback has already returned 204.

This change accepts 5-6 digit short codes as Twilio SMS source IDs while only storing E.164 senders in the contact phone number field. It adds model and service regressions covering complete message creation with synthetic values.

Closes #15283

### DIFF
--- a/app/services/twilio/incoming_message_service.rb
+++ b/app/services/twilio/incoming_message_service.rb
@@ -1,4 +1,6 @@
 class Twilio::IncomingMessageService
+  E164_REGEX = /\A\+[1-9]\d{1,14}\z/
+
   include ::FileTypeHelper
   include ::Twilio::WhatsappIdentifierHelper
   include ::Twilio::ReferralParamsHelper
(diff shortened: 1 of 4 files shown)


ericsjones:
Bump. This affects us from time to time.

PR #15295, opened 2026-08-03
fix: encode contact search params so emails with a plus sign match

iamsivin:
# Pull Request Template

## Description

This PR fixes an issue where contact search failed for emails containing a `+`, such as `jane+shop@gmail.com`. We were building the search URL manually, so `+` was treated as a space in the query string and the server searched for `jane shop@gmail.com` instead.

This also fixed an issue in the compose window. When the search returned no results, the dropdown offered to create a new contact. Since the contact already existed, the API returned an error, but it was swallowed and the **To:** field was cleared without any feedback.

The search request now passes parameters through Axios `params`, allowing Axios to handle URL encoding correctly. This also fixes searches containing other reserved characters like `&`, `#`, and `%`, where `%` could previously cause a `400` response. The redundant `encodeURIComponent` on the Contacts page has been removed to avoid double encoding, and the compose window now surfaces the actual API error when contact creation fails.

### How to reproduce

1. Create a contact with an email containing a `+`, for example `jane+shop@gmail.com`.
2. Open the compose window and search for the full email.
3. Before, the contact was not found, the dropdown offered to create a new contact, and selecting it silently failed while clearing the **To:** field. After this change, the existing contact is found correctly, and any contact creation error is shown to the user.

Fixes https://linear.app/chatwoot/issue/CW-7812/contact-search-sends-q-unencoded-so-emails-containing-never-match-and

## Type of change

### Screenshots

**Before**
image
image
image

**After**
image
image
image

## Checklist:

### DIFF
--- a/app/javascript/dashboard/api/contacts.js
+++ b/app/javascript/dashboard/api/contacts.js
@@ -1,16 +1,13 @@
 /* global axios */
 import ApiClient from './ApiClient';
 
-export const buildContactParams = (page, sortAttr, label, search) => {
-  let params = `include_contact_inboxes=false&page=${page}&sort=${sortAttr}`;
-  if (search) {
-    params = `${params}&q=${search}`;
-  }
-  if (label) {
-    params = `${params}&labels[]=${label}`;
-  }
-  return params;
-};
+export const buildContactParams = (page, sortAttr, label, search) => ({
+  include_contact_inboxes: false,
+  page,
+  sort: sortAttr,
+  ...(search ? { q: search } : {}),
+  ...(label ? { labels: [label] } : {}),
+});
 
 class ContactAPI extends ApiClient {
   constructor() {
@@ -18,13 +15,9 @@ class ContactAPI extends ApiClient {
   }
 
   get(page, sortAttr = 'name', label = '') {
-    let requestURL = `${this.url}?${buildContactParams(
-      page,
-      sortAttr,
-      label,
-      ''
-    )}`;
-    return axios.get(requestURL);
+    return axios.get(this.url, {
+      params: buildContactParams(page, sortAttr, label, ''),
+    });
   }
 
   show(id) {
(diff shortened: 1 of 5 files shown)


sony-mathew (approved):
Looks good to me. Please take a look at this finding:

## Findings

[P3] The localized creation-error fallback does not work for generic failures.

[ComposeConversation.vue](/worktrees/53f2/chatwoot/app/javascript/dashboard/components-next/NewConversation/ComposeConversation.vue:117) uses:

```js
parseAPIErrorResponse(error) || t(...)
```

However, [parseAPIErrorResponse](/worktrees/53f2/chatwoot/app/javascript/dashboard/store/utils/api.js:90) returns the original `Error`/`AxiosError` object when the response has no API message. Because that object is truthy, the new translated fallback is unreachable. It then reaches [Snackbar.vue](/worktrees/53f2/chatwoot/app/javascript/dashboard/components/Snackbar.vue:3), whose `message` prop requires a string.

The targeted duplicate-email `422` works because its response includes `message`, so this does not block the stated fix. A network or unexpected error can still display a technical error and trigger a Vue prop warning rather than “We couldn’t create the contact.” Normalize the parsed result to a string or use the translated fallback.

iamsivin:
> Looks good to me. Please take a look at this finding:
> 
> ## Findings
> [P3] The localized creation-error fallback does not work for generic failures.
> 
> [ComposeConversation.vue](/worktrees/53f2/chatwoot/app/javascript/dashboard/components-next/NewConversation/ComposeConversation.vue:117) uses:
> 
> ```js
> parseAPIErrorResponse(error) || t(...)
> ```
> 
> However, [parseAPIErrorResponse](/worktrees/53f2/chatwoot/app/javascript/dashboard/store/utils/api.js:90) returns the original `Error`/`AxiosError` object when the response has no API message. Because that object is truthy, the new translated fallback is unreachable. It then reaches [Snackbar.vue](/worktrees/53f2/chatwoot/app/javascript/dashboard/components/Snackbar.vue:3), whose `message` prop requires a string.
> 
> The targeted duplicate-email `422` works because its response includes `message`, so this does not block the stated fix. A network or unexpected error can still display a technical error and trigger a Vue prop warning rather than “We couldn’t create the contact.” Normalize the parsed result to a string or use the translated fallback.

@sony-mathew Fixed

PR #15296, opened 2026-08-03
chore: add divider to the article editor slash menu

iamsivin:
# Pull Request Template

## Description

This PR adds a **Divider** option to the Help Center article editor slash menu. Dividers were already supported by typing `---`, but there was no discoverable way to insert them. They can now be added directly from the `/` menu alongside Table, Image, and Video.

### How to test

1. Open any Help Center article in the editor.
2. Type `/` on a new line and select **Divider** (or type `/div` to filter).
3. Verify a divider is inserted and the cursor moves to the line below it.
4. Save and reload the article to confirm the divider is persisted.
5. Verify the existing `---` shortcut still inserts a divider.

Fixes https://linear.app/chatwoot/issue/CW-7797/insert-a-horizontal-divider-in-articles

## Type of change

- [x] New feature (non-breaking change which adds functionality)

### Screenshots
image
image

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #15291, opened 2026-08-02
perf: load i18n messages on demand in vitest setup

RianPolonini:
## Description

`vitest.setup.js` runs for every spec file and builds the i18n instance from the full message catalogue, so each of the 389 spec files resolves and transforms the 2537 JSON files under `app/javascript/dashboard/i18n/locale/`. With `pool: 'threads'`, that cost is paid per worker. The result is that `setup` takes about **60x longer than the tests themselves**.

Only one spec in the suite asserts on translated copy. This PR leaves the global i18n instance without messages and adds `withFullI18n`, an opt-in helper for the specs that need the real catalogue.

**On this repository's own CI, `setup` drops from ~629s to ~71s (-89%) and total Vitest duration from ~388s to ~196s (-49%).**

This is a performance and testability change, not a cosmetic one. It touches only how the test harness loads messages — **no translatable string and no locale file is modified**, so nothing changes for Crowdin contributors.

Worth noting: `vitest.config.ts` already excludes `**/i18n/**/*` from coverage, so loading the catalogue in every spec was not serving any metric.

Fixes #15290

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

### On this repository's CI

The numbers below come from the `test` job of `frontend-fe.yml` on GitHub Actions — this repo's own runners, not a local machine. Baseline is **60 successful `develop` runs** (23–31 Jul 2026), parsed from the Vitest summary line in each job log.

| Vitest metric | `develop` (n=60) | This PR | Delta |
|---|---|---|---|
| **setup** | median **629s** (min 329 / p25 607 / p75 641 / max 680) | **71s** | **-559s (-89%)** |
| **duration** | median **388s** (min 205 / p25 371 / p75 393 / max 418) | **196s** | **-192s (-49%)** |

None of the 60 `develop` runs beat this PR on either metric.

Same commit, both on this repo's CI:

```
develop @ bc7ae88   Test Files 389 passed (389)
                    Duration 323.35s (transform 13.75s, setup 522.48s, collect 42.53s,
                                      tests 9.02s, environment 187.05s, prepare 32.77s)

this PR             Test Files 389 passed (389)
                    Duration 196.20s (transform 18.18s, setup 70.67s, collect 65.97s,
                                      tests 12.22s, environment 251.28s, prepare 43.90s)
```

`setup` is the metric that isolates this change: it is a sum of work, so it is not distorted by how fast a given runner happens to be. `duration` improves less because it also covers transform, collect and environment, which this change does not touch.

[...]

PR #15298, opened 2026-08-03
fix(spec): i18n assertion failing for full editor spec

scmmishra:
Fixes the three `FullEditor.spec.js` slash-menu failures that appeared after #15291 and #15296 crossed on develop: #15296 added slash-menu specs that assert real translated labels ("Divider", "Heading 1"), while #15291 emptied the global i18n catalogue in the vitest setup, so `t()` started returning raw keys like `SLASH_COMMANDS.DIVIDER`.

The spec now calls `withFullI18n()` from the `test-i18n` helper introduced in #15291, opting into the full message catalogue the same way `MacroProperties.spec.js` does. Test-only change, no production code touched.

Related: https://github.com/chatwoot/chatwoot/pull/15291


iamsivin (approved):

sony-mathew (commented):
lgtm

PR #15300, opened 2026-08-03
feat(captain): add inactivity handling policies (3/3)

aakashb95:
Captain assistants can now choose how inactive conversations are handled: let Captain review the context, always resolve after a timer, or wait for the customer to return. Review mode can also send a context-aware follow-up before resolving, while keeping the existing resolution message setting separate.

## Closes

- [AI-163](https://linear.app/chatwoot/issue/AI-163)

## Depends on

This is a stacked draft PR based on and dependent on draft PR [#15215](https://github.com/chatwoot/chatwoot/pull/15215). Merge #15215 first.

## What changed

- Stores assistant-level inactivity policy, review timing, follow-up behavior, and second-stage resolution timing with backward-compatible defaults.
- Lets Captain resolve, generate a contextual follow-up, or hand off after reviewing conversation context.
- Cancels inactivity handling when the customer replies and uses locked, idempotent follow-up handling for retries.
- Adds always-resolve and wait-for-customer behavior, plus scheduler support for assistant-level policies.
- Adds the matching Captain settings UI with one inactivity save action and separate resolution and handoff message controls.

## How to test

1. Open a Captain V2 assistant's Settings page and expand **Inactive conversations**.
2. Select **Let Captain review it**, enable **Follow up before resolving**, set both durations, save, and reload. Confirm the saved summary and controls are restored.
3. Select **Always resolve** and confirm the warning and **Resolve after** timing are shown.
4. Select **Captain will handle it** and confirm inactivity timing, follow-up, and resolution message controls are hidden.
5. Exercise an inactive conversation in review mode and confirm Captain can resolve, follow up from context, or hand off; reply as the customer before the next stage and confirm the pending inactivity action is cancelled.

### DIFF
--- a/app/javascript/dashboard/components-next/captain/pageComponents/assistant/settings/AssistantSystemSettingsForm.vue
+++ b/app/javascript/dashboard/components-next/captain/pageComponents/assistant/settings/AssistantSystemSettingsForm.vue
@@ -8,6 +8,7 @@ import { useAccount } from 'dashboard/composables/useAccount';
 
 import Button from 'dashboard/components-next/button/Button.vue';
 import Editor from 'dashboard/components-next/Editor/Editor.vue';
+import RadioCard from 'dashboard/components-next/radioCard/RadioCard.vue';
 import Select from 'dashboard/components-next/select/Select.vue';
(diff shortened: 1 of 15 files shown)


aakashb95:
Closing for better tracking

PR #15310, opened 2026-08-03
fix(captain): prevent AgentBot inbox overlap

aakashb95:
Prevents Captain and AgentBot from being connected to the same inbox, so only one bot can own new conversations.

This PR is intended to merge before #14902. Existing inboxes that already have both bots connected are left unchanged.

## How to reproduce

1. Connect an AgentBot to an inbox.
2. Try to connect Captain to the same inbox.
3. Captain currently connects and both bots can reply to the same customer message.

The same conflict can be created in reverse by connecting an AgentBot to an inbox that already has Captain.

## What changed

- Rejects both conflicting connection paths in the backend.
- Hides inboxes with an active AgentBot from the Captain inbox picker.
- Shows the backend conflict message when an AgentBot connection is rejected.
- Allows existing connections to disconnect either bot.

### DIFF
--- a/app/controllers/api/v1/accounts/inboxes_controller.rb
+++ b/app/controllers/api/v1/accounts/inboxes_controller.rb
@@ -9,7 +9,7 @@ class Api::V1::Accounts::InboxesController < Api::V1::Accounts::BaseController
 
   def index
     @inboxes = policy_scope(Current.account.inboxes)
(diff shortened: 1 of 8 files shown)


aakashb95:
closing in favour of https://github.com/chatwoot/chatwoot/pull/15324

PR #15317, opened 2026-08-04
refactor(captain): route conversation completion by feature

aakashb95:
Conversation completion evaluations now use a dedicated internal LLM feature with GPT 4.1 as the default. The internal route keeps the completion model separate from the installation wide Captain model override and from the assistant route, which can use GPT 5.2 for Captain V2 accounts. Evaluations continue to use the installation API key and do not consume Captain response credits.

## What changed

Added an internal `conversation_completion` feature to the LLM model config and excluded internal features from account preferences, the Captain settings API, and Super Admin model overrides.

Updated `Captain::ConversationCompletionService` to resolve its model through `Llm::FeatureRouter`.

Added focused service and request coverage for model routing and settings visibility.


sony-mathew (approved):
Looks good. small finding (please check if it is valid):

- **P2 — New: the fixed internal model removes the operator compatibility seam.** [conversation_completion_service.rb:18](/worktrees/8403/chatwoot/enterprise/lib/captain/conversation_completion_service.rb:18) now ignores `CAPTAIN_OPEN_AI_MODEL`, while [llm.yml:72](/worktrees/8403/chatwoot/config/llm.yml:72) fixes the hidden route to `gpt-4.1`. An installation using another configured model—or a custom `CAPTAIN_OPEN_AI_ENDPOINT` that does not expose `gpt-4.1`—will start failing completion requests. Provider errors become `complete: false`, causing stale conversations to be handed off rather than evaluated normally.

  The isolation goal is valid, but the internal route needs an operator-level seam, such as a dedicated `CAPTAIN_CONVERSATION_COMPLETION_MODEL` defaulting to `gpt-4.1`, or an explicit support decision that every Captain endpoint must provide that model. The current test deliberately proves the installation override is ignored, but does not address endpoint compatibility.

aakashb95:
@codex review?

PR #15319, opened 2026-08-04
fix: handle ActionController::Parameters in days_before filter

vishnu-narayanan:
## Description

The `days_before` filter operator in `FilterService#days_before_filter_query` calls `with_indifferent_access` directly on the query hash. In production the filter payload arrives as `ActionController::Parameters` (the controller passes `params.permit!`), which does not respond to `with_indifferent_access`, so every conversation filter request using `days_before` raises `NoMethodError` and returns a 500.

Existing specs pass plain hashes with `with_indifferent_access`, which is why this was never caught. This PR normalizes the query hash via `to_h` first (permitted parameters convert to a `HashWithIndifferentAccess`, plain hashes are unaffected) and adds a regression spec that builds the payload as `ActionController::Parameters`, mirroring the controller.

Fixes https://linear.app/chatwoot/issue/CW-7832

## Type of change

## How Has This Been Tested?

- Added a regression spec that passes the filter payload as permitted `ActionController::Parameters` with the `days_before` operator. It reproduces the `NoMethodError` on the current develop branch and passes with this change.
- `bundle exec rspec spec/services/conversations/filter_service_spec.rb` (31 examples, 0 failures)
- `bundle exec rspec spec/services/conversations/filter_service_frontend_alignment_spec.rb` (10 examples, 0 failures)

## Checklist:

### DIFF
--- a/app/services/filter_service.rb
+++ b/app/services/filter_service.rb
@@ -99,7 +99,7 @@ class FilterService
 
   def days_before_filter_query(query_hash, current_index)
     date = Time.zone.today - query_hash['values'][0].to_i.days
-    updated_query_hash = query_hash.with_indifferent_access.merge(
+    updated_query_hash = query_hash.to_h.with_indifferent_access.merge(
       values: [date.strftime],
       filter_operator: 'is_less_than'
     )
--- a/spec/services/conversations/filter_service_spec.rb
+++ b/spec/services/conversations/filter_service_spec.rb
@@ -639,6 +639,23 @@ describe Conversations::FilterService do
           result = filter_service.new(params, user_1, account).perform
           expect(result[:conversations].length).to eq expected_count
         end
+
+        it 'filter by last_activity_at days_before when payload is ActionController::Parameters' do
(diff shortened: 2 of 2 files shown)


sony-mathew (approved):
Looks good. There are some non blocking things we can do as a followup to lock this down better:


- **P3 — Existing, matches Ito QA:** malformed `days_before` values are still coerced with `to_i`; `-1` produces a future cutoff and `"abc"` becomes zero/today, potentially returning an overly broad—but still tenant- and permission-scoped—result. This behavior exists on the PR base and is unchanged by the patch. See [filter_service.rb](/worktrees/9084/chatwoot/app/services/filter_service.rb:101).

Non-blocking follow-ups:

- Validate `days_before` as a strict integer within the UI-supported `1..998` range and raise `InvalidValue` otherwise.
- Correct the existing weak spec at [filter_service_spec.rb](/worktrees/9084/chatwoot/spec/services/conversations/filter_service_spec.rb:626), which sends `3` days but calculates its expectation using `2` days.

PR #15321, opened 2026-08-04
perf: reduce per-request work in conversation filter endpoint

vishnu-narayanan:
## Description

Every conversation filter request runs three separate unbounded COUNT queries over the filtered set (mine, unassigned, all) and eager-loads every message of every conversation on the page. On large accounts this adds a fixed 1-2s of latency per request regardless of the filter.

This PR trims both:

- The three counts are now computed in a single pass using `COUNT(*) FILTER (...)` aggregates. Response shape and count semantics are unchanged.
- The `:messages` eager-load is removed from the filter base relation. The list payload fetches messages through scoped queries (last message, last non-activity message, unread messages), which never read the preloaded collection, so it was loaded and discarded on every request.

Fixes https://linear.app/chatwoot/issue/CW-7830

## Type of change

## How Has This Been Tested?

- New service specs assert the mine/assigned/unassigned/all counts are unchanged, that the counts run as a single query, and that an empty permission scope returns zero counts.
- `bundle exec rspec spec/services/conversations/filter_service_spec.rb spec/services/conversations/unread_counts/filter_query_counter_spec.rb` (34 examples, 0 failures) and the conversations controller filter specs (4 examples, 0 failures).

## Checklist:

### DIFF
--- a/app/services/conversations/filter_service.rb
+++ b/app/services/conversations/filter_service.rb
@@ -24,8 +24,10 @@ class Conversations::FilterService < FilterService
   end
 
   def base_relation
+    # :messages is deliberately not preloaded: the list payload fetches messages through
+    # scoped queries (last message, last_non_activity_message), which bypass the preload.
     conversations = @account.conversations.includes(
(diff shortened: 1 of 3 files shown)


sony-mathew (approved):
lgtm

PR #15322, opened 2026-08-04
fix: prevent command bar crash and hide inaccessible commands

iamsivin:
# Pull Request Template

## Description

The command bar (`⌘K`) no longer crashes on accounts that have hit their plan limits, and now only shows destinations the user can actually access. Commands are filtered using the same route policy as the sidebar, so they're hidden when the user lacks permission, the feature is disabled, or the route isn't available for the current installation. It also adds the remaining sidebar destinations, including My Inbox, Captain, Calls, Campaigns, Help Center, Templates, Data, CSAT/Bot/SLA reports, Automation, Macros, Custom Attributes, Audit Logs, and Billing, using the same icons as the sidebar.

### How to reproduce

1. On a Cloud account that has exceeded a plan limit (for example, agents), open the dashboard with DevTools open.
2. Notice the console logs `Cannot read properties of undefined (reading 'focus')` from `@chatwoot/ninja-keys`.
3. Press `⌘K`. The command bar shows pages like Conversations, Contacts, and Reports even though they're inaccessible behind the upgrade page.
4. Sign in as a custom role without Automation access, open `⌘K`, and search for `au`.
5. "Go to Automation" and "Go to Audit Logs" are still listed even though the role cannot access them.

Fixes https://linear.app/chatwoot/issue/CW-7725/typeerror-cannot-read-properties-of-undefined-reading-focus
https://chatwoot-p3.sentry.io/issues/6471888378/?project=4507182691975168&referrer=Linear

## Type of change

#### Screenshots (showing console error)

Screenshot 2026-08-04 at 3 10 31 PM

#### Loom videos (after the fix)
https://www.loom.com/share/6d2fac642430468d95f5dd0238b1727b
https://www.loom.com/share/3644aeed469b4b2eb45a9adf828a5097

#### Steps
* **Paywalled Cloud account:** Open the dashboard and verify there are no `focus` errors in the console. Press `⌘K` and confirm only accessible destinations are listed.
* **Custom role:** Sign in with a role that lacks Automation or Audit Logs permissions. Verify those commands are not shown.
* **Feature flags:** Disable a feature such as Help Center and confirm its command disappears.
* **Community edition:** Verify Cloud/Enterprise-only destinations are not listed.
* **Application Settings:** Confirm "Go to Application Settings" navigates correctly.

## Checklist:


sony-mathew (approved):
lgtm.

One non-blocking finding:

**[P3] Add coverage for the actual component lifecycle regression.**

The fix depends on keeping `<CommandBar>` mounted when the upgrade page replaces the routed dashboard content in [Dashboard.vue](/worktrees/12a6/chatwoot/app/javascript/dashboard/routes/dashboard/Dashboard.vue:145). The new tests thoroughly cover command filtering, but none mount `Dashboard.vue`, toggle the paywall state, and press `Cmd/Ctrl+K`.

The implementation is correct, but a future move back inside the conditional could restore the crash without failing the current suite. A component or browser test covering that transition would close the gap. This is **non-blocking**.

PR #15331, opened 2026-08-05
fix: validate days_before filter values before date arithmetic

vishnu-narayanan:
## Description

Follow-up to #15319, addressing the non-blocking review notes and the Ito QA finding there.

The `days_before` filter value went straight through `to_i`, which never fails: `-1` moves the cutoff into the future and matches every conversation, and a non-numeric string becomes `0`. The value is now parsed as a base-10 integer and must fall within the UI-supported `1..998` range; anything else raises the existing `CustomExceptions::CustomFilter::InvalidValue`, which the controller already turns into a client error.

Also corrects an existing weak spec that sent `3` days but computed its expectation with `2` days, passing only because the seeded data made both counts equal. It now sends `2` and exercises the exclusive boundary.

Refs https://linear.app/chatwoot/issue/CW-7832

## Type of change

## How Has This Been Tested?

- New specs: invalid values (`-1`, `abc`, `0`, `999`) raise `InvalidValue`; string values parse as base 10 (`'02'` means 2 days, not octal).
- `bundle exec rspec spec/services/conversations/filter_service_spec.rb` (36 examples, 0 failures).

## Checklist:

### DIFF
--- a/app/services/filter_service.rb
+++ b/app/services/filter_service.rb
@@ -98,7 +98,12 @@ class FilterService
   end
 
   def days_before_filter_query(query_hash, current_index)
-    date = Time.zone.today - query_hash['values'][0].to_i.days
+    days = Integer(query_hash['values'][0].to_s, 10, exception: false)
+    # UI supports 1..998 days; anything else would silently coerce into an
+    # unintended cutoff (negative => future date matching everything)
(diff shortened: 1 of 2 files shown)


sony-mathew (approved):
lgtm

PR #15332, opened 2026-08-05
fix: prevent avatar upload overlay from showing in conversation list

iamsivin:
# Pull Request Template

## Description

This PR fixed the conversation list was incorrectly rendering the avatar upload overlay on every conversation card, even though uploads aren't supported there. Clicking it could throw a `TypeError: Cannot read properties of null (reading 'click')`. Conversation cards now only show the selection checkbox, while avatar uploads continue to work everywhere they're supported.

### Cause

`ConversationCard` always passed the `#overlay` slot, but the checkbox inside it was wrapped in `v-if`. When the checkbox wasn't rendered, Vue treated the slot as empty and fell back to the default upload overlay from `Avatar`.

That overlay's click handler expects a file input, but the file input is only rendered when `allowUpload` is enabled. Since conversation cards never enable uploads, clicking the overlay could dereference a null file input and throw.

### How to reproduce

This isn't reliably reproducible manually. It only happens when the upload overlay becomes visible while the card's internal hover state is out of sync with the browser's CSS `:hover` state. In normal interaction, entering the card immediately updates the hover state and shows the checkbox instead, so the issue effectively self-recovers.

The new test reproduces this state directly and verifies the fix.

## What changed

* Moved `v-if="allowUpload"` from the hidden file input to the upload overlay itself, so the overlay and file input are always mounted together.
* Added `Avatar.spec.js` coverage for the overlay slot, including the empty-slot case that triggered this bug, along with the existing upload, delete, badge, sizing, initials, and image fallback behavior.

Fixes https://linear.app/chatwoot/issue/CW-7726/typeerror-cannot-read-properties-of-null-reading-click
https://chatwoot-p3.sentry.io/issues/7291677410/?project=4507182691975168&referrer=Linear

## Type of change

### Screenshots

**Before**
image

**After**
image

## Checklist:

### DIFF
--- a/app/javascript/dashboard/components-next/avatar/Avatar.vue
+++ b/app/javascript/dashboard/components-next/avatar/Avatar.vue
@@ -280,6 +280,7 @@ watch(
         :handle-image-upload="handleImageUpload"
       >
         <div
+          v-if="allowUpload"
           class="absolute inset-0 z-10 flex items-center justify-center invisible w-full h-full transition-all duration-300 ease-in-out opacity-0 bg-n-alpha-black1 group-hover/avatar:visible group-hover/avatar:opacity-100"
           :class="borderRadiusClass"
           @click="handleUploadAvatar"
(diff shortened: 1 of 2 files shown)


muhsin-k (approved):

sony-mathew (approved):
lgtm

PR #15358, opened 2026-08-07
feat: back off IMAP fetch for repeatedly failing inboxes

vishnu-narayanan:
# Pull Request Template

## Description

The IMAP fetch scheduler enqueues a fetch job for every IMAP-enabled inbox each minute. When a remote IMAP server is unreachable or misconfigured (connection refused, timeouts, SSL or protocol errors), the job logs the error and retries on the next tick indefinitely, roughly 1,440 failed connection attempts per day per broken inbox. Mail providers can read this pattern of repeated failed logins as brute force and block or report the sender IP.

This PR adds a per-channel backoff for failing IMAP fetches:

- New columns on `channel_email`: `imap_fetch_error_count` and `imap_fetch_paused_till`.
- Connection-level fetch failures increment the counter and pause scheduling with an escalating backoff: first two failures are tolerated (no pause), 3rd pauses 5 minutes, 4th pauses 15 minutes, 5th onwards 60 minutes (cap, which also acts as the probe cadence for permanently broken remotes).
- `Inboxes::FetchImapEmailInboxesJob` skips paused channels at enqueue time, and `Inboxes::FetchImapEmailsJob` re-checks the pause so already-queued jobs also skip.
- Backoff state is persisted with `update_columns` so internal retry bookkeeping does not generate audit log entries.
- OAuth fetch failures also count toward the backoff (the existing reauthorization flow is unchanged).
- A successful fetch resets the backoff state. Lock contention (`LockAcquisitionError`) is not counted as a failure.
- Updating IMAP settings or reauthorizing the channel clears the backoff, so a fixed configuration resumes syncing immediately. Routine OAuth token refreshes do not clear it.

Fixes https://linear.app/chatwoot/issue/CW-7881

## Type of change

## How Has This Been Tested?

- `bundle exec rspec spec/jobs/inboxes/ spec/models/channel/email_spec.rb` covering: skip when paused, no pause on first two failures, 5/15/60 minute escalation and cap, reset on success, lock failures not counted, reset on IMAP settings change.

## Checklist:

### DIFF
--- a/app/jobs/inboxes/fetch_imap_email_inboxes_job.rb
+++ b/app/jobs/inboxes/fetch_imap_email_inboxes_job.rb
@@ -18,6 +18,7 @@ class Inboxes::FetchImapEmailInboxesJob < ApplicationJob
     return false if inbox.account.suspended?
     return false unless inbox.channel.imap_enabled
     return false if inbox.channel.reauthorization_required?
+    return false if inbox.channel.imap_fetch_paused?
 
     return true unless ChatwootApp.chatwoot_cloud?
     return false if default_plan?(inbox.account)
--- a/app/jobs/inboxes/fetch_imap_emails_job.rb
+++ b/app/jobs/inboxes/fetch_imap_emails_job.rb
@@ -10,7 +10,7 @@ class Inboxes::FetchImapEmailsJob < MutexApplicationJob
 
     fetch_mails_with_lock(channel, interval)
   rescue *ExceptionList::IMAP_EXCEPTIONS => e
-    Rails.logger.error "Authorization error for email channel - #{channel.inbox.id} : #{e.message}"
+    Rails.logger.error "Fetch error (network/protocol) for email channel - #{channel.inbox.id} : #{e.message}"
     channel.imap_fetch_error!
   rescue IOError, OpenSSL::SSL::SSLError, Net::IMAP::NoResponseError, Net::IMAP::BadResponseError, Net::IMAP::InvalidResponseError,
          Net::IMAP::ResponseParseError, Net::IMAP::ResponseReadError, Net::IMAP::ResponseTooLargeError => e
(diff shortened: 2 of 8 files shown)


sony-mathew (approved):
Looks good overall. Please take a look at these findings:

## Findings

1. **P2 — Concurrent failures can lose increments and miss the pause threshold** — matches the existing unresolved same-head thread. [`imap_fetch_error!`](/worktrees/2252/chatwoot/app/models/channel/email.rb:75) reads and writes in separate queries without locking or atomic incrementing. My coordinated probe reproduced `expected_count: 3, actual_count: 2, pause_present: false`. The worker mutex reduces normal overlap, but does not eliminate it: its five-minute TTL can expire during an unbounded IMAP operation, and the mutex implementation explicitly permits overlapping ownership after expiry.

2. **P2 — The PR does not implement the customer-facing recovery requirements from CW-7881** — new finding. Linear requires showing the inbox as unhealthy and providing “sync now” to reset/retry. The PR persists and enforces the pause at [`email.rb:70`](/worktrees/2252/chatwoot/app/models/channel/email.rb:70), but the inbox response at [`_inbox.json.jbuilder:79`](/worktrees/2252/chatwoot/app/views/api/v1/models/_inbox.json.jbuilder:79) exposes no backoff/health state, and there is no manual-sync endpoint or dashboard action. Consequently, an inbox can silently stop receiving mail for up to an hour while administrators see no reason and cannot immediately retry after fixing an external server.

3. **P3 — Updating an unrelated SMTP setting clears IMAP backoff** — new finding. `Channel::Email#reauthorized!` unconditionally clears backoff at [`email.rb:87`](/worktrees/2252/chatwoot/app/models/channel/email.rb:87), while the public inbox update path invokes `reauthorized!` after every channel update at [`inboxes_controller.rb:135`](/worktrees/2252/chatwoot/app/controllers/api/v1/accounts/inboxes_controller.rb:135). A request changing only `smtp_address` reset a stored count of 5 to 0 and removed its future pause. This contradicts the new model spec’s assertion that unrelated attribute updates preserve backoff.

PR #15361, opened 2026-08-07
feat: classify outbound email send errors and add deferred message state [1/2]

vishnu-narayanan:
# Pull Request Template

## Description

Outbound email send failures are currently all treated the same way: the sender rescues `StandardError`, reports it, and marks the message failed with no distinction between a temporary "slow down" and a permanent "wrong password / bad address". This PR lays the foundation for handling those cases differently.

It adds:

- `ProviderErrorClassifier`, a small pure class that buckets a provider error (SMTP / OAuth / HTTP) into one of `:throttle`, `:transient`, `:auth`, `:permanent`, or `:unknown`. It is robust to `nil`/unknown input and never raises. SMTP reply codes are parsed only from SMTP errors, so an HTTP/OAuth message that happens to contain a status code is not misread as an SMTP code. Note: Gmail returns "454 4.7.0 Too many login attempts" as an authentication error class but it is semantically a throttle, so a 454 classifies `:throttle` while a 535 classifies `:auth`.
- A new `deferred` message status, appended to the enum with an explicit integer so existing values keep their numbers, plus the frontend status mapping (constant, icon, colour, tooltip) so the UI is ready to render it.

The `deferred` status is introduced as foundation only. Nothing sets it in this change and it is intentionally rejected by the public status-update path; the transition into it is wired up by a follow-up (CW-7882), which also consumes the classifier. No retry behaviour is implemented here.

**Merge order:** this is PR 1 of 2. The follow-up that wires up retry (CW-7882) is stacked on top of this branch. Merge this one first; the follow-up will retarget to `develop` automatically once this is merged.

Fixes https://linear.app/chatwoot/issue/CW-7880

## Type of change

## How Has This Been Tested?

- New RSpec unit specs for `ProviderErrorClassifier` covering one example per branch, the 454-vs-535 distinction, the HTTP/OAuth-code regression cases, and `nil` input.
- Added an `Messages::StatusUpdateService` spec asserting `deferred` cannot be set through the update path.
- `bundle exec rspec` on the changed specs and the message model spec, all green; `bundle exec rubocop` clean on changed Ruby files.

## Checklist:

### DIFF
--- a/app/javascript/dashboard/components-next/message/MessageMeta.vue
+++ b/app/javascript/dashboard/components-next/message/MessageMeta.vue
@@ -123,6 +123,7 @@ const isRead = computed(() => {
 });
 
 const statusToShow = computed(() => {
+  if (status.value === MESSAGE_STATUS.DEFERRED) return MESSAGE_STATUS.DEFERRED;
   if (isRead.value) return MESSAGE_STATUS.READ;
   if (isDelivered.value) return MESSAGE_STATUS.DELIVERED;
   if (isSent.value) return MESSAGE_STATUS.SENT;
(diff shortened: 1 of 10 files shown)


salmonumbrella:
@vishnu-narayanan can you let us forward email in Chatwoot?

PR #15371, opened 2026-08-10
fix: Use dark text on amber solid buttons so the disabled Add Note button stays visible

### LINKED ISSUE #15221: Private Note "Add Note" button hard to see (poor contrast)
### Describe the bug

In **Private Note** mode (dark theme), the yellow **Add Note** button blends into the brown note composer background and is hard to see.

Screenshot attached.

screenshot

### To Reproduce

1. Open a conversation
2. Switch to **Private Note**
3. Type a note
4. Try to find **Add Note** (bottom right)

### Expected behavior

The button should have clear contrast against the note background.

### Environment

Self-hosted, Chatwoot **4.16.1**, browser on Linux

limehawk:
## Description

In the private note composer (dark theme), the disabled **Add Note** button is almost invisible. Two things combine to cause this:

1. The amber solid button variant uses `text-white` on `bg-n-amber-9` (`#FFC53D`). White on that yellow is a **1.58:1** contrast ratio.
2. The base button class applies `disabled:opacity-50`. The half-transparent amber fill then composites into the amber-tinted note background (`--solid-amber`, `rgb(56 50 41)` in dark). The fill and the background share the same hue, so the button camouflages and the dimmed white label becomes illegible.

This PR changes the amber solid variant from `text-white` to `text-n-black` (an existing fixed token, already used in `components-next`). Black on `amber-9` is **13.31:1** (12.29:1 on the `amber-10` hover state). The disabled state keeps the standard 50% dim, but the dark label stays readable against the golden fill in both themes. Bright ambers take dark foreground text — this matches Radix Colors guidance for amber solid backgrounds.

This is a design-token-level fix, so it also corrects the other amber solid buttons (call mute state, Slack channel warning) which had the same white-on-yellow text.

Fixes #15221

## Type of change

## How Has This Been Tested?

- Reproduced the reported state: dark theme, private note mode, empty composer (disabled button).
- Rendered the exact token values (`--amber-9`, `--solid-amber` dark) with the current and fixed classes in headless Chromium and compared enabled + disabled states in dark mode.
- Verified contrast ratios (WCAG relative luminance): white on `#FFC53D` = 1.58:1 (fails AA); black on `#FFC53D` = 13.31:1 (passes AAA); black on hover `amber-10` = 12.29:1 light / 14.88:1 dark.

## Checklist:


app/javascript/dashboard/components-next/button/Button.vue
```diff
@@ -122,7 +122,7 @@ const STYLE_CONFIG = {
     },
     amber: {
       solid:
-        'bg-n-amber-9 text-white hover:enabled:bg-n-amber-10 focus-visible:bg-n-amber-10 outline-transparent',
+        'bg-n-amber-9 text-n-black hover:enabled:bg-n-amber-10 focus-visible:bg-n-amber-10 outline-transparent',
```
iamsivin:
```suggestion
        'bg-n-amber-9 text-n-amber-12 dark:text-n-amber-3 hover:enabled:bg-n-amber-10 focus-visible:bg-n-amber-10 outline-transparent',
```

iamsivin (approved):
@limehawk Thanks for the PR.

Image

Image

Image

Image

PR #15374, opened 2026-08-10
feat(telegram): track business connection state

muhsin-k:
Telegram inboxes can process Business messages today, but Chatwoot does not retain whether a bot supports Business connections or whether its account connection is currently enabled. This change records that provider state on the Telegram channel so future inbox UI and health handling can use a reliable source of truth. Existing message delivery and conversation routing remain unchanged.

### Things to know

- This is a backend-only change; it does not add a Business status indicator to the inbox UI.
- A bot's `can_connect_to_business` capability is distinct from an active Business account connection.
- Connection IDs remain on conversations as routing snapshots.
- If Telegram reports multiple active Business connections, Chatwoot retains them and records a channel configuration error.

### How to test

1. Create a Telegram inbox using a bot with Business mode enabled in BotFather.
2. Connect that bot to a Telegram Business account.
3. Send a customer message to the Business account and confirm it appears in Chatwoot.
4. Reply from Chatwoot and confirm the response is delivered through the Business account.
5. Disconnect or change the Business connection and confirm subsequent Telegram connection events are accepted without affecting standard Telegram messages.

### What changed

- Stores the bot's Business capability during Telegram token validation.
- Handles Telegram `business_connection` lifecycle webhooks.
- Recovers unknown connection state through `getBusinessConnection` when the first Business message arrives.
- Persists normalized connection identity, enabled state, and rights on `channel_telegram`.


sony-mathew (approved):
Looks good to me overall. please check the following findings:

## Findings

- **P2 — New: valid messages from separate Business connections can collide and be silently dropped.** The deduplication key in [incoming_message_service.rb](/worktrees/2fc7/chatwoot/app/services/telegram/incoming_message_service.rb:97) uses only `message.source_id` within the customer’s `ContactInbox`. It ignores the conversation’s `business_connection_id`. Telegram message IDs are scoped to a chat, and private Business messages use the connected business user’s message-ID sequence—not a bot-global sequence. [Telegram Bot API](https://core.telegram.org/bots/api?source=post_page), [Telegram update documentation](https://core.telegram.org/api/updates).

  I reproduced this with two otherwise valid Business messages:

  - Same customer/chat and `message_id: 77`
  - Different connections: `business-a` and `business-b`
  - Different bot-wide updates: `1001` and `1002`
  - Different contents

  Head stored only the first: `{message_count: 1, contents: ["hello"], connection_ids: ["business-a"]}`. This contradicts the PR’s explicit support for retaining multiple connections and can lose customer messages. Namespace the duplicate lookup by `business_connection_id` plus `message_id`; keep `source_id` itself provider-compatible because it is also used for Telegram reply IDs.

- **P2 — Matching the current open thread: malformed messages persist with a blank identity and then collide.** [incoming_message_service.rb](/worktrees/2fc7/chatwoot/app/services/telegram/incoming_message_service.rb:26) persists `telegram_params_message_id.to_s`, while [param_helpers.rb](/worktrees/2fc7/chatwoot/app/services/telegram/param_helpers.rb:97) can return `nil`. The deduplication query consequently treats every missing ID as `source_id: ""`.

  Independent reproduction produced `{message_count: 1, source_ids: [""]}` for two malformed deliveries with different `update_id`s. This corroborates the [existing open Ito thread](https://github.com/chatwoot/chatwoot/pull/15374#discussion_r3755903036). Reject or skip blank message identities before contact/conversation creation and persistence.

PR #15386, opened 2026-08-10
fix: scope Captain assistant responses to account

aakashb95:
## Description

Fixes account scoping for Captain assistant responses.

Create and update accepted an `assistant_id` from the request. The model then set the response account from that assistant. The controller lookup read the top level parameter, while the API sends the ID inside `assistant_response`, and create did not use the lookup result.

The controller now resolves the nested assistant ID through `Current.account`, removes `assistant_id` before assigning request fields, and assigns the scoped assistant directly. The model now fills the account only when it is blank and rejects a response when its account and assistant do not match.

Linear issue: [CW-7913](https://linear.app/chatwoot/issue/CW-7913/ghsa-phpm-m2mf-r8r9-captain-assistant-responses-writes-into-another)

## Type of change

- [x] Bug fix

## How has this been tested?

- Ran `bundle exec rspec spec/enterprise/controllers/api/v1/accounts/captain/assistant_responses_controller_spec.rb spec/enterprise/models/captain/assistant_response_spec.rb`. All 19 examples passed.
- Ran the two new account isolation examples against the original code. Both failed and reproduced the create and update issue. Both pass with this fix.
- Ran RuboCop on the five changed Ruby files. It found no offenses.

## Checklist

- [x] I have performed a self review of my code
- [x] I have added tests that prove the fix is effective


aakashb95:
@codex review please

sony-mathew (approved):
lgtm

sony-mathew:
@aakashb95 , we can run a script to check if there are any occurrences of responses belonging to assistant not in the same account after this is deployed, just to be sure.

aakashb95:
> @aakashb95 , we can run a script to check if there are any occurrences of responses belonging to assistant not in the same account after this is deployed, just to be sure.

will do

PR #15391, opened 2026-08-10
fix: allow resizing the reply editor while a Copilot suggestion is active

iamsivin:
# Pull Request Template

## Description

This PR fixes an issue where the reply editor's drag handle did nothing while a Copilot suggestion or the "Copilot is thinking" loader was visible.

The handle now resizes the Copilot section the same way it resizes the normal editor. The reply box also keeps the same height when switching between the editor, loader, and suggestion states.

Fixes https://linear.app/chatwoot/issue/CW-7902/i-cant-change-the-size-of-the-editor-when-the-ai-suggestion-is-active

## Type of change

### Screencast

https://github.com/user-attachments/assets/9ef45f12-0b97-4e2b-9bcb-453e4d4773fd

### Steps to reproduce

1. Open a conversation and trigger a Copilot reply suggestion.
2. Drag the handle above the reply box. Nothing moves.
3. Discard or accept the suggestion. The reply box jumps to a different height.

## Checklist:

### DIFF
--- a/app/javascript/dashboard/components/widgets/WootWriter/CopilotEditor.vue
+++ b/app/javascript/dashboard/components/widgets/WootWriter/CopilotEditor.vue
@@ -203,16 +203,16 @@ onMounted(() => {
 </script>
 
 <template>
-  <div class="space-y-2 mb-4">
-    <div class="overflow-y-auto max-h-56">
+  <div class="resizable-editor-body flex flex-col gap-2 mb-3">
+    <div class="flex-1 min-h-0 overflow-y-auto">
       <p
         v-dompurify-html="formatMessage(generatedContent, false)"
-        class="text-n-iris-12 text-sm prose-sm font-normal !mb-4"
+        class="text-n-iris-12 text-sm prose-sm font-normal"
       />
     </div>
-    <div class="editor-root relative editor--copilot space-x-2">
+    <div class="editor-root relative editor--copilot shrink-0">
       <div ref="editor" />
-      <div class="flex items-center justify-end absolute right-2 bottom-2">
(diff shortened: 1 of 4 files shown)


sony-mathew (approved):
looks good to me. please check this finding - not sure if it's valid. (not sure if we support 80px width body)

- **P2 — New: minimum resize hides the generated Copilot suggestion.** [CopilotEditor.vue](/worktrees/381a/chatwoot/app/javascript/dashboard/components/widgets/WootWriter/CopilotEditor.vue:206) places the suggestion and the fixed 80px follow-up editor inside one resizable body. [ResizableEditorWrapper.vue](/worktrees/381a/chatwoot/app/javascript/dashboard/components/widgets/conversation/ResizableEditorWrapper.vue:12) permits that body to shrink to 80px.

  Browser reproduction on the exact head:

  1. Generate a long Copilot suggestion.
  2. Drag the reply handle downward to its minimum.
  3. The body becomes 80px, the follow-up editor consumes all 80px, and the suggestion pane becomes 0px high.
  4. The generated reply disappears and the follow-up placeholder is visibly clipped.

  Measurements at minimum: `bodyHeight=80px`, `suggestionClientHeight=0`, `suggestionScrollHeight=924`, `editorHeight=80px`.

  This is PR-introduced: base kept the suggestion as a separately visible `max-h-56` block and did not resize the composite Copilot state. The effective minimum for the generated-suggestion state needs to accommodate both a usable suggestion region and the follow-up editor, with the JavaScript resize bounds and CSS kept synchronized. Add a regression test covering the minimum boundary.

aakashb95 (approved):

PR #15408, opened 2026-08-11
fix(inboxes): show facebook inbox creation errors

muhsin-k:
Facebook inbox creation failures in Settings could leave users on the same screen without explaining what went wrong. This change preserves the backend error response and displays it through the existing toast mechanism, making account-limit and other creation failures visible and actionable.

The shared action now passes the original API error through, so the newer onboarding flow also receives the specific backend message. Successful Facebook inbox creation remains unchanged.

### How to reproduce

1. Use an account that has reached its inbox limit.
2. Go to Settings → Inboxes → Add Inbox → Facebook.
3. Select a Facebook Page and create the inbox.
4. Previously, the loading state ended without any visible error.

### How to test

1. Attempt Facebook inbox creation while the backend returns an HTTP 402 account-limit error.
2. Confirm the returned error appears as a toast message.
3. Retry after increasing the account limit and confirm inbox creation continues normally.

### Things to know

The toast falls back to a generic localized creation error when the API response does not contain a displayable message.


iamsivin (approved):
Code LGTM

PR #15412, opened 2026-08-11
perf: scale sidekiq worker MemoryMax with host memory

vishnu-narayanan:
## Description

This systemd unit is installed on every self-hosted Linux install (copied by `deployment/setup_18.04.sh` / `setup_20.04.sh`) as well as our own worker hosts, so the memory cap has to be correct across a wide range of box sizes.

It is currently a fixed `MemoryMax=1.2G`, chosen when workers ran on 2GB hosts (#12915). On larger hosts that fixed cap sits at a high fraction of the limit under normal load, so transient spikes hit the hard limit and the worker is OOM-restarted more often than necessary (each restart drops the jobs in-flight at that instant).

Switching to `MemoryMax=60%` makes the cap scale with the host's physical memory instead of being fixed:

- 2GB host: 60% = 1.2G, identical to today (no regression for small installs)
- 4GB host: 60% = ~2.4G, room to absorb spikes without a premature restart
- larger hosts scale proportionally

`MemoryHigh=infinity` is unchanged, so the worker is still never throttled by cgroup reclaim (kept intentionally since #12871 to avoid reclaim-induced worker stalls). Only the hard cap changes, and only its scaling.

Fixes https://linear.app/chatwoot/issue/INF-100

## Type of change

- [x] Performance (non-breaking change which improves performance)

## How Has This Been Tested?

Verified on a running 4GB worker host: the worker cgroup sits at ~1.0G under normal load against the current 1.2G cap (~80%), with the rest of box memory unused. `60%` yields ~2.4G on that host while resolving to exactly 1.2G on a 2GB host, matching the value the fixed cap was originally chosen for. No change to throttling behaviour.

## Checklist:

### DIFF
--- a/deployment/chatwoot-worker.1.service
+++ b/deployment/chatwoot-worker.1.service
@@ -17,7 +17,7 @@ StandardInput=null
 SyslogIdentifier=%p
 LimitNOFILE=65536
 
-MemoryMax=1.2G
+MemoryMax=60%
 MemoryHigh=infinity
 MemorySwapMax=0
 OOMPolicy=stop


sony-mathew (approved):
Looks good. One question though: why 60%? Why not 75% or 80%. The free space becomes pronounced in larger machines.

vishnu-narayanan:
Good question. I used  60% because this same unit ships to self-hosted single-box installs where Postgres, Redis, web and the worker all share one host, so the shared default has to stay conservative (75-80% would starve the co-resident services). 60% of 2GB is also exactly the current 1.2G cap, so small installs don't regress.
 
On dedicated sidekiq hosts you're right that it leaves headroom, but 60% of 4GB is already ~2.4G vs the ~1.0G a worker uses at load, so a higher % wouldn't change anything. We can do a followup based on how this go.

PR #15413, opened 2026-08-11
fix(voice): sync Twilio call accept/end state across tabs

tds-1:
## Summary

When an agent has multiple tabs/windows open and picks up (or ends) a Twilio voice call in one of them, the other tabs never find out — they keep ringing and showing the call as still incoming, even though it's already been answered or has ended elsewhere.

Chatwoot already broadcasts this kind of accept/end state for WhatsApp calls so every other tab clears its ringing UI, but the equivalent never existed for Twilio — neither the backend event nor the frontend listener.

## How to reproduce

1. Log into the same agent account in two browser tabs.
2. Receive an inbound Twilio call (rings in both tabs).
3. Accept the call in one tab.
4. The other tab keeps ringing and shows the call as still incoming.

Same issue in reverse when discarding/ending a call from one tab — the others keep showing it as active/ringing.

## What changed

- `Api::V1::Accounts::ConferenceController#create`/`#destroy` now broadcast `voice_call.accepted`/`voice_call.ended` account-wide for Twilio calls, mirroring the existing `Whatsapp::CallService` broadcast shape.
- `onVoiceCallAccepted`/`onVoiceCallEnded` in `actionCable.js` no longer ignore non-WhatsApp providers — they now route through the right local-ownership check per provider before deciding whether to tear the call down in the current tab.
- Added a Twilio-side `markLocalCall`/`isLocalCall`/`clearLocalCall` (in `dashboard/helper/voice.js`), the equivalent of WhatsApp's `activeCallId`, set synchronously before the join API call so the account-wide broadcast can't race ahead of the joining tab's own state and tear down its live call.


iamsivin (approved):
@tds-1 Code LGTM, see if you can add a loom video.
cc @sony-mathew

sony-mathew (commented):
please check the following 2 P2 findings:

- **P2 — still-present previously resolved concern: an ended call can emit a later accepted event.**  
  [`ConferenceController#destroy`](/worktrees/03d0/chatwoot/enterprise/app/controllers/api/v1/accounts/conference_controller.rb:30) broadcasts `voice_call.ended` after provider teardown without first persisting a terminal status for an in-progress call. A delayed participant-join then passes the non-terminal check in [`mark_accepted_broadcast!`](/worktrees/03d0/chatwoot/enterprise/app/services/voice/conference/manager.rb:62) and emits `voice_call.accepted`.  
  I reproduced the sequence on the exact head: `events: ["voice_call.ended", "voice_call.accepted"]`, with the call still persisted as `in_progress`. This contradicts the resolved Ito teardown thread claiming commit `d7923f5` fixed it.

- **P2 — new: simultaneous acceptance in two tabs can leave the losing tab ringing.**  
  Each tab records ownership using only the shared call SID in [`voice.js`](/worktrees/03d0/chatwoot/app/javascript/dashboard/helper/voice.js:30). If both tabs begin joining before the callback arrives, both consider the accepted event local and return without removing the call at [`actionCable.js`](/worktrees/03d0/chatwoot/app/javascript/dashboard/helper/actionCable.js:391). The backend permits repeated claims by the same agent in [`ConferenceService`](/worktrees/03d0/chatwoot/enterprise/app/services/voice/provider/twilio/conference_service.rb:28), which I confirmed in the isolated database.  
  Twilio requires participant labels to be unique, so one same-agent leg is rejected once the other becomes active. The non-409 catch path clears local ownership but leaves the existing ringing card in the store at [`useCallSession.js`](/worktrees/03d0/chatwoot/app/javascript/dashboard/composables/useCallSession.js:167). Thus the PR’s original stale-tab symptom survives this realistic race. [Twilio’s error documentation confirms duplicate participant labels are rejected.](https://www.twilio.com/docs/api/errors/16025)

tds-1:
@sony-mathew #1: Fixed — destroy now finalizes the call to a terminal status under call.with_lock before broadcasting :ended, closing the terminal? gate a delayed join webhook needs.
#2: Not fixed — same-agent double-claim + tab-local isLocalCall let the losing tab miss its own cleanup; needs cross-tab coordination or a per-attempt token, will track as follow-up.

PR #15414, opened 2026-08-11
fix: prevent reopening resolved conversations from widget

iamsivin:
# Pull Request Template

## Description

This PR fixes resolved conversations being reopened by customers from the live chat widget when **Allow messages after conversation resolved** is disabled.

The check is now handled on the server when selecting the target conversation. If the last conversation is resolved and replies are not allowed, a new conversation is created instead of reopening the old one.

It also handles blocked contacts correctly, prevents duplicate conversations when multiple messages are sent at the same time, and creates the new conversation and first message in a single transaction so failed messages do not leave empty conversations behind.

On the widget side, the previous thread is dropped and refetched when the server assigns the message to a new conversation. Failed messages are also marked as `in_progress` so retries replace the existing message instead of creating duplicates.

### What changed

* `Api::V1::Widget::MessagesController#set_conversation` now starts a new conversation when the last conversation is resolved and `allow_messages_after_resolved` is disabled. This covers both text messages and attachments since they use the same endpoint.
* Blocked contacts are excluded from this check, since their conversations are always created as resolved.
* The conversation lookup and creation are wrapped in `@contact.with_lock` to prevent multiple messages sent at the same time from creating duplicate conversations.
* The new conversation and its first message are created in a single transaction, so failed messages do not leave an empty conversation behind.
* The widget drops the previous thread and refetches when the server assigns the message to a different conversation, so the new message is not rendered under the old resolved conversation.
* Messages are pushed as `in_progress` while sending, so retrying a failed message replaces the existing bubble instead of creating a duplicate.

Fixes https://linear.app/chatwoot/issue/CW-7846/known-issue-with-customers-being-able-to-open-resolved-chats

## Type of change

### Screencast

**Before**

https://github.com/user-attachments/assets/92617b37-2ec0-42fe-a208-98f9a8dfbab7

**After**

https://github.com/user-attachments/assets/93fbf896-2d8c-4068-9e5a-d91b3af642a9

### Steps to reproduce

The reply box is correctly hidden on a resolved conversation, so the message needs to come from somewhere that bypasses it. The retry button on a failed message is the simplest way to reproduce this.

1. Go to **Inbox settings** and set **Allow messages after conversation resolved** to **Disabled**.
2. Open the widget, start a conversation, and send a message.
3. Open DevTools → Network and set the network to **Offline**.
4. Send another message. It fails and shows a **Send message again** retry icon.
5. Set the network back to **Online**.
6. Resolve the conversation from the dashboard or using the widget's **End Conversation** button.
7. Click the retry icon on the failed message.

**Before:** The message is added to the resolved conversation and reopens it.

**After:** The message starts a new conversation and the old conversation stays resolved.

## Checklist:


iamsivin:
@codex review

app/controllers/api/v1/widget/messages_controller.rb
```diff
@@ -43,10 +46,23 @@ def build_attachment
   end
 
   def set_conversation
-    return unless conversation.nil?
+    return unless conversation.nil? || conversation_closed_for_replies?
+
+    # Serialize concurrent sends so they cannot each start a thread. Locks the contact, not the
+    # contact inbox, since HMAC sessions share conversations across contact inboxes.
+    @contact.with_lock do
```
scmmishra: Do we need a row lock on the contact here? That can block potential row updates during this time

app/controllers/api/v1/widget/messages_controller.rb
```diff
@@ -43,10 +46,23 @@ def build_attachment
   end
 
   def set_conversation
-    return unless conversation.nil?
+    return unless conversation.nil? || conversation_closed_for_replies?
```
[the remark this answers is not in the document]
iamsivin: I think unrelated

app/controllers/api/v1/widget/messages_controller.rb
iamsivin: @scmmishra Good call. Moved the lock from the contact to the resolved conversation, so concurrent sends still serialize without blocking other contact updates.

sony-mathew (approved):
looks good. tested in browser also. Please take a look at this one finding:

- **P2 — Confirmed, matches the current unresolved thread:** [`set_conversation` returns without locking when the selected conversation is initially open](/worktrees/25eb/chatwoot/app/controllers/api/v1/widget/messages_controller.rb:49). If an agent resolves it before `@message.save!`, the widget request still returns 200 and stores the message in that resolved conversation instead of creating a replacement. My deterministic probe produced: original conversation still `resolved`, one new visitor message in it, and total conversations still `1`. This race predates the PR, but leaves CW-7846’s server-enforcement guarantee incomplete. The selected conversation must be locked and rechecked through message creation, including the initially-open path.

iamsivin:
@sony-mathew @scmmishra PTAL

sony-mathew (commented):
@iamsivin got this fidnings, please take a look:

1. **P2 — Still present from a resolved/outdated thread: late responses are added to the replacement thread.**  
   When conversation `55` is superseded by `99`, `resetStaleThread` correctly avoids deleting conversation `99`, but both send paths still commit the delayed `55` response unconditionally ([actions.js](/worktrees/0d59/chatwoot/app/javascript/widget/store/modules/conversation/actions.js:98), attachment path at [line 155](/worktrees/0d59/chatwoot/app/javascript/widget/store/modules/conversation/actions.js:155)). The store and UI do not filter messages by active conversation, so the old message appears in the new thread. The committed “ignores a late response” test only checks deletion/refetch behavior and misses this insertion ([actions.spec.js](/worktrees/0d59/chatwoot/app/javascript/widget/store/modules/specs/conversation/actions.spec.js:312)). My adversarial probe failed with `expected false, received true`, confirming conversation `55` remained in the active `99` store.

2. **P2 — New: a failed attribute refresh leaves the widget attached to the resolved conversation.**  
   After a successful replacement POST, the switch depends on `getAttributes` ([actions.js](/worktrees/0d59/chatwoot/app/javascript/widget/store/modules/conversation/actions.js:32)). That action swallows failures without updating or retrying the active ID ([conversationAttributes.js](/worktrees/0d59/chatwoot/app/javascript/widget/store/modules/conversationAttributes.js:18)). The message/history can switch to conversation `99` while attributes remain resolved conversation `55`; consequently the reply box remains hidden ([ChatFooter.vue](/worktrees/0d59/chatwoot/app/javascript/widget/components/ChatFooter.vue:45)) and ActionCable discards events for `99` ([actionCable.js](/worktrees/0d59/chatwoot/app/javascript/widget/helpers/actionCable.js:9)). The exact-head probe failed with `expected 99, received 55`.

muhsin-k:
@iamsivin Please let me know following comments are addressed or not.

P2] Failed attribute refresh leaves the widget on the resolved conversation.
[resetStaleThread](https://github.com/chatwoot/chatwoot/blob/30dae798ccfb6b037cf6408057e3cbf8bb08c17d/app/javascript/widget/store/modules/conversation/actions.js#L22-L40) continues after getAttributes, but that action [swallows failures](https://github.com/chatwoot/chatwoot/blob/30dae798ccfb6b037cf6408057e3cbf8bb08c17d/app/javascript/widget/store/modules/conversationAttributes.js#L18-L26). Messages may switch to the new conversation while the active ID/status remains on the resolved one. The reply box stays hidden and ActionCable rejects new-thread events.

[P2] A delayed response from the superseded conversation is inserted into the replacement thread.
Both [text](https://github.com/chatwoot/chatwoot/blob/30dae798ccfb6b037cf6408057e3cbf8bb08c17d/app/javascript/widget/store/modules/conversation/actions.js#L93-L100) and [attachment](https://github.com/chatwoot/chatwoot/blob/30dae798ccfb6b037cf6408057e3cbf8bb08c17d/app/javascript/widget/store/modules/conversation/actions.js#L148-L157) paths commit the response unconditionally. When conversation 99 is already active and a delayed response for 55 arrives, the reset is correctly skipped—but message 55 is still rendered in the active store. The existing “ignores a late response” test does not assert that insertion is rejected.

[P2] An already-running history request can restore deleted old-thread messages.
The reset deletes currently stored stale messages, but [fetchOldConversations](https://github.com/chatwoot/chatwoot/blob/30dae798ccfb6b037cf6408057e3cbf8bb08c17d/app/javascript/widget/store/modules/conversation/actions.js#L167-L180) merges every returned payload without checking its conversation ID or request generation. A pagination request started before the switch can finish afterward and mix the resolved conversation back into the new thread.

iamsivin:
> @iamsivin Please let me know following comments are addressed or not.
> 
> P2] Failed attribute refresh leaves the widget on the resolved conversation. [resetStaleThread](https://github.com/chatwoot/chatwoot/blob/30dae798ccfb6b037cf6408057e3cbf8bb08c17d/app/javascript/widget/store/modules/conversation/actions.js#L22-L40) continues after getAttributes, but that action [swallows failures](https://github.com/chatwoot/chatwoot/blob/30dae798ccfb6b037cf6408057e3cbf8bb08c17d/app/javascript/widget/store/modules/conversationAttributes.js#L18-L26). Messages may switch to the new conversation while the active ID/status remains on the resolved one. The reply box stays hidden and ActionCable rejects new-thread events.
> 
> [P2] A delayed response from the superseded conversation is inserted into the replacement thread. Both [text](https://github.com/chatwoot/chatwoot/blob/30dae798ccfb6b037cf6408057e3cbf8bb08c17d/app/javascript/widget/store/modules/conversation/actions.js#L93-L100) and [attachment](https://github.com/chatwoot/chatwoot/blob/30dae798ccfb6b037cf6408057e3cbf8bb08c17d/app/javascript/widget/store/modules/conversation/actions.js#L148-L157) paths commit the response unconditionally. When conversation 99 is already active and a delayed response for 55 arrives, the reset is correctly skipped—but message 55 is still rendered in the active store. The existing “ignores a late response” test does not assert that insertion is rejected.
> 
> [P2] An already-running history request can restore deleted old-thread messages. The reset deletes currently stored stale messages, but [fetchOldConversations](https://github.com/chatwoot/chatwoot/blob/30dae798ccfb6b037cf6408057e3cbf8bb08c17d/app/javascript/widget/store/modules/conversation/actions.js#L167-L180) merges every returned payload without checking its conversation ID or request generation. A pagination request started before the switch can finish afterward and mix the resolved conversation back into the new thread.

@muhsin-k Done

muhsin-k (approved):

iamsivin:
@codex Review?

PR #15418, opened 2026-08-11
fix: Alt+E resolve shortcut also opens Chrome's built-in menu on Windows

### LINKED ISSUE #15258: Default keyboard shortcuts clash with Chrome
### Describe the bug

The default windows shortcut for resolving a conversation is ALT+e. When using Chatwoot in chrome, ALT+e opens Chrome's built in menu.

### To Reproduce

Use Chrome on Windows.
Open a unresolved converstation.
Hit ALT+e
Ticket wil resolve, but also open Chrome's built-in menu

### Expected behavior

_No response_

### Environment

Linux VM

### Cloud Provider

_No response_

### Platform

_No response_

### Operating system

_No response_

### Browser and version

_No response_

### Docker (if applicable)

_No response_

### Additional context

_No response_

vjymisal0:
## Description

The `Alt+KeyE` keybinding for resolving a conversation, registered in `app/javascript/dashboard/components/buttons/ResolveAction.vue`, never calls `event.preventDefault()`. On Chrome/Windows, `Alt+E` is a legacy built-in shortcut that opens the "Customize and control Google Chrome" menu. Since our handler doesn't stop propagation of the default browser action, both things happen: the conversation resolves *and* Chrome's menu pops open on top of it.

Root cause, traced through `useKeyboardEvents` (`app/javascript/dashboard/composables/useKeyboardEvents.js`): it wires bindings via `tinykeys`'s `createKeybindingsHandler`, which does not call `preventDefault()` automatically - that's left to each binding's own `action`. The sibling combo `$mod+Alt+KeyE` in the very same file already does this correctly (`event.preventDefault()` right before/around its logic), which is what tipped me off that the plain `Alt+KeyE` binding was simply missing the same call, not a different underlying issue.

## Type of change

## How Has This Been Tested?

Manually traced the reported repro against the current code: `Alt+KeyE`'s `action` calls `onCmdResolveConversation()` only, with no `event.preventDefault()`, while `$mod+Alt+KeyE`'s handler in the same object explicitly calls it. This matches the reported symptom exactly (ticket resolves, Chrome menu also opens) and the fix mirrors the existing working pattern in the same component.

I was not able to run the JS test suite in this sandbox (no Node/Vitest available in the environment I worked in) or manually reproduce in an actual Chrome/Windows browser session, so I could not execute automated or manual browser verification for this change. I also checked for existing test coverage of this class of keybinding (e.g. `ChatTypeTabs.vue`'s `Alt+KeyN` binding, `useSidebarKeyboardShortcuts.js`) and found none of the per-component `Alt+Key*` bindings in this codebase have dedicated specs today, so I did not add a new spec pattern in isolation - happy to add one if a reviewer points me at the preferred way to unit-test a `useKeyboardEvents` binding registered inside a `<script setup>` SFC.

**Note on scope:** several other components register their own `Alt+Key*` bindings (e.g. `Alt+KeyM`, `Alt+KeyJ`/`Alt+KeyK`, `Alt+KeyC`/`Alt+KeyV`/`Alt+KeyR`/`Alt+KeyS`, `Alt+KeyN`, `Alt+KeyP`/`Alt+KeyL`) that also don't call `preventDefault()`. `Alt+E` is the one explicitly reported here because it collides with a well-known Chrome/Windows legacy shortcut; I kept this PR scoped to the reported binding rather than auditing/changing every `Alt+Key*` handler in the app, since that's a broader decision (whether *all* single-letter Alt shortcuts should suppress default) that's out of scope for this bug report. Flagging it here as a possible follow-up.

Fixes #15258

## Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works (see note above - no existing precedent for testing this binding pattern; happy to add if pointed at a preferred approach)
- [ ] New and existing unit tests pass locally with my changes (could not run locally, see note above)


iamsivin (approved):
@vjymisal0 Thanks for the PR. Tested it locally and LGTM.

bahaduralihashmi:
hello

PR #15423, opened 2026-08-12
fix: route inbound reply addresses from Cc

### LINKED ISSUE #15365: Inbound reply silently discarded when `reply+<uuid>@` is in `Cc` rather than `To`
### Describe the bug

On a self-hosted install using the email channel, a customer reply that carries the `reply+<uuid>@` address in `Cc` instead of `To` is discarded entirely — no message appended to the conversation, no new conversation, no contact, and nothing written to the log.

The message is delivered successfully to the ingress (Rails logs the POST as `200`), so the loss happens inside routing. Neither the sender nor the operator gets any signal.

### To Reproduce

1. Configure an email inbox (forwarding/ingress; IMAP off).
2. From an outside address, email the inbox. A conversation is created. ✅
3. Reply from the Chatwoot dashboard. The outbound message carries `Reply-To: reply+<conversation-uuid>@<MAILER_INBOUND_EMAIL_DOMAIN>`.
4. In Gmail, reply to that message. `reply+<uuid>@` lands in `To`, and the reply appears in the conversation. ✅
5. Now in Gmail open **your own sent message from step 4** and click Reply on it — a normal way to follow up on an earlier mail. Gmail places `reply+<uuid>@` in `Cc`.

**Actual:** nothing happens. The message reaches the ingress and is dropped.

### Expected behavior

The reply is appended to the same conversation, as in step 4.

### Environment

Docker

### Cloud Provider

AWS

#### Root cause

Two reads consider `to` only, and the fallback mailbox is a no-op.

`app/mailboxes/application_mailbox.rb` — the routing gate:

```ruby
def reply_uuid_mail?(inbound_mail)
  inbound_mail.mail.to&.any? do |email|
    conversation_uuid = email.split('@')[0]
    conversation_uuid.match?(REPLY_EMAIL_UUID_PATTERN)
  end
end
```

`app/presenters/mail_presenter.rb` — read by `Mailbox::ConversationFinderStrategies::ReceiverUuidStrategy`:

```ruby
def mail_receiver
  if @mail.to.blank?
    return [email_forwarded_for] if email_forwarded_for.present?

    []
  else
    @mail.to
  end
end
```

`app/mailboxes/default_mailbox.rb` — where everything unmatched ends up:

```ruby
class DefaultMailbox < ApplicationMailbox
  def process; end
end
```

Meanwhile `app/finders/email_channel_finder.rb` **already** considers `Cc`:

```ruby
def primary_recipient_emails
  (@email_object.to.to_a + @email_object.cc.to_a + [@email_object['X-Original-To'].try(:value)]).flatten.compact
end
```

So `Cc` is honoured when identifying the **channel** and ignored when identifying the **conversation**. In this scenario neither `channel.email` nor `forward_to_email` is on the thread — the customer is replying to their own message, so the inbox address is not among the recipients — and `EmailChannelFinder` finds nothing either. Both routing branches miss, and the message lands in a mailbox whose `process` is empty.

#### Suggested fix

Consider `cc` alongside `to` in both reads, for symmetry with `EmailChannelFinder`:

* `ApplicationMailbox.reply_uuid_mail?` — scan `mail.to` and `mail.cc`
* `MailPresenter#mail_receiver` — include `@mail.cc`

Separately, it may be worth logging in `DefaultMailbox#process`. Mail that cannot be routed is currently invisible; a single line would turn a silent drop into something an operator can find.

#### Version and setup

* Chatwoot **v4.16.2-ce**, self-hosted (Docker)
* Email channel via Action Mailbox `ses` ingress (`aws-actionmailbox-ses`), IMAP disabled
* Verified the message reached the application: SES delivered it and Rails logged `POST /rails/action_mailbox/ses/inbound_emails` returning `200`

#### Impact

Replying to one's own sent message is an ordinary way to chase up an earlier email, so this is not an exotic path. The failure is silent in both directions: the sender gets no bounce, and the operator gets no log line. For an inbox where a missed reply matters — support, admissions, anything with a deadline — the message is simply lost.

There does not appear to be a configuration workaround: no setting changes which headers are scanned, and the inbox address is not present on the thread for `EmailChannelFinder` to match.

#### Related

* [#9927](<https://github.com/chatwoot/chatwoot/issues/9927>) — a CC'd recipient's reply is not processed. Different mechanism (IMAP fetch, rejected as an auto-reply) and a different actor (a third party rather than the original correspondent), but the same theme: `Cc` is handled inconsistently across the email channel.
* chatwoot/chatwoot#15092 — CC'd participants' mail is attributed to the conversation contact.

Both suggest `Cc` handling in the email channel is worth a look more broadly than this one fix.

---

Diagnosis assisted by Claude.

jbdrak:
Fixes #15365.

## Summary

Inbound replies can put the `reply+<conversation-uuid>@...` address in `Cc` instead of `To`. The mailbox routing gate only inspected `To`, and the presenter later exposed only `To` to the conversation finder, so those replies were dropped.

This change:

- checks both `To` and `Cc` when identifying reply UUID mail
- passes both recipient lists to the receiver UUID strategy
- adds fixture, mailbox-routing, presenter, and receiver-strategy regression coverage
- preserves the `X-Forwarded-For` fallback when neither recipient header exists

## Verification

- `bundle exec rubocop` on all 5 changed Ruby files - passed
- isolated Mail fixture/routing check - passed; reply UUID in `Cc` is detected
- `git diff --check` - passed

The Rails specs were attempted but couldn't boot because the local PostgreSQL 16 server is missing the `vector` extension. Homebrew's available pgvector bottle targets PostgreSQL 17/18, so I didn't replace the active database installation. No application test failure was observed beyond that environment prerequisite.

### DIFF
--- a/app/mailboxes/application_mailbox.rb
+++ b/app/mailboxes/application_mailbox.rb
@@ -19,8 +19,10 @@ class ApplicationMailbox < ActionMailbox::Base
 
   class << self
     # checks if follows this pattern: reply+<conversation-uuid>@<mailer-domain.com>
+    # the reply address can appear in either To or Cc (e.g. replying to your own sent message)
     def reply_uuid_mail?(inbound_mail)
-      inbound_mail.mail.to&.any? do |email|
+      recipients = Array(inbound_mail.mail.to) + Array(inbound_mail.mail.cc)
+      recipients.any? do |email|
         conversation_uuid = email.split('@')[0]
         conversation_uuid.match?(REPLY_EMAIL_UUID_PATTERN)
       end
--- a/app/presenters/mail_presenter.rb
+++ b/app/presenters/mail_presenter.rb
@@ -156,13 +156,12 @@ class MailPresenter < SimpleDelegator
   end
 
   def mail_receiver
-    if @mail.to.blank?
-      return [email_forwarded_for] if email_forwarded_for.present?
+    receivers = Array(@mail.to) + Array(@mail.cc)
+    return receivers if receivers.present?
 
-      []
-    else
-      @mail.to
-    end
+    return [email_forwarded_for] if email_forwarded_for.present?
+
+    []
   end
 
   def auto_reply?
--- a/spec/fixtures/files/reply_uuid_in_cc.eml
+++ b/spec/fixtures/files/reply_uuid_in_cc.eml
@@ -1,0 +1,20 @@
+From: Sony Mathew <sony@chatwoot.com>
+Mime-Version: 1.0 (Apple Message framework v1244.3)
+Content-Type: multipart/alternative; boundary="Apple-Mail=_33A037C7-4BB3-4772-AE52-FCF2D7535F74"
+Subject: Discussion: Let's debate these attachments
+Date: Tue, 20 Apr 2020 04:20:20 -0400
+In-Reply-To: <4e6e35f5a38b4_479f13bb90078178@small-app-01.mail>
+References: <4e6e35f5a38b4_479f13bb90078178@small-app-01.mail>
+Message-Id: <0CB459E0-0336-41DA-BC88-E6E28C697DDB@chatwoot.com>
+X-Mailer: Apple Mail (2.1244.3)
+To: Sony Mathew <sony@chatwoot.com>
+Cc: "Replies" <reply+6bdc3f4d-0bec-4515-a284-5d916fdde489@example.com>
+
+--Apple-Mail=_33A037C7-4BB3-4772-AE52-FCF2D7535F74
+Content-Transfer-Encoding: quoted-printable
+Content-Type: text/plain;
+  charset=utf-8
+
+Let's talk about these images:
+
+--Apple-Mail=_33A037C7-4BB3-4772-AE52-FCF2D7535F74--
--- a/spec/mailboxes/application_mailbox_spec.rb
+++ b/spec/mailboxes/application_mailbox_spec.rb
@@ -7,6 +7,7 @@ RSpec.describe ApplicationMailbox do
     let(:welcome_mail) { create_inbound_email_from_fixture('welcome.eml') }
     let(:reply_mail) { create_inbound_email_from_fixture('reply.eml') }
     let(:reply_cc_mail) { create_inbound_email_from_fixture('reply_cc.eml') }
+    let(:reply_uuid_in_cc_mail) { create_inbound_email_from_fixture('reply_uuid_in_cc.eml') }
     let(:reply_mail_without_uuid) { create_inbound_email_from_fixture('reply.eml') }
     let(:reply_mail_with_in_reply_to) { create_inbound_email_from_fixture('in_reply_to.eml') }
     let(:support_mail) { create_inbound_email_from_fixture('support.eml') }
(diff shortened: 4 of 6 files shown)


scmmishra:
Hey @jbdrak is this ready for review? The changes seem largely good, I can review it and get it merged

psloh:
Thanks for picking this up — I reported #15365, and the fix looks right to me: both reads that considered `to` only, plus the receiver-strategy coverage.

On the local blocker you mentioned: I think you may be more unblocked than you realise. **CI has already run this and it is green** — 34 checks passing, including the whole `backend-tests` 16-shard matrix, so the specs you added did execute in an environment that has the pgvector prerequisite. The PostgreSQL 16 + vector extension problem looks like a local-setup issue rather than anything holding the change back.

(The one red check, `auto-assign`, is unrelated — it tries to assign the author and cannot, because `GITHUB_TOKEN` is read-only on pull requests from forks. It fails the same way on any outside contribution.)

**Would you consider marking this ready for review?** Drafts do not generally show up in maintainer review queues, so it may simply not have been seen yet.

For whatever it adds as a second data point: we hit this on a self-hosted install and it cost us real messages. A customer replying to their own earlier message is an ordinary way to follow up, and Gmail puts `reply+<uuid>@` in `Cc` when they do. There is no configuration workaround we could find, and the failure is silent in both directions — the sender gets no bounce and the operator gets no log line, so a missed reply is invisible until someone notices the customer went quiet.

We are heading to production with an email inbox where a lost reply matters, so we would be glad to see this land. Happy to help if anything else is needed.

psloh:
@scmmishra one small thing that may be exactly what's holding this up: **anyone with write access can mark a draft PR as ready for review** — it isn't restricted to the author. So if @jbdrak is away, you can flip it yourself and unblock your own review. `itoqa` skips drafts, which is why nothing has reviewed this since the 12th.

Two things that might help when you do:

**This makes Chatwoot consistent with itself.** `EmailChannelFinder#primary_recipient_emails` already reads Cc:

```ruby
(@email_object.to.to_a + @email_object.cc.to_a + [@email_object['X-Original-To'].try(:value)]).flatten.compact
```

So the inbox-matching half of the routing lambda honours Cc today, while the `reply_uuid_mail?` half does not. A reply whose `reply+<uuid>@` lands in Cc is dropped purely because of that asymmetry — which is what this PR removes.

**One gap, which I'd suggest as a follow-up rather than a blocker here.** `valid_to_address?` still guards on `to` alone:

```ruby
to_address_class = inbound_mail.mail.to&.class
return true if to_address_class == Mail::AddressContainer
```

A message with no `To` header at all, carrying the reply address only in `Cc`, still logs "malformed" and falls through to the catch-all. That's rarer — most clients set `To` — and it's outside what #15365 described, so I don't think it should hold this PR up.

For context on why the failure mode is worth closing: `DefaultMailbox#process` is an empty method, so an unroutable message is marked `delivered` and disappears with nothing in the log. On our install that meant a customer reply that never appeared in the dashboard, with no bounce and no error to trace. It took reading the routing code to work out where it had gone.

@jbdrak — thank you for picking this up. The fix and the specs both look right to me.

jbdrak:
Apologies for the slow turnaround. This is ready for review now - I'd left it in draft, which is why Ito skipped it on the 12th.

CI is green apart from `auto-assign`, which fails because the token can't assign the author on a fork PR - unrelated to the change. The whole `backend-tests` matrix passes.

Also thanks for confirming the pgvector/Postgres 16 prerequisite is a local setup issue rather than anything holding this back.

PR #15440, opened 2026-08-13
fix(whatsapp): load health status for regional locales

muhsin-k:
The WhatsApp health status page fails to render for accounts using underscore-based regional locales such as `pt_BR`. The browser rejects that locale when formatting the last-onboarded timestamp, leaving the page blank even though health data loaded successfully.

This change formats the timestamp with Chatwoot’s shared resolved locale, which normalizes and validates the locale before passing it to the browser. Other health status behavior remains unchanged.

### How to reproduce

1. Set the dashboard language to Portuguese (Brazil).
2. Open an inbox’s WhatsApp health status page when the response includes `last_onboarded_time`.
3. Observe that the page remains blank with `RangeError: Invalid language tag: pt_BR` in the browser console.

### How to test

1. Use an account with Portuguese (Brazil) selected.
2. Open the WhatsApp health status page for an inbox with onboarding data.
3. Confirm the complete health status page loads and the last-onboarded timestamp is formatted in the selected locale.

### DIFF
--- a/app/javascript/dashboard/routes/dashboard/settings/inbox/components/AccountHealth.vue
+++ b/app/javascript/dashboard/routes/dashboard/settings/inbox/components/AccountHealth.vue
@@ -2,6 +2,7 @@
 import { computed } from 'vue';
 import { useI18n } from 'vue-i18n';
 import { copyTextToClipboard } from 'shared/helpers/clipboard';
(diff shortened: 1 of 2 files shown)


iamsivin (approved):
Code LGTM

PR #15469, opened 2026-08-13
docs: document message read status APIs for API channels

vishnu-narayanan:
## Description

Customers building on the API channel could not find how to push message read status into Chatwoot. Two endpoints are involved, one was entirely missing from the API reference and one was undocumented in what it actually does:

- `PATCH /api/v1/accounts/{account_id}/conversations/{conversation_id}/messages/{message_id}` accepts `status` (`sent`/`delivered`/`read`/`failed`, plus `external_error` for failures) and is restricted to API channel inboxes. The path previously documented only DELETE; this adds the PATCH operation.
- The public `update_last_seen` endpoint marks agent messages sent before the contact's last seen time as read, but its description only mentioned updating the timestamp, so searching the docs for read receipts found nothing. The summary and description now state the read-receipt behavior.

Derived swagger files regenerated via `rake swagger:build`.

Fixes https://linear.app/chatwoot/issue/CW-7937

## Type of change

## How Has This Been Tested?

Ran `bundle exec rake swagger:build` and verified the generated `swagger.json`, `application_swagger.json`, and `client_swagger.json` contain only the new PATCH operation and the updated summary/description strings. Documented behavior cross-checked against the controller, status update service, and dashboard read-indicator logic.

## Checklist:

### DIFF
--- a/swagger/paths/application/conversation/messages/update.yml
+++ b/swagger/paths/application/conversation/messages/update.yml
@@ -1,0 +1,60 @@
+tags:
+  - Messages
+operationId: update-message-status
+summary: Update message status
+security:
+  - userApiKey: []
+description: |
+  Updates the delivery status of a single message. Available only for messages in
+  API channel inboxes, where your own system is responsible for delivering messages
+  and reporting what happened. Use this to push `sent`, `delivered`, and `read`
+  receipts into Chatwoot, for example, to mark an agent's message as read when the
+  end user views it in your client, so the agent sees the read indicator on the
+  message.
+
+  A message already marked `read` cannot be moved back to `delivered`. When
+  reporting a failure, include `external_error` with the reason, which is shown to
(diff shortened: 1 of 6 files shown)


sony-mathew:
@vishnu-narayanan should we document this on the docs repo? eg: https://github.com/chatwoot/docs/pull/580

sony-mathew (approved):
lgtm

vishnu-narayanan:
@sony-mathew Not needed for the reference itself. `docs.json` pulls the specs straight from develop (`swagger/tag_groups/application_swagger.json` and `client_swagger.json`), and this PR regenerates both, so the new PATCH operation and the updated update_last_seen description publish as soon as this lands.

Will raise a follow-up on the docs repo for the read-receipt behaviour though, since that spans both this endpoint and update_last_seen, and isn't obvious from the spec alone.

PR #15472, opened 2026-08-13
feat: add conditions to delayed automations

aakashb95:
## Summary

Delayed automations can now include additional conditions joined with `AND`, including a Pending status condition. The form keeps these conditions inside the wait section and uses the existing automation controls.

Delayed jobs recheck all saved conditions before running an action. The action is skipped when a condition no longer matches.

Saved delayed rules are not normalized merely by opening the editor. API-created operators such as `inbox_id not_equal_to` remain unchanged until an administrator edits a wait control.

The Captain timer conflict warning is split into the stacked follow-up #15482.

## Validation

- 39 focused frontend tests passed.
- 75 focused backend examples passed.
- ESLint, Prettier, RuboCop, JSON parsing, and `git diff --check` passed.


aakashb95:
@codex review

sony-mathew (commented):
looks good overall. please take a look at these findings:

1. **P1 — New: label exclusions can still execute delayed customer-facing actions.**  
   The new filter exposure in [AutomationWaitCondition.vue](/worktrees/b074/chatwoot/app/javascript/dashboard/routes/dashboard/settings/automation/components/AutomationWaitCondition.vue:69) includes Labels from [constants.js](/worktrees/b074/chatwoot/app/javascript/dashboard/routes/dashboard/settings/automation/constants.js:84), including `not_equal_to`. However, [ConditionsFilterService](/worktrees/b074/chatwoot/app/services/automation_rules/conditions_filter_service.rb:167) evaluates joined tag rows using `tags.name != :value`.

   I reproduced this at the exact head:

   - Conversation labels: `bug`, `feature`
   - Delayed condition: `Labels != feature`
   - Expected: skip
   - Actual: the pending execution ran its action

   The `bug != feature` joined row makes the entire condition pass. The fire-time recheck uses the same predicate, so it does not prevent the unintended message, webhook, assignment, or other delayed action. Until correct `EXISTS`/`NOT EXISTS` semantics land, Labels should be excluded from these additional filters—or the backend correction must be a tested dependency of this PR.

2. **P2 — New: valid saved `OR` rules are displayed as `AND`.**  
   API-created delayed rules may legally contain `or`; [AutomationRule](/worktrees/b074/chatwoot/app/models/automation_rule.rb:141) accepts it and [useEditableAutomation](/worktrees/b074/chatwoot/app/javascript/dashboard/composables/useEditableAutomation.js:43) preserves it. But every additional wait condition is rendered with a hard-coded `AND` in [AutomationWaitCondition.vue](/worktrees/b074/chatwoot/app/javascript/dashboard/routes/dashboard/settings/automation/components/AutomationWaitCondition.vue:315).

   An adversarial component test confirmed that a saved `message_type OR status` rule:

   - remained persisted as `OR`;
   - emitted no normalization update on opening;
   - visibly claimed the connector was `AND`.

   An administrator can therefore believe the rule is narrower than it really is. Render the saved connector/order, or explicitly reject and migrate unsupported legacy `OR` rules.

3. **P2 — New: condition values and deletion controls are clipped on narrow screens.**  
   The outer row and [ConditionRow](/worktrees/b074/chatwoot/app/javascript/dashboard/components-next/filter/ConditionRow.vue:183) both use non-wrapping flex layouts. In a real 390×844 browser session, selecting `Labels → Not equal to` placed:

   - the value control at right edge ≈446px;
   - the delete button at right edge ≈486px;
   - inside a 390px viewport with a 366px dialog that hides overflow.

   There was no usable horizontal scrolling, leaving both controls inaccessible. Stack or wrap the row at narrow breakpoints and add viewport coverage.

sony-mathew:
@tds-1 and @iamsivin Can you guys take a look at this PR please?

app/javascript/dashboard/routes/dashboard/settings/automation/components/AutomationWaitCondition.vue
```diff
@@ -44,11 +54,40 @@ const selectedTrigger = ref(DEFAULT_TRIGGER);
 const triggerStatus = ref(DEFAULT_TRIGGER_STATUS);
 // No inbox selected means the rule applies to every inbox.
 const triggerInboxes = ref([]);
+const conditionsRef = useTemplateRef('conditionsRef');
+const unsupportedDelayedAttributes = new Set(['labels']);
```
iamsivin: Let's move `DELAYED_UNSUPPORTED_ATTRIBUTES` to `app/javascript/dashboard/routes/dashboard/settings/automation/constants.js`.

```
export const DELAYED_UNSUPPORTED_ATTRIBUTES = ['labels'];
```

app/javascript/dashboard/components-next/filter/inputs/MultiSelect.vue
iamsivin, line 1: Do we need this one? New prop on a shared component + a new translatable string (CONDITION_PLACEHOLDER: "Select...") purely to shorten "Select an option…" so it fits the narrow card.

app/javascript/dashboard/components-next/filter/inputs/MultiSelect.vue
```diff
@@ -155,7 +159,9 @@ const toggleOption = option => {
         <template #icon>
           <Icon icon="i-lucide-plus" class="text-n-slate-11" />
         </template>
-        <span class="text-n-slate-11">{{ t('COMBOBOX.PLACEHOLDER') }}</span>
+        <span class="text-n-slate-11 whitespace-nowrap">
```
iamsivin: Have you verified that `whitespace-nowrap` is needed for all four `MultiSelect` usages?

app/javascript/dashboard/routes/dashboard/settings/automation/components/AutomationWaitCondition.vue
```diff
@@ -44,11 +54,40 @@ const selectedTrigger = ref(DEFAULT_TRIGGER);
...
+});
+
+const additionalConditionIndexes = computed(() =>
+  conditions.value.reduce((indexes, condition, index) => {
+    if (!managedAttributeKeys.value.has(condition.attribute_key)) {
+      indexes.push(index);
+    }
+    return indexes;
+  }, [])
+);
```
iamsivin, line 86: Could we extract `isAdditionalCondition` since the same check is repeated here, at `199`, and `204`? Also, `flatMap` would simplify the `reduce` and make this easier to read.

```js
const isAdditionalCondition = condition =>
  !managedAttributeKeys.value.has(condition.attribute_key);
```

app/javascript/dashboard/routes/dashboard/settings/automation/components/AutomationWaitCondition.vue
```diff
@@ -131,30 +170,108 @@ const buildCondition = (attributeKey, values) => ({
...
+};
+
+const connectorLabel = conditionIndex =>
+  connectorBeforeCondition(conditionIndex) === 'or'
+    ? t('FILTER.QUERY_DROPDOWN_LABELS.OR')
+    : t('FILTER.QUERY_DROPDOWN_LABELS.AND');
+
+const applyTrigger = ({ preserveAdditional = true } = {}) => {
+  const firstAdditionalConditionIndex = conditions.value.findIndex(
+    condition => !managedAttributeKeys.value.has(condition.attribute_key)
```
iamsivin: https://github.com/chatwoot/chatwoot/pull/15472/changes#r3842105025

app/javascript/dashboard/routes/dashboard/settings/automation/components/AutomationWaitCondition.vue
```diff
@@ -131,30 +170,108 @@ const buildCondition = (attributeKey, values) => ({
...
+  );
+  const additionalConditions =
+    preserveAdditional && firstAdditionalConditionIndex !== -1
+      ? conditions.value.filter(
+          condition => !managedAttributeKeys.value.has(condition.attribute_key)
```
iamsivin: https://github.com/chatwoot/chatwoot/pull/15472/changes#r3842105025

app/javascript/dashboard/routes/dashboard/settings/automation/components/AutomationWaitCondition.vue
```diff
@@ -131,30 +170,108 @@ const buildCondition = (attributeKey, values) => ({
...
+  // trigger condition. OR is commutative for this two-condition edge case, so show its connector
+  // rather than claiming the saved rule uses AND.
+  if (conditionIndex === 0) {
+    return normalizedQueryOperator(
+      conditions.value[conditionIndex]?.query_operator
+    );
+  }
+
+  return normalizedQueryOperator(connector);
+};
```
iamsivin: Could we inline `normalizedQueryOperator` here since it only has one caller? This also avoids looking up `previousCondition` before the `conditionIndex === 0` branch.

```js
const connectorBeforeCondition = conditionIndex => {
  const connectorSource =
    conditions.value[conditionIndex === 0 ? 0 : conditionIndex - 1];

  return connectorSource?.query_operator?.toLowerCase() === 'or' ? 'or' : 'and';
};
```

app/javascript/dashboard/routes/dashboard/settings/automation/components/AutomationWaitCondition.vue
```diff
@@ -131,30 +170,108 @@ const buildCondition = (attributeKey, values) => ({
...
+  );
+  const additionalConditions =
+    preserveAdditional && firstAdditionalConditionIndex !== -1
+      ? conditions.value.filter(
+          condition => !managedAttributeKeys.value.has(condition.attribute_key)
+        )
+      : [];
+  const additionalConditionsConnector = connectorBeforeCondition(
+    firstAdditionalConditionIndex
+  );
```
iamsivin: `firstAdditionalConditionIndex !== -1` is redundant here since `filter` already returns an empty array when there are no matches. We can just check `additionalConditions.length`.

```js
  const additionalConditions = preserveAdditional
    ? conditions.value.filter(isAdditionalCondition)
    : [];
```

app/javascript/dashboard/routes/dashboard/settings/automation/components/AutomationWaitCondition.vue
```diff
@@ -131,30 +170,108 @@ const buildCondition = (attributeKey, values) => ({
...
+    waitConditions.push(
       buildCondition(
         'inbox_id',
         triggerInboxes.value.map(inbox => inbox.id)
       )
     );
   }
-  conditions.value = nextConditions;
+  if (additionalConditions.length) {
+    waitConditions.at(-1).query_operator = additionalConditionsConnector;
```
iamsivin:
```suggestion
    waitConditions.at(-1).query_operator = connectorBeforeCondition(
      conditions.value.indexOf(additionalConditions[0])
    );
```

app/javascript/dashboard/routes/dashboard/settings/automation/components/AutomationWaitCondition.vue
```diff
@@ -163,20 +280,33 @@ hydrateFromRule();
...
+watch(selectedTrigger, (nextTrigger, previousTrigger) => {
+  const nextEvent = DELAYED_TRIGGERS.find(
+    trigger => trigger.key === nextTrigger
+  ).eventName;
+  const previousEvent = DELAYED_TRIGGERS.find(
+    trigger => trigger.key === previousTrigger
+  ).eventName;
+  applyTrigger({ preserveAdditional: nextEvent === previousEvent });
+});
+watch([triggerStatus, triggerInboxes], () => applyTrigger());
```
iamsivin: Could we avoid the repeated `DELAYED_TRIGGERS.find(...)` calls here? We can use a single watcher and get both the current and previous values, keeping the logic closer to `develop` while adding only the `preserveAdditional` decision.

iamsivin (changes requested):
@aakashb95 Added some comments, PTAL
1. `ALL` in the status is repeated twice. See if we can fix this easily.

Image
2. The **Add condition** button seems to be missing in the Add Automation rule panel. Please check this as well.

Image

Image

app/models/automation_rule.rb
```diff
@@ -111,9 +111,14 @@ def query_operator_value
...
 
-    errors.add(:execution_delay, 'cannot be used with attribute_changed conditions.')
+    if conditions.any? { |obj| obj['filter_operator'] == 'attribute_changed' }
+      errors.add(:execution_delay, 'cannot be used with attribute_changed conditions.')
+    end
+
+    return if conditions.none? { |obj| obj['attribute_key'] == 'labels' }
+
+    errors.add(:execution_delay, 'cannot be used with label conditions.')
   end
```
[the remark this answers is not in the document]
aakashb95: non issue since it doesn't exist on prod

iamsivin (approved):
@aakashb95 FE code LGTM! Made a few small layout adjustments. PTAL and LMK if you need any help or run into any issues.
cc @sony-mathew

sony-mathew (commented):
Can you please check these findings:

1. **[P1] Existing delayed label rules cannot be opened or deactivated after upgrading.**  
   [automation_rule.rb](/worktrees/pr-15472/chatwoot/app/models/automation_rule.rb:112) now rejects every delayed rule containing a label condition, including updates that only set `active: false`. The UI also excludes labels from supported filter definitions in [AutomationWaitCondition.vue](/worktrees/pr-15472/chatwoot/app/javascript/dashboard/routes/dashboard/settings/automation/components/AutomationWaitCondition.vue:70) but still tries to render persisted conditions, causing the editor to throw.

   I reproduced both failures and verified that the exact base commit accepts and saves the same delayed label rule. The base feature shipped in v4.17.0/v4.17.1, so “doesn’t exist on prod” does not cover upgraded self-hosted installations. The affected population is unverified, but the compatibility break is confirmed. This matches the [existing unresolved thread](https://github.com/chatwoot/chatwoot/pull/15472#discussion_r3842311547).

2. **[P2] Editing a mixed AND/OR rule can silently broaden its scope.**  
   [AutomationWaitCondition.vue](/worktrees/pr-15472/chatwoot/app/javascript/dashboard/routes/dashboard/settings/automation/components/AutomationWaitCondition.vue:185) collects all managed wait conditions, moves them ahead of additional conditions, and then reapplies connectors by position.

   An API-created rule equivalent to:

   `status=open OR (message_type=outgoing AND private_note=false AND priority=high)`

   became:

   `(message_type=outgoing AND private_note=false AND inbox=…) OR status=open OR priority=high`

   after changing only the inbox. My backend probe confirmed a conversation that did not match the original expression matched the rewritten one. Because the delayed job uses this expression before executing customer-facing actions, this can send an action to unintended conversations. This matches the [existing unresolved thread](https://github.com/chatwoot/chatwoot/pull/15472#discussion_r3841640193).

app/javascript/dashboard/routes/dashboard/settings/automation/components/AutomationWaitCondition.vue
```diff
@@ -163,65 +366,119 @@ hydrateFromRule();
...
+            {{ connectorLabel(conditionIndex) }}
+          </li>
+          <ConditionRow
+            ref="conditionsRef"
+            v-model:attribute-key="conditions[conditionIndex].attribute_key"
+            v-model:filter-operator="conditions[conditionIndex].filter_operator"
+            v-model:values="conditions[conditionIndex].values"
+            class="w-full min-w-0"
+            :filter-types="additionalFilterTypes"
+            @remove="removeFilter(conditionIndex)"
```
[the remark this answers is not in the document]
aakashb95: This is the existing behaviour in normal automation conditions as well. Better to take it up as a separate PR.

aakashb95:
> Can you please check these findings:
> 
> 1. **[P1] Existing delayed label rules cannot be opened or deactivated after upgrading.**
>    [automation_rule.rb](/worktrees/pr-15472/chatwoot/app/models/automation_rule.rb:112) now rejects every delayed rule containing a label condition, including updates that only set `active: false`. The UI also excludes labels from supported filter definitions in [AutomationWaitCondition.vue](/worktrees/pr-15472/chatwoot/app/javascript/dashboard/routes/dashboard/settings/automation/components/AutomationWaitCondition.vue:70) but still tries to render persisted conditions, causing the editor to throw.
>    I reproduced both failures and verified that the exact base commit accepts and saves the same delayed label rule. The base feature shipped in v4.17.0/v4.17.1, so “doesn’t exist on prod” does not cover upgraded self-hosted installations. The affected population is unverified, but the compatibility break is confirmed. This matches the [existing unresolved thread](https://github.com/chatwoot/chatwoot/pull/15472#discussion_r3842311547).
> 2. **[P2] Editing a mixed AND/OR rule can silently broaden its scope.**
>    [AutomationWaitCondition.vue](/worktrees/pr-15472/chatwoot/app/javascript/dashboard/routes/dashboard/settings/automation/components/AutomationWaitCondition.vue:185) collects all managed wait conditions, moves them ahead of additional conditions, and then reapplies connectors by position.
>    An API-created rule equivalent to:
>    `status=open OR (message_type=outgoing AND private_note=false AND priority=high)`
>    became:
>    `(message_type=outgoing AND private_note=false AND inbox=…) OR status=open OR priority=high`
>    after changing only the inbox. My backend probe confirmed a conversation that did not match the original expression matched the rewritten one. Because the delayed job uses this expression before executing customer-facing actions, this can send an action to unintended conversations. This matches the [existing unresolved thread](https://github.com/chatwoot/chatwoot/pull/15472#discussion_r3841640193).

Fixed the P1.
P2 is a consistent behaviour with the existing automations, the fix will increase scope of this PR. It should be addressed separetely.

aakashb95:
@codex review

PR #15480, opened 2026-08-14
fix: verify Slack webhook request signatures

vishnu-narayanan:
## Description

Adds Slack signing-secret verification to the Slack events webhook (`POST /api/v1/integrations/webhooks`) so inbound events are confirmed to originate from Slack before they are processed. The `X-Slack-Signature` HMAC is checked against the request body with a timestamp-freshness window.

Verification only runs when `SLACK_SIGNING_SECRET` is configured; when it is not set, requests are processed as before so existing installations are not affected. **To enable verification, set `SLACK_SIGNING_SECRET`** (via env or the super-admin config) to the Slack app's signing secret.

Ref https://linear.app/chatwoot/issue/CW-7465

## Type of change

## How Has This Been Tested?

Added request specs in `spec/requests/api/v1/integrations/webhooks_request_spec.rb`: valid signature is processed, invalid signature / missing headers / stale (replayed) timestamp are rejected, the request is processed unchanged when no signing secret is configured, and verification is still enforced when the secret is only available via env. Spec file green, rubocop clean.

## Checklist:


sony-mathew (approved):
Looks good. Please check the following finding and see if it's valid:

## Findings

[P1] Environment-configured signing secrets are ignored after normal config reconciliation.

The new config entry defaults to blank in [installation_config.yml](/worktrees/pr-15480/chatwoot/config/installation_config.yml:431). During migrations, [ConfigLoader](/worktrees/pr-15480/chatwoot/lib/config_loader.rb:40) creates that blank database row. [GlobalConfigService](/worktrees/pr-15480/chatwoot/lib/global_config_service.rb:12) then finds the environment value but returns the existing row’s blank value, causing [signature verification](/worktrees/pr-15480/chatwoot/app/controllers/api/v1/integrations/webhooks_controller.rb:16) to be skipped.

Request-level reproduction on the exact head:

- `SLACK_SIGNING_SECRET=env-secret`, no database row: unsigned webhook → `401`
- Run the real `ConfigLoader.new.process`, which creates a blank row
- Same environment and same unsigned webhook → `200`

This contradicts the PR’s documented environment-variable setup and can leave the original webhook-forgery vulnerability exploitable. The request specs stub `GlobalConfigService.load`, so CI does not exercise this lifecycle.

This independently confirms the existing unresolved [same-head P1 review thread](https://github.com/chatwoot/chatwoot/pull/15480#discussion_r3782211329). Add a regression test that reconciles configuration before making a real request, then either make `GlobalConfigService` populate/return the environment value when the existing row is blank or implement an equivalent scoped fallback.

vishnu-narayanan:
Valid, confirmed. `GlobalConfigService` does read ENV, but `first_or_create` then returns the existing blank row and discards it, so the secret reads blank and verification is skipped. Fixed with a scoped ENV fallback here rather than changing `GlobalConfigService`, since that would affect every config key. Spec now creates the blank row and exercises the real lookup instead of stubbing it. Raising the `GlobalConfigService` behaviour separately.

PR #15481, opened 2026-08-14
feat: expose Captain outcome statistics

scmmishra:
Captain outcome reporting is now available through dedicated assistant stats endpoints. The new overview summary analyzes outcome, resolution-flow, and trend data to surface up to three specific, evidence-backed insights using rounded approximations suited to an hourly cached summary.

## Closes

Follow-up to https://github.com/chatwoot/chatwoot/pull/15425

## How to test

1. Select an assistant with recorded conversation outcomes.
2. Request the overview, resolution flow, and resolution trend endpoints for different reporting ranges.
3. Confirm the overview summary returns no more than three distinct points backed by rounded, softened report values rather than exact figures.
4. Confirm low-volume reports qualify rate-based conclusions and reports without meaningful activity return no points.
5. Request the overview summary again with the same assistant, range, timezone, and locale, and confirm the successful summary is reused from the one-hour cache.
6. Confirm raw statistics remain uncached and continue reflecting the selected reporting range and timezone.

## Chatwoot CLI

View the 30-day statistics for assistant ID `1` using a UTC+5:30 viewer timezone:

```bash
chatwoot api "/captain/assistants/1/stats/overview?range=30&timezone_offset=5.5" -o json
chatwoot api "/captain/assistants/1/stats/resolution_flow?range=30&timezone_offset=5.5" -o json
chatwoot api "/captain/assistants/1/stats/resolution_trend?range=30&timezone_offset=5.5" -o json
chatwoot api "/captain/assistants/1/stats/overview_summary?range=30&timezone_offset=5.5" -o json
```

## What changed

- Added a dedicated assistant stats controller for overview, resolution-flow, and resolution-trend data.
- Added a server-generated overview summary based on all three reporting datasets.
- Added a structured LLM response schema and a focused prompt for up to three standout insights.
- Rounded and softened metric values in generated insights so cached summaries do not present stale figures as exact.
- Cached successful LLM summaries for one hour while leaving raw statistics uncached and avoiding caching generation failures.
- Added a frontend API client for the new endpoints.
- Preserved the existing assistant metrics and summary endpoints for compatibility.

### DIFF
--- a/enterprise/app/controllers/api/v1/accounts/captain/assistant_stats_controller.rb
+++ b/enterprise/app/controllers/api/v1/accounts/captain/assistant_stats_controller.rb
@@ -11,7 +11,8 @@ class Api::V1::Accounts::Captain::AssistantStatsController < Api::V1::Accounts::
   end
 
   def overview_summary
-    result = Rails.cache.read(overview_summary_cache_key) || generate_overview_summary
+    cached_result = Redis::Alfred.get(overview_summary_cache_key)
+    result = cached_result ? JSON.parse(cached_result, symbolize_names: true) : generate_overview_summary
 
     if result[:error]
       render json: { error: result[:error] }, status: :unprocessable_content
@@ -38,7 +39,7 @@ class Api::V1::Accounts::Captain::AssistantStatsController < Api::V1::Accounts::
       timezone_offset: params[:timezone_offset]
     ).perform
 
-    Rails.cache.write(overview_summary_cache_key, result, expires_in: OVERVIEW_SUMMARY_CACHE_TTL) unless result[:error]
+    Redis::Alfred.set(overview_summary_cache_key, result.to_json, ex: OVERVIEW_SUMMARY_CACHE_TTL.to_i) unless result[:error]
     result
   end
 
@@ -51,7 +52,7 @@ class Api::V1::Accounts::Captain::AssistantStatsController < Api::V1::Accounts::
       stats_window.range,
       stats_window.timezone,
       Current.account.locale
-    ]
+    ].join(':')
   end
 
   def stats_window
--- a/enterprise/app/controllers/api/v1/accounts/captain/assistant_stats_controller.rb
+++ b/enterprise/app/controllers/api/v1/accounts/captain/assistant_stats_controller.rb
@@ -1,4 +1,7 @@
 class Api::V1::Accounts::Captain::AssistantStatsController < Api::V1::Accounts::BaseController
+  OVERVIEW_SUMMARY_CACHE_VERSION = 'v1'.freeze
+  OVERVIEW_SUMMARY_CACHE_TTL = 1.hour
+
   before_action -> { authorize(Captain::Assistant, :metrics?) }
   before_action :set_assistant
 
@@ -7,12 +10,7 @@ class Api::V1::Accounts::Captain::AssistantStatsController < Api::V1::Accounts::
   end
 
   def overview_summary
-    result = Captain::AssistantOverviewSummaryService.new(
-      account: Current.account,
-      assistant: @assistant,
-      range: params[:range],
-      timezone_offset: params[:timezone_offset]
-    ).perform
+    result = Rails.cache.read(overview_summary_cache_key) || generate_overview_summary
 
     if result[:error]
       render json: { error: result[:error] }, status: :unprocessable_content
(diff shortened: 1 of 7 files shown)


enterprise/app/services/captain/assistant_overview_summary_service.rb
```diff
@@ -0,0 +1,64 @@
...
+      resolution_flow: Captain::AssistantResolutionFlowBuilder.new(assistant, range, timezone_offset).build,
+      resolution_trend: Captain::AssistantResolutionTrendStatsBuilder.new(assistant, range, timezone_offset).metrics
+    }
+  end
+
+  def stats_window
+    @stats_window ||= Captain::AssistantStatsWindow.new(range, timezone_offset)
+  end
+
+  def extract_points(message)
```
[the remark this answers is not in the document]
scmmishra: Stupid comment

enterprise/app/controllers/api/v1/accounts/captain/assistant_stats_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+  before_action -> { authorize(Captain::Assistant, :metrics?) }
+  before_action :set_assistant
+  before_action :validate_timezone_offset, only: :overview_summary
+
+  def overview
+    render json: Captain::AssistantOverviewStatsBuilder.new(@assistant, params[:range], params[:timezone_offset]).metrics
+  end
+
+  def overview_summary
+    result = Rails.cache.read(overview_summary_cache_key) || generate_overview_summary
```
sony-mathew: Since this Rails.cache puts things in local folder `tmp/cache/` I guess this is applicable temporarily to the specific pod. Should we change config.cache_store to redis. Does not make sense to store this in a specific pod as we do not have any specific routing for users/accounts to go to the same pod.

enterprise/app/controllers/api/v1/accounts/captain/assistant_stats_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def generate_overview_summary
+    result = Captain::AssistantOverviewSummaryService.new(
+      account: Current.account,
+      assistant: @assistant,
+      range: params[:range],
+      timezone_offset: params[:timezone_offset]
+    ).perform
+
+    Rails.cache.write(overview_summary_cache_key, result, expires_in: OVERVIEW_SUMMARY_CACHE_TTL) unless result[:error]
```
sony-mathew: Same as above comment ^^

enterprise/app/controllers/api/v1/accounts/captain/assistant_stats_controller.rb
sony-mathew: Or are we overriding this in production builds? @vishnu-narayanan  do you know?

sony-mathew (approved):
Looks good overall. Please take a look at the comment added and the following findings:

- ### [P2] Resolved thread remains unfixed: malformed provider values become cached insights

[The extraction logic](/worktrees/pr-15481/chatwoot/enterprise/app/services/captain/assistant_overview_summary_service.rb:50) calls `to_s` on every provider-supplied point. On the exact head, the probe produced:

```ruby
["valid insight", "{unsafe: \"provider-field\"}", "42"]
```

It also accepted a 500-character string despite the response schema’s 300-character limit.

RubyLLM sends the schema to the provider and parses JSON, but does not locally validate the returned object. This matters for configurable OpenAI-compatible endpoints and malformed upstream responses. The resulting values are rendered as successful insights and cached for one hour.

This matches the [resolved malformed-output thread](https://github.com/chatwoot/chatwoot/pull/15481#discussion_r3783007525) and Ito’s latest current-head failure. Resolution state does not reflect the code.

Accept only strings satisfying the schema—or validate the complete returned object locally—before caching.

- ### [P2] Concurrent cache misses duplicate unmetered LLM requests

[The read-generate-write sequence](/worktrees/pr-15481/chatwoot/enterprise/app/controllers/api/v1/accounts/captain/assistant_stats_controller.rb:13) is not atomic. Two simultaneous identical requests from an ordinary account agent both observed a miss and invoked `AssistantOverviewSummaryService#perform`; the exact reproduction recorded two provider calls and two successful responses.

This endpoint:

- Is available to agents because `metrics?` permits them.
- Performs a synchronous external LLM call.
- Explicitly does not count toward Captain usage.
- Has no endpoint-specific throttling or single-flight protection.

The hourly cache therefore limits sequential calls but not bursts, exposing avoidable provider cost and request-capacity consumption. Ito also reported this in its latest issue comment, though no inline thread tracks it.

Use a distributed per-cache-key single-flight lock, with errors remaining uncached, and add a concurrency regression spec.

scmmishra:
@sony-mathew both P2 suggestions are technically valid. About the first 1, it's extremely rare, that this happens, so we can safely ignore it. No need to be defensive for this. OpenAI ensures that the response schema is respected. 
About the second one, it's fine, and acceptable.

PR #15482, opened 2026-08-14
feat(captain): warn about delayed follow-up conflicts

aakashb95:
## Summary

This is stacked on #15472. Review this pull request against `codex/delayed-automation-pending`.

Captain System settings now warn administrators when Captain may resolve or hand off a conversation before a matching delayed follow-up automation runs. A singular warning links directly to that automation, while multiple matches link to the delayed automation list.

The warning uses only automations that can affect an inbox where Captain inactivity resolution can run. The Assistant API exposes the matching automation IDs and delays only to administrators.

## Review fixes

- Handle included, excluded, present, and missing inbox filters when finding overlap.
- Ignore Assistant fetch and save responses after the selected Assistant changes.
- Keep cached settings when a detail refresh fails and never expose an editable default form.
- Wait for automation filter options before opening a deep-linked automation.

## Validation

- 16 focused frontend tests passed.
- 47 focused backend examples passed.
- Three deterministic Playwright scenarios passed for the warning link, out-of-order Assistant responses, and failed refresh fallback.
- ESLint, Prettier, RuboCop, JSON parsing, and `git diff --check` passed.

### DIFF
--- a/app/javascript/dashboard/components-next/captain/pageComponents/assistant/settings/AssistantSystemSettingsForm.spec.js
+++ b/app/javascript/dashboard/components-next/captain/pageComponents/assistant/settings/AssistantSystemSettingsForm.spec.js
@@ -12,7 +12,10 @@ vi.mock('vue-i18n', () => ({
 }));
 
 vi.mock('dashboard/composables/useAccount', () => ({
-  useAccount: () => ({ isCloudFeatureEnabled: () => true }),
+  useAccount: () => ({
+    accountScopedRoute: (name, params, query) => ({ name, params, query }),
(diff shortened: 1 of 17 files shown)


aakashb95:
Closing this PR for now because:
- The warning only works when the delayed automation already exists.
- Users get no warning if they configure Captain Auto resolve timer first and then create the delayed automation later.
- The warning only checks if the delayed automation sends messages or attachments.
- We have not defined the expected behaviour for other automation actions.
- The warning informs users but does not prevent the conflict.
- The change adds too much code for a small UX improvement.
- We can add simple guidance or fix the behaviour later if customers face this problem.

PR #15493, opened 2026-08-16
fix: skip business email validation on self-hosted installations (#14805)

### LINKED ISSUE #14805: Work email validation blocking signup on self-hosted installation
### Describe the bug

On my self-hosted Chatwoot instance (Docker, `chatwoot/chatwoot:latest`), users get a validation error "Please enter a valid work email" immediately when typing a Gmail or personal email address in the signup form. The error appears before the form is even submitted.

According to discussion #12690, maintainer @pranavrajs confirmed: *"You can already do this on self-hosted installations. There's no restriction on using any valid email."* However, my installation is clearly enforcing this restriction.

### To Reproduce

1. Set up a self-hosted Chatwoot instance 
2. Go to `/app/auth/signup`
3. Type a Gmail address (e.g. `test@gmail.com`) in the email field
4. Observe the error "Please enter a valid work email" appears immediately, before clicking submit

### Expected behavior

_No response_

### Environment

Docker

### Cloud Provider

### Platform

None

### Operating system

_No response_

### Browser and version

_No response_

### Docker (if applicable)

- Chatwoot version: `chatwoot/chatwoot:latest`

### Additional context

_No response_

wakqasahmed:
## What
The signup form's `businessEmailValidator` (backed by `company-email-validator`) ran unconditionally in `Form.vue`, blocking personal/free-provider email addresses (e.g. Gmail) at the client level before the form was even submitted — including on self-hosted installations.

This PR gates the validator so it only runs when `globalConfig.installationName === 'Chatwoot'`, reusing the existing `isAChatwootInstance` pattern already present in the parent `Index.vue`.

## Why
Closes #14805.

On discussion #12690, maintainer @pranavrajs confirmed the intended behavior:

> You can already do this on self-hosted installations. There's no restriction on using any valid email.
>
> On the cloud side, our experience has been different. When we allowed signups using Gmail or Yahoo addresses, we saw a lot of abuse. After we restricted those domains, both spam and low-quality signups dropped significantly. Because of that, we don't plan to re-enable those domains for new signups.

So the free-email restriction is meant to be a Chatwoot Cloud-only policy, not a self-hosted default. The reporter's self-hosted instance was hitting the restriction unexpectedly because `Form.vue` had no installation gate at all (unlike `Index.vue`, which already conditions cloud-only UI on `isAChatwootInstance`).

I did not touch server-side validation — `SignUpEmailValidationService`/`AccountBuilder` (see closed PR #14493) already scope free-email enforcement to specific paths and were not part of this bug report; this fix is limited to the client-side validator that was surfacing the wrong error immediately on self-hosted installs.

## Test plan
- Added `Form.spec.js` (new file, no prior spec existed for this component) with two cases:
  - a personal Gmail address is flagged invalid when `installationName === 'Chatwoot'`
  - the same address passes when `installationName` is anything else (self-hosted)
- Verified the new test fails against the pre-fix code (`git stash` the `Form.vue` change) and passes once the fix is applied, confirming it actually exercises the regression.
- Ran `npx vitest run app/javascript/v3/views/auth/signup/components/Signup/Form.spec.js` — 2/2 passing.
- Ran `npx eslint` on both changed files — clean.
- Did not run the full frontend/backend suite (out of scope for this change) or a live self-hosted signup flow in a browser.

### DIFF
--- a/app/javascript/v3/views/auth/signup/components/Signup/Form.spec.js
+++ b/app/javascript/v3/views/auth/signup/components/Signup/Form.spec.js
@@ -1,8 +1,9 @@
 import { mount } from '@vue/test-utils';
 import Form from './Form.vue';
 
-const { globalConfigState } = vi.hoisted(() => ({
+const { globalConfigState, isOnChatwootCloudState } = vi.hoisted(() => ({
   globalConfigState: { value: {} },
+  isOnChatwootCloudState: { value: false },
 }));
 
 vi.mock('vuex', () => ({
@@ -11,6 +12,13 @@ vi.mock('vuex', () => ({
       'globalConfig/get': globalConfigState.value,
     },
   }),
+}));
+
+vi.mock('dashboard/composables/store.js', () => ({
+  useMapGetter: getter =>
+    getter === 'globalConfig/isOnChatwootCloud'
+      ? isOnChatwootCloudState
+      : { value: undefined },
 }));
 
 vi.mock('vue-i18n', () => ({
@@ -60,8 +68,9 @@ describe('Signup Form', () => {
     window.chatwootConfig = { allowedLoginMethods: ['email'] };
   });
 
-  it('rejects a personal email address when installationName is Chatwoot', async () => {
+  it('rejects a personal email address on Chatwoot Cloud', async () => {
     globalConfigState.value = { installationName: 'Chatwoot' };
+    isOnChatwootCloudState.value = true;
     const wrapper = mountForm();
     const emailInput = wrapper.find('input[name="email_address"]');
 
@@ -71,8 +80,33 @@ describe('Signup Form', () => {
     expect(emailInput.classes()).toContain('error');
   });
 
-  it('skips the business email restriction on self-hosted installations', async () => {
+  it('skips the business email restriction on a rebranded self-hosted installation', async () => {
     globalConfigState.value = { installationName: 'My Self Hosted Chatwoot' };
+    isOnChatwootCloudState.value = false;
+    const wrapper = mountForm();
+    const emailInput = wrapper.find('input[name="email_address"]');
+
+    await emailInput.setValue('personal@gmail.com');
+    await emailInput.trigger('blur');
+
+    expect(emailInput.classes()).not.toContain('error');
+  });
+
+  it('skips the business email restriction on a stock self-hosted installation with the default installation name', async () => {
+    globalConfigState.value = { installationName: 'Chatwoot' };
+    isOnChatwootCloudState.value = false;
+    const wrapper = mountForm();
+    const emailInput = wrapper.find('input[name="email_address"]');
+
+    await emailInput.setValue('personal@gmail.com');
+    await emailInput.trigger('blur');
+
+    expect(emailInput.classes()).not.toContain('error');
+  });
+
+  it('skips the business email restriction when globalConfig is empty', async () => {
+    globalConfigState.value = {};
+    isOnChatwootCloudState.value = false;
     const wrapper = mountForm();
     const emailInput = wrapper.find('input[name="email_address"]');
 
--- a/app/javascript/v3/views/auth/signup/components/Signup/Form.spec.js
+++ b/app/javascript/v3/views/auth/signup/components/Signup/Form.spec.js
@@ -1,0 +1,84 @@
+import { mount } from '@vue/test-utils';
+import Form from './Form.vue';
+
+const { globalConfigState } = vi.hoisted(() => ({
+  globalConfigState: { value: {} },
+}));
+
+vi.mock('vuex', () => ({
+  useStore: () => ({
+    getters: {
+      'globalConfig/get': globalConfigState.value,
+    },
+  }),
+}));
+
+vi.mock('vue-i18n', () => ({
+  useI18n: () => ({
+    t: key => key,
+  }),
+}));
+
+vi.mock('vue-router', () => ({
+  useRouter: () => ({ push: vi.fn() }),
+}));
+
+vi.mock('dashboard/composables', () => ({
+  useAlert: vi.fn(),
+}));
+
+vi.mock('../../../../../api/auth', () => ({
+  register: vi.fn(),
+}));
+
+vi.mock('@hcaptcha/vue3-hcaptcha', () => ({
+  default: { name: 'VueHcaptcha', template: '<div />' },
+}));
+
+vi.mock('../../../../../components/GoogleOauth/Button.vue', () => ({
+  default: { name: 'GoogleOAuthButton', template: '<div><slot /></div>' },
+}));
+
+const mountForm = () =>
+  mount(Form, {
+    global: {
+      stubs: {
+        FormInput: {
+          props: ['modelValue', 'name', 'hasError'],
+          emits: ['update:modelValue'],
+          template:
+            '<input :name="name" :class="{ error: hasError }" :value="modelValue" @input="$emit(\'update:modelValue\', $event.target.value)" />',
+        },
+        NextButton: true,
+        PasswordRequirements: true,
+      },
+    },
+  });
+
+describe('Signup Form', () => {
+  beforeEach(() => {
+    window.chatwootConfig = { allowedLoginMethods: ['email'] };
+  });
+
+  it('rejects a personal email address when installationName is Chatwoot', async () => {
+    globalConfigState.value = { installationName: 'Chatwoot' };
+    const wrapper = mountForm();
+    const emailInput = wrapper.find('input[name="email_address"]');
+
+    await emailInput.setValue('personal@gmail.com');
+    await emailInput.trigger('blur');
+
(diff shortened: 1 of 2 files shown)


app/javascript/v3/views/auth/signup/components/Signup/Form.vue
```diff
@@ -31,12 +31,18 @@ const credentials = reactive({
...
+  () => globalConfig.value.installationName === 'Chatwoot'
+);
+
 const rules = {
   credentials: {
     email: {
       required,
       email,
       businessEmailValidator(value) {
+        if (!isAChatwootInstance.value) return true;
```
wakqasahmed: This gate does not fix the case reported in #14805.

`INSTALLATION_NAME` defaults to `Chatwoot` for every self-hosted installation (`config/installation_config.yml`, `- name: INSTALLATION_NAME / value: 'Chatwoot'`). The reporter is running stock `chatwoot/chatwoot:latest` with no branding override, so `globalConfig.installationName === 'Chatwoot'` on their instance too, `isAChatwootInstance` is `true`, and the work-email error still fires. Only installations that have explicitly rebranded would benefit.

The free-email restriction is a *cloud* anti-abuse policy (per @pranavrajs in #12690), and the store already exposes exactly that distinction:

```js
isOnChatwootCloud: $state => $state.deploymentEnv === 'cloud',
```
(`app/javascript/shared/store/globalConfig.js`)

Suggest gating on `globalConfig/isOnChatwootCloud` instead. `isAChatwootInstance` in `Index.vue` answers a different question (branding/layout), so reusing it here reads consistent but is semantically the wrong predicate.

app/javascript/v3/views/auth/signup/components/Signup/Form.vue
```diff
@@ -31,12 +31,18 @@ const credentials = reactive({
   hCaptchaClientResponse: '',
 });
 
+const globalConfig = computed(() => store.getters['globalConfig/get']);
+const isAChatwootInstance = computed(
+  () => globalConfig.value.installationName === 'Chatwoot'
```
wakqasahmed: Minor: this duplicates an existing getter. `app/javascript/shared/store/globalConfig.js` already ships `isAChatwootInstance: $state => $state.installationName === 'Chatwoot'`. If the installationName-based predicate is kept, prefer `store.getters['globalConfig/isAChatwootInstance']` (or `useMapGetter('globalConfig/isAChatwootInstance')`, the pattern used in `shared/composables/useBranding.js`) rather than re-deriving the string comparison in a third place.

app/javascript/v3/views/auth/signup/components/Signup/Form.spec.js
```diff
@@ -0,0 +1,84 @@
...
+    const emailInput = wrapper.find('input[name="email_address"]');
+
+    await emailInput.setValue('personal@gmail.com');
+    await emailInput.trigger('blur');
+
+    expect(emailInput.classes()).toContain('error');
+  });
+
+  it('skips the business email restriction on self-hosted installations', async () => {
+    globalConfigState.value = { installationName: 'My Self Hosted Chatwoot' };
```
wakqasahmed, line 84: This is the gap that hides the issue flagged on `Form.vue`: the "self-hosted" case is modelled as `installationName: 'My Self Hosted Chatwoot'`, i.e. a *rebranded* install. The bug report is a stock Docker install, where `installationName` is the default `'Chatwoot'`. A test case for "self-hosted with default installation name" would fail against this implementation.

Two other test-quality notes:
- The assertion goes through a stubbed `FormInput`'s `error` class, which couples the test to the stub template. Asserting on the vuelidate state directly (e.g. `v$.credentials.email.businessEmailValidator.$invalid`) exercises the actual gate rather than rendering.
- Neither case covers `globalConfig` being `{}` / `installationName` undefined. Today that fails open (validator skipped); worth pinning down explicitly since it is the shape the getter returns before/without `window.globalConfig`.

wakqasahmed (commented):
Independent review of the diff against #14805 and discussion #12690.

**Summary:** the direction is right (this is a client-only UX gate; the server never enforced free-provider blocking), but as written the fix does not resolve the reported scenario.

1. **Blocking — wrong predicate.** `installationName === 'Chatwoot'` is the *default* for self-hosted installs (`config/installation_config.yml` sets `INSTALLATION_NAME: 'Chatwoot'`). The reporter runs stock `chatwoot/chatwoot:latest`, so the gate evaluates true on their instance and the "valid work email" error still appears. Only rebranded installs are fixed. `globalConfig/isOnChatwootCloud` (`deploymentEnv === 'cloud'`) is the predicate that matches the maintainer-stated policy in #12690. Inline comment on `Form.vue`.

2. **Reuse.** `globalConfig/isAChatwootInstance` already exists in the store; the local computed re-derives it. Inline comment.

3. **Test coverage.** `Form.spec.js` models "self-hosted" as a renamed install, so it cannot catch (1). Missing: default-named self-hosted, and the empty-`globalConfig` shape. The assertion also runs through a stubbed `FormInput`'s class binding rather than the vuelidate state. It would catch a straight revert of the gate, so it is not vacuous — just narrower than the bug.

**Non-issues checked:**
- No crash risk from an unloaded `globalConfig`: the store builds state synchronously from `window.globalConfig || {}`, so `globalConfig.value` is always an object; `installationName` merely comes back `undefined` (fails open).
- Reactivity is fine — the validator body reads `isAChatwootInstance.value` inside vuelidate's computed, so the dependency is tracked.
- No security regression: `Account::SignUpEmailValidationService` only enforces `ValidEmail2` validity, disposable-domain and the configurable `BLOCKED_EMAIL_DOMAINS`; it never blocked free providers. Relaxing the client check does not bypass any server-side control (and correspondingly, the cloud restriction remains client-only, unchanged by this PR).
- Style is consistent with the surrounding `<script setup>` composition API code.

Not approving — leaving merge judgement to maintainers.

app/javascript/v3/views/auth/signup/components/Signup/Form.vue
wakqasahmed: Fixed. Gated on `globalConfig/isOnChatwootCloud` (deploymentEnv === 'cloud') instead of installationName, via `useMapGetter`. Added a test case for a default-named self-hosted install (installationName left at 'Chatwoot', deploymentEnv not 'cloud') which now correctly skips the validator, plus an empty-globalConfig case. Verified via a revert-and-restore regression check that the new default-named test fails against the old implementation and passes with the fix. See d15f3839a.

app/javascript/v3/views/auth/signup/components/Signup/Form.vue
wakqasahmed: Addressed — switched to `useMapGetter('globalConfig/isOnChatwootCloud')` (the composable used pervasively elsewhere, e.g. `dashboard/composables/useAccount.js`) instead of a locally re-derived computed. This also resolves the blocking predicate issue at the same time.

app/javascript/v3/views/auth/signup/components/Signup/Form.spec.js
wakqasahmed: Added: a case for a stock self-hosted install with the default installation name ('Chatwoot') and deploymentEnv not 'cloud' (now correctly skips the validator, previously would have failed), and a case for an empty globalConfig object. Confirmed via revert-and-restore that the new default-named case fails on the old implementation and passes on the fix.

wakqasahmed:
Hi @chatgpt-codex-connector — just checking in on this one, it's been about 4 days with CI green and mergeable. Is there anything needed from our side to help move it forward, or any rough idea on timing? No rush, just following up. Thanks for the review!

wakqasahmed:
Checked back in on this — all inline review threads (installation-name vs. isOnChatwootCloud gating, duplicate getter, test coverage for the default-branded self-hosted case) already have replies and were addressed in d15f3839a. No open review feedback remains as far as I can tell; let me know if I'm missing something.

wakqasahmed:
Hi @chatgpt-codex-connector — following up again — still green and mergeable, about 4 days since the last activity. Happy to change anything that's blocking it, otherwise just don't want it lost. Thanks!

app/javascript/v3/views/auth/signup/components/Signup/Form.vue
```diff
@@ -31,12 +31,18 @@ const credentials = reactive({
...
+);
```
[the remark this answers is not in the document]
wakqasahmed: Confirmed, thanks — you're right that `installationName === 'Chatwoot'` is the self-hosted default, so this gate didn't actually fix #14805. Switched to `globalConfig/isOnChatwootCloud` (`deploymentEnv === 'cloud'`), which matches the maintainer-stated cloud-only policy in #12690, in commit d15f3839a. Added a regression test for a stock self-hosted install (default installation name, non-cloud deploymentEnv) that fails on the old gate and passes with the fix. See the reply threads below for the reuse and test-coverage points too.

PR #15503, opened 2026-08-17
fix: dropdown clipping and condition label in automation rules

iamsivin:
# Pull Request Template

## Description

This PR fixes condition and action dropdowns in the automation rule panel being clipped at the edges of the panel body. The dropdowns now open at their full height, flip above the trigger when there isn’t enough space below, and stay within the viewport on shorter screens. Removed the height restriction that was limiting the action dropdown to around 120px.

It also renames the **Message Contains** condition to **Message Content**. Previously, the row could read **Message Contains → Equal to → hello**, which made the attribute and operator confusing. The new label matches the other conditions such as Message Type, Private Note, Email, and Inbox, where the attribute names the field and the operator provides the comparison.

Fixes https://linear.app/chatwoot/issue/CW-7900/dropdown-gets-cut-off-in-the-new-automation-rule-modal

## Type of change

### How to reproduce

1. Settings → Automation → Create Automation.
2. Set the condition attribute to **Inbox** and open its value dropdown. The list is clipped at the edge of the Conditions box.
3. Open the **Assign to Agent** value dropdown. The list is limited to around 120px, and dropdowns near the bottom of the panel are cut off by the footer instead of opening upwards.
4. Add the **Message Content** condition and verify that the attribute is labelled correctly and the operator provides the comparison.

## Checklist:


sony-mathew (approved):
@iamsivin  please check this finding:

- **P2 — Dropdowns can remain off-screen after a viewport resize** — Matches the [existing open thread](https://github.com/chatwoot/chatwoot/pull/15503#discussion_r3796782806). The new floating path applies `fixedPosition` directly in [DropdownFloating.vue](/worktrees/9c29/chatwoot/app/javascript/dashboard/components-next/dropdown-menu/base/DropdownFloating.vue:15), using the above/below calculations in [useDropdownPosition.js](/worktrees/9c29/chatwoot/app/javascript/dashboard/composables/useDropdownPosition.js:71).

  Exact-head reproduction:

  1. Open the 19-item action dropdown at 1280×720.
  2. Resize to 900×240 while it remains open.
  3. The menu remains 378px tall at y=137–515, with `max-height: 499px` and `bottom: -275px`.
  4. Reopening at 900×240 recalculates correctly to y=144–224 with an 80px scroll area.

  Therefore, initial rendering on a short viewport works, but an active viewport transition can leave lower options unreachable. Window resizing is a direct path; mobile keyboard or orientation changes are plausible equivalents. The fix should remeasure/clamp after layout settles or close/reposition on viewport changes. Merely imposing a positive minimum height could still force overflow.

## Verdict

The underlying CW-7900 bug is valid, and the PR correctly fixes ordinary panel clipping, restores the action list from its previous 120px cap to a 320px scroll list, and renders “Message Content → Equal to → value” clearly.

muhsin-k (approved):

PR #15506, opened 2026-08-17
fix(i18n): register missing Ukrainian locale files (contentTemplates, yearInReview)

AndriyRyabchenko:
### Summary

`app/javascript/dashboard/i18n/locale/uk/index.js` does not import two locale files that exist and are already translated into Ukrainian, so their strings fall back to English in the dashboard:

- `contentTemplates.json`
- `yearInReview.json`

### Changes

Add the two missing `import` statements and their spreads, in alphabetical order to match the existing convention. No string content is changed (handled via Crowdin).

### Note on snooze

`snooze.json` is also unregistered for `uk`, but it is intentionally **left out** of this PR: its `SNOOZE_PARSER.UNITS` are mistranslated (`WEEK` → `день`, `MONTH` → `тиждень`, `YEAR` → `місяць`, plus untranslated plurals). Registering it as-is would make the snooze parser misread typed input. It should be corrected via Crowdin first, then registered separately. Thanks to the automated review for catching this.

### DIFF
--- a/app/javascript/dashboard/i18n/locale/uk/index.js
+++ b/app/javascript/dashboard/i18n/locale/uk/index.js
@@ -39,7 +39,6 @@ import sessionLimit from './sessionLimit.json';
 import settings from './settings.json';
 import signup from './signup.json';
 import sla from './sla.json';
-import snooze from './snooze.json';
 import teamsSettings from './teamsSettings.json';
 import whatsappTemplates from './whatsappTemplates.json';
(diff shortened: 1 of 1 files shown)


AndriyRyabchenko:
Good catch on the snooze parser — confirmed. I've dropped `snooze` from this PR and kept only `contentTemplates` and `yearInReview` (display strings, no parser impact). `uk/snooze.json` needs its `SNOOZE_PARSER.UNITS` corrected via Crowdin before it can be safely registered.

PR #15508, opened 2026-08-18
feat: add CSV FAQ import

aakashb95:
## Summary

- Add CSV import under the existing Create new FAQ action.
- Validate UTF-8 CSV files with question and answer columns, up to 1,000 rows, and show a read-only preview.
- Skip invalid and repeated rows, download invalid rows, and let administrators skip or overwrite existing FAQs.
- Save imported FAQs as manual FAQs and rebuild embeddings before search can use them.
- Show the latest import status and recover stalled row or embedding work without queuing duplicate recovery jobs.
- Keep import history when an importing user is deleted, and clean up imports when their account or Assistant is deleted.
- Handle concurrent preview replacement and confirmation safely.
- Retry temporary embedding failures and overwrite only the FAQ selected during preview.
- Reject files over the configured upload limit and make invalid-row downloads safe for spreadsheets.

## Tests

- 49 focused RSpec examples
- 18 focused Vitest tests
- Deterministic Playwright coverage for preview, UTF-8 upload, overwrite, invalid-row download, import progress, and cleanup
- RuboCop, ESLint, Prettier, and git diff --check


aakashb95:
closing this since it has become too big for review. will split.

Issue #15359, opened 2026-08-07
[Bug]: Merged WhatsApp contacts can route replies to the wrong destination


jochenstu:
## Problem

When two WhatsApp contacts with different routing identities are merged in the dashboard, the surviving Chatwoot contact owns both sets of ContactInbox rows and conversations. Contact-wide conversation lookup can append an event from identity A to identity B's conversation. Replies then use the stale conversation.contact_inbox.source_id on B's conversation and can be sent to the wrong recipient.

The dashboard can hide the mismatch because both conversations display the merged base contact's fields.

Reproduction

1. In one WhatsApp inbox, create identity A with its own ContactInbox and conversation.
2. Create identity B with a different phone/BSUID and conversation.
3. Merge B into A in the dashboard.
4. Receive a message or call from identity A.
5. Observe whether Chatwoot selects a conversation belonging to identity B.
6. Reply and compare the inbound identifier with conversation.contact_inbox.source_id.

Expected behavior

An inbound message, echo, or call from identity A must resolve the correct eligible ContactInbox for A and may reuse only a conversation already attached to that row. If no eligible conversation exists there, Chatwoot creates one.

The selected routing model does not require every phone and BSUID alias on the same Chatwoot contact to share one historical conversation. An older phone conversation may remain under Previous Conversations while a later BSUID-only conversation becomes active. Mixed phone+BSUID payloads should preserve phone-backed conversation continuity when the phone is present.

Acceptance criteria

* A merged contact containing phone A/BSUID A and phone B/BSUID B never reuses B's conversation for an event from A, or A's conversation for an event from B.
* Outbound replies use the exact source_id selected for the inbound identity.
* Meta Cloud message, SMB echo, and inbound-call paths apply the same provider-shaped selected-ContactInbox rule.
* Mixed phone+BSUID payloads preserve phone-backed conversation continuity when the phone is present.
* BSUID-only payloads route through the exact BSUID ContactInbox.
* No conversation is repointed across ContactInbox rows.
* Display name, username, phone field on the base contact, conversation recency, or shared contact ownership never determines routing.
* Cover locked and unlocked conversation modes and repeated/concurrent events.
* Preserve non-Cloud provider behavior while preventing cross-identity routing in each supported provider path.

[...]

Issue #14931, opened 2026-07-05
Route WhatsApp events through the correct phone or BSUID ContactInbox


slmartinez:
## Outcome

Route WhatsApp Cloud messages, SMB echoes, and inbound calls through the correct provider-shaped `ContactInbox` without moving existing conversations or searching conversations across the whole contact.

This issue implements the corrected decision recorded in ⁠ [CW-7915](https://linear.app/chatwoot/issue/CW-7915/decision-preserve-phone-conversations-for-mixed-bsuid-payloads) Decision: Preserve phone conversations for mixed BSUID payloads. The selected model is phone-continuity-first for mixed phone+BSUID payloads, and exact-BSUID routing when the phone number is absent.

## Problem

A customer can already have a phone-backed conversation when Meta begins including a regular or parent BSUID. Meta can send both phone and BSUID in the same webhook. If Chatwoot always prefers BSUID whenever present, it creates a second conversation even though the phone-backed thread is still valid and expected by the customer.

At the same time, contact-wide conversation lookup is unsafe after a dashboard merge because the surviving contact can own conversations for unrelated WhatsApp identities. Reusing a conversation merely because it belongs to the same Chatwoot contact can route replies to the wrong destination.

## Selected behavior

For Meta Cloud:

When a webhook includes a usable phone number, preserve existing phone-backed conversation continuity.

If this is a brand-new mixed phone+BSUID customer, create the first conversation on the phone ContactInbox and sync BSUID aliases to the same contact.

## Acceptance criteria

When the phone number is absent, resolve the exact BSUID ContactInbox and reuse or create a conversation on that row.

Leave existing phone and BSUID conversations unchanged under Previous Conversations.

Never repoint a conversation from one ContactInbox to another.

Never select a conversation merely because it belongs to the same Chatwoot contact.

Apply the same provider-shaped routing rule to incoming messages, SMB message echoes, and inbound WhatsApp calls.

Twilio retains its provider-specific behavior and remains phone-first when a phone number is present. 360Dialog is unsupported and excluded from this issue.

Acceptance criteria

A mixed Cloud event containing phone, regular BSUID, and parent BSUID for a contact with existing phone-backed history reuses the phone-backed conversation.

A brand-new mixed Cloud event creates the first conversation on the phone ContactInbox and syncs regular/parent BSUID aliases to the same contact.

A BSUID-only Cloud event creates or reuses the exact BSUID ContactInbox conversation.

[...]

Issue #13837, opened 2026-03-18
WhatsApp Business-Scoped User IDs (BSUID) support needed before June 2026


m-ret:
## Current Chatwoot direction

Chatwoot is implementing BSUID support in stages. The current routing decision is:

- When Meta sends both phone and BSUID, preserve phone-backed conversation continuity when phone history exists.
- For a brand-new mixed phone+BSUID customer, create the first conversation on the phone `ContactInbox` and sync the BSUID aliases to the same contact.
- When the phone number is absent, route through the exact BSUID `ContactInbox`.
- Do not repoint conversations across `ContactInbox` rows.
- Do not use contact-wide conversation lookup, because dashboard-merged contacts can contain unrelated WhatsApp identities.
- Twilio keeps its provider-shaped, phone-first behavior when a phone number is present.

The selected routing decision is tracked in CW-7915/CW-7536 and corrected by chatwoot/chatwoot#15630 after chatwoot/chatwoot#15175 used a broader identifier-first interpretation.

The original report below is retained for context.

## Summary

Meta is launching [WhatsApp Usernames](https://developers.facebook.com/documentation/business-messaging/whatsapp/business-scoped-user-ids) — users will be able to set an `@username` and **hide their phone number** from businesses. To support this, Meta is adding a **Business-Scoped User ID (BSUID)** to all webhook payloads via a new `user_id` field.

**Starting June 2026**, the `wa_id` / `from` fields that currently always contain a phone number **may contain a BSUID instead** (format: `CC.alphanumeric`, e.g. `BR.1A2B3C4D5E6F7G8H9I0J...`). This will break Chatwoot's contact resolution for username-adopting users.

## Timeline

| Date | What happens |
|------|-------------|
| Early April 2026 | Contact Book feature — Meta auto-stores phone+BSUID pairs from prior interactions |
| May 2026 | BSUIDs start appearing in **all** webhook payloads (testing phase) |
| **June 2026** | Usernames launch in test countries — `wa_id` may be BSUID, not phone |
| August 2026+ | Global rollout |

## What breaks in Chatwoot

Chatwoot currently uses the phone number (`wa_id`) as the primary identifier for WhatsApp contacts:

1. **`Whatsapp::IncomingMessageBaseService`** resolves contacts via `wa_id` assuming it's always a phone number
2. **`ContactInbox`** uses `source_id` (phone number in E.164) with a unique index on `(inbox_id, source_id)`
[...]

PR #15533, opened 2026-08-19
feat(whatsapp): add an identity group for BSUID and phone contact inboxes

marcoazcabral:
> **Status:** draft, inert, and staying that way. @muhsin-k asked in #15357 to keep the identity-group work separate from the lifecycle PR, and this description has been corrected to match what that conversation established.

**This is a draft, and it is not a request to merge.** In #13837 I said I was posting the identity group as a proposal rather than a PR on purpose, because it touches shared routing code across CE and Enterprise and the migration is yours to weigh. That is still true. What changed is that a proposal has been easier to agree with than to react to, so this puts the smallest useful amount of code behind it to make the separate evaluation you mentioned concrete.

The decision is yours in both directions: take it forward, or keep it open as the place where the approach gets discussed and close it once a direction exists. Either outcome works for me.

### What this assumes

That the direction settled in #14931 holds: resolve the exact BSUID `ContactInbox`, reuse or create a conversation on it, no contact-wide lookup, no repointing. Nothing here reopens that, and nothing here changes any behaviour at all.

If #15175 lands on a different shape, this should be closed rather than adapted.

### Why this does not ask you to revert anything, and where it stops

Each step in this area fixed the cost of the one before it:

| Step | What it fixed | What it left behind |
|---|---|---|
| contact-wide reuse (#15098) | coexistence stopped splitting the conversation | routing breaks after a dashboard merge |
| exact contact inbox (#14931, #15175) | routing is correct after a merge | a coexistence contact splits into two conversations again |
| identity group, as drafted here | records who is who, from evidence | **cannot be used for conversation lookup** |

That last row is a correction to what this description claimed before, and the credit for it goes to @muhsin-k. The write path groups phone, regular and parent together, because a payload carrying them together does prove they are one person. That makes the grouping accurate as a statement about identity, and unusable as a conversation scope: a lookup reading it would reuse the phone-backed conversation for a BSUID event, which is exactly the outcome #14931 decided against.

So the group records identity evidence only. It must not, on its own, imply conversation reuse or outbound routing. Those are separate questions, and this draft does not answer them.

### What is in the diff

- a nullable `identity_group_id` on `contact_inboxes`

- a backfill that gives every existing row a group of its own

- the write path in `Whatsapp::IdentifierSyncService`, which is the one place that already sees every co-occurrence

The rules are the ones from the proposal. Identifiers carried by one payload join the same group. An existing group is never overwritten. A merge never assigns one, which holds by construction rather than by a check, since the write only runs from an inbound payload. Two identifiers that arrive together while already belonging to different groups are left alone: joining groups needs a deliberate path, and it is the one place where being wrong recreates the routing bug this exists to avoid.

The backfill guesses nothing. One group per row degrades to the behaviour that predates #15098, where one source id was one identity by construction, and groups then form as payloads arrive.

### Nothing reads the column yet, and that is the honest state

The consumer I wanted to include is the rotation path, where the group forms without any inference: `user_id_update` names the previous and the current identifier in the same payload, so the evidence and the grouping arrive together. That path is not upstream. It is #15357, still open.

The call sites that are upstream are `set_conversation` and `Voice::InboundCallBuilder`. The voice builder already scopes to the exact contact inbox. `set_conversation` on develop is still contact-wide, since the narrowing agreed in #14931 lives in #15175 and has not landed yet. I left both alone because converting them changes behaviour, and that belongs with whoever owns the decision rather than with a draft meant to stay inert.

That leaves a schema change with no reader, which I would normally not send. I am sending it because the alternative was another paragraph of prose, and because the shape is easier to argue with than the description was.

### The two cases this is really about

**#15359.** `ContactMergeAction#merge_contact_inboxes` reassigns `contact_id` and preserves every row, while `merge_conversations` leaves each conversation's `contact_inbox_id` untouched. Any rule that treats *same contact* as *same identity* inherits that, which is why the merged-contact routing test added to #15175 has to exist. With groups, an agent merge never assigns one, so rows brought together stay distinct by construction.

**#15357.** A rotation is the opposite case: two identifiers that genuinely are one person, proven by the event itself. The group survives the change, and the conversation does not need to be moved for the reply to reach the right place.

### What would have to change before this is useful

Not the consumer this description used to offer. Scoping `set_conversation` by the group as it stands would cross the routing boundary described above, so that offer is withdrawn.

The shape worth revisiting is narrower: regular and parent identifiers grouped as aliases of one identity, phone left out of that grouping entirely, and outbound routing covered explicitly rather than inherited from whichever row a conversation happens to be anchored to.

That needs evidence that does not exist yet: a real payload where a previously supplied parent identifier is later omitted, which is the case the narrower grouping would repair. Worth knowing how high that bar is. Parent BSUIDs are not automatic, they are opt-in ("If you are a managed business and want to enroll your business portfolios to receive parent BSUIDs, you can ask your Meta point-of-contact to check if you are eligible"), and they exist to share one identifier across separate enrolled portfolios. Installations that never enrolled will never see one, so the evidence can only come from an enrolled business.

### Open questions, unchanged from the proposal

- **Joining groups.** Evidence can arrive later proving two groups are one person. That needs a deliberate path.

- **Enterprise.** The voice path would need the same treatment, and it is not in this draft.

### Verification

RuboCop is clean at the version the CI pins. The new spec passes, and `spec/services/whatsapp`, `spec/models/contact_inbox_spec.rb` and `spec/actions/contact_merge_action_spec.rb` pass together with it, which is the surface the write path touches. I have not run the full suite or anything under `spec/enterprise`.

### Credit

The term is @jochenstu's, from #15359. The proposal built on it is mine, and this draft is the code behind it.


marcoazcabral:
One cost worth raising before anyone has to ask about it.

The backfill writes every row of `contact_inboxes`, and on a large installation that table is not small, since the web widget creates a row for every anonymous visitor. A full rewrite with `gen_random_uuid()` means hours of batched updates, the WAL that goes with them and the replication lag behind that, for a column that nothing reads yet.

The alternative is to drop the backfill entirely and let `identity_group_id IS NULL` mean "not grouped yet, treat as a group of one". The migration then becomes a nullable column and an index, which is close to free, and groups form only as payloads arrive.

It is a real trade rather than a clear win. The backfill buys query simplicity: `WHERE identity_group_id = ...` never matches NULL, so without it every consumer has to handle the ungrouped case explicitly, and that handling moves into each call site instead of being paid once.

I left the backfill in because it is what I described in #13837, but I would rather ship the one you would actually be willing to run. Say which and I will change it, or drop the migration to the column alone and leave the grouping of existing rows out of scope.

marcoazcabral:
I built the conditional consumer locally and measured it, so the offer above is no longer hypothetical.

First a correction to the description, which I got wrong: `set_conversation` on develop is still contact-wide. The narrowing agreed in #14931 lives in #15175 and has not landed. So scoping that lookup by group is not a change to the agreed direction, it is a change to what runs today.

The consumer is one method:

```ruby
def conversations_in_identity_group
  group_id = @contact_inbox.identity_group_id
  return @contact_inbox.conversations if group_id.blank?

  Conversation.where(contact_inbox_id: @inbox.contact_inboxes.where(identity_group_id: group_id).select(:id))
end
```

Two measurements, same database, nothing different but that method.

**No regression.** `spec/services/whatsapp`, `spec/models/contact_inbox_spec.rb` and `spec/actions/contact_merge_action_spec.rb` are 319 examples together, and they pass identically with the contact-wide lookup and with the group one. Nothing in the current suite tells the two apart.

**The case from #15359 does tell them apart.** Two people write in, an agent merges the two contacts by mistake, and then the first person writes again. Its thread is not the most recent one, which is where a contact-wide `.last` lands:

```ruby
it 'does not answer one identity through the source id of another' do
  deliver(from: '5511900000001', wamid: 'wamid.ig-a1')
  deliver(from: '5511900000002', wamid: 'wamid.ig-b1')

  first = inbox.contact_inboxes.find_by(source_id: '5511900000001')
  second = inbox.contact_inboxes.find_by(source_id: '5511900000002')
  first_conversation = first.conversations.last
  second_conversation = second.conversations.last
  expect(first_conversation).not_to eq(second_conversation)

  ContactMergeAction.new(account: account, base_contact: first.contact, mergee_contact: second.contact).perform

  deliver(from: '5511900000001', wamid: 'wamid.ig-a2')

  message = Message.find_by(source_id: 'wamid.ig-a2')
  expect(message.conversation).to eq(first_conversation)
end
```

On develop that fails:

```
reply thread is 282, expected 281 (the other identity's thread is 282)
```

With the group it passes. A person wrote in and the message landed in someone else's thread, which is the same defect #15359 reports, reached through the message path rather than through a reply.

Worth noting what this does not do: it fixes which conversation is found, not which address the reply goes out to. That is still `conversation.contact_inbox.source_id` and is unchanged here.

It also removes the need for the provider conditional in #15175. The lookup stops depending on which row the message arrived on, so 360Dialog needs no special case to keep landing on a thread it can answer.

None of this is pushed to this PR. Say which shape you want and I will push it, either here or separately.

marcoazcabral:
One correction to the description above. It names `user_id_update` as the event that carries a rotation, and that is no longer accurate.

Meta corrected its own documentation on 11 August 2026. `user_id_update` is not a subscribable webhook field and does not appear in the App Dashboard. BSUID changes are delivered in the `messages` field as a `user_changed_user_id` system message carrying `system.previous_user_id` and `system.previous_parent_user_id`, and `user_changed_number` now carries the same two fields, since a phone number change also regenerates the BSUID. Details and the full changelog entry are in #15357.

Nothing in the migration or the backfill depended on that field, so the design here is unchanged. The motivating example is actually stronger now: two generally available system messages carry the lineage explicitly, instead of one field that turned out not to exist.

PR #15538, opened 2026-08-20
fix(sdk): refresh users when contact details change

### LINKED ISSUE #13448: [SDK] setUser() does not update existing contact fields (phone_number, country_code, etc.)
## Describe the bug

The `window.$chatwoot.setUser()` method from the Web SDK does not update existing contact fields when called with new data. Only new contacts get the fields properly set.

## Steps to reproduce

1. Install Chatwoot SDK on a website
2. Create a contact using `setUser()` **without** `phone_number` and `country_code`:
```javascript
window.$chatwoot.setUser("user@example.com", {
  name: "John Doe",
  email: "user@example.com"
});
```
3. Later, call `setUser()` again with the **same identifier** but with additional fields:
```javascript
window.$chatwoot.setUser("user@example.com", {
  name: "John Doe",
  email: "user@example.com",
  phone_number: "+212666778899",
  country_code: "MA",
  city: "Casablanca"
});
```

## Expected behavior

The contact should be **updated** with the new fields:
- `phone_number` should show `+212666778899`
- `country_code` should be set to `MA` (Morocco)
- `city` should be set to `Casablanca`

## Actual behavior

The contact fields **remain empty/unchanged**:
- `phone_number` stays empty
- `country_code` stays empty
- `city` stays empty

Only **new contacts** (created after the code update) get these fields properly set.

## Workaround

The only way to update existing contact fields is to:
1. Manually edit the contact in Chatwoot dashboard
2. Change any field and save
3. Change it back and save again
4. Then the fields sync correctly

OR delete the contact and let it be recreated.

## Environment

- **Chatwoot version**: SaaS (support.nawat.ma)
- **Chatwoot SDK**: Latest (loaded from `/packs/js/sdk.js`)
- **Browser**: Chrome 131
- **Integration**: Odoo 17 backend with custom session data

## Additional context

This appears to be related to issues:
- #7065 - setUser not work anymore
- #7822 - country_code not appearing in contact list
- #8341 - update contact_inboxes.source_id when update contact.phone_number

The SDK seems to only **create** contacts but not **update** them. The API documentation shows `PUT /api/v1/accounts/{account_id}/contacts/{id}` exists for updates, but the SDK's `setUser()` doesn't seem to trigger it for existing contacts.

## Suggested fix

The `setUser()` method should:
1. Check if the contact already exists (by identifier)
2. If yes, **PATCH/PUT** the contact with new data
3. If no, **POST** a new contact

This would make `setUser()` truly idempotent and allow progressive enrichment of contact data.

scmmishra:
Updates the web widget user fingerprint to include contact information supported by setUser. Existing contacts now sync when phone, company, location, description, or social profiles change, while custom attributes remain handled separately.

## Closes

Closes chatwoot/chatwoot#13448

## How to reproduce

1. Identify a contact with setUser.
2. Change only phone_number, country_code, city, company_name, description, or social_profiles.
3. Call setUser again and confirm the existing contact is updated.

### DIFF
--- a/app/javascript/sdk/cookieHelpers.js
+++ b/app/javascript/sdk/cookieHelpers.js
@@ -3,6 +3,23 @@ import Cookies from 'js-cookie';
 
 const REQUIRED_USER_KEYS = ['avatar_url', 'email', 'name'];
 const ALLOWED_USER_ATTRIBUTES = [...REQUIRED_USER_KEYS, 'identifier_hash'];
+const CONTACT_INFORMATION_ATTRIBUTES = [
+  'phone_number',
+  'company_name',
+  'city',
+  'country_code',
+  'description',
+  'social_profiles',
+];
+
+const serializeUserAttribute = value => {
+  if (!value || typeof value !== 'object') return value || '';
+
+  const sortedEntries = Object.keys(value)
+    .sort()
+    .map(key => [key, value[key]]);
+  return JSON.stringify(sortedEntries);
+};
 
 export const getUserCookieName = () => {
   const SET_USER_COOKIE_PREFIX = 'cw_user_';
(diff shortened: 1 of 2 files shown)


muhsin-k (approved):
@scmmishra code looks good to me.

please check this comment as well

[P2] Preserve field boundaries in the fingerprint

Concatenating optional key/value fragments is ambiguous: `{ company_name: 'cityParis' }` and `{ company_name: '', city: 'Paris' }` both produce `company_namecityParis`, so the hashes match and `setUser` suppresses a real update. Serialize the complete selected structure with canonical JSON or another boundary-safe encoding, and cover this cross-field collision in a spec.

scmmishra:
@muhsin-k good catch, fixed

muhsin-k (approved):

PR #15546, opened 2026-08-21
feat(whatsapp): route BSUID calls through the active identity

muhsin-k:
WhatsApp customers can become addressable only by a BSUID after WhatsApp stops sharing their phone number. Text messages already follow the exact identity attached to the conversation, but outbound calling still assumes that the contact has a phone number. This hides or rejects the call action for BSUID-only customers and can select the wrong destination after contacts are merged.

This PR makes the current conversation's WhatsApp identity authoritative for outbound calls. Calls and call-permission requests use the exact `ContactInbox#source_id` attached to that conversation. Phone-number calling continues to work as before.

### What changed

- Show the call action for a BSUID-only conversation when its WhatsApp inbox has calling enabled.
- Send the BSUID in Meta's `recipient` field for outbound calls and call-permission requests.
- Keep phone-number recipients on the existing `to` field.
- Preserve the selected conversation instead of guessing between identities owned by a merged contact.
- Keep contact-level calling phone-only when no conversation has established the destination.

### Things to know

- This change applies to Meta Cloud WhatsApp calling. Twilio Voice does not use the WhatsApp BSUID calling API and remains unchanged.
- This PR is based on [#15175](https://github.com/chatwoot/chatwoot/pull/15175). Once that PR is merged, this PR should be retargeted to `develop`.

### How to test

1. Use a Meta Cloud WhatsApp inbox with calling enabled.
2. Open a conversation whose `ContactInbox#source_id` is a BSUID and whose contact has no phone number.
3. Confirm the call action is available and start an outbound call.
4. Confirm the call remains in the same conversation and Meta receives the BSUID as `recipient`.
5. Test a customer without call permission and confirm the permission request is sent to the same BSUID identity.
6. Confirm a normal phone-number conversation still places the call using the existing flow.

### Fixes

Fixes https://linear.app/chatwoot/issue/CW-7807


sony-mathew (commented):
please take a look at these fidnings:

1. **P2 — Contact-level phone calls are attached to the wrong identity’s conversation** *(new finding; pre-existing on base but explicitly in CW-7807 scope)*

   When a contact has a phone number and a recent BSUID-backed conversation, the Contacts-page call action dials the phone correctly but records the call in the BSUID conversation. [`set_context_from_contact`](/worktrees/d32c/chatwoot/enterprise/app/controllers/api/v1/accounts/whatsapp_calls_controller.rb:73) accepts the contact-wide conversation returned by [`CallConversationBuilder#contact_conversations`](/worktrees/d32c/chatwoot/enterprise/app/services/whatsapp/call_conversation_builder.rb:15), while [`call_recipient`](/worktrees/d32c/chatwoot/enterprise/app/controllers/api/v1/accounts/whatsapp_calls_controller.rb:144) independently chooses the contact phone.

   My request-spec probe confirmed:

   - Meta recipient: `15551234567`
   - Resulting `Call#conversation.contact_inbox.source_id`: `IN.2081978709342942`

   The call bubble, permission activity, and permission-granted broadcast therefore land in identity B’s conversation while identity A was dialed. This violates [CW-7807](https://linear.app/chatwoot/issue/CW-7807/support-bsuid-addressing-in-whatsapp-calling-flows): identities A and B on a merged contact must not route A’s call through B’s conversation.

   The contact-level path should first resolve the phone `ContactInbox`, then reuse or create a conversation only for that identity.

2. **P2 — Legacy phone throttling still suppresses an unrelated BSUID request** *(matches the current open review thread; confirmed)*

   [`permission_request_attrs`](/worktrees/d32c/chatwoot/enterprise/app/services/whatsapp/call_permission_request_service.rb:73) falls back to the legacy top-level permission state for every recipient whenever the new nested map is empty. Before this PR, that state always represented `conversation.contact.phone_number`.

   Immediately after deployment, a recent legacy phone request therefore makes a conversation-scoped BSUID call return `permission_pending` without sending anything to the BSUID. My direct probe produced:

   ```json
   {"bsuid_status":"permission_pending","phone_status":"permission_pending"}
   ```

   The same transition can also lose a late legacy phone reply’s WAMID when the first new BSUID request overwrites the top-level fields. The fix should associate or migrate legacy state specifically to the phone recipient while preserving its reply context—not merely disable the fallback for BSUIDs.

enterprise/app/services/whatsapp/call_conversation_builder.rb
```diff
@@ -26,7 +28,7 @@ def perform!
     contact_inbox = ContactInboxBuilder.new(contact: contact, inbox: inbox).perform
 
     contact_inbox.with_lock do
-      existing_conversation || new_conversation.tap { |conversation| conversation.update!(contact_inbox: contact_inbox) }
+      existing_conversation || new_conversation.tap { |conversation| conversation.update!(contact_inbox: contact_inbox) }.reload
```
tds-1: @muhsin-k This runs before the with_lock transaction commits, so it nils @mutations_before_last_save. When the commit callbacks fire, AssignmentHandler's after_commit :notify_assignment_change, :process_assignment_changes both see saved_change_to_assignee_id? == false — so every thread opened by an outbound call loses the assignee.changed dispatch (no ActionCable broadcast, other agents' lists don't update), the "Assigned to <agent>" activity message, the conversation_participants row, and the assignment notification. Verified through POST /whatsapp_calls/initiate: the new conversation comes back with conversation_participants == [] and no activity messages.

Not redundant though — dropping it 500s the keeps the contact-level phone recipient… test with Locking a record with unpersisted changes is not supported. Changed attributes: "display_id", since load_attributes_created_by_db_triggers is an after_create_commit that leaves the record dirty and CallPermissionRequestService then calls conversation.with_lock. The reload is genuinely fixing that, just early enough to take the change set with it.

Moving it past the block keeps both:

conversation = contact_inbox.with_lock do
  existing_conversation || new_conversation.tap { |conversation| conversation.update!(contact_inbox: contact_inbox) }
end

- display_id/uuid come from a DB trigger via an after_create_commit that leaves the record dirty,
- which makes a later with_lock raise. Discard that once the commit callbacks have run.
conversation.tap { |record| record.reload if record.changed? }

muhsin-k: Fixed in 1eafdbe283. Moved the reload outside contact_inbox.with_lock and made it conditional on changed?, so assignment commit callbacks run before clearing the trigger-populated attributes. This also preserves the clean record needed by the permission-request lock. Validated with the existing WhatsApp calls controller and call permission request service specs: 27 examples, 0 failures. RuboCop passed.

tds-1 (approved):

PR #15570, opened 2026-08-24
perf: avoid per-membership queries in user auth payload

vishnu-narayanan:
# Pull Request Template

## Description

The user payload rendered by `api/v1/models/_user.json.jbuilder` (shared by sign-in, token validation, and profile endpoints) iterates the user's account memberships and previously issued one `accounts` lookup per membership, one `custom_role` lookup per membership on enterprise, and re-ran the unmemoized `active_account_user` ordering query three times per render. For users belonging to many accounts this makes the auth response cost linear in membership count and can push `/auth/sign_in` and `/auth/validate_token` past the request timeout.

This change preloads the memberships once (`includes(:account)`, plus `:custom_role` on enterprise) and derives the active account in memory from the preloaded rows, keeping full timestamp precision and `NULLS LAST` semantics.

Measured by counting SQL queries per request at increasing membership counts:

| Memberships | `GET /auth/validate_token` on develop | with this change |
|---|---|---|
| 1 | 10 | 9 |
| 10 | 19 | 9 |
| 100 | 109 | 9 |
| 500 | 509 | 9 |

`POST /auth/sign_in` drops from 2016 to 1516 queries at 500 memberships; its remaining growth comes from the per-account audit writes addressed in #15572. With both changes applied, sign-in settles at 19 queries and token validation at 9, independent of membership count.

Fixes https://linear.app/chatwoot/issue/INF-103

## Type of change

## How Has This Been Tested?

- Tested locally with users in 1, 10, 100, and 500 accounts; the query counts above are from those runs.
- New request spec asserts the SQL query count for `/auth/validate_token` does not grow with the number of account memberships (fails on develop, passes with this change).
- Existing specs covering all render sites of the partial (devise sessions/token validations, profiles, enterprise profiles) pass locally.

## Checklist:

### DIFF
--- a/app/views/api/v1/models/_user.json.jbuilder
+++ b/app/views/api/v1/models/_user.json.jbuilder
@@ -1,7 +1,8 @@
 # loaded once here; per-row lookups below would otherwise cost a query per membership
 account_users = resource.account_users.includes(:account)
 account_users = account_users.includes(:custom_role) if ChatwootApp.enterprise?
-active_account_user = account_users.max_by { |account_user| account_user.active_at&.to_i || 0 }
+# full-precision timestamps, nils last, matching ORDER BY active_at DESC NULLS LAST
+active_account_user = account_users.max_by { |account_user| [account_user.active_at ? 1 : 0, account_user.active_at || Time.at(0).utc] }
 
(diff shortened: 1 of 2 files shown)


sony-mathew (approved):
lgtm

PR #15574, opened 2026-08-24
fix(conversations): clear unread after staff reply

nawal-hilal:
Conversations no longer stay marked unread after staff have replied. The list now hides the unread badge when the last message is a public outgoing reply, or when the customer is no longer waiting.

Closes the inbox mismatch where replied threads still showed 9+ / unread counts.

## How to test
- Open All Conversations on an account with replied WhatsApp/email threads.
- Threads whose last message is a staff reply should have no unread badge.
- A new customer message after that reply should show unread again.

## What changed
- Backend unread count ignores incoming messages that already have a later public staff reply, and treats `waiting_since` as cleared after a reply.
- Dashboard cards use that answered state instead of the stale stored unread count.


nawal-hilal:
Opened against the wrong repository; this change belongs on the Samersadek1/chatwoot fork.

PR #15575, opened 2026-08-24
fix(instagram): restore business login authorization

muhsin-k:
Meta deprecated part of the existing Instagram Business Login flow, which could prevent customers from reaching the login page even when their app configuration was correct. This hotfix updates the authorization flow to use Metas current endpoint and reauthentication parameter.

The callback URL, requested permissions, and token exchange remain unchanged.

### How to reproduce

1. Configure an Instagram app with valid credentials and callback URL.
2. Start the Instagram inbox connection flow.
3. Observe that the previous authorization flow can show a page unavailable error before login.

### How to test

1. As an administrator, start connecting a new Instagram inbox.
2. Confirm the browser opens the current Instagram Business Login page.
3. Complete authorization and confirm the callback returns to Chatwoot successfully.

### Things to know

This is limited to the Instagram authorization endpoint and reauthentication parameter.


sony-mathew (approved):
lgtm

PR #15599, opened 2026-08-27
chore: remove Captain V2 rollout eligibility marker

aakashb95:
All eligible paid plans now enable Captain V2 during plan reconciliation. The Hacker or configured default plan still disables V2. Captain V1 runtime and the captain_integration_v2 feature flag remain in place.

## Closes

No linked issue.

## How to test

1. Reconcile an existing or new cloud account on a Startups, Business, or Enterprise plan.
2. Confirm that captain_integration and captain_integration_v2 are enabled.
3. Reconcile an account on the Hacker plan.
4. Confirm that captain_integration_v2 is disabled.

## What changed

1. Removed the transitional captain_v2_default_eligible account attribute and its new account write.
2. Enabled captain_integration_v2 for every non-default paid plan without checking account attributes.
3. Kept the default plan and self-hosted Enterprise behavior unchanged.
4. Updated the focused account and billing reconciliation specs.


sony-mathew (approved):
maybe a non-blocking followup:

1. **P2 — New: the sole behavioral change has no regression coverage.**

   Before this PR, paid accounts with an absent or `true` marker already received V2. The only changed state is a legacy account with `captain_v2_default_eligible: false`, which [reconciliation now ignores](/worktrees/pr-15599/chatwoot/enterprise/app/services/enterprise/billing/reconcile_plan_features_service.rb:43).

   The PR deletes the explicit-false example and replaces it with [existing/new account examples](/worktrees/pr-15599/chatwoot/spec/enterprise/services/enterprise/billing/handle_stripe_event_service_spec.rb:225) that exercise behavior already present on the base. Consequently, the tests do not prove that the former hold is intentionally overridden.

   Add a legacy account containing the literal stale key with `false`, reconcile it to Startups, and assert V2 becomes enabled. This should fail on the base and pass on the PR head.

aakashb95:
> Add a legacy account containing the literal stale key with false, reconcile it to Startups, and assert V2 becomes enabled. This should fail on the base and pass on the PR head.

I checked in prod, there are no accounts with this key enabled.

PR #15605, opened 2026-08-27
feat(slack): add alerts only mode for the Slack integration

pranavrajs:
Earlier, every reply posted there was delivered to the customer. Teams that use the channel for alerts had to remember a `note:` prefix on every message, and a single slip reached the customer.

The integration now has a thread behavior setting. 2-way sync keeps the current behavior and stays the default, so existing installs are untouched. Alerts only makes the integration one way: conversations are still pushed to Slack, but nothing posted in the thread is synced back, leaving the thread free for internal discussion.


sojan-official (approved):
code LGTM

PR #15606, opened 2026-08-28
fix(conversation): always show TO field in email reply composer

### LINKED ISSUE #15588: Email reply drops TO field when last message was sent without TO address
### Describe the bug

When an employee (or automatic integration) replies to a conversation with an empty TO. Any subsequent replies don't have a TO field. The TO field is also removed from the GUI, so the conversation is in some kind of dead lock. 

### To Reproduce

1. Send an email to the inbox
2. Open the conversation, remove the TO address, and reply (although this will not be sent, this is not the bug report).
3. The reply box now never shows the TO field anymore

### Expected behavior

The TO field is always shown (empty or not)
or
The TO field is always populated with the contact's email address.

### Environment

Linux VM

### Cloud Provider

None

### Platform

Browser

### Operating system

Ubuntu 24.04

### Browser and version

Chrome 150.0.7871.187 

### Docker (if applicable)

_No response_

### Additional context

_No response_

ayam04:
When an agent or integration replies to an email conversation with an empty TO address, the TO field disappears from the reply composer permanently. `ReplyEmailHead.vue` only rendered the TO row when `toEmails` was truthy, so after `setCCAndToEmailsFromLastChat` left it empty there was no way to re-add a recipient — every subsequent reply went out without a TO.

This renders the TO row unconditionally, matching the always-visible CC row. Empty values already pass `validEmailsByComma`, so no other behavior changes.

Closes #15588

### DIFF
--- a/app/javascript/dashboard/components/widgets/conversation/ReplyEmailHead.vue
+++ b/app/javascript/dashboard/components/widgets/conversation/ReplyEmailHead.vue
@@ -88,7 +88,7 @@ export default {
 
 <template>
   <div>
-    <div v-if="toEmails">
+    <div>
       <div class="input-group small" :class="{ error: v$.toEmailsVal.$error }">
         <label class="input-group-label">
           {{ $t('CONVERSATION.REPLYBOX.EMAIL_HEAD.TO') }}


ayam04:
Hi @iamsivin — gentle nudge on this one (open since Aug 28, CI green): the email reply composer now always renders the TO row, so a reply no longer hides the recipient field when the previous message had an empty TO. It is a one-line template change and the frontend suite passes. Could you take a look?

PR #15621, opened 2026-08-30
feat(voice): per-inbox call recording and transcription settings

tds-1:
Admins can now decide, per voice inbox, whether calls are recorded and whether recordings are transcribed. Both settings live on the inbox's Voice (Twilio) and Calls (WhatsApp Cloud) configuration pages: **Record calls** switches recording off entirely for teams operating where call recording isn't permitted, and **Transcribe recordings** — shown only while recording is on — controls whether Captain turns each recording into a transcript. Both default to on, so existing inboxes keep their current behaviour.

Supersedes #15598, which is closed. Same feature, rebased on current `develop` and reduced to a single reviewable commit.

## Closes

- https://linear.app/chatwoot/issue/CW-8001

## Screenshots

**Twilio voice inbox → Voice**

Twilio voice recording settings

**WhatsApp Cloud inbox → Calls**

WhatsApp calls recording settings

## How to test

1. Open a Twilio voice inbox → Settings → **Voice**, or a WhatsApp Cloud inbox with calling enabled → **Calls**. Two new toggles appear below "Allow incoming calls": Record calls and Transcribe recordings.
2. Turn **Record calls** off. The Transcribe toggle disappears and the setting saves immediately.
   - Twilio: place or receive a call — no recording appears on the call bubble or the Calls page afterwards.
   - WhatsApp: accept or place a call — the browser never starts recording, and no recording is attached when the call ends.
3. Turn **Record calls** back on and turn **Transcribe recordings** off. Complete a call: the recording is attached but no transcript is generated (including when Captain reads the conversation). An ordinary voice note on the same inbox still transcribes.
4. Turn both back on (the defaults) — recording and transcription behave exactly as before this change.
5. As an agent (non-admin), the toggles are read-only: the API rejects changes with 401.

## What changed

**One decision point per setting.** Recording is decided once, when the `Call` row is created: `before_create` snapshots the inbox's flag into `meta`, and every consumer — Twilio's per-leg conference TwiML, the ring broadcast, `/initiate`, `GET /whatsapp_calls/:id` — reads `call.recording_enabled?`. Because Twilio asks for TwiML separately for each leg and the agent's browser reads a payload of its own, a live lookup could hand the same call two different answers; the snapshot means it can't, and flipping the toggle mid-call never changes a call already under way.

Transcription is deliberately *not* snapshotted: it is read when a transcript is actually produced. `Messages::AudioTranscriptionService` is re-entrant (Captain calls it every time it reads a conversation), so a snapshot would permanently block transcribing a recording an admin later decides they want transcribed.

- `Concerns::CallRecordingSettings` mixed into `Enterprise::Channel::TwilioSms` and `Enterprise::Channel::Whatsapp`: `recording_enabled?` / `transcription_enabled?` read from `provider_config` and default on (only an explicit `false` disables). Kept in the EE overlay since calling is Enterprise-only; exposed on the inbox payload.
- New admin-only endpoint `POST /api/v1/accounts/:id/inboxes/:id/set_call_recording`, mirroring `set_inbound_calls` — including the `validate: false` save so WhatsApp's remote credential re-check can't reject a toggle. It posts both flags together, so neither can clobber the other.
- Twilio: `<Conference record>` switches between `record-from-start` and `do-not-record`. With `do-not-record` Twilio produces no recording at all, so no status callback fires and there is nothing to download, attach, or discard.
- WhatsApp: the agent's browser is what records, so the decision rides the same three payloads `sdp_offer`/`ice_servers` already ride — the `voice_call.incoming` ring broadcast, the `/initiate` response, and `GET /whatsapp_calls/:id` (the fallback the accept path already uses when the ring broadcast was missed). When recording is off the `MediaRecorder` is never created, so no audio is captured and nothing is uploaded.
- Transcription is gated in `Voice::CallTranscriptionService` (Twilio's automatic path) and in `Messages::AudioTranscriptionService` (the WhatsApp recording attachment path, which Captain also calls directly). The second is scoped to `voice_call` messages, so ordinary voice notes are unaffected.
- Shared `CallRecordingSettings.vue` used by both settings pages, following the existing inbound-calls toggle: optimistic update, rollback on failure.
- `WhatsappCallsController#upload_recording` refuses audio for a call whose snapshot says no. The browser gate ships in JS, so a tab loaded before the setting changed would otherwise still be able to store a recording — the client gate avoids *capturing* audio, this one enforces the policy.

## Verified against a running instance

Beyond the specs, this was exercised end to end on a local server:

- Both toggles render on the Twilio **Voice** and WhatsApp **Calls** pages, default to on for an inbox whose `provider_config` is `{}`, and survive a reload. Turning Record calls off hides the Transcribe toggle.
- Toggling writes both flags together (`{"recording_enabled" => true, "transcription_enabled" => false}`) — neither clobbers the other.
- The real Twilio webhook returns `record="do-not-record"` with recording off and `record="record-from-start"` with it on. **Snapshot confirmed:** a call created while recording was off keeps `do-not-record` even after the inbox is switched back on, while a call created afterwards gets `record-from-start`.
- `Messages::AudioTranscriptionService` returns the disabled error for a `voice_call` message when transcription is off, and does *not* block an ordinary voice note on the same inbox, nor a `voice_call` message when transcription is on.
- Endpoint authorization: agent → 401, admin → 200, non-voice inbox → 422 `Inbox does not support calling`.
- `upload_recording` → 422 `Call recording is disabled for this inbox` for a call snapshotted off, 200 `uploaded` for one snapshotted on.


muhsin-k (commented):
Please check the following comments

P1 — Reject Twilio recordings for calls with recording disabled
Call#before_create snapshots recording_enabled: false, but the Twilio callback and attachment pipeline never checks that value. A completed recording callback still downloads and attaches the audio, then enqueues Voice::CallTranscriptionJob.
I reproduced this at 189e390: a call with recording disabled still enqueued the transcription job. Please guard Voice::Provider::Twilio::RecordingAttachmentService before SafeFetch and add a regression spec confirming that no download, attachment, or transcription job occurs.

P3 — Remove the unrelated scripts/ ignore
The new scripts/ entry is unrelated to the call-recording feature and broadly hides future top-level scripts from Git. Please remove it from this PR.

iamsivin (approved):
Code LGTM

PR #15655, opened 2026-09-02
fix: improve Google OAuth signup error feedback

scmmishra:
Improves the Google OAuth signup error shown when someone uses a personal email address. The message now explains that a work email is required and stays visible for six seconds.


iamsivin (approved):
Lgtm

Issue #15253, opened 2026-07-30
Make allowed attachment file types configurable


kme84:
### Is your feature or enhancement related to a problem? Please describe.

Currently, the list of file types that users can attach in conversations is hardcoded in the application. This makes it impossible for administrators to allow or restrict specific file types without modifying the source code and maintaining custom patches.

This limitation reduces flexibility for organizations with different security policies or business requirements. For example, some deployments may need to allow additional document or archive formats, while others may want to restrict the default list.

### Describe the solution you'd like

Introduce a configuration option that allows administrators to define the list of permitted attachment file types.

The configuration could be exposed through an environment variable or an application setting, for example by specifying allowed MIME types or file extensions.

This would allow deployments to:

customize the list of allowed file types without modifying the code;
adapt the application to their security policies;
simplify upgrades by eliminating the need for custom patches.

The default behavior should remain unchanged if no custom configuration is provided.

### Describe alternatives you've considered

The current workaround is to modify the hardcoded list of allowed file types in the source code. However, this approach requires maintaining a custom fork or patch and reapplying the changes after every upgrade, which increases maintenance effort.

### Additional context

Making this configurable would improve flexibility for self-hosted installations while preserving the current default behavior for existing users.

PR #15656, opened 2026-09-02
feat(attachments): make file type validation toggleable in widget

drnextgis:
## Description

Add a new `ENABLE_FILE_TYPE_VALIDATION` installation setting that lets administrators completely disable the website-widget attachment file-type validation. When disabled, any file type is accepted in both the widget's file picker (accept attribute) and backend validation, when enabled (default) the hardcoded allow-list behaviour is unchanged.

This partially addresses #15253. Where that issue asks for the ability to customize the allowed file-type list, this PR takes the simpler complement: a boolean toggle to disable validation entirely when an organization needs to accept any file type. The default behavior and the existing hardcoded `ACCEPTABLE_FILE_TYPES` allow-list are preserved, so existing users see no change out of the box. The configurable-list direction remains open as follow-up work.

Additionally fixes the classification of direct-uploaded widget attachments: when `DIRECT_UPLOADS_ENABLED` is on, the widget sends a signed blob ID instead of an uploaded file, which the messages controller previously skipped when assigning `file_type` - leaving non-image files stuck as the default `:image` and shown as broken images. Now resolved from the signed blob ID, mirroring `MessageBuilder#attachment_file_type`.

## Type of change

## How Has This Been Tested?

- Added a new block in `attachment_spec.rb` covering content-type validation.

## Checklist:


drnextgis:
Hi @iamsivin, could you please take a look at this when you get a chance? Since you worked on the widget file-type handling before, your review would be really appreciated. Thanks!

PR #15660, opened 2026-09-03
fix: sync embedded app context

scmmishra:
Dashboard apps now receive custom attribute definitions and the resolved Chatwoot theme in their context. The context stays synchronized as attributes and theme change, and refresh requests are routed to the requesting iframe.


iamsivin (approved):
Code LGTM

PR #15663, opened 2026-09-03
feat(email): park bounces, auto-replies and notification senders in inbound triage

YJack0000:
Inbound email triage now also silences the three biggest remaining sources of inbox noise: delivery bounces, out-of-office / auto-generated replies, and mail from no-reply notification senders. Like newsletters, these conversations are still created and searchable, but arrive resolved, labelled by type (`bounce`, `auto-reply`, `notification`), and never notify agents or trigger automations.

## How to test

1. On an email inbox, keep **Newsletter and automated email filtering** enabled (Settings → Inbox → Configuration; it is on by default).
2. Send the inbox a mail with an `X-Failed-Recipients` header (or a real bounce) → conversation arrives resolved with the `bounce` label and no agent notification.
3. Send a mail with `Auto-Submitted: auto-generated` or `X-Auto-Response-Suppress: OOF, AutoReply` → resolved with `auto-reply`.
4. Send from `no-reply@…` / `donotreply@…` / `mailer-daemon@…` / `postmaster@…` → resolved with `notification`.
5. Add the sender to the inbox's **allowed** or **vip** sender list and resend → the mail lands as a normal open conversation.
6. Turn the inbox toggle off → all of the above arrive as normal open conversations.

## What changed

- `Email::SenderTriageService` gains three lanes ahead of the newsletter check (bounce first, since bounces often carry the original message's list/auto-reply headers). All lanes share the existing per-inbox `newsletter_filter_enabled` toggle and respect allowed/vip bypass — no new migration.
- `MailPresenter#auto_reply?` now also recognises Exchange/Outlook's `X-Auto-Response-Suppress` header, and the previously unused `bounced?` finally has a caller.
- Notification senders are matched on the local part only (`no-reply`/`do-not-reply` variants incl. prefixed forms like `comments-noreply@`, exact `mailer-daemon`/`postmaster`), so VERP-style `bounce+token@` system senders are untouched.
- New triage labels registered on `Conversation::SENDER_TRIAGE_LABELS`; inbox settings copy updated to describe the wider scope (en only, per Crowdin policy).


YJack0000:
Apologies — opened against upstream by mistake; this belongs on our fork.

PR #15666, opened 2026-09-03
feat(billing): cancel subscriptions at period end and show the cancellation date

tds-1:
The billing page now shows when a subscription is set to end. After a customer cancels from the Stripe portal, the plan card shows "Cancels on <date>" instead of "Renews on", and goes back to "Renews on" if they resume.

## Closes

- [CW-8034](https://linear.app/chatwoot/issue/CW-8034/mark-the-subscription-as-cancelled-at-the-end-of-the-period)

## How to test

1. On a cloud account with a paid plan, cancel the subscription from the billing portal.
2. Settings → Billing shows "Cancels on <date>", and the plan stays active until that date.
3. Resume the subscription, or change the seat count, in the portal — the card goes back to "Renews on".

## What changed

- `subscription_cancels_on` is read from Stripe's `cancel_at` (falling back to the period end for older subscriptions that only set `cancel_at_period_end`), stored on the account and exposed in the account payload.
- Account deletion now schedules the cancellation with an explicit `cancel_at` timestamp. `cancel_at_period_end` is deprecated and is not used by Stripe's flexible billing mode, where it stays `false` even while a cancellation is scheduled.

### DIFF
--- a/app/helpers/billing_helper.rb
+++ b/app/helpers/billing_helper.rb
@@ -34,4 +34,12 @@ module BillingHelper
 
     Time.zone.at(period_end)
   end
+
+  # Stripe schedules cancellations on cancel_at; older subscriptions only set cancel_at_period_end.
+  def subscription_cancels_on(subscription)
+    cancel_at = subscription['cancel_at']
+    return Time.zone.at(cancel_at) if cancel_at.present?
+
+    subscription_ends_on(subscription) if subscription['cancel_at_period_end']
+  end
 end
--- a/app/helpers/billing_helper.rb
+++ b/app/helpers/billing_helper.rb
@@ -24,8 +24,9 @@ module BillingHelper
   end
 
   # current_period_end moved to the subscription item in newer Stripe API versions; read both.
+  # Stripe objects don't respond to dig, so the items path is walked with [].
   def subscription_period_end(subscription)
-    subscription['current_period_end'] || subscription['items']['data'].first&.[]('current_period_end')
+    subscription['current_period_end'] || subscription['items']&.[]('data')&.first&.[]('current_period_end')
   end
 
   def subscription_ends_on(subscription)
--- a/app/javascript/dashboard/i18n/locale/en/settings.json
+++ b/app/javascript/dashboard/i18n/locale/en/settings.json
@@ -585,6 +585,7 @@
       "PLAN_NOTE": "You are currently subscribed to the **{plan}** plan with **{quantity}** licenses",
       "SEAT_COUNT": "Number of seats",
       "RENEWS_ON": "Renews on",
+      "CANCELS_ON": "Cancels on",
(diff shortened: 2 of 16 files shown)


sony-mathew (commented):
Please take a look at this finding:

- **P2 — Blocking — Existing QA concern, independently confirmed:** out-of-order Stripe events can invert the newly displayed cancellation state. [`HandleStripeEventService`](</worktrees/pr-15666/chatwoot/enterprise/app/services/enterprise/billing/handle_stripe_event_service.rb:71>) persists `subscription_cancels_on` directly from each webhook snapshot; [`Index.vue`](</worktrees/pr-15666/chatwoot/app/javascript/dashboard/routes/dashboard/settings/billing/Index.vue:85>) then treats that value as current truth.

  On the exact head, I reproduced both sequences:

  - Current Stripe state resumed → delayed older cancellation event → Chatwoot saved `2030-03-17T17:46:40.000Z`.
  - Current Stripe state cancelling → delayed older resume event → Chatwoot saved `nil`.

  Customers can therefore see “Cancels on” for a renewing subscription, or “Renews on” for one that will terminate. Reloading the billing page only reloads the stale database value. Stripe explicitly states that webhook delivery order is not guaranteed and that `event.created` cannot safely establish ordering; it recommends retrieving the current resource when needed. [Stripe webhook ordering documentation](https://docs.stripe.com/webhooks#event-ordering)

  This is rooted in pre-existing webhook architecture, but the incorrect cancellation display is introduced by this PR’s changed persistence/UI path and directly violates its acceptance criteria. The latest Ito report also raises it, although Ito classifies it as unrelated.

  The blocking fix should reconcile against the current Stripe subscription—preferably asynchronously—or another durable source of current state. A timestamp-only guard would not be sufficient. Add regression cases for both orderings above.

tds-1:
@sony-mathew Agreed the mechanism is real, nothing in the webhook path guards ordering, and we write the whole snapshot on every event.

But this PR just adds one line to a merge that was already unconditional. The same stale event today flips plan_name, subscription_status and subscription_ends_on, and those hurt more — a stale plan_name toggles ~28 features and rewrites Captain credits, while subscription_cancels_on only renders a label. It also sorts itself out on the next subscription.updated.

sony-mathew:
> @sony-mathew Agreed the mechanism is real, nothing in the webhook path guards ordering, and we write the whole snapshot on every event.
> 
> But this PR just adds one line to a merge that was already unconditional. The same stale event today flips plan_name, subscription_status and subscription_ends_on, and those hurt more — a stale plan_name toggles ~28 features and rewrites Captain credits, while subscription_cancels_on only renders a label. It also sorts itself out on the next subscription.updated.

As long as we are intentional about these changes, we are good.

sony-mathew (approved):
lgtm

Issue #15530, opened 2026-08-19
WhatsApp contacts identified only by a business-scoped user id are missing from the contacts list, search, filters and export


marcoazcabral:
### Describe the bug

A contact whose only identity on a WhatsApp Cloud inbox is a business-scoped user id (BSUID) exists, holds conversations and exchanges messages normally, but it is absent from the contacts list, from contact search, from contact filters and from the contacts export.

`Contact.resolved_contacts` decides which contacts are shown, and it recognises exactly three fields:

```ruby
# app/models/contact.rb
def self.resolved_contacts(use_crm_v2: false)
  return where(contact_type: 'lead') if use_crm_v2

  where("contacts.email <> '' OR contacts.phone_number <> '' OR contacts.identifier <> ''")
end
```

A contact created from a BSUID-only payload has none of the three. Its identity lives in `contact_inboxes.source_id`, which this predicate does not consider. The contact is therefore filtered out of every surface that goes through the scope:

* `Api::V1::Accounts::ContactsController#index`
* the same controller's `search`
* `Contacts::FilterService`
* `Account::ContactsExportJob`

The contact is still reachable by direct id, and its conversations appear normally in the inbox, so the data is intact. It is only unreachable from the contacts side of the dashboard.

### To Reproduce

Observed on a self-hosted instance running a WhatsApp Cloud inbox, on code where `resolved_contacts` is unmodified.

1. Have a contact whose only `contact_inbox.source_id` on the WhatsApp Cloud inbox is a BSUID, for example `BR.1234567890123456789`, with `phone_number`, `email` and `identifier` all empty. This is what the coexistence payload produces when Meta sends `contacts[].user_id` without `wa_id`.
2. `GET /api/v1/accounts/{account_id}/contacts/{contact_id}` returns the contact, with the BSUID under `contact_inboxes`.
3. `GET /api/v1/accounts/{account_id}/contacts` does not include it.
4. `GET /api/v1/accounts/{account_id}/contacts/search?q={contact name}` returns an empty payload.

In the instance where this was reproduced, a direct fetch by id returned the contact with its conversation intact, while the same contact was absent from both the list and the search results.

### Expected behavior

A contact that has a usable identity on a channel should be visible to the agent who is already talking to it. Today an agent can answer that person in the inbox and then fail to find the same person in Contacts, which also means the contact cannot be labelled, filtered, segmented into a campaign audience or exported.

### Environment

Docker

### Why this has probably not been reported yet

[...]

PR #15670, opened 2026-09-03
fix(whatsapp): classify BSUID contacts as leads

muhsin-k:
WhatsApp customers who only expose a business-scoped user ID are reachable in the inbox, but they should also be classified as CRM leads once CRM v2 is enabled. This change promotes visitor contacts to leads when WhatsApp identity sync sees a BSUID, so future CRM v2 contact lists treat them like real reachable channel contacts.

The change keeps BSUIDs in ContactInbox aliases and does not write them into contacts.identifier.

### Things to know

- This is scoped to WhatsApp BSUID identities, including Meta Cloud IDs and Twilio's whatsapp:<BSUID> shape.
- This does not change the legacy resolved_contacts predicate when crm_v2 is disabled.
- Existing BSUID-only contacts may still need a backfill before enabling CRM v2 broadly.
- Plain phone source IDs and anonymous widget visitors are unaffected.

### How to test

1. Receive a WhatsApp message where the contact identity is a BSUID and no phone number is available.
2. Confirm the contact remains attached to the WhatsApp ContactInbox identity.
3. Confirm the contact is classified as a lead.
4. Enable CRM v2 for the account and confirm the contact appears in the Contacts list.
5. Repeat with a Twilio WhatsApp BSUID payload and confirm the same lead classification.
6. Confirm a phone-only WhatsApp source ID does not get promoted by this BSUID-specific rule.

### What changed

- Promotes visitor contacts to leads when WhatsApp identifier sync sees a BSUID source ID.
- Supports both native Cloud BSUIDs and Twilio-prefixed WhatsApp BSUIDs.
- Adds focused service coverage for CRM v2 visibility and non-BSUID behavior.

Related: https://linear.app/chatwoot/issue/CW-7982/whatsapp-contacts-identified-only-by-a-business-scoped-user-id-are
Related: https://github.com/chatwoot/chatwoot/issues/15530

### DIFF
--- a/app/services/whatsapp/identifier_sync_service.rb
+++ b/app/services/whatsapp/identifier_sync_service.rb
@@ -3,7 +3,7 @@ class Whatsapp::IdentifierSyncService
 
   def perform(source_ids: [], username: nil, phone_number: nil)
     create_contact_inboxes(source_ids)
-    update_contact(username, phone_number)
+    update_contact(source_ids, username, phone_number)
   end
 
   private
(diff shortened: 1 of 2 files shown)


sony-mathew (approved):
lgtm

PR #15672, opened 2026-09-03
feat(integrations): MUToday FAQ First-Reply — foundation (T1–T8)

mennwebs:
Lays the foundation for the MUToday FAQ First-Reply feature: an AI-labelled acknowledgment on the first LINE message of a new conversation, routing to a human for everything else. **Nothing user-facing yet** — the integration can be connected from Settings → Integrations, but the job it triggers only logs a line and returns. No reply is sent to any customer on this branch.

Spec and implementation plan: [docs/mutoday-faq-first-reply-spec.md](docs/mutoday-faq-first-reply-spec.md), [docs/mutoday-faq-first-reply-plan.md](docs/mutoday-faq-first-reply-plan.md) (merged separately, already on `develop`). This PR covers tasks T1–T8 of the plan's 21-task list; T9 (the first shippable increment, where a reply actually gets composed and sent) is next.

## What changed

- **Integration registration** — `mutoday_faq_reply` app in `apps.yml` + `en.yml` + icons. Inbox-scoped, single-hook, Connect form asks for an API token and a Shadow/Live/Off mode.
- **Upstream wiring** — one line in `HookListener`, one processor entry + method in `HookJob`, and the job shell itself (`MutodayFaqReplyJob`), with its full retry policy: fails closed on `Redis::BaseError`, stops (not retries) on `ActiveRecord::RecordInvalid`, everything else bubbles to Sidekiq's normal retries.
- **`Copy`** — the frozen Thai reply constants and the two AI-disclosure labels, chosen by route (model vs. not) rather than by content, so the label is never something the model could omit or the admin could blank.
- **`DenyList`** — six hard-stop topics (money, legal, complaint, crisis, human, credential) that route straight to a human handoff with no model call, with mechanical guardrails so a deny term can never swallow a real corpus question.
- **`Eligibility`** — the full guard chain (loop protection, sender/channel checks, kill switch, "never talk over a human," the greeting-deferral so a bare "สวัสดีครับ" doesn't burn the conversation's one reply, and the atomic once-per-conversation Redis claim), all failing closed if Redis is unreachable.
- **`RateLimiter`** — three-state Redis reservation (`ok` / `limited` / `contended`) so ordinary concurrency on a shared counter is never mistaken for a breach.
- **`Telemetry`** — one structured log line per job with a closed field set that cannot carry customer text, plus deduped Sentry alerting.

## Corrections made along the way

A few places where the merged spec didn't hold up under an actual test run, fixed in the commits that touch them (see individual commit messages for details):

- The corpus id pattern's `\A`/`\z` anchors are inert under `json_schemer`'s ECMA regexp resolver — switched to `^...$`. The same bug pre-exists in the Lark integration's `webhook_url` pattern (unrelated to this feature, flagged separately).
- `retry_on ActiveRecord::RecordInvalid, attempts: 1` needs a block to actually stop retrying — without one it re-raises into Sidekiq's normal 3 retries.
- Capping the whole assembled reply at 1200 chars (rather than just the answer) could silently drop the handoff line off a long approved answer; truncation now applies to the answer only.
- Two deny-list terms were still shadowing real corpus questions after the spec's own fix (`ต่อเจ้าหน้าที่` still matched `ติดต่อเจ้าหน้าที่`; `ขอเงินคืน` was dead weight under `เงินคืน`) — caught by a mechanical "no term may swallow an approved question" spec rather than by eye.

## How to test

Nothing is live yet. On a local dev server: Settings → Integrations → search "MUToday" → the card renders with real copy (no `translation missing`) → Connect opens a form for an API token, inbox picker and mode (defaults to Shadow) → connecting against a `Channel::Line` inbox creates the hook and leaves `inbox.active_bot?` false. Sending a LINE message afterward produces exactly one `[mutoday_faq_reply] outcome=noop` log line and no reply.


mennwebs:
Opened in error — this branch and its target repository are MUTodayTeam/chatwoot, not upstream. Closing immediately; apologies for the noise.

Issue #15625, opened 2026-08-30
Allow external integrations to persist, display and reuse processed attachment text


jonasgpayer:
### Is your feature or enhancement related to a problem? Please describe.

Attachments often contain important conversational information that is not directly available as text.

For example:

- Audio → transcription
- Image → AI vision analysis or OCR
- PDF/document → extracted text
- Video → transcription or content analysis

External AI agents and automation platforms such as n8n can already process these attachments using OpenAI, Gemini, OCR services, document parsers, local models or custom AI pipelines.

However, after an external service processes an attachment, there is currently no supported Chatwoot API mechanism to persist that textual representation back into the original attachment.

This creates several problems:

- AI agents may need to process the same attachment again whenever conversation history is loaded.
- AI/API costs and processing latency are unnecessarily repeated.
- Integrations need an external database just to cache attachment analysis.
- The processed information becomes separated from the original Chatwoot message.
- Multiple AI agents or integrations cannot easily reuse the same processed result.
- Human agents cannot see the textual interpretation of an attachment that an AI agent is using as conversational context.

Chatwoot already solves this very well for audio transcriptions: the transcription is stored with the attachment, displayed directly below the audio message in the conversation UI, and can be reused as textual context.

It would be very useful to generalize this concept to images, documents and other attachment types, and allow external integrations to write this information through the API.

### Describe the solution you'd like

Provide a supported API mechanism that allows authenticated integrations to store processed textual content for an existing attachment.

The exact schema is up to the Chatwoot team, but conceptually it could look like:

{
  "processed_text": "Textual representation of the attachment",
  "processing_type": "vision"
}

Possible processing types could include:

- transcription
- vision
- ocr
- document_extraction
- video_analysis

The existing transcribed_text behavior for audio attachments could remain unchanged for backwards compatibility.

The processed content should then:

1. Be persisted with the original attachment.
2. Be returned through the Messages/Attachments API.
3. Be available in relevant webhook attachment payloads.
4. Be displayed directly alongside the attachment in the conversation UI, similarly to how audio transcriptions are currently displayed.
5. Be available as textual context for AI/LLM integrations.

For example, an image attachment could appear in the conversation as:

payment_receipt.jpg
[image preview]

AI analysis:
Payment receipt for ARS 52,000.
Date: August 30, 2026.
Reference: 123456.

A PDF could similarly show:

invoice.pdf

Extracted text:
Invoice A 0004-00001234
Total: ARS 128,450
[...]

PR #15676, opened 2026-09-04
feat(audio): allow audio transcription to use account-level OpenAI integration (BYOK)

### LINKED ISSUE #15624: Allow audio transcription to use the account-level OpenAI integration (BYOK)

Audio transcription currently depends on Captain. The Speech-to-Text implementation requires Captain, uses Captain's OpenAI credentials, and consumes Captain response credits.

However, Chatwoot already provides an account-level OpenAI integration under Settings → Integrations → OpenAI, where each account can configure its own OpenAI API key.

For multi-account and self-hosted installations, this creates an unnecessary limitation: individual accounts cannot use and pay for their own transcription usage, even when they already provide their own OpenAI credentials.

This is especially relevant in multi-tenant installations, where different Chatwoot accounts may belong to different customers. Ideally, each customer should be able to own their credentials, usage and billing instead of relying on a shared instance-level Captain/OpenAI configuration.

It also differs from the original audio transcription feature request (#10182), which specifically proposed using the OpenAI integration to transcribe audio. The feature was eventually implemented through Captain instead.

Allow audio transcription to use the OpenAI integration configured for the current Chatwoot account as an alternative to Captain.

The existing Captain behavior should remain available and unchanged.

For example, Chatwoot could provide an account-level setting such as:

Audio transcription provider:
- Captain
- Account OpenAI integration

When "Account OpenAI integration" is selected:

- The account's own OpenAI API key is used.
- Usage is billed directly to that account's OpenAI account.
- Captain response credits are not consumed.
- Credentials and costs remain isolated between Chatwoot accounts.
- The resulting transcription behaves exactly like the current transcription feature in the conversation UI.

Since the OpenAI integration is already configurable per account, this could also provide a BYOK (Bring Your Own Key) transcription option for Community Edition and self-hosted installations.

The goal is not to remove or replace Captain, but to allow the existing account-level OpenAI integration to be used for this task.

Audio attachments can currently be sent through webhooks to an external automation platform such as n8n, transcribed through OpenAI or another Speech-to-Text provider, and then processed externally.

This works, but duplicates functionality Chatwoot already has internally and requires additional workflows, API calls and infrastructure.

It also does not provide the same native experience as Chatwoot's built-in transcription, where the transcription is stored and displayed directly with the audio message.

### Additional context

The main use case is multi-account/self-hosted Chatwoot installations.

A typical setup could be:

Account A → OpenAI API key A → transcription usage billed to customer A
Account B → OpenAI API key B → transcription usage billed to customer B
Account C → Captain → existing Captain behavior

This provides better tenant isolation, credential ownership, API usage accounting and cost separation.

Related:
- #10182 — Original audio transcription feature request, which proposed using the OpenAI integration.
- #11670 — Audio transcription implementation.

AbhiPra24:
Fixes #15624

### Description
Allows audio transcription (for voice messages and voice call recordings) to use account-level OpenAI integration credentials (BYOK) when configured on the account, falling back to system-level `CAPTAIN_OPEN_AI_API_KEY` if not present.

### Changes
- Check account-level OpenAI integration hook credentials before falling back to system keys in `Llm::SpeechToTextService`.
- Update `Llm::SpeechToTextService.available_for?` so accounts with BYOK configured are not blocked by exhausted system response credits.
- Skip incrementing response usage for BYOK transcriptions.
- Allow custom `api_key` injection into `Llm::LegacyBaseOpenAiService`.
- Add comprehensive RSpec test coverage for account-level BYOK audio transcription.

### DIFF
--- a/enterprise/app/services/llm/legacy_base_open_ai_service.rb
+++ b/enterprise/app/services/llm/legacy_base_open_ai_service.rb
@@ -11,9 +11,10 @@ class Llm::LegacyBaseOpenAiService
 
   attr_reader :client, :model
 
-  def initialize
+  def initialize(api_key: nil)
+    api_key ||= InstallationConfig.find_by!(name: 'CAPTAIN_OPEN_AI_API_KEY').value
     @client = OpenAI::Client.new(
-      access_token: InstallationConfig.find_by!(name: 'CAPTAIN_OPEN_AI_API_KEY').value,
+      access_token: api_key,
       uri_base: uri_base,
       log_errors: Rails.env.development?
     )
--- a/enterprise/app/services/llm/speech_to_text_service.rb
+++ b/enterprise/app/services/llm/speech_to_text_service.rb
@@ -12,12 +12,18 @@ class Llm::SpeechToTextService < Llm::LegacyBaseOpenAiService
 
   attr_reader :blob, :account, :transcription_model
 
-  # Transcription runs on Captain's OpenAI credentials and consumes its response credits.
+  # Transcription runs on Captain's OpenAI credentials and consumes its response credits,
+  # or uses account-level OpenAI integration (BYOK).
   def self.available_for?(account)
     return false unless account.feature_enabled?('captain_integration')
     return false if account.audio_transcriptions.blank?
+    return true if account_openai_hook_configured?(account)
 
     account.usage_limits[:captain][:responses][:current_available].positive?
+  end
+
+  def self.account_openai_hook_configured?(account)
+    account&.hooks&.find_by(app_id: 'openai', status: 'enabled')&.settings&.dig('api_key').present?
   end
 
   def self.too_large?(blob)
@@ -25,10 +31,10 @@ class Llm::SpeechToTextService < Llm::LegacyBaseOpenAiService
   end
 
   def initialize(blob:, account:)
-    super()
     @blob = blob
     @account = account
     @transcription_model = Llm::FeatureRouter.resolve(feature: 'audio_transcription', account: account)[:model]
+    super(api_key: api_key)
   end
 
   def perform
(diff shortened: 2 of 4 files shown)


jonasgpayer:
@AbhiPra24 
Thanks for working on this — the account-level BYOK behavior looks great.

One thing I wanted to clarify: it looks like transcription still requires captain_integration to be enabled, even when the account has its own OpenAI integration configured.

One of the main goals of [#15624]( https://github.com/chatwoot/chatwoot/issues/15624) was to make account-level BYOK independent from the instance-level Captain/OpenAI setup: if an account provides and pays for its own OpenAI credentials, it should ideally be able to use transcription without depending on Captain.

This would also make the feature usable in self-hosted/Community installations that provide their own OpenAI credentials, but the main point is keeping the BYOK path independent from Captain.

Would it make sense for an account-level OpenAI integration to be enough to enable transcription, with Captain remaining as the fallback when no account-level OpenAI integration is configured?

AbhiPra24:
Hi @jonasgpayer,

Thanks for raising this point!

You are completely right regarding decoupling BYOK from Captain. In the initial revision of the PR, `SpeechToTextService.available_for?` kept the `account.feature_enabled?('captain_integration')` guard and only allowed bypassing the Captain quota/usage check when account-level OpenAI credentials existed. In addition, the frontend toggle for audio transcription in settings (`showAudioTranscriptionConfig`) was guarded behind `FEATURE_FLAGS.CAPTAIN`.

Making account-level OpenAI integration independently sufficient to enable transcription makes total sense:
1. **Backend Availability (`available_for?`)**: We can update the check so an account is eligible for transcription if `captain_integration` is enabled OR if the account has an active OpenAI integration hook configured (`account_openai_hook_configured?(account)`), while still requiring `account.audio_transcriptions.present?`. When falling back to Captain, the responses limit check continues to apply.
2. **Settings UI**: We can update `showAudioTranscriptionConfig` to display the Audio Transcription toggle if either Captain is enabled or an OpenAI integration hook is present for the account, allowing self-hosted/community users with BYOK to enable the feature without requiring the Captain enterprise flag.

Would you prefer expanding this within this PR so transcription can run entirely on account BYOK without requiring `captain_integration`? If so, I'd be happy to push the updates.

jonasgpayer:
Hi @AbhiPra24 
Yes, please — that would be exactly what I had in mind.

Thanks a lot for taking this into account and for updating the implementation.

jonasgpayer:
@AbhiPra24 
By the way, I also opened a related feature request: [#15625](https://github.com/chatwoot/chatwoot/issues/15625) allowing external integrations to persist processed attachment text (audio transcriptions, image analysis, PDF text, etc.) back into Chatwoot via API, display it alongside the attachment, and reuse it as AI context.

It touches a similar area around attachment processing, so I thought I'd share it in case you find it interesting.

PR #15692, opened 2026-09-05
fix(mail,reports): support multi-delimiter email lists and prepend UTF-8 BOM to CSV downloads

### LINKED ISSUE #15609: Recipients separated by spaces or semicolons are silently dropped and the email fails to send
### Describe the bug

`Messages::MessageBuilder#process_email_string` strips all whitespace from the recipient string and then splits on commas only:

```ruby
# app/builders/messages/message_builder.rb:102-106 (v4.17.1)
def process_email_string(email_string)
  return [] if email_string.blank?

  email_string.gsub(/\s+/, '').split(',')
end
```

Any separator other than a comma fuses adjacent addresses into a single unusable token. Both addresses are lost and the agent gets no warning — the message simply never arrives.

This matters more than it might look because **a semicolon is the default separator in Outlook**, so any recipient list copied out of Outlook and pasted into CC hits it every time.

**Impact:** silent data loss on the send path. From the agent's point of view a failed send is indistinguishable from a successful one, so a customer or supplier communication just never happens.

### To Reproduce

1. Open an email conversation.
2. In the CC field enter two valid addresses separated by a space: `a@example.com b@example.com`
3. Send.

**Result:** recipients parse to `["a@example.comb@example.com"]` — one invalid address. Neither recipient receives the message. No error is surfaced.

Identical behaviour with a semicolon: `a@example.com;b@example.com`

Comma-plus-space works, which is why this is easy to miss in testing.

### Expected behavior

Commas, semicolons and whitespace should all be accepted as separators, as in every mainstream mail client. Genuinely malformed addresses should still be rejected rather than silently passed through.

Note that naive whitespace splitting would break display-name forms such as `Jane Smith <jane@example.com>, John Doe <john@example.com>`, so whitespace inside angle-bracket forms has to be preserved.

### Environment

Self-hosted, Docker Compose, image `chatwoot/chatwoot:v4.13.0`.

### Platform

Web. Not browser-specific — the fault is server-side in Ruby.

### Additional context

Reproduced on v4.13.0; the code is **unchanged as of v4.17.1**.

We are running a fix in production and have a 13-case test script covering comma, semicolon, whitespace, display-name and mixed forms — **including a case that must still be rejected**, so a pass means it works rather than that it merely stopped complaining. Happy to open a PR if that would be useful.

### LINKED ISSUE #15684: CSV report downloads (reports, CSAT, SLA) lack UTF-8 BOM, non-ASCII names garbled in Excel
### Describe the bug

CSV files downloaded from the reports pages (Agents, Inboxes, Labels, Teams, Conversations summary, Conversation traffic), the CSAT responses download and the SLA "breached conversations" download do not start with a UTF-8 byte order mark. Microsoft Excel does not auto-detect UTF-8 for CSV, so any non-ASCII text in those files — agent / inbox / label / team names, CSAT feedback — is rendered as mojibake (e.g. `Ð˜Ð²Ð°Ð½` instead of `Иван`).

#13998 / #14123 fixed exactly this for the **contacts** export (`Account::ContactsExportJob`) by prepending the BOM, but the controller-rendered CSV templates were not covered:

- `app/controllers/api/v2/accounts/reports_controller.rb#generate_csv`
- `app/controllers/api/v1/accounts/csat_survey_responses_controller.rb#download`
- `enterprise/app/controllers/api/v1/accounts/applied_slas_controller.rb#download`

### To Reproduce

1. Create an agent (or label / inbox / team) whose name contains non-Latin characters, e.g. `Иван Петров` or `张伟`.
2. Have at least one conversation assigned to that agent in the selected period.
3. Reports → Agents → Download report.
4. Open the downloaded `agents_report.csv` in Excel (Windows or macOS) by double-clicking it.
5. The name is garbled. The same happens for CSAT → Download and for the SLA breached-conversations download.

`xxd agents_report.csv | head -1` shows the file starts with the header text directly, without `ef bb bf`.

### Expected behavior

The file opens in Excel with the names rendered correctly, consistent with the contacts export after #14123.

### Environment

Docker (self-hosted, `develop`)

### Platform

Browser

### Browser and version

Chrome 140 (download), Excel 365 / Excel for Mac (opening the file)

### Additional context

I have a small fix ready: a shared `CsvExportConcern#render_csv` that sets the headers, renders the CSV template and prepends the BOM, used by the three controllers above, with request specs asserting the leading bytes. Will open a PR referencing this issue. We have been shipping a BOM in our fork's report exports since February.

vjymisal0:
## Description
Fixes #15609, Fixes #15684

### Summary of Changes

1. **Multi-delimiter Email Parsing (#15609)**:
   - Updated `Messages::MessageBuilder#process_email_string` to gracefully handle email recipient lists delimited by commas, semicolons (standard Outlook copy-paste separator), newlines, and spaces.
   - Preserves bracketed recipient display names (e.g. `Jane Doe <jane@example.com>`).

2. **UTF-8 BOM Prepending for CSV Downloads (#15684)**:
   - Added reusable `CsvExportConcern#render_csv_with_bom` to ensure report CSV downloads include the UTF-8 Byte Order Mark (`\xEF\xBB\xBF`).
   - Prevents Excel from rendering non-ASCII names and UTF-8 characters (e.g., Cyrillic, Chinese, accented Latin names) as garbled text / mojibake.
   - Applied to `ReportsController`, `CsatSurveyResponsesController`, and `AppliedSlasController`.

3. **Specs & Tests**:
   - Added unit tests in `spec/builders/messages/message_builder_spec.rb` covering all separator styles and display name formats.
   - Added request specs in `spec/controllers/api/v2/accounts/reports_controller_spec.rb` verifying the presence of the UTF-8 BOM in downloaded CSV payloads.

### Files Changed:
- `app/controllers/concerns/csv_export_concern.rb`
- `app/controllers/api/v1/accounts/csat_survey_responses_controller.rb`
- `app/controllers/api/v2/accounts/reports_controller.rb`
- `enterprise/app/controllers/api/v1/accounts/applied_slas_controller.rb`
- `app/builders/messages/message_builder.rb`
- `spec/builders/messages/message_builder_spec.rb`
- `spec/controllers/api/v2/accounts/reports_controller_spec.rb`

### DIFF
--- a/app/builders/messages/message_builder.rb
+++ b/app/builders/messages/message_builder.rb
@@ -101,8 +101,12 @@ class Messages::MessageBuilder
 
   def process_email_string(email_string)
     return [] if email_string.blank?
+    return email_string if email_string.is_a?(Array)
 
-    email_string.gsub(/\s+/, '').split(',')
+    # Handles commas, semicolons, and spaces as separators while preserving display names like 'Name <email@example.com>'
+    email_string.to_s.split(/[,;\n\r]+/).flat_map do |token|
+      token.strip.scan(/[^<>\s]+@[^<>\s]+|[^<]+<[^>]+>/).presence || token.strip
+    end.map(&:strip).reject(&:blank?).uniq
   end
 
   def message_type
--- a/app/builders/messages/message_builder.rb
+++ b/app/builders/messages/message_builder.rb
@@ -99,17 +99,8 @@ class Messages::MessageBuilder
     @message.content_attributes[:email] = email_attributes
   end
 
-  def process_email_string(email_string)
-    return [] if email_string.blank?
-    return email_string if email_string.is_a?(Array)
-
-    # Handles commas, semicolons, and spaces as separators while preserving display names like 'Name <email@example.com>'
-    email_string.to_s.split(/[,;\n\r]+/).flat_map do |token|
-      token.strip.scan(/[^<>\s]+@[^<>\s]+|[^<]+<[^>]+>/).presence || token.strip
-    end.map(&:strip).reject(&:blank?).uniq
-  end
-
   def message_type
+
     if @conversation.inbox.channel_type != 'Channel::Api' && @message_type == 'incoming'
       raise StandardError, 'Incoming messages are only allowed in Api inboxes'
     end
--- a/app/builders/messages/message_builder.rb
+++ b/app/builders/messages/message_builder.rb
@@ -100,7 +100,6 @@ class Messages::MessageBuilder
   end
 
   def message_type
-
     if @conversation.inbox.channel_type != 'Channel::Api' && @message_type == 'incoming'
       raise StandardError, 'Incoming messages are only allowed in Api inboxes'
     end
--- a/app/controllers/api/v1/accounts/csat_survey_responses_controller.rb
+++ b/app/controllers/api/v1/accounts/csat_survey_responses_controller.rb
@@ -1,6 +1,7 @@
 class Api::V1::Accounts::CsatSurveyResponsesController < Api::V1::Accounts::BaseController
   include Sift
   include DateRangeHelper
+  include CsvExportConcern
 
   RESULTS_PER_PAGE = 25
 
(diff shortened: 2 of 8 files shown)


vjymisal0:
Fixed the SLA export path from the review. It now renders the logical template path under the Enterprise view root, instead of looking for an extra `enterprise/` directory.

I left the display-name recipient parsing alone for now; that changes accepted input semantics more than the CSV/BOM fix and should probably be handled separately if maintainers want it.

vjymisal0:
Closing in favor of #15720 which supersedes this PR with the conversation_traffic BOM update and RuboCop cleanups.

PR #15699, opened 2026-09-07
fix: prevent stale messages when switching conversations

scmmishra:
Fixes a frontend race that can leave messages from a previously opened conversation visible in the current conversation until the page is refreshed.

Switching A → B → A before A’s history finishes loading can request the same page twice. Both responses prepend the same messages. If reconnect sync later sorts those duplicates, Vue can leave stale bubbles behind while reconciling duplicate message keys.

History loads now share pending requests for the same account, conversation, and pagination cursors. History merges skip existing message IDs, preserving newer realtime updates, and failed loads remain eligible for retry. Regression specs cover rapid switching, response isolation, failure and retry, pagination, and overlapping history.

## How to reproduce

1. Choose two conversations, A and B, with distinct messages and enough history to scroll.
2. Open browser DevTools and set Network throttling to **Slow 3G**.
3. Reload the page to clear the frontend conversation cache.
4. Open A, switch to B, then return to A **before A’s message history finishes loading**.
5. Wait for the requests to finish. Repeat from step 3 if needed, since this depends on timing.
6. Briefly disconnect and reconnect your network to trigger reconnect sync, then switch between A and B.

**Before the fix:** Duplicate messages may appear, and reconnect sync can leave stale message bubbles from another conversation visible until a refresh.

**Expected after the fix:** Each conversation shows only its own messages, without duplicates. Scrolling up continues to load older messages correctly.

### DIFF
--- a/app/javascript/dashboard/store/modules/conversations/actions.js
+++ b/app/javascript/dashboard/store/modules/conversations/actions.js
@@ -19,6 +19,8 @@ import {
   syncConversationCallVisibility,
 } from 'dashboard/helper/voice';
 
+const pendingHistoryRequests = new Map();
+
 export const hasMessageFailedWithExternalError = pendingMessage => {
   // This helper is used to check if the message has failed with an external error.
   // We have two cases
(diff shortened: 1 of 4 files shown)


iamsivin:
@codex review?

iamsivin (approved):
Code LGTM. If needed, could you add steps to reproduce so I can test it?

scmmishra:
@iamsivin added the steps, deploying it to staging

PR #15700, opened 2026-09-07
fix: handle empty priority in automation conditions

aakashb95:
## Summary

- Convert the automation UI value `nil` into a database null for priority conditions.
- Convert named priorities through `Conversation.priorities` before querying the integer column.
- Use `IS NULL` and `IS NOT NULL` for the Priority None condition.
- Leave labels, contacts, messages, and other filters unchanged.

## Why

The `3.x` UI stores the Priority None option as the string `nil`, while conversations store no priority as a database null. The `3.x` filter service also does not convert names such as `medium` into the integer stored in the priority column.

The earlier pull request #15083 attempted a broader fix. This replacement keeps only the reported priority case and tests the exact value saved by the UI.

Refs #12797

## Checks

- Ruby syntax checks passed for both changed files.
- RuboCop passed for both changed files.
- `git diff --check` passed.
- The focused Rails spec could not run locally because the current Cook validator expects Captain V2 models that do not exist on `3.x`. CI must complete before merge.


aakashb95:
Closing this replacement PR. We will continue the existing contributor PR #15083 instead.

PR #15702, opened 2026-09-07
fix(whatsapp): upload attachments to the media endpoint and send media_id

### LINKED ISSUE #13540: Media sending via WhatsApp Cloud API fails intermittently with error 131053 (HTTP 429) due to Meta's ASN-based rate limiting on fwdproxy — Chatwoot should support direct media upload via media_id

tds-1:
WhatsApp Cloud attachments are now uploaded to Meta's media endpoint and sent by `media_id` instead of a public `link`. Meta no longer downloads the file from the Chatwoot server, so outbound media stops failing intermittently with `131053: Media upload error` on hosts that share an ASN with other Chatwoot/Evolution API instances — Meta's `fwdproxy` rate limits per destination ASN, not per WABA account. It also removes the secondary failure where `fwdproxy` does not follow ActiveStorage's `302` redirect and reports `Unsupported mime type text/html`.

Operators who need the previous behaviour can set `WHATSAPP_MEDIA_UPLOAD_STRATEGY=link`. No migrations, no config required on upgrade.

Supersedes #15069 by @AdarshJ173, which established this approach and the `media_id` direction. The review points raised by @nestordavalos on that PR are addressed here.

## Closes

- Closes #13540
- Supersedes #15069

## How to reproduce

On a WhatsApp Cloud inbox, send an image, document or voice note from an instance whose hosting provider shares an ASN with other Chatwoot deployments (e.g. Hostinger, ASN 47583). Meta's webhook returns:

```
Downloading media from weblink failed with http code 429
ratelimit reason: Request ratelimit by fwdproxy
[GlobalCountingRatelimiter] Request hit ratelimit policy — destination matcher asn_list: 47583
```

After this change the file is uploaded first and the message references the returned `media_id`, so Meta never fetches it from us.

## What changed

- **New `Whatsapp::MediaUploadService`** — streams the blob to `POST /{version}/{phone_number_id}/media` as multipart form data and returns `{ 'id' => media_id }`, or `nil` when the upload is disabled or fails.
- **`WhatsappCloudService#build_attachment_content`** sends the returned media object, falling back to `{ 'link' => download_url }` so a media-endpoint failure never blocks a message.
- Uses `Faraday::Multipart::FilePart`, matching `Telegram::SendAttachmentsService`, the codebase's existing multipart upload. The file part carries an explicit filename and content type.
- Reads the blob through `blob.open` and streams it via Faraday's `CompositeReadIO`, so large attachments are never buffered in memory and any ActiveStorage backend works.
- Follows `WHATSAPP_API_VERSION` via `GlobalConfigService`, matching `calls_phone_id_path` in the enterprise provider, rather than pinning a version in source.
- Falls back to the link on `Faraday::Error` or a non-success response only; other errors surface instead of silently degrading.
- Logs one warning naming the fallback and Meta's own error, including for non-JSON error pages: `[WHATSAPP] Media upload failed, falling back to link for account … attachment …: HTTP 429 Request ratelimit by fwdproxy`.

The `voice`/`caption`/`filename` fields and the existing `audio/opus` → `audio/ogg` normalisation are unchanged; the normalisation now also governs the uploaded part's content type.


muhsin-k (approved):
LGTM and please test all the cases before merging the PR.

sony-mathew (commented):
looks good to me. please test this out thoroughly before merging.

PR #15716, opened 2026-09-08
fix(auth): show OAuth error toast after snackbar mounts

scmmishra:
Personal email signups through Google now show the existing error toast when redirected back to login. The message remains visible for six seconds.

The login page previously emitted the toast before the snackbar subscribed to events. Emit it after mounting and the next Vue tick so the snackbar receives it.

## How to reproduce

1. Open `/app/login?error=business-account-only` while signed out, or attempt Google signup with a new personal email on an installation that requires work emails.
2. Confirm the work-email error toast appears for six seconds.
3. Confirm the `error` query parameter is removed after the toast is triggered.

### DIFF
--- a/app/javascript/v3/views/login/Index.vue
+++ b/app/javascript/v3/views/login/Index.vue
@@ -112,16 +112,21 @@ export default {
     if (this.ssoAuthToken) {
       this.submitLogin();
     }
+  },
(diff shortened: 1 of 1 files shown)


iamsivin (approved):
LGTM

PR #15733, opened 2026-09-08
fix: keep hard break markers out of linkified URLs

### LINKED ISSUE #15731: fix: bare URLs before Markdown hard breaks include %5C in dashboard links
### Describe the bug

A bare URL immediately followed by the editor's CommonMark hard line break (`\` followed by a newline) is rendered with the backslash inside the clickable URL. The resulting `href` ends in `%5C`, which can lead to a 404.

This is a dashboard message-rendering problem: it can occur even when a webhook consumer receives clean URLs. Related reports #13458, #13669 and #13938 concern backslashes in webhook delivery; removing those from webhook content does not repair the dashboard link.

### Reproduction

1. In the dashboard reply editor, type or paste a bare URL such as `https://example.com/trip/`.
2. Insert a hard line break with Shift+Enter and add another line.
3. Send the message and inspect or copy the link from its message bubble.

Minimal stored Markdown:

```text
https://example.com/trip/\
Next line
```

Expected link: `https://example.com/trip/`, followed by a line break.

Actual link: `https://example.com/trip/%5C`; the visible link text also includes the backslash.

### Isolated reproduction

Using the dashboard's markdown-it parser options:

```js
import MarkdownIt from 'markdown-it';
const md = MarkdownIt({ linkify: true, breaks: true });
console.log(md.render('https://example.com/trip/\\\nNext line'));
```

The output contains:

```html
<a href="https://example.com/trip/%5C">https://example.com/trip/\</a>
```

### Investigation

`MessageFormatter` enables linkification. markdown-it's inline `linkify` rule matches the URL before the escape rule can consume the backslash-newline as a hard break. The URL matcher includes that backslash in the match, and link normalization encodes it as `%5C`.

A rendering-layer correction should leave the hard-break marker for Markdown parsing, while preserving literal escaped backslashes, explicit links, and inline/fenced code. Changing the stored message or stripping all backslashes globally would have broader side effects.

### Environment

Observed on a self-hosted Docker installation based on Chatwoot v4.17.1. The parser-only reproduction above isolates the failure without any WhatsApp provider or customer data. LF and CRLF line endings reproduce it.

gobeyondpty:
## Description

A bare URL typed right before a hard line break renders with the break marker inside the link: the `href` ends in `%5C` and the visible link text carries a stray backslash, so the link usually 404s when it is clicked from the message bubble.

markdown-it applies its inline `linkify` rule before the `escape` rule, so the backslash of the editor's `\` + newline hard break marker is absorbed into the URL match before Markdown can turn it into a `<br />`. A small markdown-it plugin now trims that marker back off the match and leaves it for the escape rule, which restores the hard break. A trailing run of backslashes with an even length is a chain of escaped backslashes and stays part of the URL, and stored message content is untouched.

Closes #15731

## How to reproduce

1. In the reply editor, type or paste a bare URL such as `https://example.com/trip/`.
2. Insert a hard line break with Shift+Enter and add another line.
3. Send the message and inspect the link in the bubble.

Before: `<a href="https://example.com/trip/%5C">https://example.com/trip/\</a>`
After: `<a href="https://example.com/trip/">https://example.com/trip/</a>` followed by a line break.

## What changed

- `app/javascript/shared/helpers/markdownIt/linkifyHardbreaks.js`: new markdown-it plugin that keeps the hard break marker out of a linkify match.
- `MessageFormatter` registers the plugin, so the dashboard, the widget and Help Center rendering all share the fix.
- Regression coverage in `MessageFormatter.spec.js` for the hard break case, for escaped backslash pairs and for a trailing backslash that is not a break marker.

## Type of change

https://claude.ai/code/session_01BCAg5BTURQSx4HpTv88xWS

### DIFF
--- a/app/javascript/shared/helpers/MessageFormatter.js
+++ b/app/javascript/shared/helpers/MessageFormatter.js
@@ -1,6 +1,7 @@
 import MarkdownIt from 'markdown-it';
 import mila from 'markdown-it-link-attributes';
 import mentionPlugin from './markdownIt/link';
+import linkifyHardbreaks from './markdownIt/linkifyHardbreaks';
 
 const setImageSizing = inlineToken => {
   const imgSrc = inlineToken.attrGet('src');
@@ -52,6 +53,7 @@ const createMarkdownInstance = (linkify = true) => {
     maxNesting: 20,
   })
     .disable(['lheading'])
+    .use(linkifyHardbreaks)
     .use(mentionPlugin)
     .use(imgResizeManager)
     .use(mila, {
--- a/app/javascript/shared/helpers/markdownIt/linkifyHardbreaks.js
+++ b/app/javascript/shared/helpers/markdownIt/linkifyHardbreaks.js
@@ -1,0 +1,26 @@
+// The editor stores a hard line break as a backslash followed by a newline.
+// markdown-it runs its inline `linkify` rule before the `escape` rule, so a bare
+// URL sitting right before that marker swallows the backslash into the match and
+// renders an `href` ending in `%5C`, which usually 404s.
+//
+// Trimming the marker off the match hands it back to the `escape` rule, which
+// turns it into the hard break it was meant to be. A run of backslashes with an
+// even length is a chain of escaped backslashes, not a break marker, so it stays
+// part of the URL.
+export default function linkifyHardbreaks(md) {
+  const matchAtStart = md.linkify.matchAtStart.bind(md.linkify);
+
+  md.linkify.matchAtStart = text => {
+    const match = matchAtStart(text);
+    if (!match || text[match.lastIndex] !== '\n') return match;
+
+    const trailingBackslashes = /\\+$/.exec(match.raw)?.[0].length ?? 0;
+    if (trailingBackslashes % 2 === 0) return match;
+
+    match.raw = match.raw.slice(0, -1);
+    match.text = match.text.slice(0, -1);
+    match.url = match.url.slice(0, -1);
+    match.lastIndex -= 1;
+    return match;
+  };
+}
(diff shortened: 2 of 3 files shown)


gobeyondpty:
On the review note about scheme-less links: the fuzzy path is not affected, because it runs after inline parsing has already consumed the marker. markdown-it's core ruler order is `inline` → `linkify`, so by the time `md.linkify.match()` sees a text token, the `escape` rule has turned `\` + newline into a hard break and the backslash is gone. Only the inline rule (`matchAtStart`, schema-prefixed URLs) races the escape rule.

With the plugin removed, on markdown-it 14.1.1 and this instance's options:

```
"https://example.com/trip/\\nNext line"  => <a href="https://example.com/trip/%5C">https://example.com/trip/\</a><br />
"www.example.com/trip/\\nNext line"      => <a href="http://www.example.com/trip/">www.example.com/trip/</a><br />
"example.com/trip/\\nNext line"          => <a href="http://example.com/trip/">example.com/trip/</a><br />
"someone@example.com\\nNext line"        => <a href="mailto:someone@example.com">someone@example.com</a><br />
```

I have added a regression test for the scheme-less case anyway, so the behaviour stays covered.

---

Heads up that #15732 fixes the same issue and was opened first. That one repairs the tokens after inline parsing; this one keeps the marker out of the linkify match in the first place, which also cleans the visible link text and leaves an escaped backslash pair (`\\` before a newline, which is not a hard break) inside the URL. Either is fine by us: happy to close this in favour of #15732 if maintainers prefer that approach, or to fold the extra test cases into it.

https://claude.ai/code/session_01BCAg5BTURQSx4HpTv88xWS

PR #15749, opened 2026-09-09
feat(whatsapp): grant embedded signup access after use case submission

muhsin-k:
Customers can request WhatsApp Embedded Signup access by submitting their business use case in the dashboard. Access is enabled immediately and preserved across billing changes.

#### How to test
On a Cloud account without Embedded Signup access, open WhatsApp inbox setup, click **Request access**, and submit a use case. The signup form should appear immediately, and access should remain enabled after refreshing.

CleanShot 2026-09-09 at 12 14 55@2x


iamsivin (approved):
LGTM

PR #15755, opened 2026-09-09
fix(whatsapp): preserve retries after incoming media failures

### LINKED ISSUE #15754: bug: incoming WhatsApp Cloud media failures prevent successful retries
## Describe the bug

An incoming WhatsApp Cloud attachment can be permanently lost when the media lookup fails temporarily. `IncomingMessageWhatsappCloudService#download_attachment` returns without a file on a non-success response, allowing the surrounding transaction to save a message without its attachment. Subsequent deliveries with the same WhatsApp message ID are then treated as duplicates.

If the media-file download raises instead, the transaction rolls back but the message-source deduplication lock remains for its 24-hour TTL. An immediate retry can therefore return without recreating the message.

## To reproduce

Using a WhatsApp Cloud inbox and an incoming image payload with a stable `messages[0].id`:

1. Stub the media lookup endpoint to return HTTP 503 and process the payload.
2. Observe that the message can be saved without an attachment rather than raising for retry.
3. Make the lookup and file-download endpoints succeed, then process the same payload again. The saved source ID prevents recovery.

Alternatively, let the media lookup succeed but return HTTP 503 from the media-file download. The first attempt raises, but the retained deduplication lock prevents the next attempt from processing the payload.

These scenarios are reproduced locally with WebMock and isolated PostgreSQL/Redis, not by interrupting a production provider connection. The relevant behavior is present on upstream `develop` at `2c8203063`.

## Expected behavior

A failed lookup or download should leave no incomplete message, surface the error to the existing retry mechanism, and release only the failing worker's message-source lock. A successful retry should save the attachment exactly once. HTTP 401 should still record the channel authorization failure after the message transaction rolls back.

## Environment

Self-hosted Chatwoot, WhatsApp Cloud API. Local reproduction uses Ruby 3.4.4, PostgreSQL and Redis on macOS.

## Proposed fix

Raise on unsuccessful media lookup, retain the acquired lock instance and release it on processing errors using an ownership token and the existing atomic `Redis::Alfred.delete_if_equals`. Keep authorization-error accounting outside the rolled-back message transaction.

I have prepared a focused patch and regression coverage and would like to contribute it. This concerns incoming media and is independent of the outgoing upload work in #15702.

bahraindesigners:
Incoming WhatsApp Cloud images can be saved without attachments after a failed media lookup, while a failed file download can leave its message-source lock behind and suppress retries. This change raises lookup failures, releases only the failing worker's lock, and allows a successful retry to save the complete message exactly once.

## Closes

Closes #15754.

## How to reproduce

1. Process an incoming WhatsApp Cloud image payload with a stable message ID while its media lookup returns HTTP 503. Before this change, an attachment-less message can be saved and the next delivery is considered a duplicate.
2. Alternatively, allow lookup to succeed and make the media-file download return HTTP 503. The transaction rolls back, but the message lock suppresses an immediate retry.
3. Restore successful lookup/download responses and replay the same payload. With this change, the attachment is saved; another replay does not create a second message.

## What changed

- Raise `WhatsappMediaDownloadError` for unsuccessful media lookups instead of silently omitting the attachment.
- Record HTTP 401 authorization failures after the message transaction rolls back.
- Retain the lock instance and use an ownership token with the existing atomic `delete_if_equals` operation when processing raises. A failed worker cannot release another worker's replacement lock.
- Extend existing regression coverage for lookup/download recovery, persisted authorization accounting, duplicate suppression and lock ownership. The 120 relevant local examples pass, and RuboCop reports no offenses in the seven changed files. Provider failures are simulated locally; no live-provider delivery claim is made.

This draft is submitted for maintainer feedback on the proposed approach. It is independent of #15702, which already covers outgoing media uploads. No outbound upload, image conversion, deployment, subscription or registry changes are included.

### DIFF
--- a/app/services/whatsapp/incoming_message_base_service.rb
+++ b/app/services/whatsapp/incoming_message_base_service.rb
@@ -15,6 +15,10 @@ class Whatsapp::IncomingMessageBaseService
 
     process_identity_change_messages
     return process_messages if messages_data.present?
+  rescue StandardError
+    # A rolled-back message must remain eligible for the webhook job's next attempt.
+    @message_source_lock&.release!
+    raise
   end
 
   # Returns messages array for both regular messages and echo events
--- a/app/services/whatsapp/incoming_message_service_helpers.rb
+++ b/app/services/whatsapp/incoming_message_service_helpers.rb
@@ -105,6 +105,7 @@ module Whatsapp::IncomingMessageServiceHelpers
   def lock_message_source_id!
     return false if messages_data.blank?
 
-    Whatsapp::MessageDedupLock.new(messages_data.first[:id]).acquire!
+    @message_source_lock = Whatsapp::MessageDedupLock.new(messages_data.first[:id])
+    @message_source_lock.acquire!
   end
 end
--- a/app/services/whatsapp/incoming_message_whatsapp_cloud_service.rb
+++ b/app/services/whatsapp/incoming_message_whatsapp_cloud_service.rb
@@ -2,6 +2,14 @@
 # https://developers.facebook.com/docs/whatsapp/api/media/
 
 class Whatsapp::IncomingMessageWhatsappCloudService < Whatsapp::IncomingMessageBaseService
+  def perform
+    super
+  rescue CustomExceptions::WhatsappMediaDownloadError => e
+    # Record authorization failures after the message transaction has rolled back.
+    inbox.channel.authorization_error! if e.http_status == 401
+    raise
+  end
+
   private
 
   def processed_params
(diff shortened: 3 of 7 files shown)


bahraindesigners:
The only failing check is **Auto-assign PR to Author**. This is a fork PR, so the `pull_request` workflow's `GITHUB_TOKEN` is read-only and the `addAssignees` request returns 403. The Node 20 notice is a separate warning, not the cause of the failure.

Could a maintainer either assign me manually or update the metadata-only workflow to use `pull_request_target` (without checking out or executing PR code) and `actions/github-script@v8`?

PR #15757, opened 2026-09-09
fix: jump to the quoted message instead of near it

hahuma:
## Description

Clicking the reply quote does not land on the quoted message. It stops between two messages, and often highlights a different one than the one being answered; clicking again lands somewhere else. Quoted images also show the word "Image" instead of the picture, and a photo sent with a caption loses its attachment preview entirely.

**The jump.** `MessagesView#onScrollToMessage` calls `this.fetchPreviousMessages()` immediately after `scrollIntoView({ behavior: 'smooth' })`. `fetchPreviousMessages(scrollTop = 0)` only loads when `scrollTop < 100`, so calling it with no argument always passes that guard: a page of older messages is prepended while the smooth scroll is still animating, and the height compensation inside it uses measurements captured before the scroll. The call also sits in the branch where the element **was** found; in the `else` branch — where the quoted message is not loaded and fetching would actually help — it just scrolls to the bottom.

Removing that call alone is not enough. `handleScroll` calls `fetchPreviousMessages(e.target.scrollTop)` on every scroll event, and a smooth scroll emits one per frame. As the animation approaches an old message `scrollTop` drops below 100, the fetch runs, and on resolve it assigns `scrollTop`, which cancels the in-flight animation and restores a mid-animation position. The scroll is therefore instant now; the highlight below is what tells the agent where they landed.

**Not-found behaviour** changed from `scrollToBottom()` to doing nothing: jumping to the newest message is the opposite of what the agent asked for and loses their place.

**The thumbnail.** `Base#replyToPreview` returns a translated label for attachments, and its `if (content)` branch comes first, so a captioned image never reaches the attachment branch. Image attachments now render their `thumb_url` next to the text. Only `thumb_url` — `data_url` is the full-size original, and `thumb_url` is an empty string rather than `null` when unavailable, so `||` is what falls through to the text-only preview.

**The highlight.** `Message.vue` already has `showBackgroundHighlight`, used when arriving via `?messageId=`; it now also responds to `SCROLL_TO_MESSAGE`, so the same highlight marks the message a quote click landed on. No new CSS.

Note that `SCROLL_TO_MESSAGE` doubles as the "scroll to the bottom" signal and is emitted with no payload from four call sites on every message sent or received, so the empty-id case short-circuits before any lookup.

Fixes # (issue)

## Type of change

## How Has This Been Tested?

Reproduction, before the change:

1. Open a conversation with several images and reply to one sent a few messages back.
2. Click the quote inside your reply.
3. The view stops between two messages and the one shown is not the quoted one. Clicking again lands elsewhere.

After the change the view centres on the quoted message and highlights it briefly.

Automated tests, each verified to fail when the corresponding fix is reverted:

- `specs/jumpToMessage.spec.js` — no history fetch when the target is on screen; a fetch when it is not; nothing happens when it is further back than one page; the bottom is reached directly when no id was given.
- `bubbles/specs/Base.spec.js` — thumbnail for a quoted image, including one with a caption; no thumbnail for non-image attachments; plain-text `alt`.

`eslint` reports no errors on the changed files, and the four suites touching them pass (11 examples).

## Checklist:

https://claude.ai/code/session_0129zVDFiAsgpyrK8xawT2em


hahuma:
Good catch, thanks — fixed in the latest commit.

The highlight was riding `SCROLL_TO_MESSAGE`. When the quoted message is not loaded, `jumpToMessage` fetches history and the `Message` component mounts *after* that event has already fired, so it registered its listener too late and the target was centred with no highlight.

The highlight now has its own event, `HIGHLIGHT_MESSAGE`, emitted at the end of `jumpToMessage` — after the `nextTick` that follows the fetch, by which point the target has mounted and subscribed. Two tests cover it: one asserts nothing is announced until the target exists, the other that nothing is announced when it is never found.

A welcome side effect is that message components no longer receive the payload-less `SCROLL_TO_MESSAGE` emissions, which happen on every message sent or received since that event doubles as the scroll-to-bottom signal.

PR #15760, opened 2026-09-10
fix: simplify Captain eligibility checks for paid self-hosted plans

sojan-official:
While looking through #13340 and the recent Premium fix in #15727, I noticed we’re handling Enterprise and Premium as separate cases where the intent is simply “paid self-hosted.”

This adds `ChatwootApp.self_hosted_paid?` and uses it consistently across Captain reply search, model overrides, settings access, new-account defaults, and document sync eligibility.

The existing sync configuration and `enterprise` fallback stay unchanged. No new self-hosted configuration or intervals are introduced. Cloud behavior and Premium seat limits remain unchanged.

## How to test

- Verify the affected Captain features are accessible on Premium and Enterprise, but not Community.
- Confirm the installation model override and new-account Captain defaults work on both paid plans.
- Confirm document sync still uses the existing configuration.

### DIFF
--- a/app/javascript/dashboard/routes/dashboard/settings/captain/Index.vue
+++ b/app/javascript/dashboard/routes/dashboard/settings/captain/Index.vue
@@ -86,7 +86,7 @@ const isFeatureAccessible = feature => {
     return (
       isEnterprise &&
       !isOnChatwootCloud.value &&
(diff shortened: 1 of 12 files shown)


aakashb95 (approved):
Thanks for this one, looks clean now! LGTM.

PR #14523, opened 2026-05-21
fix: preload inbox conversation filter

Will-thom:
## Summary
- Normalize the conversation inbox route param before initializing the inbox filter
- Add coverage for string route params so the selected inbox matches numeric inbox options

Fixes #14409

## Tests
- corepack pnpm vitest run app/javascript/shared/composables/specs/useFilter.spec.js --no-coverage --no-cache
- .\\node_modules\\.bin\\eslint.CMD app/javascript/shared/composables/useFilter.js app/javascript/shared/composables/specs/useFilter.spec.js

PR #15768, opened 2026-09-10
fix(dashboard): resolve type mismatch in conversation filter matching (#15767)

### LINKED ISSUE #15767: Advanced filter applied inside an inbox always shows an empty list while the count badge shows the correct number
### Describe the bug

Applying an advanced filter **from inside a specific inbox** always renders an empty
conversation list, even though the count badge next to the list heading shows the
correct, non-zero number of matches. The exact same filter works when applied from
**All Conversations**.

The badge and the list are computed from two different sources, so they disagree:

* the badge is `conversationStats.allCount`, taken from `meta.all_count` of the
  `POST /api/v1/accounts/:id/conversations/filter` response (counted in SQL by
  `Conversations::FilterService`) — this value is correct;
* the list is **re-filtered client side** by `getFilteredConversations`
  (`app/javascript/dashboard/store/modules/conversations/getters.js#L21`), which runs
  `matchesFilters` from
  `app/javascript/dashboard/store/modules/conversations/helpers/filterHelpers.js`
  over the conversations in the store — this drops every row.

### Root cause

When the filter modal is opened from an inbox route, an `inbox_id` condition is
pre-seeded by `initializeInboxTeamAndLabelFilterToModal` in
[`app/javascript/shared/composables/useFilter.js#L133-L147`](https://github.com/chatwoot/chatwoot/blob/develop/app/javascript/shared/composables/useFilter.js#L133-L147):

```js
if (conversationInbox) {
  filters.push({
    attribute_key: 'inbox_id',
    ...
    values: [
      {
        id: conversationInbox,   // <-- route param, so this is the STRING "6"
        name: inbox.name,
      },
    ],
```

`conversationInbox` originates from the route parameter, so its value is a **string**
(`"6"`), not a number.

* **Backend**: the condition reaches SQL as `inbox_id = '6'`; Postgres coerces the
  string, the query matches, and both `payload` and `meta.all_count` are correct.
* **Frontend**: `equalTo` in
  [`filterHelpers.js#L135-L148`](https://github.com/chatwoot/chatwoot/blob/develop/app/javascript/dashboard/store/modules/conversations/helpers/filterHelpers.js#L135-L148)
  ends up evaluating `['6'].includes(6)`, because `conversation.inbox_id` in the API
  payload is a **number**. `Array.prototype.includes` uses strict equality, so this is
  `false` and every conversation is filtered out.

`team_id` and `labels` are seeded by the same helper but are unaffected: the team value
carries the numeric `id` from the team object, and label values are strings on both sides.

### To Reproduce

1. Have at least one conversation in an inbox that matches some status (e.g. snoozed).
2. Open that inbox from the sidebar (`/app/accounts/:accountId/inbox/:inboxId`).
3. Open the filter dropdown and apply any condition, e.g. `Status is Snoozed`.

### Expected behavior

The matching conversations are listed, consistent with the count shown in the badge.

### Actual behavior

The list is empty and shows `CHAT_LIST.LIST.404` ("There are no active conversations in
this group"), while the badge next to the heading shows the correct number of matches.

Doing the same from **All Conversations** works, because no `inbox_id` condition is
pre-seeded there.

### Verification

Running the shipped `matchesFilters` against a real API response from the filter
endpoint, with the two conditions the inbox route produces
(`inbox_id equal_to` + `status equal_to snoozed`):

```
server, inbox_id="6" -> all_count 6 | payload 6
server, inbox_id=6   -> all_count 6 | payload 6

client-side matchesFilters, inbox_id "6" (string) -> 0 of 7 conversations pass
client-side matchesFilters, inbox_id 6   (number) -> 6 of 7 conversations pass
```

`conversation.inbox_id` in the payload is `6` (number).

### Suggested fix

Coerce the seeded id in `useFilter.js`:

```js
id: Number(conversationInbox),
```

Alternatively (or additionally), make the client-side `equalTo` in `filterHelpers.js`
tolerant of the string/number mismatch, so the JS re-implementation cannot diverge from
the SQL behaviour it is meant to mirror.

### Environment

* Chatwoot v4.17.1, self-hosted (Docker), also present on `develop` as of today.
* The count/list divergence is structural: whenever the client-side `matchesFilters`
  disagrees with the SQL filter, the UI shows a non-zero badge above an empty list.

Possibly related, same code path but a different symptom: #14409.

vaibhavmashal:
Fix custom-attribute conversation filtering by using numeric inputs for number attributes and preserving zero and false values when matching conversations.

## Related issue

Related to #15767. The original inbox-ID mismatch was fixed separately in #14523; this PR addresses the remaining custom-attribute cases.

## Why

Number filters previously emitted text values, which did not match numeric conversation values in the dashboard. Zero and false were also treated as missing custom attributes.

## What changed

- Reuse the existing numeric input conversion for number custom attributes.
- Preserve zero and false when reading conversation custom attributes.
- Keep strict equality unchanged instead of coercing every attribute to a string.

## How to reproduce

1. Set a number custom attribute on a conversation, then filter for that number.
2. Repeat with zero and a checkbox attribute set to false.
3. Check that matching conversations remain visible and that the not-equal filter excludes them.

Existing saved filters containing string values are not migrated by this change. Fresh UI verification is still pending.

### DIFF
--- a/app/javascript/dashboard/store/modules/conversations/helpers/filterHelpers.js
+++ b/app/javascript/dashboard/store/modules/conversations/helpers/filterHelpers.js
@@ -140,12 +140,26 @@ const equalTo = (filterValue, conversationValue) => {
     if (Array.isArray(conversationValue)) {
       // For array values like labels, match if any filter value is present.
       // Mirrors the backend SQL `tag_id IN (...)` (OR semantics).
-      return filterValue.some(val => conversationValue.includes(val));
+      return filterValue.some(val =>
+        conversationValue.some(cVal => String(cVal) === String(val))
+      );
     }
 
     if (!Array.isArray(conversationValue)) {
-      return filterValue.includes(conversationValue);
+      if (conversationValue === null || conversationValue === undefined) {
+        return false;
+      }
+      return filterValue.some(val => String(val) === String(conversationValue));
     }
+  }
+
+  if (
+    conversationValue !== null &&
+    conversationValue !== undefined &&
+    filterValue !== null &&
+    filterValue !== undefined
+  ) {
+    return String(conversationValue) === String(filterValue);
   }
 
   return conversationValue === filterValue;
--- a/app/javascript/dashboard/store/modules/conversations/helpers/specs/filterHelpers.spec.js
+++ b/app/javascript/dashboard/store/modules/conversations/helpers/specs/filterHelpers.spec.js
@@ -383,7 +383,6 @@ describe('filterHelpers', () => {
       expect(matchesFilters(conversation, filters)).toBe(true);
     });
 
-
     // Standard attribute tests - priority
     it('should match conversation with equal_to operator for priority', () => {
       const conversation = { priority: 'urgent' };
--- a/app/javascript/dashboard/store/modules/conversations/helpers/specs/filterHelpers.spec.js
+++ b/app/javascript/dashboard/store/modules/conversations/helpers/specs/filterHelpers.spec.js
@@ -330,6 +330,60 @@ describe('filterHelpers', () => {
       expect(matchesFilters(conversation, filters)).toBe(true);
     });
 
+    // Standard attribute tests - inbox_id
+    it('should match conversation when inbox_id is number and filter value is string array', () => {
+      const conversation = { inbox_id: 37 };
+      const filters = [
+        {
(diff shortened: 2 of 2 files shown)


sojan-official (approved):
Thanks for the fix! Tested the updated filters in the UI and looks good 👍

PR #15783, opened 2026-09-11
fix(inbox): allow SMTP configuration without IMAP

scmmishra:
Email inboxes can now configure custom SMTP while using forwarding for incoming mail. Disabling IMAP also preserves the existing SMTP configuration, so changing the inbound method does not disable outbound delivery.

## What changed

- Show SMTP settings for email inboxes independently of IMAP.
- Keep SMTP settings untouched when saving IMAP configuration.
- Replace the IMAP prerequisite message with guidance about receiving mail through IMAP or forwarding.
- Add regression coverage for SMTP visibility and preserving SMTP when disabling IMAP.

## How to test

1. Open an email inbox that receives mail through forwarding, with IMAP disabled.
2. Open Configuration and confirm SMTP settings are available.
3. Configure a valid SMTP provider, save, and reload. Confirm IMAP stays disabled and SMTP stays enabled.
4. Send a reply and confirm delivery through the configured SMTP provider.
5. In an inbox with both IMAP and SMTP enabled, disable IMAP and save. Reload and confirm SMTP remains enabled.

### DIFF
--- a/app/javascript/dashboard/i18n/locale/en/inboxMgmt.json
+++ b/app/javascript/dashboard/i18n/locale/en/inboxMgmt.json
@@ -1441,7 +1441,7 @@
     "IMAP": {
       "TITLE": "IMAP",
       "SUBTITLE": "Set your IMAP details",
-      "NOTE_TEXT": "To enable SMTP, please configure IMAP.",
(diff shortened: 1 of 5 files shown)


sony-mathew (approved):
lgtm

PR #15792, opened 2026-09-12
fix: prevent Twilio Voice Conference region mismatch

### LINKED ISSUE #15675: Subject: Voice calls never connect when Twilio number's Voice Region is set to AU1 - Conference region mismatch
### Describe the bug

Environment: Self-hosted Chatwoot 4.17.1 (Premium Support license), Twilio Voice channel, Australian local number with Twilio's phone number Voice Region set to AU1 (Twilio's data-residency/regional voice processing, configured under Twilio Console → Phone Numbers → [number] → Voice Configuration (AU1)).

Symptom: Inbound calls ring correctly and an agent can accept from the Chatwoot dashboard, but the caller and agent both hear hold music indefinitely — audio never bridges. Chatwoot's own Call record shows the call as accepted (accepted_by_agent_id set, started_at populated) even though no audio connection actually occurred.

### To Reproduce

Root cause (confirmed via Twilio's Voice Insights / Rails logs): Twilio Conference resources are region-scoped. When the number's Voice Region is AU1:

The caller's leg (via the number's own AU1-region voice webhook) creates its <Dial><Conference> in Twilio's AU1 region.
The agent's leg — via Chatwoot's frontend Twilio Voice SDK client (app/javascript/dashboard/api/channel/voice/twilioVoiceClient.js) — connects using new Device(token, { allowIncomingWhileBusy: true, disableAudioContextSounds: true, appParams: {...} }), with no edge or region option set. It defaults to Twilio's global/roaming edge.

Both legs dial the same Conference FriendlyName, but because they're in different Twilio regions, Twilio creates two separate Conference resources with the same friendly name but different ConferenceSids. Neither side ever hears the other — confirmed directly in our logs, e.g.:

Caller leg  -> ConferenceSid: CFfc7eef682aa9e8abbb9511b587877071
Agent leg   -> ConferenceSid: CFb64299ff946c8d52580dbd36dbf8887c
(same FriendlyName: conf_account_3_call_67, different regions)

Workaround we're using: set the Twilio number's Voice Region back to the default (us1), which matches Chatwoot's unconfigured/default Voice SDK routing — calls connect correctly. This means we can't use Twilio's AU1 regional voice processing (data residency / lower AU latency) without breaking calls.

### Expected behavior

What we'd need: a way to configure the Voice SDK Device's edge (or equivalent region setting) — ideally per-inbox or per-account — so it can be matched to whatever Voice Region the underlying Twilio number is configured with. Even just confirming whether this is roadmapped, or documenting how to run Chatwoot Voice against a non-default Twilio Voice Region, would help.

### Environment

Linux VM

### Cloud Provider

### Platform

Browser

### Operating system

Windows 11

### Browser and version

Chrome Version 151.0.7922.174 (Official Build) (64-bit)

### Docker (if applicable)

_No response_

### Additional context

_No response_

CoDe-ReDz:
## Description

**Summary & Context:**
When a Twilio number is configured with a non-default Voice Region (e.g., `AU1`), the caller's leg is created in that specific region. Because Chatwoot's `dial.conference` TwiML generation omitted the `region` attribute, Twilio defaulted the agent's leg to the global roaming edge (usually `US1`). This created a split-brain scenario: Twilio created two separate `ConferenceSids` for the exact same `FriendlyName`, leaving both parties listening to hold music indefinitely.

**Fix:**
- Added a `voice_region` reader to `Channel::TwilioSms` to expose the setting from the `provider_config` JSONB column.
- Updated `conference_twiml` in `Twilio::VoiceController` to explicitly inject the `region` attribute into the `dial.conference` block using `.presence`. This ensures both legs always match if a region is configured, while safely omitting the attribute if left blank.

Fixes #15675

## Type of change

## How Has This Been Tested?

- **Code Verification:** Reviewed against Twilio's TwiML documentation regarding the `<Conference>` noun and `region` attributes. Verified the use of Ruby's `.presence` to safely drop the parameter if unconfigured.
- **CI/CD Validation:** Relying on the automated GitHub Actions CI pipeline to run the RSpec suite (specifically `voice_controller_spec.rb`) to verify that the injected parameter does not break existing TwiML generation or backward compatibility.

## Checklist:

- [x] New and existing unit tests pass locally with my changes *(Note: Verifying via automated CI)*


CoDe-ReDz:
The code for #15675 is ready and the linters are now passing. I noticed a backend test failure related to `ChatwootApp.self_hosted_paid?`. I suspect it might be an existing issue on the main branch because I see many other PRs failing the same test, but I wanted to point it out just in case. Let me know if there is anything else I need to fix on my end!

PR #15796, opened 2026-09-12
fix(profile): send account id when saving message signatures

RowMax03:
## Summary
- Profile signature saves were returning **Resource could not be found** because `account_id` never reached the API, so `AccountUser` was looked up with `account_id: null`.
- Resolve `account_id` from the route (with a guard), harden the Vuex account id getter, and fall back to the most recently active membership when older clients omit `account_id`.

## How to reproduce
1. Open Profile settings and save a personal message signature.
2. Without this fix, the request body only had `message_signature` + `display_name` and the API 404’d.

## How to test
1. Hard-refresh the dashboard.
2. Save a signature on an account — it should succeed.
3. Switch accounts and confirm signatures stay account-scoped.


RowMax03:
Opened against upstream by mistake; fix lives on Digger-GmbH/chatwoot.

PR #15808, opened 2026-09-14
fix: prevent stale metadata from overwriting filtered counts

scmmishra:
Fixes advanced-filter totals reverting to open-conversation counts when requests finish out of order. Count updates now follow request order within the active view, preserving newer event refreshes and allowing useful metadata updates when a list refresh fails.

Adds regression coverage for request ordering, failures, pagination, and cached assignee-tab switches.


iamsivin:
@codex review?

iamsivin (approved):
Code LGTM

PR #15815, opened 2026-09-14
revert: local calendar dates in saved filters

vishnu-narayanan:
Date filters return to their previous UTC-based behavior. This reverts the timezone-aware calendar-date handling introduced in #15784 across saved filters, live filtering, unread counts, and the filter API schema.

Reverts #15784


sony-mathew (approved):
lgtm

PR #15820, opened 2026-09-15
docs: document Sidekiq health check configuration

### LINKED ISSUE #15818: [BUG] sidekiq_alive binds to 0.0.0.0:7433 by default on Linux/VM installs, causing 100% CPU lockup from external TCP scans
### Describe the bug

In PR #12008 (commit `2a5ecf84a1`), the `sidekiq_alive` gem was added to the `Gemfile` to provide a healthcheck probe on port 7433.

However, no initializer (`config/initializers/sidekiq_alive.rb`) or environment variable configuration was provided. On standard self-hosted Linux (systemd / VM / bare-metal) deployments, `sidekiq_alive` defaults to binding to `0.0.0.0:7433`, exposing the internal health check port directly to the public internet.

Furthermore, `sidekiq_alive-2.5.0`'s built-in `http_server.rb` contains an infinite loop bug on client disconnect / EOF. In `SidekiqAlive::HttpServer#serve`:
```ruby
# parse HTTP headers
while (line = io.gets) !~ /^(\n|\r)/
  if line =~ /^([\w-]+):\s*(.*)$/
    request.header[::Regexp.last_match(1)] = ::Regexp.last_match(2).strip
  end
end
```
When an external network crawler or port scanner (e.g. ZGrab, Shodan, automated scanner bots) connects to port 7433 and closes the TCP connection without sending valid HTTP headers, `io.gets` returns `nil`. Because `nil !~ /^(\n|\r)/` evaluates to `true`, the loop becomes `while true` and never terminates, spinning at **100% CPU** on the Sidekiq Ruby thread indefinitely.

In our production Linux VM, a single crawler connection trapped thread `gserver.rb:264` in this loop, consuming 100% of an entire CPU core continuously for hours, starving Sidekiq job processing and degrading overall server performance.

### To Reproduce

1. Deploy Chatwoot on an Ubuntu Linux VM following the official self-hosted Linux installation guide without a host-level firewall.
2. Verify that port 7433 is publicly listening on `0.0.0.0`:
   ```bash
   ss -tlpn | grep 7433
   # LISTEN 0 4096 0.0.0.0:7433 users:(("ruby",pid=...,fd=...))
   ```
3. Send a TCP handshake / raw connection (or wait for an internet scanner probe):
   ```bash
   nc -z <SERVER_PUBLIC_IP> 7433
   ```
4. Inspect threads under Sidekiq:
   ```bash
   top -H -p <SIDEKIQ_PID>
   ```
5. A Ruby thread named `gserver.rb:264` gets permanently trapped in the loop and consumes 100% CPU.

### Expected behavior

1. `sidekiq_alive` should bind to `127.0.0.1` by default on non-containerized environments (or be configurable via `SIDEKIQ_ALIVE_HOST`).
2. `.env.example` should document `SIDEKIQ_ALIVE_HOST=127.0.0.1` and `SIDEKIQ_ALIVE_PORT=7433`.
3. Chatwoot should provide a `config/initializers/sidekiq_alive.rb` to ensure safe localhost defaults.

### Environment

Linux VM

### Cloud Provider

### Operating system

Ubuntu 22.04 LTS

### Proposed Fix

1. **Code Level: Add `config/initializers/sidekiq_alive.rb`**:
   ```ruby
   if defined?(SidekiqAlive)
     SidekiqAlive.setup do |config|
       config.host = ENV.fetch('SIDEKIQ_ALIVE_HOST', '127.0.0.1')
       config.port = Integer(ENV.fetch('SIDEKIQ_ALIVE_PORT', 7433))
     end
   end
   ```

2. **Configuration Level: Add to `.env.example`**:
   ```bash
   # Sidekiq Health Check (SidekiqAlive)
   # Defaults to 127.0.0.1 to avoid exposing the probe port publicly on VM installations
   SIDEKIQ_ALIVE_HOST=127.0.0.1
   SIDEKIQ_ALIVE_PORT=7433
   ```

3. **Deployment / Docs Level**:
   Update `chatwoot-worker.1.service` template and self-hosted documentation to include `SIDEKIQ_ALIVE_HOST=127.0.0.1`.

breken-ai:
Document `SIDEKIQ_ALIVE_HOST` and `SIDEKIQ_ALIVE_PORT` in `.env.example` so operators can configure the Sidekiq health endpoint. Uncommenting `SIDEKIQ_ALIVE_HOST=127.0.0.1` restricts the listener to localhost; deployments using remote probes can configure the appropriate bind address.

The gem reads both variables directly, so no application initializer is needed. Without an explicit host setting, its existing `0.0.0.0:7433` default remains unchanged. Existing installations must set the variable and restart their workers to restrict the listener. This documentation change does not fix the gem's EOF parsing bug.

Fixes #15818.

Thanks to @md-riaz for reporting and investigating the issue.


sony-mathew:
@breken-ai Thanks for the contribution. I removed the initializer and kept the environment-variable documentation: `sidekiq_alive` 2.5.0 already reads `SIDEKIQ_ALIVE_HOST` and `SIDEKIQ_ALIVE_PORT` directly in its [configuration defaults](https://github.com/arturictus/sidekiq_alive/blob/v2.5.0/lib/sidekiq_alive/config.rb#L26-L29), and its [server configuration converts the port to an integer](https://github.com/arturictus/sidekiq_alive/blob/v2.5.0/lib/sidekiq_alive/server/base.rb#L39-L41). We do not need an application initializer to support those settings.

For localhost-only health checks, operators can uncomment `SIDEKIQ_ALIVE_HOST=127.0.0.1` in their `.env` and restart the worker. Chatwoot's Linux systemd worker loads that file through `dotenv`.

To be explicit about the scope: removing the initializer preserves the gem's `0.0.0.0:7433` default when the variables are unset. The examples are commented out, so they do not change existing deployments automatically. I updated the PR title and description to reflect this documentation-only change; the EOF parser bug still needs a separate fix.

Verified the environment overrides and unset defaults directly against the pinned gem after removing the initializer.

sony-mathew (approved):
lgtm

sony-mathew:
Thank you for the contribution @breken-ai !

PR #15827, opened 2026-09-15
fix(contacts): default the contacts list to most recent activity first

tds-1:
The Contacts page could fail to load on larger accounts. When a user has never picked a sort, the list asked for contacts ordered by **oldest** activity first. No index can serve that order, so every page load read every contact in the account just to return 15. When those rows aren't cached, or the database disk is under load, that runs past the 14s statement timeout and the page never loads. The page now defaults to **most recent** activity first. The existing `(account_id, last_activity_at DESC NULLS LAST)` index serves that order, so a page load reads about 20 pages whatever the account's size. Users who picked a sort keep it.

## Closes

- No existing issue. Found via recurring `PG::QueryCanceled` / `Rack::Timeout` errors on `contacts#index`.

## How to reproduce

1. On a large account, open **Contacts** as a user who has never changed the sort.
2. The page sends `GET /api/v1/accounts/:id/contacts?sort=last_activity_at`, which becomes `ORDER BY last_activity_at ASC NULLS LAST LIMIT 15`.
3. Postgres fetches and sorts every listed contact of the account. When those pages come from disk, the request hits the statement timeout.

## What changed

- `ContactsIndex.vue`: when no `contacts_sort_by` UI setting is saved, the default sort is now `-last_activity_at` instead of `last_activity_at`.
- Measured on production for a ~57k-contact account:
  - old default: 57,768 pages read per load (56 ms cached, about 24s from disk at the measured ~0.42 ms per page);
  - new default: an index scan that stops after 15 rows.
- Contact search uses the same plan for either sort direction (trigram bitmap scan, then a sort of the matches), so it is unaffected.
- Known trade-off: on accounts made up mostly of anonymous visitors, the new order walks more of the activity index before finding 15 listed contacts. A partial index restricted to listed contacts removes that cost if it shows up.


iamsivin (approved):
Code LGTM

PR #15842, opened 2026-09-16
test(email): reproduce OAuth refresh and auth counter failures

scmmishra:
Google inboxes can save refreshed OAuth credentials while returning `nil` to the current IMAP authentication attempt, and later successful fetches leave the authorization error counter intact. Together, these behaviors allow intermittent, self-recovering failures to accumulate toward reauthorization.

This PR adds two intentionally red regression specs requiring:

- `access_token` to return the newly issued token from the same refresh call.
- A successful email fetch to clear previous authorization errors.

The current failures reproduce both defects:

```text
expected: <new access token>
got: nil

expected authorization_error_count: 0
got: 1
```

Related: #15812, #15823


sony-mathew (approved):
lgtm

PR #15882, opened 2026-09-18
docs(swagger): document team icon and icon_color fields

no-hup:
The teams API already accepts `icon` and `icon_color` when you create or update a team, and returns both on the team resource, but the API reference never documented them. Anyone reading the docs wouldn't know they can set or read a team's icon. This adds the two fields so the spec matches what the endpoint actually does.

The fields were introduced with team icons (#14767); the swagger definitions were missed at the time.

### What changed
- Added `icon` and `icon_color` (both `string`) to the team create/update request payload and to the team resource in `swagger/definitions/`.
- Regenerated `swagger.json` from the definitions.


sojan-official (approved):
Thanks for adding these! Verified the fields against the local API. Looks good 👍

PR #15895, opened 2026-09-18
fix: enforce custom role permissions in search

### LINKED ISSUE #11885: Bug: Search Bypasses Custom Role Permissions to Expose Restricted Conversations
### Describe the bug

The global search functionality does not respect the access restrictions configured in a Custom Role. We have set up a role to limit an agent's visibility to only their assigned or participating conversations. However, any agent can use the search bar to find and access any conversation on the platform, even those they are not assigned to. This creates a significant security and privacy vulnerability, as it makes the permission system ineffective.

### To Reproduce

1 - Go to Settings > Roles and create a Custom Role.

2 - Set the conversation permission for this role to Manage conversations they participate in and those assigned to them.

3 - Assign this Custom Role to an agent (let's call them Agent A).

4 - Create or select a conversation and assign it exclusively to a different agent (Agent B).

5 - Log in as Agent A.

6 - Use the search bar and type a word or phrase from the conversation that is assigned only to Agent B.

### Expected behavior

The search results should not show any conversations that Agent A does not have permission to view. The search functionality should honor the permissions set in the Custom Role, and therefore, the conversation assigned to Agent B should not be found.

### Environment

app.chatwoot.com

### Cloud Provider

### Platform

Browser

### Operating system

Debian 11

### Browser and version

Edge

### Docker (if applicable)

_No response_

### Additional context

VPS Hetzner

aakashb95:
## Summary

- Apply the existing conversation permission filter to conversation and message search.
- Use the database search path for restricted custom roles because the search index does not store assignment and participant access.
- Keep the existing fast path for administrators and users with full conversation access.

Fixes #11885

## Verification

- Reproduced the leak locally before the fix. Both the assigned and restricted records were returned from the same inbox.
- `bundle exec rspec spec/services/search_service_spec.rb spec/controllers/api/v1/accounts/search_controller_spec.rb`
- `bundle exec rubocop app/services/search_service.rb enterprise/app/services/enterprise/search_service.rb spec/services/search_service_spec.rb`
- `git diff --check`


sony-mathew (commented):
Please take a look at this finding:

1. **P2 — Restricted agents lose call-transcript search.** The [new fallback query](codex://review?pr=https%3A%2F%2Fgithub.com%2Fchatwoot%2Fchatwoot%2Fpull%2F15895&path=enterprise%2Fapp%2Fservices%2Fenterprise%2Fsearch_service.rb&line=33&side=right) searches attachment metadata but omits `Call#transcript`. Previously, the indexed path included call transcripts through the search presenter.

   Reproduced with a voice-enabled Twilio inbox, an assigned conversation, and a participating-only custom role: searching a term present solely in the call transcript returned **HTTP 200 with no messages**, although the permission scope included that conversation and its search document contained the transcript.

   This confirms the [existing review thread](https://github.com/chatwoot/chatwoot/pull/15895#discussion_r4051432915).

PR #15899, opened 2026-09-19
fix(shopify): read the feature gate without config write side effects

vishnu-narayanan:
## Description

`Shopify::FeatureGate.globally_enabled?` resolved the flag through `GlobalConfigService.load`,
which is a write-capable helper. When the stored value is boolean `false` (the shipped default
for `ENABLE_SHOPIFY_INTEGRATION`), it never takes the cached early return, so every call runs
`InstallationConfig.first_or_create` and `GlobalConfig.clear_cache`. Since the gate is invoked
from the account payload and integration surfaces added in
https://github.com/chatwoot/chatwoot/pull/15756, every dashboard request paid for a DB
roundtrip, a full config cache wipe, and the blocking Redis `KEYS` scan inside `clear_cache`.

The gate now checks `ENV` first and falls back to the cached, side-effect-free
`GlobalConfig.get`. Flag behavior is unchanged: disabled by default, enabled via super admin.
The ENV override now genuinely works as a kill switch; previously it was unreachable once a
config row existed, even though the specs asserted it.

Specs that enabled the gate by stubbing `GlobalConfigService.load` now configure a real
installation config row instead, since the stubbed call no longer happens.

## Type of change

## How Has This Been Tested?

Feature gate spec rewritten against real config rows, covering the disabled/enabled states,
cache survival across calls, no config writes from the read path, and both ENV overrides.
All OSS and Enterprise specs referencing the flag were run locally and pass.

## Checklist:


sony-mathew (approved):
lgtm

PR #15900, opened 2026-09-19
fix: stop clearing the config cache on every read of falsy global configs

vishnu-narayanan:
## Description

`GlobalConfigService.load` returned early only when the cached value was `present?`. A stored
boolean `false` (any `type: boolean` config that is switched off) is not "present", so every
call fell through to the one-time ENV migration path: an `InstallationConfig.first_or_create`
plus an unconditional `GlobalConfig.clear_cache`. On a hot caller this means a DB roundtrip and
a full config cache wipe per call, and `clear_cache` itself ran a blocking Redis `KEYS` command
followed by one `EXPIRE` per key.

Three changes:

- `load` returns cached nonblank values and boolean `false`. Stored blank strings are still
  treated as unconfigured, preserving current behavior for installations that saved empty
  values (for example Firebase settings) from super admin.
- After a cached blank, the cache is cleared when the database lookup returns a configured
  value, including boolean `false`. This also handles a row created by a concurrent request;
  existing blank rows do not trigger a cache clear.
- `clear_cache` uses cursor-based `SCAN` followed by one multi-key `DEL` instead of `KEYS`,
  avoiding a single blocking keyspace scan.

The hot caller that surfaced this is fixed independently in
https://github.com/chatwoot/chatwoot/pull/15899.

## Type of change

## How Has This Been Tested?

New specs cover: a stored boolean false served from cache without clearing other cached
configs, an existing row with a blank value not clearing the cache, a stored empty string
still treated as unconfigured, and the create path clearing a stale cached nil. Existing
`GlobalConfig` cache specs pass unchanged. A deterministic two-reader probe also confirms
that concurrent configuration creation clears the stale cached nil and subsequent reads
are served from cache.

## Checklist:


sony-mathew (commented):
**P2 — New: concurrent config creation can leave stale `nil` cached.** At [global_config_service.rb:14](codex://review?pr=https%3A%2F%2Fgithub.com%2Fchatwoot%2Fchatwoot%2Fpull%2F15900&path=lib%2Fglobal_config_service.rb&line=14&side=right), `previously_new_record?` skips invalidation when another request created the row during this request’s cache miss.

I reproduced two overlapping first reads using real PostgreSQL and Redis:

- **Base:** the cache recovers; five subsequent reads execute **zero database queries**.
- **PR head:** `nil` remains cached with a **24-hour TTL**; five subsequent reads execute **five database queries**.

The service still returns the correct database value, but the shared cache remains stale until expiry or another invalidation.

sony-mathew (approved):
lgtm

PR #15930, opened 2026-09-21
fix(widget): surface failures instead of discarding them silently

### LINKED ISSUE #15929: Widget silently discards sync and conversation-creation failures, leaving a stale chat with no signal
## Describe the bug

Two paths in the widget catch their errors and discard them, so a failure
produces no signal anywhere: no retry, no state change, nothing rendered, and
nothing logged.

**1. `syncLatestMessages` (`app/javascript/widget/store/modules/conversation/actions.js`)**

This is the only routine that recovers messages missed while the websocket was
down, and it runs from `onReconnect`. Its catch block is `// IgnoreError`. A
single failed fetch therefore leaves the conversation permanently stale while
the widget continues to look perfectly healthy. Only a page reload recovers it,
and the visitor has no reason to think a reload is needed.

**2. `createConversation` (same file)**

The same shape. On failure it clears `isCreating` and returns, so a visitor
whose conversation could not be created sees the spinner stop and nothing else.
No error, no indication that retrying is worth doing.

## Why this matters

This is not theoretical. We hit the first one on a self-hosted instance:
real-time delivery stopped for several days, visitors saw a chat that looked
normal but silently omitted agent replies, and nothing in the product reported
it. We only found it by going looking.

The failure mode is the problem rather than the underlying outage. An outage
that shows an error is recoverable by the visitor. An outage that renders as a
working chat is not, and it is invisible to the operator too.

## To Reproduce

1. Open the widget and start a conversation.
2. Make the messages endpoint fail for that conversation, for example by
   returning a 500 from `/api/v1/widget/messages`, or by blocking the request
   in devtools.
3. Drop and restore the websocket connection so `onReconnect` fires.

**Expected:** the visitor is told that the conversation could not be brought up
to date, and is offered a retry.

**Actual:** nothing happens. The widget looks healthy and is missing messages.

The same applies to step 2 against conversation creation: the spinner stops and
no error appears.

## Expected behavior

A failure in either path should be surfaced to the visitor and should be
retryable, rather than discarded.

## Environment

- Chatwoot: reproduced on `v4.18.0`, also present on `v4.6.0`
- Self-hosted, web widget

## Additional context

We have a fix and will open a PR against this issue. It adds a small
exponential-backoff retry helper, two `uiFlags` defaulting to false so the happy
path is unchanged, and an aria-live banner with a retry action. 733 widget and
shared tests pass on `v4.18.0` with it applied, 13 tests are added, and each new
test was validated by reintroducing the fault it guards, with seven mutations
tried and all seven detected.

One thing worth flagging from that work: the existing spec
`#createConversation > sends correct mutations` calls the action without a
`dispatch`, so it threw a `TypeError` on every run that the catch-and-ignore
swallowed, and still passed. Removing the swallow surfaces it. The suite had
been hiding a real error for as long as the silent catch existed.

DhanuOJC:
Closes #15929

## Description

Two paths in the widget catch their errors and do nothing with them, so a failure produces no signal at all: no retry, no state change, and nothing shown to the visitor.

`syncLatestMessages` is the only routine that recovers messages missed while the socket was down, and it is called from `onReconnect`. Its catch block is `// IgnoreError`, so a single failed fetch leaves the conversation permanently stale while the widget continues to look healthy. The visitor sees a working chat that is quietly missing messages, and only a page reload recovers it.

`createConversation` has the same shape: on failure it clears `isCreating` and returns, so a visitor whose conversation could not be created sees the spinner stop and nothing else, with no error and no indication that a retry is worth trying.

This is not theoretical. We hit the first one in production: real-time delivery stopped for several days, customers saw a chat that looked normal but silently omitted replies, and nothing anywhere reported it.

### What this adds

- `retryWithBackoff`, a small helper that retries an async call with exponential backoff. `syncLatestMessages` runs immediately after a reconnect, which is exactly when a transient failure is most likely, so most of these now resolve without the visitor noticing anything.
- `isSyncFailed` and `isCreateFailed` in the existing `uiFlags`, both defaulting to false so nothing changes on the happy path. `isSyncFailed` is cleared only when it was previously set, so a healthy sync emits no state change at all.
- A banner in `ConversationWrap` with a retry action when recovery has failed, and an alert in `PreChatForm` when the conversation could not be started. Both are aria live regions. A flag that nothing renders would leave the failure exactly as invisible as it was before.

### A note on one existing spec

`#createConversation > sends correct mutations` calls the action with `{ commit }` and no `dispatch`. The action calls `dispatch(...)`, so it threw a `TypeError` on every run and the old `// Ignore error` swallowed it; the test asserted the resulting three commits and passed.

Removing the swallow surfaces that `TypeError`, so the spec now supplies the `dispatch` the action actually uses. Worth flagging explicitly: the suite had been hiding a real error for as long as the catch-and-ignore existed.

## Type of change

## How Has This Been Tested?

Against `v4.18.0`:

- Full widget and shared suites: **734 tests across 73 files pass**, and eslint is clean on every file touched.
- 14 tests added: 6 for the retry helper, 3 for the two action paths, 5 for the rendered surface.
- Measured against a baseline on unpatched `v4.18.0`, so that nothing pre-existing is attributed to this change:

  | | Test files | Tests |
  |---|---|---|
  | unpatched `v4.18.0` | 71 | 720 pass |
  | with this patch | 73 | 734 pass |

  The patch adds two spec files and fourteen tests, and the set of passing tests is otherwise identical.
- Each new test was checked by reintroducing the fault it guards against. Eight mutations were tried and all were detected: restoring either original swallow, removing the retry, making the retry resolve `undefined` instead of rethrowing, removing the banner, removing the retry button, showing the banner when healthy, cutting the retry budget, and declaring `retrySync` in a second `methods` block.

```
npx vitest run app/javascript/widget app/javascript/shared
```

### One thing to know if you run this outside UTC

Those figures are with `TZ=UTC`. On my machine, which is `Asia/Kolkata` (+05:30),
**15 tests fail both with and without this patch**, in four spec files:
`availabilityHelpers`, `timeHelper`, `useExactTimestamp` and `ReportsDataHelper`.
They pass again under `TZ=UTC`. None of them is touched by this change, and the
set of failures is identical before and after it, so this is a pre-existing
timezone assumption in the suite rather than anything to do with this PR.
Mentioning it only so the numbers reconcile if you reproduce them somewhere
other than CI. Happy to raise it separately if it is useful.

## Checklist

### Not done here

New strings are added to `en.json` only; the other locales need translating. Happy to follow whatever process you prefer for that.


DhanuOJC:
CI caught a real bug rather than a style nit, so thank you. Force-pushed a corrected commit.

`retrySync` and the `syncLatestMessages` mapping were declared in a **second `methods` block**. The object literal keeps the last duplicate key, so the component's existing `methods` silently discarded both: the banner rendered, the retry button rendered, and clicking it did nothing because `retrySync` was `undefined`. That is `no-dupe-keys` in your config and it was right to stop the build.

Every one of my tests passed over it, because they all asserted the button *existed*. There is now one that clicks it and asserts the action fires. With the duplicate restored, the three original assertions still pass and only that one fails, which is the check I should have written first.

Also fixed in the same push:
- `no-promise-executor-return` in the default `sleep` for `retryWithBackoff`.
- The five `prettier/prettier` findings.

`npx eslint` is now clean on every file the patch touches, and the suite is 734 passing across 73 files under `TZ=UTC`.

PR #15937, opened 2026-09-21
fix: point to the emailed code in the device verification error

vishnu-narayanan:
## Description

Device verification returns its errors as server-rendered strings, and both the web login screen and the mobile app display them verbatim (inline on web, toast plus inline text on mobile). When a wrong code is submitted, the current message, "Invalid verification code", does not tell the user where the correct code actually is.

This matters most on mobile builds that predate [feat(auth): verify new devices with an emailed code on sign-in (chatwoot-mobile-app#1183)](https://github.com/chatwoot/chatwoot-mobile-app/pull/1183). Those builds reuse the MFA screen for the device verification challenge, so the on-screen copy points at an authenticator app. Device verification only triggers for users without MFA enabled, so those users have no authenticator code, and the emailed code is the only valid input. Since the error string comes from the server, updating it here reaches already-shipped app versions without an app update.

The message now reads: "Incorrect code. Enter the 6-digit code we sent to your email."

## Type of change

## How Has This Been Tested?

Validated the locale file parses and the key resolves through I18n. No specs assert this string literally, the controller specs compare against `I18n.t` and the service specs use error symbols, so behavior coverage is unchanged.

## Checklist:


sony-mathew (approved):
lgtm

devi-r (approved):

